PR #7, opened 2023-01-25
No longer assume we are always running in a notebook

cbrew:
If notebook is set to False, it will return a string instead of displaying a styled dataframe.


okhat:
Ah, this is helpful. Thanks!

The evaluate function is pretty basic at the moment. But if you have uses for it outside the notebooks, then it's worth merging this in!

PR #9, opened 2023-01-27
Fix typo in template_v2.py

eltociear:
seperator -> separator


okhat:
Thanks! We need to fix this in all files though. I see it in template v3 also and maybe elsewhere?

eltociear:
@okhat 
Thank you for reaching out! I've used grep to search and fixed everything.

okhat:
Merged. Thank you!

PR #12, opened 2023-02-04
Add KNN for DEMONSTRATE stage

stalkermustang:
Hi Omar & the team,
I was impressed by the DSM paper. Very excited about future applications and research!

It turned out that some tools/primitives/transformations weren't released. The paper says "Though most of the functionality in this section is implemented, the primitives branch, knn, and crossval are currently
work-in-progress". 

So here's my vision and naive implementation for KNN op. It isn't finished though - I'm thinking rn about vectorization process. RN, it's implemented as follows: each Example in the train should have the `vectorized` field. But I have a draft of how it should be organized. I'm thinking about, like, hf/sentence-tranformers vectorization function, that will take a name of an embeddder, download and apply it. Also, it's easy to add OpenAI embeddings as an additional vectorization function. And the last one - read from the memmap-file. What do you think about these?

Any ideas and comments are appreciated!


dsp/primitives/demonstrate.py
```diff
@@ -108,3 +109,44 @@ def answer_match(prediction, answers):
...
+    need_add_one = len(train) % vectorize_train_bs != 0
+    n_batches = len(train) // vectorize_train_bs + int(need_add_one)
+    emb_dim = vectorize(train[:1]).shape[1]
+    all_vectors = np.empty((len(train), emb_dim), dtype=np.float32)
+    # batched vectorization for train Examples
+    for batch_idx in range(n_batches):
+        batch_start_idx  = batch_idx * vectorize_train_bs
+        batch_end_idx = (batch_idx + 1) * vectorize_train_bs
+        cur_batch = train[batch_start_idx : batch_end_idx]
+        # cur_batch = [cast(cur_elem) for cur_elem in cur_batch]
```
stalkermustang, line 138: This is commented for demonstration, and because "vectorization" happens outside this block of code.

okhat:
This is an impressive PR, @stalkermustang — thanks!

I wonder if we'll need faiss in, though. It doesn't work reliably with pip.

Most of the time people will use kNN with anything between 100 to 100k examples, which is a small enough set that maybe we won't need faiss by default.

Maybe it should be a knn_fast or something, but people will have to install faiss themselves (ideally with conda, but in any case it won't be in the main dependencies of dsp).

Thoughts?

stalkermustang:
Hey @okhat 
Here is the up-to-date CPU version (PyPI [link](https://pypi.org/project/faiss-cpu/)). As far as I know, problems with PyPi are caused because of GPU-specific libraries and stuff, so the CPU should be a safe bet. We need to change only one row in the setup.py file in this case. Everyone will use a speedy yet scalable approach - if one needs GPU, well, no code changes are required. It's flexible. 

Pros:
- easy to switch to GPU in case of need
- it's a common and popular approach with a well-written doc & tons of examples
- I haven't had any issues with this CPU version for, like, two years of active use in my work

Cons: 
- As I said, I haven't any issues with this CPU version, but I haven't checked, for example, M1-M2 Mac Chips with this (still use Intel D: )

This said I could add an option to switch the backend, say, to SKLearn or knn_fast (what it is, exactly? could you send me a link, please?).

----
By the way, what do you think about vectorization? It wasn't stated in the paper. In the following signature
`knn ( train : Examples , cast : Callable [[ Example ] , str ])`
we have no vectorization argument. I still think that there are at least three popular cases:
- set embeddings manually for each example
- vectorize in knn function using HF's pipleine
- query OpenAI

and they should be available for passing as an argument. Say, 
`knn ( train : Examples , cast : Callable [[ Example ] , str ], vectorization: Callable [[ Examples ], List [ np.ndarray ]])`

stalkermustang:
I've tested CPU-version (both KNN+vectorization), but haven't tested GPU though.  Also, need to update/add docstrings for new complicated functions.

okhat:
Really impressed by this. Will look into merging asap... we're slightly bottlenecked as we've been working on a major new feature.

Btw we didn't have a vectorize fn in the original interface because it's supposed to use the same RM (colbert, etc) as the rest of the program. But I can see the value in allowing more customization.

The DSP way for that would be:

```
dsp.settings.configure(vectorizer=....)

OR

with dsp.settings.context(vectorizer=...):
  ....
```

stalkermustang:
>  it's supposed to use the same RM (colbert, etc)

Yeah, I thought the same, but I wasn't able to understand how to use provided API endpoint for vectorizing custom queries. For regular users like me, ColBERT with its per-token embeddings (as well as pre-clustered quantized embeddings, etc.) might be not that easy to deploy/use, and might seem like overkill. Maybe I'm wrong. I believe in most cases top-scored candidates will be almost the same, given the metrics on various IR datasets (like 1-2-3% difference on huge query sets).

stalkermustang:
hey, @okhat 
image
I've changed the interface to match DSPs. What else do I need to do in this PR? I'm thinking about:
- OpenAI embedded (GPT-based)
- update docstrings for new classes and some functions

Anything else? I don't think I can integrate ColBERT-based retrieval with your endpoint, as it (probably) doesn't have a search for train examples (and it's OK).

stalkermustang:
OpenAI embeddings were added, but I decided not to cache them. First, the cache is subject to change, as @okhat mentioned in #13. Second, I added batched vectorization, and thus send not all text at once but only a part of them. This is slightly faster than sending them one by one. Please lmk if we need this caching feature.

okhat:
Awesome, thank you @stalkermustang ! I'll look into this and merge if everything looks good probably tomorrow (Pacific)

stalkermustang:
JFYI tested both vectorization + faiss-gpu on 3090, and everything works well. Can't test a multi-GPU setup, but it should be OK too. 
Only one change was added: 
`knn_func = dsp.knn(train, >>max_gpu_devices=1<<)`, because by default the amount of GPUs to store the FAISS index equals zero (we don't want to cut some vRAM without direct instruction from the user -> I decided to make default value = 0 GPUs to use)

setup.py
```diff
@@ -12,10 +12,12 @@
     python_requires='>=3.8',
     install_requires=[
         "backoff",
+        "faiss-cpu",
```
okhat, line 15: I suggest we remove `faiss-cpu` and `sentence-transformers`. These are both heavy and optional dependencies.

Instead, either the kNN notebook should instruct users to install faiss (link to https://github.com/facebookresearch/faiss/blob/main/INSTALL.md) and then sentence-transformers (pip).

OR, if these don't exist, maybe it makes sense to pip install them inside the code itself?

Basically, faiss-cpu will break repos that already have the conda version of GPU faiss. So we'd want to avoid that. Perhaps many users (who will have tiny training sets) will not need kNN right away, so it makes sense to keep the dependencies optional?

knn_demonstrations_test.ipynb
```diff
@@ -0,0 +1,323 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dsp\n",
+    "import os\n",
+    "import openai\n",
+    "# api_key = '<replace with your API key>'\n",
+    "# openai.api_key = api_key\n",
+    "\n",
+    "os.environ['KMP_DUPLICATE_LIB_OK'] = 'TRUE'"
```
okhat, line 15: I don't recognize this. Is it necessary & safe? Will people need to use it everytime they use the kNN functions?

knn_demonstrations_test.ipynb
```diff
@@ -0,0 +1,323 @@
+{
```
okhat, line 1: This is great, thanks! Can we perhaps move this file to a `tests/` directory?

It's currently good for testing but since it doesn't yet have a realistic dataset let's not have it at the top level (yet). We can still link to it from the README and, in the future, add it extensively in the docs + add a notebook that shows the benefits of kNN.

okhat (changes requested):
Thanks a ton @stalkermustang ! This is ready to merge except for the couple of notes I left. What do you think about these, particularly the dependencies comment?

knn_demonstrations_test.ipynb
stalkermustang: No, they don't. The reason I added this is that I use VS Code, and it (seems) actually runs everything under the hood from the same kernel. Thus, I just can't use FAISS if I already imported/used it during the session. 

KMP_DUPLICATE_LIB_OK=True tells OpenMP to not complain if it notices that two copies of OpenMP are loaded. Don't see how this can hurt safety. If people use Jupyter - it would work well even without this env variable. But I expect some people will get the same error as I got, thus, I decided to keep this row in the demo ipynb.

setup.py
stalkermustang: > pip install them inside the code itself?

I don't think this is a good practice to install something without notice or asking or changing env in general. 
Just mentioning that a user needs to install smth in a notebook is ok. I'll add try-except catches to the code and notes to the notebook.

stalkermustang:
@okhat fixed

okhat:
Many thanks for this!

okhat:
Hey @stalkermustang , one more fix for this. The faiss import is now imported whether or not it’s needed. This is because the ann file is imported always in the utils. 

Can we remove that import from the utils init.py?

okhat:
Or any other fix you see fit

okhat:
Just pushed a quick and dirty hot fix for now. But we may need to check how the imports are done in the knn notebook

PR #16, opened 2023-02-19
Resolve import issues for FAISS

stalkermustang:
@okhat resolves last comments in #12


okhat:
Thank you! Do we still have a similar issue for init.py of modules/?

stalkermustang:
As for now, FAISS or sentence_transformers imports won't be called unless you use the exact classes with the same purpose (build an index or use a transformer as a vectorizer).

PR #30, opened 2023-03-21
Add code for supporting arbitrary Hugging Face models

santhnm2:


okhat:
Thanks a ton @santhnm2 ! Just resolved (in an imperfect way) a conflict.

I think the only thing pending is to not have the imports require `pip install transformers` or other dependencies unless they're needed/used.

For instance, we can avoid importing the HF paths by default. If they're imported, they'll do a try/except where they try to import transformers (or other optional libs) and if that fails they print the generic command (or instructions) for installing the missing deps.

okhat:
Merging for now, with commeting out the import that might cause trouble. Some more work is needed to address the comments above more properly for the longer term

PR #37, opened 2023-04-08
enhancement(GPT-LLM): azure API support

lawliet19189:
Adds support for Azure OpenAI API Service.

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -1,13 +1,14 @@
 import functools
 import json
-from typing import Optional, Any, cast
+from typing import Any, Literal, Optional, cast
+
+import backoff
 import openai
 import openai.error
 from openai.openai_object import OpenAIObject
-import backoff
 
-from dsp.modules.lm import LM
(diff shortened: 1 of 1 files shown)


dsp/modules/gpt3.py
```diff
@@ -20,21 +21,43 @@ def backoff_hdlr(details):
...
+
+        if api_provider == "azure":
+            assert (
+                "engine" in kwargs or "deployment_id" in kwargs
+            ), "Must specify engine or deployment_id for Azure API instead of model."
+            assert "api_version" in kwargs, "Must specify api_version for Azure API"
+            assert "api_base" in kwargs, "Must specify api_base for Azure API"
+            openai.api_type = "azure"
+            openai.api_base = kwargs["api_base"]
+            if kwargs.get("api_version"):
```
okhat, line 54: This might be a bit redundant with the assert above. Not sure what's better to drop (assert or if).

lawliet19189: ah, yes. I missed removing this. Initially, I was under the impression that api_version was optional, so added this if case. Will remove this in a further PR.

Issue #47, opened 2023-04-24
text-davinci-001: majority_vote_ function does not handle cases where pred[prediction_field] is a list of str instead of str


jupiterepoch:
Link to the code snippet in question:
https://github.com/stanfordnlp/dsp/blob/12d7f1106f2f524be161d35865a81e0ea016e929/dsp/primitives/predict.py#L211

Expected behavior:
The function should work with `pred[prediction_field]` of type `Union[List[str], str]`.

Current behavior:
The function raises an error when calling `normalize_text` on `pred[prediction_field]` when it is of type `List[str]`.

okhat:
Thanks a lot for catching this! Do you have an example of where that might happen?

jupiterepoch:
![image](https://user-images.githubusercontent.com/65405767/234073937-7e2b5bd2-5f9c-46c4-98c7-1ac0c98ac1e9.png)

okhat:
Hmm, thanks! We do already do CoT+SC with dsp.majority in the intro notebook.

Here's a simplified version.

```
@dsp.transformation
def QA_predict(example: dsp.Example):
    example, completions = dsp.generate(qa_template_with_CoT, n=20, temperature=0.7)(example, stage='qa')
    completions = dsp.majority(completions)
    
    return example.copy(answer=completions.answer)
```

The only difference I see is that we extract the answer at the end. Is there anyother difference?

okhat:
Maybe you're saying we have some regression. The same code from the intro notebook no longer works on the latest branch?

jupiterepoch:
I think I know what the problem is. `text-davinci-001` sometimes returns a `List[str]`, whereas `gpt-3.5-turbo` or `text-davinci-002` consistently returns `str` in my experiments. You could try running the intro notebook with text-davinci-001.

I guess either specifying that the code does not support certain GPT models, or adding codes to handle the different LM behaviors could work.

If you want to go down the second path, the `evaluate` would probably need to change as well.

okhat:
If your PR fixes this, I'm happy to merge it. But I'm unclear on what you mean by text-davinci-001 returning a list. Don't all models return a list of strings (i.e., a list of completions)? Are you saying that text-davinci-001 returns a list of lists?

jupiterepoch:
> If your PR fixes this, I'm happy to merge it.
Please don't merge just yet. As I mentioned, the `evaluate` function has similar problems. If you want to fix it, it's better to fix both.
[...]

PR #50, opened 2023-04-26
feat: Unset Output Variable in `dsp.generate` to Prevent Leakage

tsunrise:
This PR addresses the issue raised in https://github.com/stanfordnlp/dsp/issues/47#issuecomment-1523430127.

The changes in this PR ensure that output variables in `dsp.generate` are unset if the template fails to parse the response corresponding to the output variables.

It is always recommended to pass only the necessary questions to `dsp.generate` or unset the output that users expect the function to produce. However, users often forget to do this when evaluating the dev set. In such cases, the gold answers remain in the output variable of the returned completion, which can lead to malformed output (e.g., a list) or an overestimation of performance. This PR serves as a temporary workaround for this issue.

closes: #47

### DIFF
--- a/dsp/templates/template_v2.py
+++ b/dsp/templates/template_v2.py
@@ -129,13 +129,15 @@ class TemplateV2:
         )
 
     def extract(
-        self, example: Union[Example, dict[str, Any]], raw_pred: str
+        self, example: Union[Example, dict[str, Any]], raw_pred: str, output_fields: list = []
     ) -> Example:
         """Extracts the answer from the LM raw prediction using the template structure
 
         Args:
             example (Union[Example, dict[str, Any]]): Contains the input variables that raw_pred was completed on.
             raw_pred (str): LM generated string
+            output_fields (List[str]): List of field names that we expect to extract. Those fields will be set to an empty string 
+                                       if they cannot be extracted to prevent leakage of gold answers.
 
         Returns:
             Example: The example with the output variables filled in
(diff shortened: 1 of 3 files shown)


dsp/templates/template_v2.py
```diff
@@ -142,6 +142,11 @@ def extract(
         """
         example = dsp.Example(example)
 
+        # Unset the output variables to prevent leakage of gold answers
+        for field in self.fields:
+            if field.output_variable in example:
```
okhat, line 147: Thanks a lot for opening this PR! Q: Isn't this a bit too aggressive? My quick impression is that it'll basically unset *all* values in the example

tsunrise: Thanks for the review! I have done the following changes:

- Now we only unset a part of fields. In `do_generate` and `extract` function, I've added an optional argument called `output_fields`, meaning the fields that user expect the LM to output/extract. The default value fo `output_fields` in `do_generate` is `['answer']`, which I believe is the most common use case. User can specify the output fields as well. For example, if the user expects to extract the `rationale`, it can be done by
```python
dsp.generate(my_qa_template_with_passages)(example, stage='qa', output_fields=['answer', 'rationale'])
```

- As you requested, I added some unit tests to illustrate how `extract` function behave after this change. Test cases includes some well-formed raw reponses, and a malformed one. You can run them using
```
python -m unittest tests.extract_test
```

okhat:
Thank you Tom. We've dealt with this important issue by redesigning how fields are passed around in DSPy. It completely resolves this concern more fundamentally now.

PR #53, opened 2023-05-01
bug(chatgpt-lm): remove unnecessary param passing

lawliet19189:
addresses #52 bug


tkdagdelen:
Thanks for fixing! Saw the same bug earlier this week. Has the package been updated on PyPI since pulling in this fix?

okhat:
No updates to pypi yet — do you need this currently?

tkdagdelen:
Not a blocker or anything--I just re-installed directly from the repo when I saw this ticket. But I'm sure most folks will go for the pip version first like I did, so they'll likely hit the same bug. Luckily I had the benefit of there being a fresh ticket with a well-written description at the top fo the stack to send me in the right direction :)

PR #68, opened 2023-06-07
Basic Adapter vs Chat Adapter with different inputs, flags to come

tomjoshi:


tomjoshi:
Redoing flags according to our conversation

detaos:
Small thought: It would be nice to have one abstract base case in adapter.py, then a folder with one file per model-specific adapter, like the one for Llama and another for Davinci, etc.

okhat:
Yup @detaos

PR #72, opened 2023-06-29
feature(new-caching): sqlite caching + integration into gpt3 and colbert

lawliet19189:
What the PR does:
- Creates a new caching logic using sqlite
- makes experiments reproducible using cached timeranges:
   - eg: if an example is not part of the specified time-range, it would error out instead of re-computing. This helps us to make the experiments consistent and re-producible.
- integrated the logic into gpt and colbert module.
   - usage: `dsp.settings.configure(lm=lm, rm=rm, experiment_end_timestamp=..., experiment_start_timestamp=...)`  # in timestamp
- Adds a testing notebook for understanding different behaviours of the cache. 
- adds a common logger which can be used for debugging along with custom logging level
- removes joblib caching logic


dsp/modules/colbertv2.py
```diff
@@ -24,18 +25,27 @@ def __init__(
     def __call__(
         self, query: str, k: int = 10, simplify: bool = False
     ) -> Union[list[str], list[dotdict]]:
+        cache_args: dict[str, Union[str, float]] = {
+            "worker_id": str(uuid.uuid4()),
+            "cache_end_timerange": dsp.settings.config["experiment_start_timestamp"],
+            "cache_start_timerange": dsp.settings.config["experiment_end_timestamp"],
```
lawliet19189, line 31: TODO: the mapping is incorrect. Needs to be fixed.

okhat:
cc: @lawliet19189 if it's easier for Sri to check

arnavsinghvi11:
> @arnavsinghvi11 Can we close this? I think https://github.com/stanfordnlp/dspy/pull/189/files has everything here?

ah I believe this might be the wrong PR but [this branch](https://github.com/stanfordnlp/dspy/pull/103) has all the commits from here, including the slight changes I added. If all looks correct, we can close this one (72) and merge (103).

okhat:
Yes close this one but don’t merge anything yet just ping me on the right one and close all the wrong ones

PR #91, opened 2023-08-22
updated with mlc models post-processing

arnavsinghvi11:
updated DSPy internals with post-processing for MLC Llama models - `.split('\n')[0]`


okhat:
Closing since this is a bit hacky, but we need a more general thing for MLC and HFModel. They're not in a great state yet.

PR #118, opened 2023-09-15
Add types to signatures

santhnm2:


okhat:
Thanks a lot Keshav.

@lawliet19189 Can we play with this with GPT-3.5 and see if it work well for typical tests?

thomasahle:
Do you still need this now we have TypedChainOfThought and TypedPredictor?

PR #119, opened 2023-09-16
Integrate Pyserini into DSPy

jasper-xian:
Integrate pyserini to enable users to use their own data for retrieval in dspy. I've added a small notebook to demo how it works.

### DIFF
--- a/dsp/modules/pyserini.py
+++ b/dsp/modules/pyserini.py
@@ -14,7 +14,7 @@ class PyseriniRetriever:
                  query_encoder: str = 'castorini/dkrr-dpr-nq-retriever', 
                  index: str = 'wikipedia-dpr-dkrr-nq', 
                  dataset: Dataset = None,
-                 id_field: str = 'id',
+                 id_field: str = '_id',
                  text_fields: list[str] = ['text']) -> None:
         """
         Args:
@@ -42,14 +42,14 @@ class PyseriniRetriever:
             self.searcher = FaissSearcher(index_dir=index, query_encoder=self.encoder)
 
     def __call__(
-        self, query: str, k: int = 10, simplify: bool = False, threads: int = 16,
+        self, query: str, k: int = 10, threads: int = 16,
     ) -> Union[list[str], list[dotdict]]:
         hits = self.searcher.search(query, k=k, threads=threads)
         
         topk = []
         for rank, hit in enumerate(hits, start=1):
             if self.dataset is not None:
-                # Search through user's dataset for the document (could be slow), maybe room for improvement?
+                # Search through user's dataset for the document
                 row = self.dataset[self.id_field].index(hit.docid)
                 text = ' '.join(self.dataset[field][row] for field in self.text_fields)
                 pid = self.dataset[self.id_field][row]
@@ -66,8 +66,5 @@ class PyseriniRetriever:
                 'score': hit.score,
                 'rank': rank,
             })
-
-        if simplify:
-            return [psg["long_text"] for psg in topk]
         
         return [dotdict(psg) for psg in topk]
--- a/dsp/modules/pyserini.py
+++ b/dsp/modules/pyserini.py
@@ -49,7 +49,7 @@ class PyseriniRetriever:
         topk = []
         for rank, hit in enumerate(hits, start=1):
             if self.dataset is not None:
-                # Search through user's dataset for the document
+                # Linearly search through user's dataset for the document, maybe some room for future improvement here
                 row = self.dataset[self.id_field].index(hit.docid)
                 text = ' '.join(self.dataset[field][row] for field in self.text_fields)
                 pid = self.dataset[self.id_field][row]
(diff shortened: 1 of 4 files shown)


dsp/modules/pyserini.py
```diff
@@ -0,0 +1,73 @@
...
+            topk.append({
+                'text': text,
+                'long_text': text,
+                'pid': pid,
+                'score': hit.score,
+                'rank': rank,
+            })
+
+        if simplify:
+            return [psg["long_text"] for psg in topk]
```
okhat, line 71: what is this supposed to do for Pyserini, I think we can remove the whole `simplify` argument

dsp/modules/pyserini.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __call__(
+        self, query: str, k: int = 10, simplify: bool = False, threads: int = 16,
+    ) -> Union[list[str], list[dotdict]]:
+        hits = self.searcher.search(query, k=k, threads=threads)
+        
+        topk = []
+        for rank, hit in enumerate(hits, start=1):
+            if self.dataset is not None:
+                # Search through user's dataset for the document (could be slow), maybe room for improvement?
```
okhat, line 52: what would make it slow? isn't it just a lookup? or does it go to disk?

dsp/modules/pyserini.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class PyseriniRetriever:
+    """Wrapper for retrieval with Pyserini. Supports using either pyserini prebuilt faiss indexes or your own faiss index."""
+
+    def __init__(self, 
+                 query_encoder: str = 'castorini/dkrr-dpr-nq-retriever', 
+                 index: str = 'wikipedia-dpr-dkrr-nq', 
+                 dataset: Dataset = None,
+                 id_field: str = 'id',
```
okhat, line 17: should the default be 'id' or '_id'?

okhat:
This is *great*. I left three short comments but otherwise I'm happy to merge this!

dsp/modules/pyserini.py
jasper-xian: sure, i left it there for the sake of continuity but it's not useful for pyserini (I think), so we can remove it

dsp/modules/pyserini.py
jasper-xian: yeah it's just a linear lookup through the dataset

ronakice:
Thanks a lot for this @jasper-xian and @okhat for going through this! Great stuff :)

dsp/modules/pyserini.py
ronakice: Yeah some room for the future improvements!

jasper-xian:
Changes made! I recall we discussed a step-by-step approach to integrating this (first class then indexing, etc.) and it might be useful for you to run the notebook first as a sanity check, but otherwise happy to merge.

Also, maybe I should move the notebook into the docs folder? Up to you @okhat for repo organization.

dsp/modules/pyserini.py
jasper-xian: for the indexible Dataset, it'll be faster to create a map from docid to index during initialization, so maybe we can implement that? @okhat if you have any other ideas let me know as well

jasper-xian:
Also, might be useful to add a blurb about this to the README, perhaps in the Tutorials section?

okhat:
@jasper-xian yes we should add this to the README and we should think more about more "native" indexing, but let's make these two new PRs.

okhat:
The first PR we need after this, though, is Colab runnable notebook. I'll merge your notebook as-is, but have you guys ever ran Pyserini indexing and inference inside Colab? I think it will "just work" since ColBERT just works too.

jasper-xian:
@ronakice can probably provide more insight on running pyserini in Colab, but I'd imagine it should work as is

okhat:
See our own dspy colab notebook. Basically you need an if statement at the top: if this is colab and pyserini is not pip installed, then pip install faiss , pytorch, and pyserini. That’s it really

okhat:
Btw there’s a link pattern to open any github ipynb into colab. See how our url for colab is structured in the readme and just replace that with your own ipynb branch path

dsp/modules/pyserini.py
okhat: It’s a linear lookup? Or constant time lookup?

dsp/modules/pyserini.py
```diff
@@ -0,0 +1,70 @@
+from typing import Union
+import json
+from datasets import Dataset
+
+from dsp.utils import dotdict
+from pyserini.search import FaissSearcher
```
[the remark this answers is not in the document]
jasper-xian: changes made!

jasper-xian:
I see, I can make the colab changes in a separate PR?

jasper-xian:
Running on Colab right now, getting an error `AttributeError: module 'dspy' has no attribute 'Pyserini'`

I think merging this PR and then doing `git clone` for dspy (same logic as accessing the cache in the intro) should fix this until the next dspy pip release?

okhat:
Merged! Let's get a supplement to this that makes Colab work?

jasper-xian:
yup, on it!

jasper-xian:
Colab PR: #120

PR #120, opened 2023-09-16
Colab changes for Pyserini

jasper-xian:
Continued from #119


okhat:
I added requirements.txt so you can do:

```
    !pip install -r $repo_path/requirements.txt  # adjust this line according to the location of requirements.txt in the repo
```

and remove `pip install dspy-ai` because I think this will not include dspy.Pyserini.

okhat:
btw maybe you figured this out already but this is your colab:

https://colab.research.google.com/github/jasper-xian/dspy/blob/pyserini-colab/pyserini.ipynb

jasper-xian:
Yup, thanks! I've got `pip install -e .` in colab and I think that seems to work as well?

I'm having an issue with pyserini faiss encoding on Colab, somehow after the encoding step only the `docid` file shows up but the `index` file does not. @ronakice any ideas?

okhat:
pip install -e . is better actually, makes sense

okhat:
is the example index too big? colab may be running out of disk space?

jasper-xian:
don't think so, corpus is purposefully really small and leaves 80GB of space left on colab

jasper-xian:
Did some digging, there seems to be a lot of issues with faiss when running on colab, so I'm thinking this might be a faiss issue...

I'll loop this with @ronakice once he's back

okhat:
Thanks a lot for the deep investigation! What about BM25 instead?

okhat:
Honestly BM25 will work better and faster than DPR on almost every dataset anyway

jasper-xian:
@okhat turns out it was a colab CPU issue, it ran out of memory and the faiss index didn't write. Using the free T4 GPU on colab works, does indexing in <1min for the nfcorpus dataset. This PR should be good to merge now.

With regards to BM25, @ronakice probably knows more about it and if it's better/faster we can implement it in a future PR?

okhat:
Thanks! Your work is incredible, Jasper. Merged.

ahmed-moubtahij:
Any news on using dspy.Pyserini with BM25?

PR #122, opened 2023-09-17
added ReAct module

arnavsinghvi11:
-added ReAct module support within DSPy


arnavsinghvi11:
_Next steps_
Objective: generalizing ReAct Module
-should support any signature of InputFields and OutputFields
-should use list of tools to build out model behavior
    -tool.__doc__ builds out instructions
    -tool.usage() shows description of Action
    -each tool has __call__ to run that tool
-Finish doesn't need a tool (but can be implemented as appending a Finish tool to list). also only one that breaks

Issue #128, opened 2023-09-19
Early stop when scores reach 1.0


talperetz:
`BootstrapFewShotWithRandomSearch` keeps searching even when gets to:

`Average Metric: 3.0 / 3  (100.0%)
Score: 100.0 for set: [3]
Scores so far: [0.0, 0.0, 33.33, 100.0, 100.0]
New best score: 100.0 for seed 1
Best score: 100.0
Average of max per entry across top 1 scores: 1.0
Average of max per entry across top 2 scores: 1.0
Average of max per entry across top 3 scores: 1.0
Average of max per entry across top 5 scores: 1.0
Average of max per entry across top 8 scores: 1.0
Average of max per entry across top 9999 scores: 1.0`

okhat:
haha what a good problem to have — thank you for the note, will fix

okhat:
oh actually it's a bit trickier... we can't assume the scores are out of 1.0, what if your metric returns 2 or 6 or 126.1?

okhat:
But we can add a stop_at_score keyword argument. Btw for now you can easily modify max_candidate_programs from 16 to like 3

talperetz:
`stop_at_score` would work well

okhat:
Fix merged. You can now use it if you install from git, eg with `pip install -e .`

Will be part of the next pip release too if you don’t wanna use it this way.

PR #130, opened 2023-09-20
added stop_at_score kwarg to random_search

arnavsinghvi11:
added stop_at_score per [issue request ](https://github.com/stanfordnlp/dspy/issues/128)

### DIFF
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -100,10 +100,6 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
 
             print(f"Scores so far: {scores}")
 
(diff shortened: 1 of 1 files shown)


dspy/teleprompt/random_search.py
```diff
@@ -100,6 +100,10 @@ def compile(self, student, *, teacher=None, trainset, valset=None):
 
             print(f"Scores so far: {scores}")
 
+            if self.stop_at_score is not None and score >= self.stop_at_score:
+                print(f"Stopping early because score {score} is >= stop_at_score {self.stop_at_score}")
+                break
+
```
okhat, line 106: Shouldn’t this new block be moved to the very end of the loop code. In particular: after updating the best program?

PR #131, opened 2023-09-20
Change PyseriniRetriever linear index search to lookup table

jasper-xian:
retrieval when using your own data should be faster now, let me know if you have any other ideas/changes in mind


okhat:
Looks great. Merged!

Issue #105, opened 2023-09-07
Example for using dspy from a multi threaded Flask application


hiradha:
I am trying to use spy from a Flask app server to serve api requests.
However I am getting the below error 

`File "/Users/popurir/nsx-sidekick/sidekick-app/services/vmware/dspy_vaqa_service.py", line 41, in configure_dspy
    dspy.settings.configure(lm=self.lm, rm=self.rm, inherit_config=inherit_config)

  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/dsp/utils/settings.py", line 70, in configure
    self.__append(config)

  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/dsp/utils/settings.py", line 54, in __append
    self.stack_by_thread[threading.get_ident()].append(config)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^`

    
Seems to be a case of how dspy.Settings is possibly a singleton per thread and not available across threads?
I do not know multi threading nuances in python enough.
Please provide an example of how to use dspy in multithreaded servers.

okhat:
Is this the full error message

hiradha:
> 

Full stack trace is here . 
```
2023-09-07 09:41:01,102] ERROR in app: Exception on /intelligence/sidekick/prompt [POST]
Traceback (most recent call last):
  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/popurir/nsx-sidekick/sidekick-app/venv/lib/python3.11/site-packages/flask/app.py", line 1469, in dispatch_request
[...]

PR #134, opened 2023-09-21
modified threading handling in config for multithreaded flask

arnavsinghvi11:
handling multithreading within dsp.settings in setting config
modified relevant methods that involve thread->stack dictionary

refers to this [issue](https://github.com/stanfordnlp/dspy/issues/105)

### DIFF
--- a/dsp/utils/settings.py
+++ b/dsp/utils/settings.py
@@ -39,7 +39,10 @@ class Settings(object):
 
     @property
     def config(self):
-        return self.stack_by_thread[threading.get_ident()][-1]
+        thread_id = threading.get_ident()
+        if thread_id not in self.stack_by_thread:
+            self.stack_by_thread[thread_id] = [self.main_stack[0].copy()]
+        return self.stack_by_thread[thread_id][-1]
 
     def __getattr__(self, name):
         if hasattr(self.config, name):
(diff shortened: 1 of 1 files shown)


dsp/utils/settings.py
```diff
@@ -39,7 +39,10 @@ def __new__(cls):
 
     @property
     def config(self):
-        return self.stack_by_thread[threading.get_ident()][-1]
+        thread_id = threading.get_ident()
+        if thread_id not in self.stack_by_thread:
+            self.stack_by_thread[thread_id] = [self.main_stack[0].copy()]
```
okhat, line 44: shouldn't this be main_stack[-1]?

okhat: same below etc, not 100% sure but seems so

arnavsinghvi11: ah yes if we want new threads to start with the latest configuration, self.main_stack[-1] would be the correct fix. however, if we wanted to maintain the original configuration for every thread, i defaulted it to the current setup but can change as per our use case.

okhat: It should definitely start from the parent thread, not from scratch

GuyAglionby:
Thanks for this, it works for my use case where I'm using `HFClientTGI` in multiple threads.

eggie5:
thanks for this, wasn't working w/ FastAPI

PR #138, opened 2023-09-23
added adapters for DSPy

arnavsinghvi11:
-adapters for general LMs (DavinciAdapter), chat LMs (TurboAdapter) and Llama models (LlamaAdapter)

### DIFF
--- a/docs/language_models_client.md
+++ b/docs/language_models_client.md
@@ -1,0 +1,158 @@
+# LM Modules Documentation
+
+This documentation provides an overview of the DSPy Language Model Clients.
+
+### Quickstart
+
+```python
+import dspy
+
+lm = dspy.OpenAI(model='gpt-3.5-turbo')
+
+prompt = "Translate the following English text to Spanish: 'Hi, how are you?'"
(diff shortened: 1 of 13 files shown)


darinkishore:
just a reminder to revoke that key if you haven't already!

theta-lin:
Hi @arnavsinghvi11, this looks pretty neat as it would allow a lot of flexibility for prompting for different models (or even the same model but using different ways to state your instructions). I wonder what is holding you back from merging it?

If there are issues regarding merging the entire thing, then would you consider a more immediate but less drastic change: 
As `signature_to_template()` in
https://github.com/stanfordnlp/dspy/blob/d8b8909773fc31e72cec093db2f26109590e524e/dspy/predict/predict.py#L137-L140
has an optional `adapter` parameter, which actually takes a template as opposed your adapter, I propose that we can add a setting to change the default template, so that the behavior of the template could still be customized to some extent (albeit a bit less modularized).

Additionally, I still see a bit of inflexibility with `TemplateV2.guidelines()`, which could be addressed in the future, though. I believe that adapters should also have the ability to customize the format of the guidelines.

PR #146, opened 2023-10-08
modified signatures to handle new DSPy signature format

arnavsinghvi11:
-supports new DSPy signature declarations (`dspy.Signature("{inputs}-> {outputs}"`)

methods added:
-`attach()`: defining field with prefix and description 
-`add_field()`: supports pretending and appending InputFields and OutputFields

(note for reference - currently, modifies signature in-place as the `attach()` functionality supports chaining)

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -44,7 +44,6 @@ class SignatureMeta(type):
         # Redirect attribute access to the template object when accessed on the class directly
         return getattr(cls._template, attr)
 
-
 class Signature(metaclass=SignatureMeta):
     def __init__(self, signature: str = "", instructions: str = ""):
         self.signature = signature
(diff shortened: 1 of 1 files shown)


dspy/signatures/signature.py
```diff
@@ -34,22 +34,75 @@ def __new__(cls, name, bases, class_dict):
...
+        return self
+
+    def add_field(self, field_name: str, field_type, position="append"):
+        if field_name in self.fields:
+            raise ValueError(f"{field_name} already exists in fields.")
+        if isinstance(field_type, (InputField, OutputField)):
+            field_instance = field_type
+        else:
+            raise ValueError(f"non-existent {field_type}.")
+        if position == "prepend":
```
okhat, line 85: I don't think this will handle append/prepend correctly considering that inputs must all go before outputs

okhat: Probably need to find the endpoint of the input fields (for append to inputs or prepend to outputs)

okhat:
Do we close this one now, @arnavsinghvi11

okhat:
posted important old notes above, they probably didn't show for you before

also do we close this PR since this stuff is now part of PoT's PR?

PR #148, opened 2023-10-09
Pinecone retrieval improvements

msnidal:
Firstly thank you for the great library. I was interested in using the Pinecone retrieval component and ran into several issues, including:

* Hardcoded api/env placeholders
* Undeclared dependency (reasonably to not force pinecone install for other users)
* Lack of support for multiple queries (used only results of first embedding)
* Deprecated openai embedding kwargs
etc.

This aims to address these issues by overhauling the pinecone retrieval class in many ways: fixing the hardcoded keys, including the dependency as optional so you can use `pip install dspy-ai[pinecone]`, and implementing similar behaviour to `dsp.retrieveEnsemble` to allow search against a query list, along with other code quality improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,11 @@ Or open our intro notebook in Google Colab: [<img align="center" src="https://co
 
 > _Note: If you're looking for Demonstrate-Search-Predict (DSP), which is the previous version of DSPy, you can find it on the [v1](https://github.com/stanfordnlp/dspy/tree/v1) branch of this repo._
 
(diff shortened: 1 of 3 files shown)


okhat:
This is really cool, thanks @msnidal 

@drawal1 how does this look to you, will it still support what you need

drawal1 (approved):
makes it robust and flexible. Love it! I will test it later to see if it needs any modifications for my use case.
Note: retriever documentation needs to be updated [here](https://github.com/stanfordnlp/dspy/blob/main/docs/retrieval_models_client.md)

PR #170, opened 2023-10-16
add you retriever module

sam-h-bean:
So I wasn't entirely sure on the difference between the modules in `dsp` vs the retriever module example of pinecone in `dspy`. I went ahead and followed what seemed to be the newer pattern. One thing to note about this implementation is if multiple queries are indeed passed into the new retriever we do not have the ability to rerank the results across queries so we return the passages from query 1 then the passages from query 2. I could also just raise a runtime error if that is a more desirable pattern. Let me know!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ checklist.md
 finetuning_ckpts/
 cache/
 * copy*
+.idea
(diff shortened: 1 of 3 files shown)


okhat:
Thanks a lot for opening this! Ready to merge pending one question

PR #173, opened 2023-10-18
added AnswerCorrectness and AnswerFaithfulness auto-eval programs

arnavsinghvi11:
AnswerCorrectness - verifies if predicted answer is correct for question and outputs True/False
AnswerFaithfulness - verifies if predicted answer is based off context/reasoning/rationale for question and outputs True/False


okhat:
```
        self.evaluate_correctness = dspy.ChainOfThought(self.Signature())
```

this line looks weird for two reasons: (1) why call this class Signature, that's going to be confused with dspy.Signature, (2) why invoke self.Signature (), why not just self.Signature.


```
class AnswerCorrectness(dspy.Module):
    class Signature(dspy.Signature):
        """Checks if predicted answer for question is correct"""
```

This signature is missing fundamental components in the fields and instructions: it should be comparing two answers. The gold answer and the system prediction.

The LM doesn't inherently know the right answer for evals.

PR #179, opened 2023-10-23
Allow passing of kwargs for HF TGI request

GuyAglionby:
This PR lets the user specify what arguments are provided to the HTTP call to the TGI server. My use case is HTTP basic auth, so I need to pass the relevant kwarg through.


okhat:
Thank you!

PR #185, opened 2023-10-24
OOM mitigations and Cache rationalization

ludovicofrizziero:
**Motivation**
I would like to adopt DSPy for a project, but it is not production-ready. The first problem I noticed is a possible Out Of Memory issue due to the in-memory or even on-disk cache system, plus the history collection on the LM base class.
Therefore, I updated the base LM class to change self.history to a fixed-length deque, and modified the code that utilized that.
I then changed the caching system to leave the door open to end-users to customize the cache how they like, and to avoid infinite-sized locally-memorized caches, that would kill a long-running production server otherwise.

**Problem Detection**
The issues were detected only by static analysis of the code, no runtime exceptions happened yet to confirm the analysis. 

**How to look at the diff**
First focus on the files `dsp/modules/cache_utils.py` and `dsp/modules/lm.py`, as they have the bulk of the corrections. All other files have been modified as a consequence of these two.
Pytests could be of some use also to understand usage.

**Tests**
I included some very basic unit tests, but I am a bit unable to tell how my modifications impact the overall DSPy stability.

**Remaining issue**
The folder `cache` is under source control, but I think it is a mistake and should be removed. Not done in this PR, as maybe it has a purpose, plus it would clutter the PR diff.

**Work Highlights**
- removed infinite size history from LLM, could cause OOM issues in production
- refactored LM base class, since it used knowledge leaked from derived classes to perform common functionalities
- reworked cache system: now it is more of a protocol than a ready-to-use cache
- default cache behavior is backward compatible as it still instantiates a valid cache
- added very basic pytests

### DIFF
--- a/dsp/modules/cache_utils.py
+++ b/dsp/modules/cache_utils.py
@@ -1,13 +1,11 @@
+import functools
 import os
-
-from pathlib import Path
-from joblib import Memory
 from functools import wraps
 
-from dsp.utils import dotdict
-
-
-cache_turn_on = True
+__all__ = (
+    'cache',
+    'set_cache_controller'
+)
 
 
 def noop_decorator(arg=None, *noop_args, **noop_kwargs):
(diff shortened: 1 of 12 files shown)


dsp/modules/cache_utils.py
```diff
@@ -24,20 +22,73 @@ def wrapper(*args, **kwargs):
         return decorator
 
 
-cachedir = os.environ.get('DSP_CACHEDIR') or os.path.join(Path.home(), 'cachedir_joblib')
-CacheMemory = Memory(location=cachedir, verbose=0)
+def set_cache_controller(obj: callable):
+    """
+    Caching protocol:
```
ludovicofrizziero, line 27: @okhat I added this heredoc, since , to me, this seems aligned with the spirit underlying DSPy, feel free to point out if it is unappropriate.

okhat:
Hey, thanks for this in-depth contribution! I agree this is needed. What do you think of https://github.com/stanfordnlp/dspy/pull/103

Ultimately, the core LM implementations need a revamp and caching needs that too.

That said, default caching is important too. I need to look into this more closely.

ludovicofrizziero:
Hi, thanks for the reply. In my opinion #103 tries to solve a different problem: **experiment reproducibility**, which involves saving LLM calls, but it is not actually caching per-se. A cache is much more lightweight and simple, since if the record is not there, it is ok to recompute it, it's just a cache miss. Experiment reproducibility instead has a much harder requirement on being able to always find past computations of "pure function" (otherwise side effects would negate reproducibility anyway).

If I was to guarantee exp.reproducibility, I would first notice that, as in any ML project, there are two distinct runtimes: one is inference, one is training. Most likely only training needs this kind of reproducibility. This means that I would look into `teleprompers `as candidate for using the sqlite implementation of #103, and pass explicitly the `SqliteCache` object by dependecy injection (rather than using a decorator) on those places where it is needed. On the contrary, I would avoid placing this kind of memoization on the LM itself, since it can be used both at inference and training without distinction. 

This is only an uninformed opinion based on what I can see from the code and how I usually approach problems, moreover I don't know the requirements that prompted the SQLiteCache implementation.

PR #186, opened 2023-10-25
added support for OpenAI completion API string prompting

arnavsinghvi11:
follows #3 within [Anyscale OpenAI endpoint handling](https://github.com/anyscale/endpoint-cookbook/blob/main/combine_anyscale_openai.ipynb)

allows for string prompting to models

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -113,26 +113,41 @@ class Anyscale(HFModel):
             **kwargs
         }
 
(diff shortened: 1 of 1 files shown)


okhat:
I'll merge this but we don't want [INST] and all that. We want this to be a plug-in replacement for our normal TGI llama.

okhat:
I'll play with these tweaks later.

pepo1275:

PR #187, opened 2023-10-25
added proper error handling within evaluate and bootstrap

arnavsinghvi11:
-permits up to 5 errors during 1) evaluation and 2) bootstrapping and then raises actual error, halting program for user to fix the error


okhat:
This is really good, I'm glad you did it in bootstrap too!

pepo1275:

PR #189, opened 2023-10-25
added AnswerCorrectness and AnswerFaithfulness auto-eval modules

arnavsinghvi11:
AnswerCorrectness - verifies if predicted answer is correct for question based on gold answer and outputs True/False
AnswerFaithfulness - verifies if predicted answer is based off context/reasoning/rationale for question and outputs True/False

(corrects faulty [PR](https://github.com/stanfordnlp/dspy/pull/173))

### DIFF
--- a/dspy/evaluate/auto_evaluation.py
+++ b/dspy/evaluate/auto_evaluation.py
@@ -1,7 +1,7 @@
 import dspy 
 
 class AnswerCorrectnessSignature(dspy.Signature):
-    """Determines if predicted answer matches the gold answer."""
+    """Verify that the predicted answer matches the gold answer."""
 
     question = dspy.InputField()
     gold_answer = dspy.InputField(desc="correct answer for question")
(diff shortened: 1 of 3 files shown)


dspy/evaluate/auto_evaluation.py
```diff
@@ -0,0 +1,34 @@
+import dspy 
+
+class AnswerCorrectnessSignature(dspy.Signature):
+    """Determines if predicted answer matches the gold answer."""
```
okhat, line 4: This is a strange instruction.

dspy/evaluate/auto_evaluation.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self):
+        super().__init__()
+        self.evaluate_correctness = dspy.ChainOfThought(AnswerCorrectnessSignature)
+    
+    def forward(self, question, gold_answer, predicted_answer):
+        return self.evaluate_correctness(question=question, gold_answer=gold_answer, predicted_answer=predicted_answer)
+
+
+class AnswerFaithfulnessSignature(dspy.Signature):
+    """Checks if answer for question is based on rationale."""
```
okhat, line 21: same.

okhat: please fix then merge directly!

pepo1275:

PR #191, opened 2023-10-26
Add llama-cpp python support 

DanielUH2019:
This changes allows running some large models on CPU locally. Let me know how I can improve it


dsp/modules/cohere.py
```diff
@@ -101,6 +101,9 @@ def request(self, prompt: str, **kwargs):
         """Handles retrieval of completions from Cohere whilst handling API errors"""
         return self.basic_request(prompt, **kwargs)
 
+    def get_choice_text(self, choice) -> str:
```
okhat, line 104: Is this needed?

dsp/modules/gpt3.py
```diff
@@ -117,10 +117,10 @@ def request(self, prompt: str, **kwargs) -> OpenAIObject:
         
         return self.basic_request(prompt, **kwargs)
 
-    def _get_choice_text(self, choice: dict[str, Any]) -> str:
+    def get_choice_text(self, choice: dict[str, Any]) -> str:
         if self.model_type == "chat":
-            return choice["message"]["content"]
-        return choice["text"]
+            return ' ' + choice["message"]["content"].strip()
```
okhat, line 122: Why?

dsp/modules/hf.py
```diff
@@ -99,6 +99,9 @@ def basic_request(self, prompt, **kwargs):
         self.history.append(history)
 
         return response
+    
+    def get_choice_text(self, choice: dict[str, Any]) -> str:
+        return choice["text"]
```
okhat, line 104: Is this needed?

dsp/modules/cohere.py
okhat: oh I see, I like your refactor, this is for inspect_history

dsp/modules/hf.py
okhat: got it

dsp/modules/gpt3.py
```diff
@@ -157,7 +157,7 @@ def __call__(
         if only_completed and len(completed_choices):
             choices = completed_choices
 
-        completions = [self._get_choice_text(c) for c in choices]
+        completions = [self.get_choice_text(c) for c in choices]
```
okhat, line 160: This kind of changes the behavior right? It used to be "stripped", now it has space before it.

I'm not sure if that will affect things in unexpected ways. IMO this whole refactor is great but shouldn't be part of this PR, it can be another PR though.

dsp/modules/llama.py
```diff
@@ -0,0 +1,96 @@
...
+    def __init__(
+        self,
+        model: str,
+        llama_model: Llama,
+        model_type: Literal["chat", "text"] = None,
+        **kwargs,
+    ):
+        super().__init__(model)
+
+        default_model_type = "chat" if "instruct" not in model else "text"
```
okhat, line 25: does this make sense here? this is for OpenAI models

dsp/modules/gpt3.py
DanielUH2019: you are right, now I see that it is not equivalent, I will revert the refactor, and leave only the llama_cpp implementation.

dsp/modules/llama.py
DanielUH2019: i will remove that to

dsp/modules/gpt3.py
```diff
@@ -173,7 +173,7 @@ def __call__(
                     tokens, logprobs = tokens[:index], logprobs[:index]
 
                 avglog = sum(logprobs) / len(logprobs)
-                scored_completions.append((avglog, self._get_choice_text(c)))
+                scored_completions.append((avglog, self.get_choice_text(c)))
```
okhat, line 176: Hey I'd like to merge this PR, but I think this line is incorrect?

DanielUH2019: Yes, I didn't see that when reverting the refactoring; it's fixed now

dspy/__init__.py
```diff
@@ -24,3 +24,5 @@
 
 Anyscale = dsp.Anyscale
 HFModel = dsp.HFModel
+
+LlamaCpp = dsp.LlamaCpp
```
okhat, line 28: This seems incorrect, since LlamaCpp is not imported by default?

dsp/modules/__init__.py
```diff
@@ -8,4 +8,5 @@
 from .pyserini import *
 
 from .hf_client import HFClientTGI
-from .hf_client import Anyscale
\ No newline at end of file
+from .llama import LlamaCpp
```
okhat, line 11: Hmm importing this by default means that people need a new dependency right? It should be optional

dsp/modules/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .cohere import *
 from .sbert import *
 from .pyserini import *
+from .llama import *
```
okhat, line 9: Should we remove this line?

okhat: Alternatively, we can move `from llama_cpp import Llama` into _inside_ `class LlamaCpp(LM)`.

What do you prefer?

DanielUH2019: What would be the benefit of this? In the current implementation I think it follows the same structure as cohere

arnavsinghvi11:
@DanielUH2019 just following up on old PRs :), Could you resolve relevant conflicts and lmk when ready to merge?

codysnider:
@arnavsinghvi11 there's a new PR up as a continuation of this effort. it includes the snippet used to test against openhermes with llamacpp: https://github.com/stanfordnlp/dspy/pull/862

arnavsinghvi11:
Closing this PR with #862 close to merging

PR #125, opened 2023-09-17
linter: black & isort for auto-formatting on PRs

lawliet19189:

PR #195, opened 2023-10-29
Apply ruff check, ruff formatter and use a pre-commit-config for ci

DanielUH2019:
I think this could be beneficial for the project, let me know what you think. I noticed that #125  is trying to include black and isort, but it seems inactive. I didn't want to get involved in changing the imports with wildcards in the __init__.py files due to the potential implications.


isaacbmiller:
I think safe to close this. Out of date and somewhat implemented by #538

PR #197, opened 2023-10-29
dspy assertions

manishshettym:
Adds assertions as first-class primitives in dspy.

Syntax: `dspy.Assert(fn(args), msg="feedback message")`

Semantics: On using `dspy.Assert` in a program, the compiler promises that, during inference, either _the compiler figures out how to simulate the DSPy program up to that step in a way that passes the assertion_ or it raises a DSPyAssertionError exception. For the latter, the compiler will use the assertion as a hint, trying on a best-effort basis to make it pass. If it fails, a warning is logged, but execution will continue.

TODOs:

- [x] simplify syntax from `dspy.Assert(fn, *args, msg="...")` ==> dspy.Assert(fn(args), msg="...")
- [ ] simplify syntax to make msg a non keyword argument

### DIFF
--- a/dspy/primitives/assertions.py
+++ b/dspy/primitives/assertions.py
@@ -1,0 +1,148 @@
+from typing import Any, Callable
+import dsp
+import dspy
+
+
+class DSPyAssertionError(AssertionError):
+    """Custom exception raised when a DSPy `Assert` fails.
+    """
+
+    def __init__(self, msg: str, state: Any = None) -> None:
+        super().__init__(msg)
+        self.msg = msg
+        self.state = state
+
+
+class Assert:
+    """DSPy Assertion"""
+
+    def __init__(self, assert_fun_res: bool, **kwargs):
+        self.assert_fun_res = assert_fun_res
+
+        if "msg" in kwargs:
+            self.msg = kwargs["msg"]
+            del kwargs["msg"]
+        else:
+            self.msg = ""
+
+        self.kwargs = kwargs
+        self.__call__()
+
+    def __call__(self) -> bool:
+        if isinstance(self.assert_fun_res, bool):
+            if self.assert_fun_res:
+                return True
+            else:
+                raise DSPyAssertionError(msg=self.msg, state=dsp.settings.trace)
+        else:
+            raise ValueError("Assertion function should always return [bool]")
+
+
+def assert_backtrack_policy(max_backtracks=2):
+    """Decorator that defines the backtracking policy for assertions.
+    
+    current policy: re-run the latest predictor up to `max_backtracks` times,
+    with updated signature if assertion fails. updated signature adds a new 
+    input field to the signature, which is the feedback.
+    """
+
+    def wrapper(func):
+        def inner(*args, **kwargs):
+            backtrack_to = None
+            error_msg = None
+            result = None
+            extended_predictors_to_original_forward = {}
+
+            def _extend_predictor_signature(predictor, **kwargs):
+                """Update the signature of a predictor instance with specified fields."""
+                old_signature = predictor.extended_signature
+                new_signature_kwargs = {
+                    k: v
+                    for (k, v) in old_signature.kwargs.items()
+                    if isinstance(v, dspy.InputField)
+                }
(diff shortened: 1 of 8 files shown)


dspy/primitives/assertions.py
darinkishore, line 1: Have to modify forward method to support

dspy/primitives/assertions.py
```diff
@@ -121,8 +114,12 @@ def new_forward(**kwargs):
                         backtrack_to, default_args={"feedback": feedback_msg}
                     )
 
+                if i == max_backtracks:  # Check if it's the last backtrack attempt
+                    bypass_assert = True
+                    result = func(*args, **kwargs, bypass_assert=bypass_assert)
+                    break
                 try:
-                    result = func(*args, **kwargs)
+                    result = func(*args, **kwargs, bypass_assert=bypass_assert)
```
Shangyint, line 122: Can we make `bypass_assert` part of the global execution config? Otherwise, the user needs to explicitly include this keyword into their `forward`, and check in their code explicitly.

okhat: I fully agree

okhat: But even better if both ways work

manishshettym: update: using `dspy.settings.bypass_assert` to manage this. 
User can also optionally set a default argument in forward: `def forward(..., bypass_assert=True)`
- this still checks assertions but does not raise errors (can log if needed)
- since no errors raised => disables backtracking

dspy/primitives/program.py
```diff
@@ -26,6 +27,12 @@ def __call__(self, *args, **kwargs):
         if getattr(self, "forward", False) and not getattr(
             self.forward, "_decorated", False
         ):
+            args_to_vals = inspect.getcallargs(self.forward, *args, **kwargs)
```
Shangyint: For brevity, we can move this part to teleprompter too in a future PR.

dspy/predict/retry.py
```diff
@@ -0,0 +1,58 @@
...
+from .predict import Predict
+
+
+class Retry(Predict):
+    def __init__(self, module):
+        super().__init__(module.signature)
+        self.module = module
+        self.original_signature = module.signature.signature
+        self.original_forward = module.forward
+        self.new_signature = self._create_new_signature(self.original_signature)
```
manishshettym, line 13: question: should this be
```suggestion
        self.new_signature = dsp.Template(old_signature.instructions, self._create_new_signature(self.original_signature))
```

self._create_new_signature returns a `dict` that consists of the new extended kwargs? Do we need to transform that to a template for the module? cc: @arnavsinghvi11

more generally, how is a `dsp.Template` different from a `dspy.Signature`?

arnavsinghvi11: Currently, the ChainOfThought logic handles the transformation of the signature kwargs to the template. It would be cleaner to ensure this logic is universal for any modules within Retry and can be done using the original signature's docstring which stores the instructions.  

`dsp.Template` essentially formats and structures the fields of a `dspy.Signature` to handle all of its metadata accordingly (Signature: instruction, Fields: prefix, description, separator, etc.)

dsp/utils/settings.py
```diff
@@ -33,6 +36,8 @@ def __new__(cls):
                 skip_logprobs=False,
                 trace=None,
                 release=0,
+                bypass_assert=False,
+                bypass_suggest=False,
```
okhat, line 40: Why is the default False?

dspy/predict/retry.py
okhat: This is resolved?

dspy/teleprompt/bootstrap.py
```diff
@@ -2,8 +2,10 @@
 import tqdm
 import random
 import threading
+from dspy.predict.retry import Retry
 
 from dspy.primitives import Example
+from dspy.primitives.assertions import assert_transform_module, noop_handler
```
okhat, line 8: This import is not needed?

PR #211, opened 2023-11-06
included usage logging for openai calls

arnavsinghvi11:
-logging of total_tokens usage from OpenAI requests
-thread-safe with Python logging library

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -1,3 +1,19 @@
+import logging
+from logging.handlers import RotatingFileHandler
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(message)s',
+    handlers=[
+        RotatingFileHandler(
+            'openai_usage.log',  # Filename
+            maxBytes=10000000,  # Max size of a log file before being rotated (10MB)
+            backupCount=10  # Number of log files to keep
+        )
+    ]
+)
+
 import functools
 import json
 from typing import Any, Literal, Optional, cast
@@ -78,6 +94,13 @@ class GPT3(LM):
     def _openai_client():
         return openai
 
+    def log_usage(self, response):
+            """Log the total tokens from the OpenAI API response."""
+            usage_data = response.get('usage')
+            if usage_data:
+                total_tokens = usage_data.get('total_tokens')  
+                logging.info(f'{total_tokens}')
+
     def basic_request(self, prompt: str, **kwargs) -> OpenAIObject:
         raw_kwargs = kwargs
 
@@ -150,6 +173,7 @@ class GPT3(LM):
         #         kwargs = {**kwargs, "logprobs": 5}
 
         response = self.request(prompt, **kwargs)
+        self.log_usage(response)
         choices = response["choices"]
 
         completed_choices = [c for c in choices if c["finish_reason"] != "length"]


okhat:
Looks great. Can we just activate logging only when dsp.settings.log_openai_usage is True?

okhat:
Now it will crash I think? We need the new dsp.settings flag to be set to False by default

okhat:
otherwise it will complain that it's unset

shaoyijia:
Thanks for the feature!

Adding `log_openai_usge=True` [here](https://github.com/stanfordnlp/dspy/blob/87e58d943c93766f08ccfe3b4c35d689c32bd7d8/dsp/utils/settings.py#L35) works for me.

I would suggest adding some explanation [here](https://github.com/stanfordnlp/dspy/pull/211/files#diff-f83a6b9860ba9226694bff0362b70ce0f09e4583e6a1a4484e8b5a162b41fb8bR99), e.g. `logging.info(f'#Total token in a LLM call: {total_tokens}')` rather than just logging the number - otherwise, it's hard to understand the number when you have other log info. Or, a more useful feature will be accumulating the token number in maybe the Settings class (I found it's a Singleton class); user can inspect the number at any time they want and do the calculation themselves.

shaoyijia:
This is what I change in my dsp/utils/settings.py:

```
class Settings(object):
    """DSP configuration settings."""

    _instance = None
    _token_usage_lock = threading.Lock()

    def __new__(cls):
        """
        Singleton Pattern. See https://python-patterns.guide/gang-of-four/singleton/
        """

        if cls._instance is None:
            cls._instance = super().__new__(cls)
            cls._instance.main_tid = threading.get_ident()
            cls._instance.main_stack = []
            cls._instance.stack_by_thread = {}
            cls._instance.stack_by_thread[threading.get_ident()] = cls._instance.main_stack

            #  TODO: remove first-class support for re-ranker and potentially combine with RM to form a pipeline of sorts
            #  eg: RetrieveThenRerankPipeline(RetrievalModel, Reranker)
            #  downstream operations like dsp.retrieve would use configs from the defined pipeline.
            config = dotdict(
                lm=None,
                rm=None,
                branch_idx=0,
                reranker=None,
                compiled_lm=None,
                force_reuse_cached_compilation=False,
                compiling=False,
                skip_logprobs=False,
                trace=None,
                release=0,
                log_openai_usage=False,
                token_usage={},
            )
            cls._instance.__append(config)

        return cls._instance

    """Omit unchanged functions"""

    def increment_token_usage(self, model, prompt_tokens, completion_tokens):
        with self._token_usage_lock:
            if model in self.token_usage:
                self.token_usage[model]['prompt_tokens'] += prompt_tokens
                self.token_usage[model]['completion_tokens'] += completion_tokens
            else:
                self.token_usage[model] = {'prompt_tokens': prompt_tokens, 'completion_tokens': completion_tokens}


```

arnavsinghvi11:
> Now it will crash I think? We need the new dsp.settings flag to be set to False by default

ah yes I was going to keep this but looking at [this example ](https://github.com/stanfordnlp/dspy/blob/87e58d943c93766f08ccfe3b4c35d689c32bd7d8/dsp/primitives/primitives.py#L37) on how users can set flags without including this in the settings and also since the settings __getattr__ method would check for variables in the config dictionary, I thought dsp.settings would handle this itself. But adding the flag is definitely the simpler solution :)

PR #227, opened 2023-11-21
add minimum working support for vLLM

meditans:
We had already a PR for vLLM support (https://github.com/stanfordnlp/dspy/pull/93), from which this small patch was adapted to work for the current vLLM version. Feel free to merge this or use it just as a reference to improve the other. Thanks!


okhat:
Thanks so much!

okhat:
Merged.

PR #233, opened 2023-11-24
Change the name of the GPT3 class to just GPT?

jasondotparse:
Self explanatory. The classname GPT3 resulted in confusing errors like this. A DSPy newbie like me might wonder why gpt-3 was being invoked when I was attempting to use GPT-4:

error_code_references_wrong_llm

Open to feedback on other ways to improve this!

### DIFF
--- a/dsp/modules/__init__.py
+++ b/dsp/modules/__init__.py
@@ -1,5 +1,5 @@
 from .cache_utils import *
-from .gpt3 import *
+from .gpt import *
 from .hf import HFModel
 from .colbertv2 import ColBERTv2
 from .sentence_vectorizer import *
--- a/dsp/modules/gpt.py
+++ b/dsp/modules/gpt.py
@@ -20,7 +20,7 @@ def backoff_hdlr(details):
     )
 
 
-class GPT3(LM):
+class GPT(LM):
     """Wrapper around OpenAI's GPT API. Supports both the OpenAI and Azure APIs.
 
     Args:
--- a/dsp/primitives/compiler.py
+++ b/dsp/primitives/compiler.py
@@ -155,7 +155,7 @@ def finetune(training_data, target):
     jobname, ft = openai_finetune(name, target)
     print(ft)
 
-    ft = dsp.GPT3(model=ft, stop=" </s>")
+    ft = dsp.GPT(model=ft, stop=" </s>")
     return ft
 
 # 4. Return updated program.
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -15,7 +15,7 @@ import dsp
 
 settings = dsp.settings
 
-OpenAI = dsp.GPT3
+OpenAI = dsp.GPT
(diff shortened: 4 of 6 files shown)


sutyum (commented):
This PR only introduces a renaming from GPT3 to GPT. It does indeed reduce confusion moving forward. 

We should merge it soon

okhat:
@sutyum I'm concerned about the effect on existing caches, will they continue to work? anyway a more serious refactor is in the works

okhat:
we should probably call it OpenAI eventually, instead of GPT? I'm upgrading to openai 1.0 now

okhat:
yeah I can confirm that changing the path of the file will break the joblib caches that exist now, it's probably fixable, not 100% sure

so basically I can't merge this right away

sutyum:
> we should probably call it OpenAI eventually, instead of GPT? I'm upgrading to openai 1.0 now

I had this thought too. Openai has all these multimodal models - TTS, STT, Dalle, etc. So I wonder how this would fit moving forward. 

```python
dsp.GPT3( ... ) # now

dsp.GPT( ... ) # proposed
dsp.Dalle( ... ) # would then need other model-classes moving forward

# Or, perhaps,
dsp.openai.GPT( ... ) # alternative
dsp.openai.Dalle( ... )
```

sutyum:
> we should probably call it OpenAI eventually, instead of GPT? I'm upgrading to openai 1.0 now

I tried upgrading too. The caching seems pretty extensive part of upgrading though - could you mention how and why dspy relies on this?

jhyearsley:
I assume the interface would stay the same to support multi-modal models, but not sure if that's on the roadmap (and speaking of roadmaps, is there a roadmap?), i.e.
- `dsp.OpenAI(model=<the_model>, model_type=<the_model_type>` with
  - `the_model in ['gpt-3.5', 'gpt-4', 'dall-e-3', 'tts-1', ...]` and
  - `the_model_type in ['chat', 'text', 'image', 'audio', ... ]` 

And there would be a validation step to make sure the `model` and `model_type` are correct (e.g `model='dall-e-3'` and `model_type='chat'` would be invalid). Once you know the `model` and `model_type` are correct you should be able to derive the endpoint from the `model_type`. Probably you would add more conditionals like in [basic_request](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/gpt3.py#L102) or there would be more refactoring to clean up before adding more. Just my 2 cents 😄

piotrlaczkowski:
Since you are explicitly checking the impact on caching mechanism and this means you will need to run some manual tests -> could you please build it into a dedicated pytest module, so we do not need to test manually next time ? 🙏🏻 # unit-testing
I can help set it up if needed -> just let me know.

okhat:
@piotrlaczkowski The main testing approach right now is rerunning the intro notebook in Google Colab. It's far from perfect but if it runs from cache that's a pretty solid end to end test. Yes we need to do better on unit testing in principle though...

PR #239, opened 2023-12-02
Removed conversion from Example to dict in BootstrapFewShotWithOptuna that breaks template line

bendavidsteel:
BootstrapFewShotWithOptuna currently converts demostration examples to dicts, which breaks line 210 in template_v2.py as `augmented` cannot be accessed as an attribute. I tried converting to dotdict at the line instead, but dotdict demos seem to converted to just dicts later on, which therefore would require a more complicated fix.

### DIFF
--- a/dspy/teleprompt/teleprompt_optuna.py
+++ b/dspy/teleprompt/teleprompt_optuna.py
@@ -36,7 +36,7 @@ class BootstrapFewShotWithOptuna(Teleprompter):
         for (name, compiled_predictor), (_, program2_predictor) in zip(self.compiled_teleprompter.named_predictors(), program2.named_predictors()):
             all_demos = compiled_predictor.demos
             demo_index = trial.suggest_int(f"demo_index_for_{name}", 0, len(all_demos) - 1)
-            selected_demo = dict(all_demos[demo_index])
+            selected_demo = all_demos[demo_index]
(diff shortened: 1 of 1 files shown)


dspy/teleprompt/teleprompt_optuna.py
```diff
@@ -36,7 +36,7 @@ def objective(self, trial):
         for (name, compiled_predictor), (_, program2_predictor) in zip(self.compiled_teleprompter.named_predictors(), program2.named_predictors()):
             all_demos = compiled_predictor.demos
             demo_index = trial.suggest_int(f"demo_index_for_{name}", 0, len(all_demos) - 1)
-            selected_demo = dict(all_demos[demo_index])
+            selected_demo = all_demos[demo_index]
```
piotrlaczkowski, line 39: If this causes a real problem I would rather go with an additional line below of type:

```python
if not isinstances(selected_demo, dict):
   selected_demo = dict(selected_demo)
```
what do you think ?

okhat:
We recently merged BayesianSignatureOptimizer, and we will deprecate/remove BootstrapFewShotWithOptuna.

So I'm closing this issue but please feel free to re-open if it's important for something you need. I encoruage you to try out BayesianSignatureOptimizer.

PR #241, opened 2023-12-04
README tweak, update order of links out to references

heathermiller:
This is a trivial change– I just reversed the chronological order of the list of links to references. Previously, it was: oldest -> newest, and I just swapped it: newest -> oldest.

Reason: people often click the first link in a list to read more something. So it probably makes sense to have the thing we'd actually want folks to read listed first (the most recent preprint).


okhat:
Thank you Heather! Merged!

PR #246, opened 2023-12-09
added notebook for longformqa/assertions example

arnavsinghvi11:
example for LongFormQA, including usage of new computational constraints code


dspy/datasets/hotpotqa.py
```diff
@@ -41,9 +41,9 @@ def __init__(self, *args, only_hard_examples=True, keep_details='dev_titles', un
         else:
             self._dev = None
 
-        for example in self._train:
-            if keep_details == 'dev_titles':
-                del example['gold_titles']
+        # for example in self._train:
```
okhat, line 44: hmm can this be cleaner? like follow/apply the kwargs?

okhat: in particular, old code shouldn't change behavior

okhat:
Ah this has a conflict I think?

PR #247, opened 2023-12-10
Pinecone fixes

KE7:


KE7:
Changes based on testing where DSPy was used for a different project: https://github.com/KE7/scenicNL/blob/main/notebooks/dspy.ipynb 

@okhat please send me an email (k.e@berk) if you want access to that repo. It's for a project I'm working on with Matei (he already has access)

dspy/retrieve/pinecone_rm.py
```diff
@@ -3,6 +3,9 @@
 Author: Dhar Rawal (@drawal1)
 """
 
+import torch
```
okhat, line 6: We can't import this here, torch isn't needed for other stuff. This import should be deferred, it should be inside a function call for instance.

dspy/retrieve/pinecone_rm.py
```diff
@@ -3,6 +3,9 @@
...
+from dsp.utils import dotdict
+from transformers import AutoTokenizer, AutoModel
```
okhat, line 8: Same here.

okhat:
Thank you! Left two small comments.

okhat:
@KE7 Following up after NeurIPS in case you have a chance to tweak these two issues above!

KE7:
@okhat Should be all good now :)

PR #259, opened 2023-12-26
Openai v1

okhat:


okhat:
cc @DanielUH2019 gotta check chroamdb and pinecone still work for 0.28, the probably won't? at least pinecone needs small changes

PR #260, opened 2023-12-27
Bootstrapping ScoNe CoT demos with a separate LLM from the main one

cgpotts:
Hi @okhat!

This is a simple example showing how to use a different (and presumably more powerful) model for bootstrapping examples than the one being used for the central predictions.

The example task is the ScoNe "one scoping negation" category, which is one of the hardest categories in ScoNe. Zero-shot, turbo is at chance, and using turbo to bootstrap full CoT examples seemed not to work well, but using GPT-4 for bootstrapping (a possibility you made me aware of!) took performance all the way north of 85% accuracy (and one of my runs was at 93%). This extremely high, but everything seems to be set up correctly, so this seems like a nice illustration of the power of this strategy.

Do let me know if I should make an adjustments to the way the notebook is set up. The hope is that this is code people can copy-paste for their own work (almost none of the DSPy code is even specific to ScoNe).

---Chris

### DIFF
--- a/examples/nli/scone/scone-cot_fewshot-turbo-gpt4-demos.json
+++ b/examples/nli/scone/scone-cot_fewshot-turbo-gpt4-demos.json
@@ -1,0 +1,64 @@
+{
+  "generate_answer": {
+    "lm": null,
+    "traces": [],
+    "train": [],
+    "demos": [
+      {
+        "augmented": true,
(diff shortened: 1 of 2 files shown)


okhat:
Thank you Chris! Everything looks great to me, particularly after switching the flag to False just now.

Merging!

cgpotts:
Yes, I very predictably forgot to toggle that flag!

PR #261, opened 2023-12-28
openai 0.28 backwards compatibility for pinecone and chromadb

DanielUH2019:
Hi @okhat, sorry for the late response, I think this changes should fix the compatibility with 0.28


okhat:
Thank you Daniel!

PR #262, opened 2024-01-02
Adding local model integration with Ollama

b-akshay:
This adds integration for fully local models. Integration is currently through [Ollama](https://github.com/jmorganca/ollama/) which serves the models on localhost. Follows gpt3.py.

Notes: I had issues using httpx with non-stream output so using requests instead. Tested with intro notebook and with some of the advanced notebooks.

### DIFF
--- a/dsp/modules/ollama.py
+++ b/dsp/modules/ollama.py
@@ -1,0 +1,168 @@
+from dsp.modules.lm import LM
+from typing import Any, Literal, Optional
+
+import os, multiprocessing, datetime, hashlib
+import requests
+
+import json
+
+def post_request_metadata(model_name, prompt):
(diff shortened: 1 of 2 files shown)


okhat:
Looks great, thank you!

dsp/modules/ollama.py
```diff
@@ -0,0 +1,168 @@
...
+    def __init__(
+        self,
+        model: str = "llama2", 
+        model_type: Literal["chat", "text"] = None,
+        **kwargs,
+    ):
+        super().__init__(model)
+        self.provider = "ollama"
+
+        self.base_url = "http://localhost:11434"  # where the model is hosted
```
okhat, line 42: This is hardcoded but I merged for now as it's probably going to be easy to refactor a bit

Issue #258, opened 2023-12-26
Upgrade OpenAI Python library


jhyearsley:
DSPy is using `openai<=0.28.1` and current version of OpenAI python library is `1.6.1`. There are breaking changes and with the rapid pace of development I think we should try and catch up to the latest version sooner rather than later. See the [v1.0.0 Migration Guide](https://github.com/openai/openai-python/discussions/742)

okhat:
Hey @jhyearsley , OK I agree. But this will be a new class `dspy.OpenAI_v1` with something like `pip install dspy-ai[openai_v1]`, is that fine?

We need to continue supporting the old version.

jhyearsley:
Yeah makes sense. Curious to learn more about what is depending on the older version though. Are there dependencies that would break even with a staged refactor that preserves backwards compatibility?

okhat:
Hey @jhyearsley merged to main.

Requirements still point to 0.28 but if you just do `pip install openai --upgrade`, it'll work with 1.0 or more recent just fine.

The old caches and the new caches are disjoint. e.g., rerunning the intro notebook with 1.0 will require an API key

okhat:
Thanks to @DanielUH2019 for most of this change

jhyearsley:
Awesome!

okhat:
This is on `main` now but not yet on pypi. Won't push to pypi until I get 1-2 confirmations it works fine for people.

jhyearsley:
Ok, good point, maybe better to keep open then until it is pushed to pypi

dkshjn:
I encountered an issue while using the latest update of the `dspy` package. Below are the details of my environment and the problem:
```
OS: macOS 14.2
Python Version: 3.10.12
OpenAI Version: 1.6.1
```
dspy Version: Installed using git.
```Python
!pip install git+https://github.com/stanfordnlp/dspy.git
```

Using the BasicQA class I asked a simple question but was facing an error.
```Python
class BasicQA(dspy.Signature):
    """Answer questions with short factoid answers."""

    question = dspy.InputField()
    answer = dspy.OutputField(desc="often between 1 and 5 words")

generate_answer = dspy.Predict(BasicQA)
example = "Hello, how are you?"
pred = generate_answer(question=example)
print(f"Question: {example}")
print(f"Predicted Answer: {pred.answer}")
```
**Error Message**
```
---------------------------------------------------------------------------
NotFoundError                             Traceback (most recent call last)
[...]

PR #265, opened 2024-01-08
Fix a nit for openai azure support.

shaoyijia:
Issues:

After the change to support `openai==1.x` (#258), https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/gpt3.py#L70 will always trigger assert when `api_provider == 'azure'`.

Changes:
- Fix this nit.
- Move `openai.api_type = api_provider` to better support switching api providers.

Remaining issues:
- `openai==1.x` suggests using client to manage configurations. It seems that the global configurations do not work for OpenAI Azure service any more. #258 is not fully solved.

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -53,6 +53,7 @@ class GPT3(LM):
     ):
         super().__init__(model)
         self.provider = "openai"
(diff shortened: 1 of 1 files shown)


okhat:
Thanks a lot Yijia!

PR #269, opened 2024-01-12
Update evaluate.py

bhavna3:
This modification replaces the IPython import with a try-except block. If IPython is not available, it sets ipython_display to the built-in print function and defines a dummy HTML lambda function.


okhat:
This is very clever. Thank you! Merging.

PR #274, opened 2024-01-15
Assertion-aware metrics and optimization

manishshettym:
**Tasks**
- [x] option to enable assert-aware-metric for each assert/suggest (i.e., optimizable): 
`dspy.Suggest(constraint, msg, is_metric)`
- [x] track suggest/assert failures during evaluation
- [x] propagate suggest/assert failures as an attribute of a bootstrapped program
- [x] penalize optimization score according to the number of failures
- [x] incorporate feedback traces into compilation


okhat:
Thanks so much. Should we merge this? Does this have the latest from @arnavsinghvi11 ?

PR #279, opened 2024-01-18
docs(dspy): setting up `mkdocs` based (versioned) documentation

piotrlaczkowski:
- adding required dependencies (needed for CI/CD auto-build that we can setup later)
- setting up initial example structure and some basic guidelines


okhat:
Looks amazing! Thank you SO much @piotrlaczkowski !

piotrlaczkowski:
It is just a beginning so we can start migrating the existing repo into this format, but the most impactful part of this approach is:

- code doc is generated from the doc strings (much easier to maintain and version control)

- docs deployed to pages will be versioned controlled

- I'll further add automatic CICD preview if doc changes and auto deploy

PR #282, opened 2024-01-19
ops(dspy): activating pre-commit global check (and aligning to py3.9)

piotrlaczkowski:
- aligning pre-commit version to py3.9 (do we need this version or can we start with >=3.10, which is x10 faster) and has lot's of nice features for typing etc
- activating pre-commit checks in the repo (you shall not pass principle)

### DIFF
--- a/.github/workflows/precommits_check.yml
+++ b/.github/workflows/precommits_check.yml
@@ -15,7 +15,6 @@ on:
 
 jobs:
   pre-commit-checks:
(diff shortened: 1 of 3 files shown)


piotrlaczkowski:
@okhat -> I can see that I can merge this MR but I still have a pending review, what is the procedure ?
Should we take some time together to setup (and document) the process end to end (add dev branch + branch protection rules etc ... )?

PR #285, opened 2024-01-19
ops(dspy): adding new dependency for easy logs handling (with advanced options)

piotrlaczkowski:
Adding the `loguru` package so we can easily incorporate logging culture into this package with this very useful library.


arnavsinghvi11:
Hi @piotrlaczkowski , we've moved to using structlog. We can revisit loguru if needed down the line. Thanks!

PR #289, opened 2024-01-21
Add WeaviateRM

CShorten:
## Adds WeaviateRM as a child class to `dspy.Retrieve`

Embedding Inference
- Note this assumes you are using a text2vec embedding service orchestrated in Weaviate

Hard-Coded Data Model
- The RM currently assumes 1 Collection with 1 Text Key and no metadata in the Vector DB. We believe `content` is the most common named text key for this. This is quite an interesting topic re data models for Vector DBs, but I think this model is a great point to start and keep building from here.

Batch Query API
- This integration is based on Weaviate's v3 python client, which does not yet support a batch query API. We are going to be releasing 1.24 very soon, and I will be updating this.

Aggregating Duplicate Passages
- This implementation assumes you will call `dspy.utils.deduplicate` if the module above will be sending multiple queries in one call to `dspy.Retrieve`. Maybe this is too naive, I need to learn a little more about the access patterns for this, I think in the multi-hop case, deduplicate should work well. It might make sense for us (Weaviate) to cook multiple query result aggregation in our batch query API. In addition to the data model broadly, I think this is another key opportunity for future work.

Small Note
- Interfaces with `dspy.Prediction` via a list of strings `passages`


okhat:
Thanks so much Connor! This looks good except for;
> Install it with `pip install dspy-ai[weaviate]`"
  
  
 I think this needs a change to setup.py too?

CShorten:
Oops, thanks for catching that! Fixed here, using 3.26.1

PR #292, opened 2024-01-22
Support convenience import for `OllamaLocal`

forwardprogress:
### Summary

**tldr;** These changes add a convenience import for the `OllamaLocal` import introduced in https://github.com/stanfordnlp/dspy/pull/262 consistent with what's done for GPT-3.5, et al. I also added a minimal test to show the before/after comparison.

I began working through [`intro.ipynb`](https://github.com/stanfordnlp/dspy/blob/main/intro.ipynb) today and wanted to use a model I was hosting locally with Ollama instead of GPT-3.5. Though the `OpenAI` symbol is actually a reference to `dsp.modules.GPT3`, there's some hoisting to make it a little more convenient to access:
```python
turbo = dspy.OpenAI(model='gpt-3.5-turbo')
```

I expected to do the same for Ollama:

```python
nous_hermes2_mixtral = dsp.OllamaLocal(model="nous-hermes2-mixtral", model_type="chat")
```

but found it wasn't supported. I've made the appropriate change to `dsp/modules/__init__.py` to make the Ollama module consistent with its siblings.

I wanted to add a test demonstrating the change, but couldn't find any unit tests. I've added a new `tests/` package at the root of the repo to host all test suites, which I understand to be common convention. I've avoided adding any dependencies by relying on the built-in unit testing library in Python.

⚠️ These changes also propose running the new unit test as part of GitHub CI checks. I'm unfamiliar with GitHub's CI system, so if the CI change is acceptable, it deserves some scrutiny to make sure I didn't make a silly mistake.

### Testing

I added a unit test that verifies the convenience import is now supported. Unit tests can be executed from the command line with this command:

```sh-session
% python -m unittest discover tests/
```

**Before**:

```sh-session
E
======================================================================
ERROR: test_convenience_import_of_dsp_ollama_client (test_dsp.DspModuleImportsTest.test_convenience_import_of_dsp_ollama_client)
----------------------------------------------------------------------
  File "/Users/forwardprogress/Developer/stanfordnlp/dspy/tests/test_dsp.py", line 6, in test_convenience_import_of_dsp_ollama_client
    from dsp import OllamaLocal
ImportError: cannot import name 'OllamaLocal' from 'dsp' (/Users/forwardprogress/Developer/stanfordnlp/dspy/dsp/__init__.py)

----------------------------------------------------------------------
Ran 1 test in 0.479s

FAILED (errors=1)
```

**After**:

```sh-session
.
----------------------------------------------------------------------
Ran 1 test in 0.484s

OK
```


.github/workflows/build_package.yml
```diff
@@ -41,4 +41,5 @@ jobs:
         shell: bash
         working-directory: ${{ vars.RUNNER_TEMP }}
         run: python -c "import dsp"
-        
+      - name: Run tests
```
piotrlaczkowski, line 44: Normally, this part will be added as a separated workflow in: https://github.com/stanfordnlp/dspy/pull/283

piotrlaczkowski: Tests are failing since you would need to install all required deps etc...

forwardprogress: Ah, I didn't think to poke around to see if another PR was already looking to bring up tests. I'll pull all the testing related changes out and wait for 283 to land to re-add the test.

forwardprogress:
> Hey thanks so much, but this may create a new dependency for everyone who doesn't need it?

Ah, that's a great point that I didn't fully consider. I think the Ollama module requires no additional dependencies but the HF change does.

If I updated the changes to only hoist `OllamaLocal`, would that be agreeable?

To be clear, I'm operating with the understanding that it's desirable for most (or all?) modules to have these convenience imports. If that's not the case, I'm happy to get that feedback.

forwardprogress:
Whelp, this PR got whittled down to a lot less than I expected. 😉

forwardprogress:
Alright @okhat @piotrlaczkowski I don't think I'm getting traction with this change so I'll plan to close it out sometime tomorrow if I don't hear back.

psykhi:
No reason to not merge this, it's a bugfix IMO!

okhat:
I’m confused. This creates a dependency on other tools that everyone now will need to install?

psykhi:
From `ollama.py`

```python
from dsp.modules.lm import LM
from typing import Any, Literal, Optional

import os, multiprocessing, datetime, hashlib
import requests

import json
```

Aren't those already there/ in the standard lib? Sorry if I'm missing the obvious 😅

PR #296, opened 2024-01-25
Updating Cohere model names and support

jalammar:
This PR points to Cohere's existing generation models (previously defined models are out of date). 

It also tweak the behavior in `predict.py` that required the llm provider to have a parameter called `n` (that OpenAI has) which is called `num_generations` on the Cohere side. The previous behavior prevented the llm from running (it would throw an error due to the additional unknown parameter `n`).


okhat:
Thank you Jay! Merged. Sorry for the delay, I had been traveling. Looks great, and it will be part of the next pypi release this week.

jalammar:
That's perfect! Thanks @okhat!

PR #297, opened 2024-01-26
Irera

KarelDO:
Comments on the individual commits.


dspy/__init__.py
```diff
@@ -26,4 +26,55 @@
...
-context = settings.context
\ No newline at end of file
+context = settings.context
+
+
+import json
+import os
+import openai
+
+class Models:
```
okhat, line 36: This is quite creative and convenient but it probably won't work in the grand scheme of things, at least not without a lot of effort to expand then maintain.

Basically there are at least 10 providers people use, not all of them in DSPy yet but many are.

dspy/dspy_default_config.json
```diff
@@ -0,0 +1,48 @@
+{
```
okhat, line 1: Interesting file hmm.

dspy/predict/predict.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, signature, **config):
         self.stage = random.randbytes(8).hex()
         self.signature = signature #.signature
         self.config = config
+        self.lm = None
```
okhat, line 15: In the grand scheme of things this makes sense. We'll need good docs about the behavior of LMs of modules.

dspy/predict/predict.py
```diff
@@ -86,7 +94,7 @@ def forward(self, **kwargs):
         if self.lm is None:
             x, C = dsp.generate(signature, **config)(x, stage=self.stage)
         else:
-            with dsp.settings.context(lm=self.lm, query_only=True):
```
okhat, line 89: Ok but we'll need to query_only=True somewhere... I guess we can have `self.query_only` in the module itself... Basically this is affecting now the finetune notebook in the examples/qa/hotpot folder

dspy/primitives/example.py
```diff
@@ -105,3 +105,17 @@ def without(self, *keys):
     
     def toDict(self):
         return self._store.copy()
+
```
okhat, line 108: This is great!

dspy/primitives/module.py
```diff
@@ -26,10 +26,8 @@ def add_parameter(param_name, param_value):
                 add_parameter(name, value)
 
             elif isinstance(value, BaseModule):
-                # When a sub-module is pre-compiled, keep it frozen.
-                if not getattr(value, '_compiled', False):
```
okhat, line 30: Ok but now we'll need to modify bootstrap.py at least so it replicates this behavior.

I guess that means: freeze the module that's like that...

dspy/__init__.py
```diff
@@ -26,4 +26,55 @@
...
+            cls._instance._configure_dspy(config_path=config_path)
+        return cls._instance
+
+    @classmethod
+    def create_lm(cls, model_config):
+        """ Create an LM based."""
+        model_config = {**model_config} # deepcopy config
+        model_name = model_config.pop('model')
+
+        if "gpt-" in model_name:
```
KarelDO, line 54: create_lm could call any model provider you want. Mainly, I wanted to solve the problem of LM configuration saving and loading. You can't serialize the LM so you need to save the kwargs and easily map it to an LM. Any other idea on how to solve this?

dspy/__init__.py
KarelDO: create_lm could call any model provider you want. Mainly, I wanted to solve the problem of LM configuration saving and loading. You can't serialize the LM so you need to save the kwargs and easily map it to an LM. Any other idea on how to solve this?

dspy/dspy_default_config.json
KarelDO: Same comment as above. Want a good way to serialize models, and save programs. For complicated pipelines, you need multiple LMs with different configs. Saving all of these in a config file and using them with a short-hand is very handy. 

We can not do this in DSPy and only do it in Irera, as a type of abstraction on top of DSPY. So, we can drop this here.

Still struggling with saving and loading LMs in DSPy though, that seems like a fundamental issue that should be handled in this repo.

dspy/predict/predict.py
KarelDO: Yes! And the codebase should be restructured to not use any with dsp.context.settings(lm=model) context managers. There should be mainly one way to keep track of LMs on modules (self.lm) and a canonical way of setting these (e.g. program.module1.to(lm)).

(We could also have a student_lm and teacher_lm object, so that you don't have to manually swap out teacher and student LMs during optimization for the teacher program. However, this is not that important.)

dspy/predict/predict.py
KarelDO: When is query_only used? Query_only seems it should be a kwarg for the forward method. I'm not sure where you'd want your model to not send demonstrations when they are set.

dspy/primitives/example.py
KarelDO: Ideally, we'd need a better paradigm for serialization. Manually adding all the dump and load information seems cumbersome.

dspy/primitives/module.py
KarelDO: Sure, I will add this.

KarelDO: I've resolved this here: https://github.com/stanfordnlp/dspy/commit/8f4a0e5ea42c8e22fc8b3180c7924becee975ec4

KarelDO: By default named_parameters should return all parameters, so I added an only_uncompiled flag and made sure all named_parameter calls in bootstrapping / finetuning set this flag to true.

KarelDO:
I've made the new save/load logic compatible with previous saves. I confirmed that 'hotpotqa_with_assertions.ipynb' and 'scone.ipynb' compile as intended with the new named_predictor logic.

Loading the saved scone state works, although there is a small bug where performance is consistently a tiny fraction lower after load than before load. Looking into this now.

dsp/modules/gpt3.py
KarelDO, line 1: only cast as tuple if a list.

dspy/predict/predict.py
KarelDO: Need a solve.

insop:
This PR is rather long, and I am not sure it will be merged soon.

However, at least the changes in `dsp/modules/ollama.py` seems interesting, which might help Ollama users.
Any thought @KarelDO ?

PR #304, opened 2024-01-29
DSPy-Clarifai : Integration for LM and retrievers

mogith-pn:
**Description:**
1. Added Clarifai as LLM provider and retrievers.
2. Added notebook examples, which shows how to call clarifai LM class and retrievers.


okhat:
Looks great @mogith-pn ! I'd rather have the notebook under examples/ so people see it there when looking for examples, but otherwise happy to merge.

okhat:
Oh yeah this is a problem! Let’s not add a dependency but now we need to change the import to be inside the class. Only if the class is used.

okhat:
Ok I edited the clarifai module to do this. Is it all good?

mogith-pn:
@okhat 
Thanks a lot for the changes. Happy to be contribute to DSPy !

PR #308, opened 2024-01-31
README: fix link for finetuning example

raphaelmerx:
The finetuning example doesn't seem to finetune anything! I think https://colab.research.google.com/github/stanfordnlp/dspy/blob/main/examples/qa/hotpot/multihop_finetune.ipynb is the relevant example here


okhat:
:joy: haha thank you

PR #309, opened 2024-01-31
Refactor `dsp.OllamaLocal`'s constructor

INF800:
Remove some hardcoded values from `__init__()`

Previous work: https://github.com/stanfordnlp/dspy/pull/262#discussion_r1440694439


INF800:
If it were up to me, I'd use litellm for ollama instead of using ollama directly.

INF800:
Tested the current code in [this notebook](https://github.com/INF800/dspy-examples/blob/main/ollama/intro_ollama.ipynb).

PR #312, opened 2024-01-31
Make the IPython dependency optional

arian:
This modification replaces the IPython import with a try-except block. If IPython is not available, it sets ipython_display to the built-in print function.

Related: https://github.com/stanfordnlp/dspy/pull/269


okhat:
Thanks so much! I had actually thought this was there already. It clearly wasn’t, not sure what I was misremembering.

PR #316, opened 2024-01-31
remove redundant line

johnowhitaker:
correct is re-defined in the next line


okhat:
I got confused by this before too. It’s actually correct though perhaps bad style

johnowhitaker:
Oh man yeah I missed that it gets used in the second line 🙈 sorry about that!

PR #318, opened 2024-01-31
Update README.md -> the `class RAG(dspy.Module)` section of the Docs

fergusmeiklejohn:
It seems more correct to describe the function flow this way, first we retrieve the context then we use it and the question to generate an answer, but I'm not sure, I'm reading the docs for the first time and may have misunderstood how DSPy works.. :-)


okhat:
Thank you @fergusmeiklejohn ! You are right!

PR #319, opened 2024-02-01
Add support for local embeddings for chroma

prasadshreyas:


dspy/retrieve/chromadb_rm.py
```diff
@@ -21,12 +25,19 @@
 except ImportError:
     chromadb = None
 
+try:
+    import torch
```
okhat, line 29: please don't import dependencies that are not in requirements.txt in the top level

you can import them inside the method you need, so people who don't need them don't see errors for them

insop:
It seems like a good change after addressing @okhat 's feedback.

insop:
I think with a new PR on this RM, https://github.com/stanfordnlp/dspy/pull/400, this PR might need to be revisted.

insop:
@prasadshreyas 

Since https://github.com/stanfordnlp/dspy/pull/400 is merged, could you rebase to the main and see?

dspy/retrieve/chromadb_rm.py
```diff
@@ -85,6 +97,20 @@ def __init__(
...
+            self._local_embed_model = AutoModel.from_pretrained(local_embed_model)
+            self._local_tokenizer = AutoTokenizer.from_pretrained(local_embed_model)
+            self.use_local_model = True
+            self.device = torch.device(
+                "cuda:0"
+                if torch.cuda.is_available()
+                else "mps"
+                if torch.backends.mps.is_available()
+                else "cpu"
+            )
```
tomaarsen, line 110:
```suggestion
            try:
                from sentence_transformers import SentenceTransformer
            except ImportError as e:
                raise ImportError(
                    "You need to install sentence_transformers library to use pretrained embedders. "
                    "Please check the official doc https://www.sbert.net/ "
                    "or simply run `pip install sentence-transformers"
                )
            self._local_embed_model = SentenceTransformer(local_embed_model)
```

tomaarsen (commented):
Additionally, I would not recommend a pure `transformers` implementation, this ends up being quite limited in what models it can support. For example, it won't work with models that use e.g. CLS pooling or extra Dense layers. Instead, you can load the model using the recommended approach, i.e. via Sentence Transformers.

PR #320, opened 2024-02-01
[Bug Fix] WeaviateRM

CShorten:
This should fix it, accidentally wrapped the return in a `dspy.Prediction`.

Sorry for the back and forth with this, pretty sure it is in good shape now.

### DIFF
--- a/dspy/retrieve/weaviate_rm.py
+++ b/dspy/retrieve/weaviate_rm.py
@@ -80,4 +80,4 @@ class WeaviateRM(dspy.Retrieve):
             parsed_results = [result["content"] for result in results]
             passages.extend(dotdict({"long_text": d}) for d in parsed_results)
 
-        return dspy.Prediction(passages=passages)
+        return passages


CShorten:
Hey @okhat, thanks for taking a look! I tested the currently 2.1.6 build and it is still looking for the `long_text` key, I think because the call to `dspy.Prediction` wraps passages in something else. I am basing these changes off of the Clarifai integration, which seems to work in the notebook (although it is not 100% clear if that was from importing the current dspy-ai package or a local build).

However, I can confirm 100% that the current WeaviateRM doesn't work.

CShorten:
@DanielUH2019 would you be interested in testing the PineconeRM and maybe uploading an example notebook to /examples/Integrations?

CShorten:
Maybe worth noting the ChromadbRM also returns passages directly - https://github.com/stanfordnlp/dspy/blob/main/dspy/retrieve/chromadb_rm.py

okhat:
Yeah a Prediction is better because you can return other fields too later

CShorten:
Ah indeed, that would be more powerful. Would it still be an idea to patch it for now with this -- and then going forward remove how `dsp.retrieveEnsemble` parses the `long_text` key?

DanielUH2019:
> @DanielUH2019 would you be interested in testing the PineconeRM and maybe uploading an example notebook to /examples/Integrations?

Hi @CShorten, thanks for reaching out! I will try to do it this weekend

PR #330, opened 2024-02-02
feat(bedrock-lm): Added a LM for Bedrock and a base AWS LM class

JamesScharf:
Created two new LM classes:
- Bedrock
- AWSLM

Bedrock can be used with any Bedrock model once you have configured your AWS credentials with the AWS CLI. AWSLM exists so that we can add support for AWS Sagemaker in the future, which has a very similar API with minor parameter differences.

### DIFF
--- a/dsp/modules/__init__.py
+++ b/dsp/modules/__init__.py
@@ -8,6 +8,7 @@ from .sbert import *
 from .pyserini import *
 from .ollama import *
 from .clarifai import *
+from .bedrock import *
 
 
 from .hf_client import HFClientTGI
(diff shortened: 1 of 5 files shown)


okhat:
Thank you! Looks good overall but please don't add a dependency in the main imports.

If you want to import something not in requirements, basically import it inside the class methods. That way it will basically crash only for those it needs to crash for (i.e., people who need the new class but don't have the right dependencies)

JamesScharf:
Yep! I believe I removed that in the most recent commit? I tried to emulate the style used in the Azure Cognitive Services class.

okhat:
Thanks so much James! Just made one change and merged

PR #339, opened 2024-02-04
Small Fix to vLLM Launch Documentation

sidjha1:
The current documentation for running vLLM with DSPy suggests that the user start an API server via ```python -m vllm.entrypoints.api_server```. However, this leads to errors since the DSPy code assumes an OpenAI-Compatible API server. So the user should instead launch the server via ```python -m vllm.entrypoints.openai.api_server```.

### DIFF
--- a/docs/guides/language_model_details/launching_vllm.md
+++ b/docs/guides/language_model_details/launching_vllm.md
@@ -11,7 +11,7 @@ Follow these steps to set up the vLLM Server:
 Example command:
 
 ```bash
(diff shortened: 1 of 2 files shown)


okhat:
Hey more than happy to merge this, but is this for usage with `dspy.OpenAI` or `dspy.VLLM`?

okhat:
Thanks so much, merged! Do you think we should just use it with dspy.OpenAI then?

PR #343, opened 2024-02-05
CoT fix + move torch dependeny when using vllm

joennlae:
Very cool project 💯 

This PR fixes a typo introduced in [#333](https://github.com/stanfordnlp/dspy/pull/333). The signature is undefined in this context.

Additionally, I moved the torch dependency to only be required when running the model locally through hf and not when a vllm endpoint is called.

### DIFF
--- a/dsp/modules/hf.py
+++ b/dsp/modules/hf.py
@@ -40,19 +40,20 @@ class HFModel(LM):
             hf_device_map (str, optional): HF config strategy to load the model. 
                 Recommeded to use "auto", which will help loading large models using accelerate. Defaults to "auto".
         """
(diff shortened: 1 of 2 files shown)


okhat:
Really appreciate it, @joennlae . Merged!

okhat:
@Shangyint worth checking to make sure we didn't break anything you intended in #333

joennlae:
I am running with these fixes :-) For me, it works :-) But I will do some more extensive testing in the upcoming days :-) I will report if there is an issue.

Shangyint:
Thank you @joennlae! Nice catch.

PR #347, opened 2024-02-06
relax openai version constraint

kujenga:
Since dspy does in fact support both 0.x and 1.x with [1], the constraints are updated here to reflect that, allowing users of the library to make use of whatever version they would like.

Intended to address https://github.com/stanfordnlp/dspy/issues/258

I opted not to update the poetry.lock file here in order to have this be a minimal change aimed at allowing for more flexible  usage by others, rather than changing the defaults for the project, though an upgrade of the default may make sense as v1 of the OpenAI lib has been out for three months now and is fairly stable.

[1] https://github.com/stanfordnlp/dspy/commit/40c519adc0b848b998d028d1b5d73c6dc631696e (and follow-on work)


okhat:
Thank you… ok we’ll live with this

kujenga:
Thanks! Appreciate the near-instantaneous review 😄

Issue #188, opened 2023-10-25
Unable to run HFModel with CPU


Fcabla:
I have been trying to run a Hugging Face model with dspy locally on my laptop.

```python
model = dspy.dsp.modules.HFModel(model ='mosaicml/mpt-7b-chat')
```

When running the code line, I get the following error.

```bash
File "<stdin>", line 1, in <module>
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/dsp/modules/hf.py", line 76, in __init__
    self.model = AutoModelClass.from_pretrained(model).to("cuda")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/transformers/modeling_utils.py", line 2179, in to
    return super().to(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1160, in to
    return self._apply(convert)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 810, in _apply
    module._apply(fn)
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 810, in _apply
    module._apply(fn)
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 833, in _apply
    param_applied = fn(param)
                    ^^^^^^^^^
  File "/Users/myuser/Library/Caches/pypoetry/virtualenvs/myvenv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1158, in convert
    return t.to(device, dtype if t.is_floating_point() or t.is_complex() else None, non_blocking)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #348, opened 2024-02-07
Allow HFModel to use CPU

tisu19021997:
Quick fix https://github.com/stanfordnlp/dspy/issues/188

I think this minor fix could be very helpful for Mac beginners as well as CPU beginners who just get started using the library.


insop:
The change looks good to me.
One thought is that it would been easier for whoever reads the PR if the changes only the fixes instead other changes such as coding style changes.

tisu19021997:
@insop thanks. I just reverted the coding style changes.

insop:
> @insop thanks. I just reverted the coding style changes.

Looks great to me! thank you.

insop:
@okhat , maybe you could merge if you are okay with the change?

okhat:
Merging but please prefer ollama over hfmodel when possible

PR #353, opened 2024-02-07
fix react.py #276 

### LINKED ISSUE #276: using "Tools" with DSPy — ReAct assumes .passages
Hello, 

I would like to know if adding _tools_ to a DSPY module is easy. 
By "tools" here I mean tools as e.g. implemented in Langchain [here](https://python.langchain.com/docs/modules/agents/tools/custom_tools) 

I've seen in the react.py modules that tools can be added, but I ignore how to implement such (e.g as simple one that give current time).  I can very naively make a module with deterministic forward pass 

```
class GiveTime(dspy.Module):
    def __init__(self):
        super().__init__()
    
    def forward(self):
        current_time = datetime.datetime.now().strftime("%H:%M:%S")
        return current_time
```

so in fact the real question is : how to I tell other modules they can use this when need to know current time. That's why I was searching towers React. 

Do you have a notebook of use of the react.py module?

many thanks in advance, and kudos for DSPy! It made my last 2 weeks :)

dannyoo:
fix #276

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -64,7 +64,12 @@ class ReAct(Module):
 
             if action_name == 'Finish': return action_val
 
-            output[f"Observation_{hop+1}"] = self.tools[action_name](action_val)
+            try: 
+                output[f"Observation_{hop+1}"] = self.tools[action_name](action_val).passages
+            except AttributeError:
+                # Handle the case where 'passages' attribute is missing
+                # TODO: This is a hacky way to handle this. Need to fix this.
+                output[f"Observation_{hop+1}"] = self.tools[action_name](action_val)
 
         except Exception as e:
             output[f"Observation_{hop+1}"] = "Failed to parse action. Bad formatting or incorrect action name."
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -64,7 +64,7 @@ class ReAct(Module):
 
             if action_name == 'Finish': return action_val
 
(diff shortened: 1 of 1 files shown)


younes-io:
Hello,
Any idea when this will be merged?
Thank you!

okhat:
Thank you @dannyoo ! I just need this to be temporarily backward compatible.

Can we do: try/except, if there's .passages use that, else do this new behavior? I know that's slightly hacky but we may need it for a while. We can make it deprecated / issue a warning.

okhat:
Will merge otherwise after that!

dspy/predict/react.py
```diff
@@ -64,7 +64,7 @@ def act(self, output, hop):
 
             if action_name == 'Finish': return action_val
 
-            output[f"Observation_{hop+1}"] = self.tools[action_name](action_val).passages
+            output[f"Observation_{hop+1}"] = self.tools[action_name](action_val)
```
smwitkowski, line 67: @dannyoo I think this would be a sufficient change. 

```suggestion
            try: 
                output[f"Observation_{hop+1}"] = self.tools[action_name](action_val).passages
            except AttributeError:
                # Handle the case where 'passages' attribute is missing
                # TODO: This is a hacky way to handle this. Need to fix this.
                output[f"Observation_{hop+1}"] = self.tools[action_name](action_val)
```

I could open a new PR with this change and bring in your commit, but since it's small I'm just adding this suggestion.

insop:
LGTM for the PR, it'd be great to take the changes from @smwitkowski .

PR #354, opened 2024-02-08
Easier api for transforming modules with assertions

Shangyint:
Now, it is possible to write `assert_transform_module(module)` to transform a module to handle suggestions and assertions. The default behavior is implemented by the `backtrack_handler`, which automatically backtracks for failing assertions and suggestions.

### DIFF
--- a/dspy/primitives/assertions.py
+++ b/dspy/primitives/assertions.py
@@ -336,9 +336,9 @@ def assert_transform_module(
         module
     )
 
-    if all(map(lambda p: isinstance(p, dspy.retry.Retry), module.named_predictors())):
+    if all(map(lambda p: isinstance(p[1], dspy.retry.Retry), module.named_predictors())):
         pass # we already applied the Retry mapping outside
-    elif all(map(lambda p: not isinstance(p, dspy.retry.Retry), module.named_predictors())):
+    elif all(map(lambda p: not isinstance(p[1], dspy.retry.Retry), module.named_predictors())):
         module.map_named_predictors(dspy.retry.Retry)
     else:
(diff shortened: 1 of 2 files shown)


okhat:
That’s so cool! Can we also shorten that function’s name?

Or even make it a module method?

program.activate_assertions()

Shangyint:
Great advice! Now we can do:
```python
a = Program()
a.activate_assertions()
# now a can backtrack and self-refine
```
or
```python
teleprompter.compile(student=Program().activate_assertions())
```

okhat:
Amazing! Before I merge,

will this break existing code?

is it an in-place or new-value method? either one is fine just checking

arnavsinghvi11 (commented):
`activate_assertions` is nice!

to clarify, `longformqa_with_assertions = assert_transform_module(LongFormQAWithAssertions().map_named_predictors(Retry), backtrack_handler) 
`
would look something like this now? 

`longformqa_with_assertions = LongFormQAWithAssertions().activate_assertions()`

Shangyint:
Yes. And it shouldn't break any existing code! Could you double check both the old and new api works @arnavsinghvi11?

okhat:
Arnav’s answer implies it is NOT in place, just to be sure right?

Shangyint:
Thanks! Pushed a fix. Both should work now

arnavsinghvi11:
Looks good @Shangyint ! Just a small [issue](https://github.com/stanfordnlp/dspy/commit/6407b3667792c384d81d2567bdb09a1a84217678) which relies on this change to be merged for `activate_assertions()` to work.

PR #358, opened 2024-02-08
Lower default `max_labeled_demos` in `BootstrapFewShot`

CShorten:
Hey team, I think it would be good to lower the default `max_labeled_demos` from 16 to 8. I have been playing with answering FAQs with 16 and it quickly results in compiling a massive prompt such that I exceed the max token limit. There is probably another way around this -- like ofc setting the argument to 8 or 4 or whatever, but I think this might slightly improve the UX for beginners starting with QA RAG and `GPT-3.5-Turbo` / ~4K input length LLMs.

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -34,7 +34,7 @@ from dspy.evaluate.evaluate import Evaluate
 
 
 class BootstrapFewShot(Teleprompter):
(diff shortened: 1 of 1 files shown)


insop:
Sounds sensible to me.

> ... like ofc...
What is ofc?

okhat:
Hey @CShorten , I agree with this! But it will affect a lot of existing runs so it's a bit of a silent breaking change...

Maybe we should keep this and other related things in place for one big breaking release?

Issue #359, opened 2024-02-09
Weaviate RM should NOT assume that "content" contains the text of the passage


gopidon:
Hi @CShorten 

dspy.retrieve.weaviate_rm, line 74

Give the flexibility to pass the the name of the text field as a parameter along with the collection name. Something like below:

WeaviateRM(
    weaviate_collection_name= "<collection_name>", 
    weaviate_collection_text_field="<field-name>",
    weaviate_client=weaviate.Client("<weaviate-url>")
)

CShorten:
Hey Gopi! Thanks for bringing this up, this will resolve it! 

https://github.com/stanfordnlp/dspy/pull/360

PR #360, opened 2024-02-09
Allow custom text key to WeaviateRM - resolve issue #359

CShorten:
Lets you specify which text key to return from a call to `dspy.Retrieve`

This does not incur a breaking change thanks to the `Optional` typing and default argument value.

This will resolve https://github.com/stanfordnlp/dspy/issues/359.


insop:
With a cursory check, it looks great!

Hi @CShorten,

You could take this suggestion to *a separate PR.*
I think it'd be great if you could add WeaviateRM description to this markdown `docs/retrieval_models_client.md`.
This is a PR that adds description for other RM to this document: https://github.com/stanfordnlp/dspy/pull/400.

okhat:
Thank you @CShorten and @insop ! Merged

PR #361, opened 2024-02-09
Fix a few small typos

svilupp:
Thank you for the package! 

When reading through the code, I noticed a few typos, so I took the liberty to fix them.

Only one of them is relevant for performance (it's in the prompt for Bayesian teleprompter).


svilupp:
Btw there is a typo in a kwarg name: https://github.com/thomasahle/dspy/blob/719a8d1805b4a59780ed9cda6840e782be39c013/dspy/teleprompt/bootstrap.py#L99
“bootsraps” vs bootstraps

But I intentionally didn’t fix it, because that could break code and would need to be versioned up.

okhat:
Ah thanks for the note. Actually please do feel free to fix `max_bootsraps` too. It's not used anywhere based on quick search (and I'm pretty sure it's not used outside the library itself, this is super internal). Happy to merge this, thank you!

dspy/teleprompt/signature_opt_bayesian.py
```diff
@@ -81,7 +81,7 @@ class ObservationSummarizer(dspy.Signature):
 
 class DatasetDescriptor(dspy.Signature):
     ("""Given several examples from a dataset please write observations about trends that hold for most or all of the samples. """
-    """Some areas you may consider in your observations: topics, content, syntax, conciceness, etc. """
+    """Some areas you may consider in your observations: topics, content, syntax, conciseness, etc. """
```
okhat, line 84: hmm this is a good fix but it may affect runs :/ so we probably will only want to merge this one at a later version cc: @XenonMolecule

XenonMolecule: Ah, whoops, good catch here, svilupp!  I'm checking with Krista, but we may both have a version of the Bayesian optimizer running off a branch for the current experiments, so merging may not be a concern for ongoing runs.  I will double-check with her and confirm.

svilupp: Should I remove this fix from the PR to merge the rest or let's keep it open for a few days for you to check?

svilupp:
> Ah thanks for the note. Actually please do feel free to fix `max_bootsraps` too. It's not used anywhere based on quick search (and I'm pretty sure it's not used outside the library itself, this is super internal). Happy to merge this, thank you!

Added to the PR + refresh with latest main

okhat:
Ok added back the typo cc: @XenonMolecule let's fix this in our own refactors soon.

Merging now @svilupp ! Thank you so much.

PR #363, opened 2024-02-09
fix(gmsk8_metric): Look for answer in entire LLM output. Not just the First Line

DrChrisLevy:
I was going through one of the initial documentation examples.
I noticed that the eval function, `gsm8k_metric`, was returning False even when the LLM (GPT) was giving the correct answer in its response output. It's because the eval metric was calling `parse_integer_answer` which has an argument `only_first_line=True`. Shouldn't the output from the LLM be considered a correct answer even if the answer is at the end of the LLM output and not on the first line? Is there a reason the answer needs to be on the first line? 

The other changes are all linting/formatting changes forced by the ruff linter/formatter when running the pre-commit check. Is that expected? Thanks!

### DIFF
--- a/dspy/datasets/gsm8k.py
+++ b/dspy/datasets/gsm8k.py
@@ -1,42 +1,45 @@
-import tqdm
 import random
 
+import tqdm
 from datasets import load_dataset
-from dspy.datasets.dataset import Dataset
+
 
 class GSM8K:
     def __init__(self) -> None:
+        """GSM8K Metric."""
         super().__init__()
         self.do_shuffle = False
 
-        dataset = load_dataset("gsm8k", 'main')
+        dataset = load_dataset("gsm8k", "main")
 
-        hf_official_train = dataset['train']
-        hf_official_test = dataset['test']
+        hf_official_train = dataset["train"]
+        hf_official_test = dataset["test"]
         official_train = []
         official_test = []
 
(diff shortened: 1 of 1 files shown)


okhat:
Thank you @DrChrisLevy ! I think you may want to consider just having a second metric function that sets that to False. Ultimately you can set any metric you want — and then the question is how do we get the LM to follow it.

I can see that this helped your run score better on that example, but there will often be other examples of the opposite nature; where the LM may generate additional unnecessary text (eg other questions and answers for example) so there’s no one right answer in general on the metric’s end.

I’m happy to reconsider this but just my current thoughts.

DrChrisLevy:
Right on. I did re-define the metric on my end to run through the initial example. But as a newcomer to the library, and as this metric was used in one of the first tutorials, I did find it an initial stumbling block that the metric was saying the LLM output was wrong. I thought others may find that confusing just from a tutorial standpoint. I totally get that this metric was added for demo purposes. I don't have a super strong opinion in it. So if you prefer to close this PR then all good :)

okhat:
Thanks a lot. I will close for now but will ponder on adding a note. This can be instructive: our metrics are never perfect and sometimes there are no right answers — but once we define a metric that’s inherently reasonable, we then want to maximize how the system fits it.

PR #366, opened 2024-02-09
Fixing bayesian optimizer bug with prompt / task model

klopsahlong:
Previously the prompt and task models were defaulting to "None". Now they are defaulting to `dspy.settings.lm`.


CShorten:
Ah I had no idea you could do - `dspy.settings.lm("hello")`. Learned a lot from this PR, thank you so much for the fix!

PR #368, opened 2024-02-10
New Signature class based on pydantic + unit tests

thomasahle:
This PR creates a new Signature class based on Pydantic, which allows for type signatures.
It doesn't actually do anything with the types for now, but it's the first step.

Maybe more importantly, the PR creates 85 unit tests, covering most of `dspy/`, which were important to make sure the new type signature code didn't break anything, and will hopefully be useful for future refactors.


dspy/predict/__init__.py
```diff
@@ -6,3 +6,4 @@
 from .aggregation import majority
 from .program_of_thought import ProgramOfThought
 from .retry import Retry
+from .knn import KNN
```
okhat, line 9: I want to check if this adds any requirements/dependencies that weren't there otherwise.

thomasahle: Ah, I guess I just assumed everything in predict/ should also be in the module.
I guess KNN uses numpy which is not in the dspy requirements?

okhat: Yeah, we can probably remove it for now (or in the long run move the numpy import to inside the class?)

okhat:
I _think_ this is good to merge (except the note above and maybe 1-2 minor things) but it's gigantic haha

dspy/predict/__init__.py
thomasahle: Hm. I also use numpy in the dummy vectorizer which is used to replace the search in ReACT. So the tests won't run without numpy.

thomasahle:
I tried to keep the changes outside of the tests as small as possible. But it does affect nearly eveything 😳.

Do you think there are any modules I may have forgotten to test that would also need to be updated?

younes-io:
Hi @thomasahle 
Thank you for this nice PR!
I have an [issue](https://github.com/stanfordnlp/dspy/issues/386#issuecomment-1943247745) using the `BayesianSignatureOptimizer` with extended signatures from different predictors, so I wondered if you considered this in your PR. I saw that you've "migrated to the new API" so not sure if it deals with this as well.

thomasahle:
Screenshot 2024-02-14 at 4 19 54 PM
I did consider it so far as I noticed it didn't work, and only tested with ChainOfThought :)
Maybe I should have just created a test for Predict too, and let it be failing. But I didn't want to create a PR with failing tests...

CyrusOfEden:
Wow, this looks fantastic! I'm particularly jazzed by the unit tests (which will make refactoring much easier). 

Thanks for lifting this huge load @thomasahle 🙏 

@okhat what would it take from your side of things to merge this?

younes-io:
@okhat thank you for the merge; could you please add release management so it can be easy to find the latest version and the release notes (commits)?

thomasnormal:
While working on the method decorators, I realized the `fields` property on `Signature` had a slight issue, as it would sometimes give output fields before input fields, which would screw up generation.
I fixed it here: https://github.com/thomasahle/dspy/commit/5794984132b1862dec6e33b19234bf03e161917d

I can make a new PR to the new branch with the fix?
But I can also just wait to the PR is merged into main, since it doesn't appear to break anything in the current codebase.

okhat:
@thomasahle you're now a maintainer so it's easier to do as you see fit (e.g., update the pydantic branch directly if that's easier)

thomasnormal:
Sounds good. I'll still let you decide when to merge it into main :-)

There's a bunch of work I want to do once it's in main, such as actually implementing the type checking.
But also setting github up to run tests automatically etc.
I don't think I should start any of that work before the PR is in main though, as it will make it harder to merge the branches eventually.

thomasnormal:
Regarding the actual type checking/fixing, we could either do it using dspy.assertions (like in my first version here https://github.com/thomasahle/dspy/blob/main/examples/types.py) or using a manual retry loop (like in my second version here https://github.com/thomasahle/dspy/blob/method/examples/functional.py#L67).

Having tried both I'm probably most fan of the second version. DSpy assertions require wrapping the Predict in Retry, while the simple checking loop could just be implemented in the forward method of Predict.

okhat:
Yeah, within-module checking is more appropriate, assertions are overpowered for this and will make composition harder here since this is a local check

dspy/teleprompt/signature_opt_bayesian.py
```diff
@@ -202,16 +205,26 @@ def _generate_first_N_candidates(self, module, N, view_data, view_examples, demo
...
+                basic_prefix = last_field.json_schema_extra["prefix"]
             else:
                 basic_instruction = predictor.extended_signature1.instructions
-                basic_prefix = predictor.extended_signature1.fields[-1].name
+                *_, last_field = predictor.extended_signature1.fields.values()
+                basic_prefix = last_field.json_schema_extra["prefix"]
             with dspy.settings.context(lm=self.prompt_model):
                 # Data & Examples
-                if view_data and view_examples:
+                if view_data and view_examples and 1 in example_sets[id(predictor)].keys():
```
okhat, line 216: what is `1 in example_sets[id(predictor)].keys()`? I  that based on the original?

dspy/teleprompt/signature_opt.py
```diff
@@ -103,12 +105,13 @@ def compile(self, student, *, devset, eval_kwargs):
...
             basic_prefix = None
+            *_, last_key = predictor.extended_signature.fields.keys()
             if (hasattr(predictor, 'extended_signature')):
                 basic_instruction = predictor.extended_signature.instructions
-                basic_prefix = predictor.extended_signature.fields[-1].name
+                basic_prefix = predictor.extended_signature.fields[last_key].json_schema_extra['prefix']
             else:
                 basic_instruction = predictor.extended_signature1.instructions
-                basic_prefix = predictor.extended_signature1.fields[-1].name
+                basic_prefix = predictor.extended_signature1.fields[last_key].json_schema_extra['prefix']
```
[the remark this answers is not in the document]
okhat: also name != prefix.... this seems to be a bit of a change, but for the better?

dspy/teleprompt/signature_opt_bayesian.py
thomasnormal: There seemed to be a bug where `example_sets[id(predictor)][i]` would fail if `example_sets[id(predictor)]` was an empty dict.
I added this, and things seemed to work.

Maybe I misunderstand what was causing the issue. Is there a reason why we'd expect `example_sets[id(predictor)]` to never be empty?

dspy/teleprompt/signature_opt.py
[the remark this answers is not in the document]
thomasnormal: Ok, thanks for flagging this. This is an important question.

The issue is that `pydantic` doesn't supports arbitrary keyword arguments to `Field`. (See 
 https://docs.pydantic.dev/2.6/migration/#changes-to-pydanticfield ).
There is a long list of allowed keywords here: https://docs.pydantic.dev/latest/api/fields/
And everything else gets moved to `json_schema_extra`.

Since we now return the pydantic field instance from `signature.fields`, we can't even overload this to force `prefix` to be a accessible as `fields[last_key].prefix`.

One option could be to redefine the `fields` property itself to not return the pydantic Field objects, but return the `json_schema_extra` dicts. So something like:
```python
@property
def fields(cls):
    return {k: f.json_schema_extra for k, f in cls.__fields__}
```
But then we have two notions of what a "field" is, rather than just cloning the pydantic notion.

I thought the conceptual overhead of this was not worth the benefit to ergonomics.
But I'll let you pick the winner.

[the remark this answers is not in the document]
thomasnormal: Regarding `name != prefix`, this comes from the previous code using `dsp.Template` instead of Signature, I believe. Since the old code still used `prefix` in Signature, but `name` in Template.
I tried to use Signature everywhere, and limit template to where it's really needed.
But it's possible Template is more correct here?

thomasnormal (commented):
Replied to @okhat's comments

thomasahle:
Honestly the signature optimizer is the piece I'm least sure of. So I'm glad you're checking it.

Also please check if the traces in test_signsture_opt.py are actually working as you would expect. 
Maybe there are even some more tests you can propose to stress test it further.

dspy/teleprompt/signature_opt.py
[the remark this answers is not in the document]
CyrusOfEden: I’m with Thomas on conceptual overhead here, lfg

dspy/teleprompt/signature_opt_bayesian.py
okhat: I can't identify the specific issue, but my main concern is that I don't understand why "1 in ..." is the right way to do it, why not `len(.) > 0`

thomasahle:
Regarding the Field thing we discussed, the way my Signature currently works is:
```python
class SignatureMeta(type(BaseModel)):
    ...
class Signature(BaseModel, metaclass=SignatureMeta):
    pass
```
The goal was to be able to do things like
```python
class MySignature(Signature):
    ...
```
as in the current code, where the subclass `MySignature` becomes the object you work with, rather than typical Pydantic use cases, where instances of `MySignature` would be the main objects to interact with.

Anyway, as you've seen `MySignature` can be used exactly like a pydantic type signature. And Input/OutputField are just defined as
```python
def InputField(**kwargs):
    return pydantic.Field(**kwargs,  __dspy_field_type="input")
def OutputField(**kwargs):
    return pydantic.Field(**kwargs, __dspy_field_type="output")
```
However, this mean that if a user of DSPy writes
```python
class MySignature(Signature):
    positive: int = InputField(gt=0)
```
The keyword `gt=0` could potentially collide with a keyword we'd like to add to DSPy.

In general pydantic classes are actually a quite rich language of validators and detailed specifications for types.
This is how it improves over something more simple like dataclasses in python.

The simple solution would be prefixing all dspy arguments:
```python
class MySignature(Signature):
    positive: str = InputField(dspy_desc="abc", description="abc")
```
But it's not that pretty.

okhat:
Maybe we should prefix the other way around — special keyword for pydantic-y things.

okhat:
Anyway : any choices that are backward compatible are fine with me in the short term until we can make a major release

CyrusOfEden:
re: description, we should move towards pulling from the pydantic description.

For the other ones, extra keywords is fine. I imagine we'll have a bigger breaking API change with 3.0 so whatever we decide here isn't gonna live forever and is reversible

thomasahle:
> Maybe we should prefix the other way around — special keyword for pydantic-y things.

Let's look at the parameters that are specific and shared between the two:

pydantic only:
- `gt`/`lt`: for integer sizes
- `pattern`: regular expression the output must match
- `min_length`/`max_length`: number of characters

dspy only:
- `prefix`: When you want to the `Name)` to be different from the variable name
- `format`: When you want to override the default data type serializtion

shared:
- `desc`/`description`: Human-readable description

It seems to me the pydantic field arguments are things we "want" users to define, while the dspy-only arguments are things that the user really shouldn't have to care about. Stuff that should "just work" and only be changed by optimizers etc.

(Actually we could even use the pydantic `title` or `alias` as an equivalent of dspy `prefix`. So that's even less stuff left to consolidate.)

> Anyway : any choices that are backward compatible are fine with me in the short term until we can make a major release

Does that mean you are ok with `field.json_schema_extra['prefix']` or `field.dspy['prefix']` or something for now?

PR #371, opened 2024-02-11
fix: manually iterate kwargs['n'] in anyscale

paxcema:
Hi,

I've noticed that `dsp.modules.hf_client.Anyscale` fails to produce multiple completions when a `dspy.predict.predict.Predict` object using it has `n>1`. The LLM provider itself does not support this option for now, so I was thinking this may be something DSPy could handle instead.

The following code snippet reproduces the issue:

```python
import os
import dspy

os.environ["ANYSCALE_API_BASE"] = 'https://api.endpoints.anyscale.com/v1'
os.environ["ANYSCALE_API_KEY"] = 'esecret_123'  # set to match your anyscale endpoints key

llm = dspy.Anyscale(model='mistralai/Mistral-7B-Instruct-v0.1', max_tokens=1000)
dspy.configure(lm=llm)

N = 3
generate_answer = dspy.Predict('question -> answer', n=N)
r = generate_answer(question='Why did the chicken cross the road?')

assert len(r.completions.answer) == N, f"mismatch! expected {N} but found {len(r.completions.answer)}"
```

With the code on this PR, the above error no longer occurs and completions work as expected. 

Please have a look and let me know if I've missed anything!

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -293,6 +293,8 @@ class Together(HFModel):
                 print(f"resp_json:{resp_json}")
             print(f"Failed to parse JSON response: {e}")
             raise Exception("Received invalid JSON response from server")
+
+
 class Anyscale(HFModel):
     def __init__(self, model, **kwargs):
         super().__init__(model=model, is_client=True)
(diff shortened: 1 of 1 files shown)


okhat:
Thank you! Merging.

insop:
@paxcema 
Nice.

Could you also update `class Together(HFModel)` as well? it is the same code?
If not I will follow up after this PR.
Thank you.

okhat:
Hmm I guess we don’t want to just repeat that logic for each provider, also in the sense that this will make these requests more expensive than if they supported automatic completions

insop:
Agree, I did not like the way it is now. If you have any suggestion, such as using a class that contains most common and use it as a base (asking you since I am still new to the codebase).
Happy to follow up by refactor providers code.

That said, @paxcema, ignore what I said, I can follow up after this PR for a better direction.

okhat:
Merged!

okhat:
Thanks so much @paxcema and @insop !

PR #372, opened 2024-02-11
documentation and QuizGen+TweetGen notebooks

arnavsinghvi11:
-documentation for DSPy Assertions
-example notebooks for the QuizGen and TweetGen tasks


arnavsinghvi11:
WIP currently since Compile+Infer strategy is running into caching errors due to some recent changes. will merge once this is resolved.

okhat:
Maybe mark as "draft" if still WIP

okhat:
Btw cache problems, are they due to openai version? 0.28.1 vs 1.16

Shangyint:
The issue might be due to multithreading in compiling. Needs to investigate further, but as for the cached notebook, it is fine to use one thread for `compile`, which solves the current issue.

Issue #373, opened 2024-02-11
Return outputs from Evaluate 


fearnworks:
### Summary
The current implementation of the `Evaluate` class within the DSPy framework provides a powerful mechanism for assessing model performance across development datasets. However, it lacks the capability to return individual output predictions alongside their corresponding evaluation metrics. This limitation hinders detailed analysis and debugging efforts, which are essential for iterative model improvement. To enhance usability and facilitate deeper insights into model behavior, we should allow `Evaluate` to return the outputs along with their evaluation scores.

### Details
The proposed enhancement involves extending the `Evaluate` class functionality to include an option that enables users to retrieve a list of output predictions generated during the evaluation process. This change would provide users with the ability to conduct thorough analyses of the model's predictions, identify patterns in errors, and make more informed decisions about further model adjustments.

#### Current Implementation Shortcomings:
- Lack of direct access to model predictions for each example in the development set.
- Difficulty in performing detailed error analysis and identifying opportunities for model improvement.

#### Proposed Solution:
Introduce an additional parameter to the `Evaluate` class (e.g., `return_outputs=False`) that, when enabled, modifies the behavior of the `__call__` method to include output predictions in its return value.

### Example Usage:
```python
evaluator = Evaluate(devset=my_devset, metric=my_metric, return_outputs=True)
evaluation_score, outputs = evaluator(my_program)
```

### Implementation Considerations:
- Ensure backward compatibility by defaulting the new parameter to False.
- Update documentation to reflect the new feature and provide examples of its use.

PR #374, opened 2024-02-11
Return evaluation results

fearnworks:
# Output Retrieval in Evaluate Class

## Overview
Related to : https://github.com/stanfordnlp/dspy/issues/373 
This PR introduces a new feature to the `Evaluate` class that allows users to optionally retrieve the raw output predictions alongside the evaluation metrics. The goal of this enhancement is to provide more transparency and facilitate a deeper analysis of model performance during the evaluation phase.

## Changes
- Added a `return_outputs` parameter to the `__call__`  and constructor method of the `Evaluate` class.
- When `return_outputs` is set to `True`, the method returns a tuple containing the evaluation score and a list of tuples, each containing an example, its prediction, and the corresponding score.
- Ensured backward compatibility by setting the default value of `return_outputs` to `False`.
- Utilized the new feature in the baleen comparison in intro.ipynb (this is on its own commit that can be rolled back if needed to support the cache logic in intro.ipynb)

## Benefits
- **Transparency:** Users can now see the exact predictions made by their models for each input example.
- **Analysis:** Facilitates detailed error analysis and helps identify patterns in model predictions that may require further investigation.
- **Iterative Development:** By examining model outputs, developers can make targeted adjustments, leading to more effective model tuning.

## Usage
```python
evaluator = Evaluate(devset=my_devset, metric=my_metric, return_outputs=True)
evaluation_score, outputs = evaluator(my_program)
```


okhat:
Looks good overall but this PR is trying to change intro.ipynb too.

Also maybe we should return a dspy.Prediction(outputs=…, scores=…,) if return output is true to avoid confusion with a tuple of items

fearnworks:
I'll look at the prediction return option. 

Should I remove the change to intro.ipynb, should be easy to revert. Was mostly using it as an example of the functionality.

okhat:
Yes please I’d revert the intro.ipynb change (it could be a separate smaller notebook but this is probably not so complex to need illustration, it’s pretty clean/direct)

PR #379, opened 2024-02-12
Add support to load from any HF dataset and CSV

krypticmouse:
Code Snippet:
```python
dl = DataLoader(
    train_size = 0.5,
    dev_size = 0.2,
    test_size = 0.3   
)

dl.from_huggingface("databricks/databricks-dolly-15k")

print(len(dl.train))
print(len(dl.dev))
print(len(dl.test))
```

### DIFF
--- a/dspy/datasets/dataloader.py
+++ b/dspy/datasets/dataloader.py
@@ -73,7 +73,6 @@ class DataLoader(Dataset):
             else:
                 raise ValueError("No splits provided and dataset has more than one split. At this moment multiple splits will be concatenated into one single split.")
         
-        print(dataset)
         if not fields:
             fields = list(dataset.features)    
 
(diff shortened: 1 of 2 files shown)


dspy/datasets/dataloader.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                raise ValueError("No splits provided and dataset has more than one split. At this moment multiple splits will be concatenated into one single split.")
+        
+        if not fields:
+            fields = list(dataset.features)    
+
+        self._process_dataset(dataset, fields)
+
+    def from_csv(self, file_path:str, fields: List[str] = None):
+        dataset = load_dataset("csv", data_files=file_path)["train"]
```
arnavsinghvi11, line 82: just curious on why to keep ["train"] here. wondering if anyone has datasets not sorted to train/dev/test yet who may want to use it.

krypticmouse: The idea was to support a single csv and create split from that. So the file_path would need to be a dict in order to support the multiple csv split. Though it's something we can iterate on!!

PR #387, opened 2024-02-14
Allow to set LM and RM per module on constructors and forward func

Neoxelox:
This PR allows to set LM and RM per module on their constructors or dynamically when calling them.

Note that I removed the `query_only=True` parameter on the existent `with dsp.settings.context` in the predictor `forward` because I checked and that parameter is only used on `dsp.TemplateV2` but dspy is using `dsp.TemplateV3`.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -32,13 +32,14 @@ class Predict(Parameter):
 
             for k, v in inputs.items():
                 v.finalize(k, infer_prefix(k))
-            
+
             for k, v in outputs.items():
                 v.finalize(k, infer_prefix(k))
 
             self.signature = dsp.Template(instructions, **inputs, **outputs)
 
-    
+        self.lm = config.get("lm")
+
     def reset(self):
         self.lm = None
         self.traces = []
@@ -55,10 +56,10 @@ class Predict(Parameter):
 
         import dspy
         self.demos = [dspy.Example(**x) for x in self.demos]
-    
+
     def __call__(self, **kwargs):
         return self.forward(**kwargs)
-    
+
     def forward(self, **kwargs):
         # Extract the three privileged keyword arguments.
         new_signature = kwargs.pop("new_signature", None)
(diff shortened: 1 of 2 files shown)


okhat:
Thank you so much @Neoxelox !

btw @KarelDO is this what you have in mind too?

dspy/predict/predict.py
```diff
@@ -88,12 +89,8 @@ def forward(self, **kwargs):
         if new_signature is not None:
             signature = dsp.Template(signature.instructions, **new_signature)
 
-        if self.lm is None:
+        with dsp.settings.context(lm=lm):
             x, C = dsp.generate(signature, **config)(x, stage=self.stage)
-        else:
-            with dsp.settings.context(lm=self.lm, query_only=True):
```
okhat, line 94: The only problem is that this part is necessary (`query_only=True`) for BootstrapFinetune. There's probably a cleaner way to do that but I'm not sure how.

okhat: If we can make sure that part of BootstrapFinetune is not affected, I'll merge!

okhat: See this and the TODO/FIXME above it: https://github.com/stanfordnlp/dspy/blob/bd82edb13492d526c0329bf0aaec9547938eb68e/dspy/teleprompt/finetune.py#L164

I think what we need is self.query_only = False in __init__ here, and basically we can set .query_only=True in teleprompt/finetune.py line 165

okhat:
Pinging @KarelDO also to check if this is how he'd do it

dspy/predict/predict.py
Neoxelox: I see, sorry for missing that one, there was a typo on that comment (`only_query`) and when I searched for `query_only` it did not appear on the results.

Neoxelox:
I'm closing this PR because I have seen that a lot of places (outside of the predictor) the code is relying on the global set dsp.settings.lm, ex.: `ChainOfThought` / `ChainOfThoughtWithHint` / `compile()` /  `majority_vote()` / `bootstrap` /  the `history` itself...

So I'm not feeling confident enough to make that change myself and risking library stability. I will wait until the big refactor is done, maybe you even add this feature yourselves.

Issue #365, opened 2024-02-09
Bedrock LM produces extraneous key error in unclear circumstances


JamesScharf:
AWS's Bedrock runtime does not accept the following parameters, however they are still being passed to the LM at some point:
`n` and `max_tokens`

At what point is this being done? I haven't specified either value in my code which uses DSPy.

Here's  the full error from boto3:
```
ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: #: extraneous key [max_tokens] is not permitted#: extraneous key [n] is not permitted, please reformat your input and try again.
```

tech4life87:
I am also getting the same exact error when trying to use  "Getting Started" example notebook. Here is my code

`bedrock = dspy.Bedrock(model='anthropic.claude-v2', region_name="us-west-2")
colbertv2_wiki17_abstracts = dspy.ColBERTv2(url='http://20.102.90.50:2017/wiki17_abstracts')
dspy.settings.configure(lm=bedrock, rm=colbertv2_wiki17_abstracts)
`
Evaluation
`from dspy.evaluate.evaluate import Evaluate

evaluate_on_hotpotqa = Evaluate(devset=devset, num_threads=1, display_progress=True, display_table=5)

metric = dspy.evaluate.answer_exact_match
evaluate_on_hotpotqa(compiled_rag, metric=metric)`

okhat:
I see. Does Bedrock support something else in place of `n` and `max_tokens`? Both are necessary in DSPy.

i.e., `num_completions` and `maximum_output_tokens`

If the former isn't supported, we can do it with a loop... but that's more expensive.

okhat:
Anyway can fix now by just ignoring n and max_tokens in the `request` function for Bedrock @JamesScharf if they can't be replaced.

Otherwise, can fix by replacing these values on the fly with the correct kwarg names

tech4life87:
@okhat It depends with which model one is using with Bedrock. If using Claude then as @JamesScharf correctly points out is`max_tokens_to_sample` if using Titan Models it is `maxTokenCount `. I am kind of curious if using something like [LiteLLM](docs.litellm.ai) or langChain would abstract these model/provider specific nuances.

JamesScharf:
I'd be hesitant to use LiteLLM for this abstraction. I've used its Sagemaker wrapper and it did not seem to support temperature.

JamesScharf:
[...]

PR #388, opened 2024-02-14
Bedrock args hotfix

JamesScharf:
Fixing bugs mentioned here where `n` and `max_tokens` throw errors when passed as `kwargs`:
https://github.com/stanfordnlp/dspy/issues/365


okhat:
Wonderful, thank you James!

PR #391, opened 2024-02-14
[BugFix] QdrantRM Patch, add  long_text key to passages

infoslack:
This PR adds the `long_text` key to the `passages` returned by the QdrantRM.

In dsp/primitives/search line 10: `passages = [psg.long_text for psg in passages]` throws an error without this.

I'm not aware of the full context, but this PR will patch the QdrantRM for now.


okhat:
Thanks so much @infoslack !

Issue #389, opened 2024-02-14
Embeddings.create() got an unexpected keyword argument 'engine', when using ChromaDB


pedrurrego:
# Context ℹ️ 💁🏼‍♂️ 
I have a VectorDB with `Chroma` using OpenAI's Embeddings and created a `retriever_model` in `dspy`:

```python
from dspy.retrieve.chromadb_rm import ChromadbRM
retriever_model = ChromadbRM(collection_name=COLLECTION_NAME, persist_directory=MY_PATH)
```

The following are my dependencies in the `requirements.txt` file. I am using `Python 3.10.12`.

```
chromadb==0.4.22
dspy-ai==2.1.9
openai==1.12.0
tiktoken==0.5.1
```

# Problem 💩 
When running
```python
dspy.settings.rm("Who is Chuck Norris?")
```

I get the following error:

```shell
File [~/Library/CloudStorage/OneDrive-Personal/Code/AI/LLM/dspy/venv/lib/python3.10/site-packages/dspy/retrieve/chromadb_rm.py:139](139), in ChromadbRM._get_embeddings(self, queries)
    [125] """Return query vector after creating embedding using OpenAI
    [126]
    [127] Args:
   (...)
    [131]     List[List[float]]: List of embeddings corresponding to each query.
    [132] """
    [134] model_arg = {"engine": self.model_name,
    [135]     "deployment_id": self.model_name,
    [136]     "api_version": self.api_version,
    [137]     "api_base": self.api_base,
    [138] }
--> [139] embedding = self.openai_ef._client.create(
    [140]     input=queries, model=self._openai_embed_model, **model_arg, api_provider=self.openai_api_type
    [141] )
    [143] return [embedding.embedding for embedding in embedding.data]

TypeError: Embeddings.create() got an unexpected keyword argument 'engine'
```

# Possible solution 👍🏼 ✅ 
I was able to solve it by using only `input` and `model` arguments in `self.openai_ef._client.create()` method, as shown below:

```python
embedding = self.openai_ef._client.create(input=queries, model=self._openai_embed_model)
```

What do you guys think about this? Thanks in advance for your help!

okhat:
Ok solved(?) by merging #400. Let us know if all is good now. Thank you @lukaskuhn-lku  and @animtel

okhat:
Will be pushed to pypi on the next release

PR #396, opened 2024-02-15
fixes _get_embeddings for ChromaDB

lukaskuhn-lku:
As described in #389 _get_embeddings is throwing an error when called via retriever(query): _TypeError: Embeddings.create() got an unexpected keyword argument 'engine'._

This is a result of the unnecessary usage of model_args and api_provider. I removed both and now the function works as expected.

### DIFF
--- a/dspy/retrieve/chromadb_rm.py
+++ b/dspy/retrieve/chromadb_rm.py
@@ -130,14 +130,8 @@ class ChromadbRM(dspy.Retrieve):
         Returns:
             List[List[float]]: List of embeddings corresponding to each query.
         """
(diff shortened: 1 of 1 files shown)


insop:
@lukaskuhn-lku, it seems like https://github.com/stanfordnlp/dspy/pull/400 update has wider scope changes.
Maybe this PR might not be necessary if the newer one (400) is reviewed and merged.

okhat:
Thank you so much @lukaskuhn-lku ! I will merge #400 so I will close this, but I really appreciate your contribution.

PR #398, opened 2024-02-15
Create Google LM

CShorten:
Followed the Cohere LM template. It doesn't look like you can sample multiple generations in one call to the API (linked the Google API docs I'm following in the code). Outside of that, I think it is a pretty standard implementation in sync with the other LMs.

Comment:
- Cool to get a better understanding of how `history` is stored as internal state to the LMs as a `list[dict[str, Any]]`. I would like to explore adding functionality around inspect_history next.

### DIFF
--- a/dsp/modules/google.py
+++ b/dsp/modules/google.py
@@ -52,7 +52,7 @@ class Google(LM):
             Additional arguments to pass to the API provider.
         """
         super().__init__(model)
-        self.google = genai.configure(api_key="")
+        self.google = genai.configure(api_key=self.api_key)
         self.provider = "google"
         self.kwargs = {
             "model_name": model,
--- a/dsp/modules/google.py
+++ b/dsp/modules/google.py
@@ -1,0 +1,105 @@
+import math
+from typing import Any, Optional
+import backoff
+
+from dsp.modules.lm import LM
+
+try:
+    import google.generativeai as genai
+except ImportError:
+    google_api_error = Exception
+    print("Not loading Google because it is not installed.")
+
+def backoff_hdlr(details):
+    """Handler from https://pypi.org/project/backoff/"""
+    print(
+        "Backing off {wait:0.1f} seconds after {tries} tries "
+        "calling function {target} with kwargs "
+        "{kwargs}".format(**details)
+    )
+
+
+def giveup_hdlr(details):
+    """wrapper function that decides when to give up on retry"""
+    if "rate limits" in details.message:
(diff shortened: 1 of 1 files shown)


dsp/modules/google.py
```diff
@@ -0,0 +1,105 @@
...
+            Which pre-trained model from Google to use?
+            Choices are [`gemini-pro-1.0`]
+        api_key : str
+            The API key for Google.
+            It can be obtained from https://cloud.google.com/generative-ai-studio
+        **kwargs: dict
+            Additional arguments to pass to the API provider.
+        """
+        super().__init__(model)
+        self.google = genai.configure(api_key="")
```
insop, line 55: could it be this? setting `api_key` instead of `""`
```
self.google = genai.configure(api_key=api_key)
```

dsp/modules/google.py
```diff
@@ -0,0 +1,105 @@
...
+        self.provider = "google"
+        self.kwargs = {
+            "model_name": model,
+            "temperature": 0.0,
```
insop, line 59: `temperature` and other configuration might be convenient to update through `kwargs`

How about something like this?

```
"temperature": 0.0 if "temperature" not in kwargs else kwargs["temperature"],

```

dsp/modules/google.py
```diff
@@ -0,0 +1,105 @@
...
+    """Wrapper around Google's API.
+
+    Currently supported models include `gemini-pro-1.0`.
+    """
+
+    def __init__(
+        self,
+        model: str = "gemini-pro-1.0",
+        api_key: Optional[str] = None,
+        stop_sequences: list[str] = [],
```
insop, line 39: would `stop_sequences` be used?

dsp/modules/google.py
CShorten: Ah great catch! Sorry about that! Updating!

dsp/modules/google.py
CShorten: Ah another great catch!

dsp/modules/google.py
CShorten: Awesome, good call!

insop:
Thank you so much @CShorten!

It'd be great if you could update this markdown doc as well.
`docs/language_models_client.md`

okhat:
This is fantastic, thank you so much @CShorten and @insop ! I'm excited about Connor's vision for when 1.5 is released!

PR #399, opened 2024-02-15
Added support for variable sized output vectors via Nomic Embed v1.5

AndriyMulyar:
Added support for Nomic Embed v1.5: https://blog.nomic.ai/posts/nomic-embed-matryoshka

This let's you specify a vectorizer that will generate much smaller but still performant embeddings for local use.

### DIFF
--- a/dsp/modules/sentence_vectorizer.py
+++ b/dsp/modules/sentence_vectorizer.py
@@ -3,6 +3,7 @@ from typing import List, Optional
 
 import numpy as np
 import openai
+import nomic
 
 
 class BaseSentenceVectorizer(abc.ABC):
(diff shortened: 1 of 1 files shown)


okhat:
Wonderful, thanks a lot. Merging after checks.

okhat:
Ok we have a small problem:

```python
import nomic
```

I guess normally we deal with this by importing optional dependencies either:
1. Inside the class
2. Outside the class, but in try/except form and on `except` just set it to None or something

AndriyMulyar:
Will update.

arnavsinghvi11:
Hi @AndriyMulyar , just following up on this update. looks good to merge after editing the import. Feel free to use [examples from other retrievers](https://github.com/stanfordnlp/dspy/blob/44f03ce2f92b663de6b8ae3437ebf86998e1e79e/dspy/retrieve/neo4j_rm.py#L17) on how to optionally import the local dependencies

PR #400, opened 2024-02-15
Refactored ChromadbRM to use generic EmbeddingFunction

animtel:
I've updated the ChromadbRM retriever to be more flexible and fix a bug:

- **Flexibility with Embeddings**: Removed the direct dependency on OpenAI, allowing for the use of various embedding functions from [Chroma](https://docs.trychroma.com/embeddings). This change makes the retriever more adaptable to different needs.

- **Bug Fix**: Fixed issue #389 where `_get_embeddings` threw a TypeError due to an unexpected keyword argument 'engine'. This ensures smoother functionality.

This revision enhances ChromadbRM by introducing flexibility in embedding selection and resolving a critical error.

### DIFF
--- a/docs/retrieval_models_client.md
+++ b/docs/retrieval_models_client.md
@@ -8,6 +8,7 @@ This documentation provides an overview of the DSPy Retrieval Model Clients.
 | --- | --- |
 | ColBERTv2 | [ColBERTv2 Section](#ColBERTv2) |
 | AzureCognitiveSearch | [AzureCognitiveSearch Section](#AzureCognitiveSearch) |
+| ChromadbRM | [ChromadbRM Section](#ChromadbRM) |
 
 ## ColBERTv2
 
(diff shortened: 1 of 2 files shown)


insop:
@animtel, thank you for the change.

Would you mind adding an example of using OpenAI embedding? This will help the existing code to migrate to the newer code. The example could go to this file: `docs/retrieval_models_client.md`.

insop:
I believe the `model_arg` was added originally to support Azure OpenAI API.

animtel:
Yes, `model_arg` was specifically introduced to maintain compatibility with the Azure OpenAI API. However, the current implementation supports OpenAI embeddings. You can find the details on how to define an embedding function [here](https://docs.trychroma.com/embeddings/openai). I've also added the documentation to showcase how this can be used.

insop:
> Yes, `model_arg` was specifically introduced to maintain compatibility with the Azure OpenAI API. However, the current implementation supports OpenAI embeddings. You can find the details on how to define an embedding function [here](https://docs.trychroma.com/embeddings/openai). I've also added the documentation to showcase how this can be used.

Excellent, thank you so much for making the change and adding the documentation.

Looks great to me.

PR #401, opened 2024-02-16
[Draft] feat(indexify): Init

mohit-raghavendra:


diptanu:
@mohit-raghavendra Might be worth making a PR on tensorlakeai/indexify-python-client first, make it stable and then send a PR here.

okhat:
Ok will wait for this. Thanks a lot @mohit-raghavendra and @diptanu !

PR #402, opened 2024-02-16
chore: Bump Qdrant dependencies

Anush008:
Updates the dependencies of the `qdrant` extra to latest.


okhat:
Ah great, maybe we can even just remove the version?

Anush008:
Qdrant versions are usually backward compatible. So yes.

PR #403, opened 2024-02-16
OpenAI v1 Migration

### LINKED ISSUE #377: Using DSPy with Azure OpenAI fails (when `openai >= 1.0.0`)
When I execute this code without `api_base`

```python
from dotenv import dotenv_values, load_dotenv

load_dotenv()

config=dotenv_values(".env")
azure_endpoint = config['AZURE_OPENAI_ENDPOINT']
openai_api_key = config['AZURE_OPENAI_API_KEY']
openai_api_version = config['AZURE_OPENAI_API_VERSION']
model_name=config['MODEL_NAME']
embedding_model=config["EMBEDDING_MODEL"]

import dspy

turbo = dspy.OpenAI(deployment_id=model_name,
                    api_key=openai_api_key,
                    # api_base=azure_endpoint,
                    api_provider= "azure",
                    model_type="chat",
                    api_version=openai_api_version)

dspy.settings.configure(lm=turbo, trace=["Test"])
```

I get this error `AssertionError: Must specify api_base for Azure API`.

When I set it up `api_base=azure_endpoint`, the error disappears.

Now, when I call the LM:

```python
    """Answer questions with short factoid answers."""
    question = dspy.InputField()
    answer = dspy.OutputField(desc="in JSON format: {'color': 'xxx'}")

generate_answer = dspy.ChainOfThoughtWithHint(BasicQA)

react_module = dspy.ReAct(BasicQA)

question='What is the color of the sky?'
hint = "It's what you often see during a sunny day."
pred = generate_answer(question=question, hint=hint)

print(f"Question: {question}")
print(f"Predicted Answer: {pred.answer}")
```

I get this error `ValueError: base_url and azure_endpoint are mutually exclusive` 

The reason is because DSPy configures the LM it this way:

https://github.com/stanfordnlp/dspy/blob/2d8db68c7258eda8db15a1b3cfa15ec3a15eb587/dsp/modules/gpt3.py#L95

To sum up, this is the behaviour of `openai` regarding `base_url`:

https://github.com/openai/openai-python/blob/7f9e85017a0959e3ba07834880d92c748f8f67ab/src/openai/lib/azure.py#L197

```python
if base_url is None:
            if azure_endpoint is None:
                azure_endpoint = os.environ.get("AZURE_OPENAI_ENDPOINT")

            if azure_endpoint is None:
                raise ValueError(
                    "Must provide one of the `base_url` or `azure_endpoint` arguments, or the `AZURE_OPENAI_ENDPOINT` environment variable"
                )

            if azure_deployment is not None:
                base_url = f"{azure_endpoint}/openai/deployments/{azure_deployment}"
            else:
                base_url = f"{azure_endpoint}/openai"
        else:
            if azure_endpoint is not None:
                raise ValueError("base_url and azure_endpoint are mutually exclusive")
```

I'm just thinking why DSPy is forcing the library consumer to use `base_url` instead of delegating this to `openai`; in my use case, I'm stuck/blocked..

KCaverly:
Migrated OpenAI GPT3 LM client to v1's OpenAI/AzureOpenAI clients.
The previous version, had kwarg inconsistencies with Azure calls.

Fixes #377

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -19,7 +19,7 @@ from dsp.modules.cache_utils import CacheMemory, NotebookCacheMemory, cache_turn
 from dsp.modules.lm import LM
 
 try:
-    OPENAI_LEGACY = int(openai.version.__version__[0]) == 0
+    OPENAI_LEGACY = int(openai.__version__[0]) == 0
 except Exception:
     OPENAI_LEGACY = True
 
@@ -44,9 +44,6 @@ def backoff_hdlr(details):
         "calling function {target} with kwargs "
         "{kwargs}".format(**details),
     )
-
-
-client = None
 
 
 class GPT3(LM):
@@ -113,14 +110,13 @@ class GPT3(LM):
         )
         self.model_type = model_type if model_type else default_model_type
 
-        if api_provider == "azure" and "model" not in kwargs:
+        if not OPENAI_LEGACY and api_provider == "azure" and "model" not in kwargs:
             if "deployment_id" in kwargs:
                 kwargs["model"] = kwargs["deployment_id"]
                 del kwargs["deployment_id"]
-            elif "engine" in kwargs:
-                kwargs["model"] = kwargs["engine"]
 
-            del kwargs["api_version"]
+            if "api_version" in kwargs:
+                del kwargs["api_version"]
 
         if "model" not in kwargs:
             kwargs["model"] = model
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -1,23 +1,20 @@
 import logging
-from logging.handlers import RotatingFileHandler
 
 # Configure logging
 logging.basicConfig(
     level=logging.INFO,
-    format='%(message)s',
-    handlers=[
-        logging.FileHandler('openai_usage.log')
-    ]
+    format="%(message)s",
+    handlers=[logging.FileHandler("openai_usage.log")],
 )
 
 import functools
 import json
 from typing import Any, Literal, Optional, cast
 
-import dsp
 import backoff
 import openai
 
+import dsp
 from dsp.modules.cache_utils import CacheMemory, NotebookCacheMemory, cache_turn_on
 from dsp.modules.lm import LM
 
@@ -27,9 +24,14 @@ except Exception:
     OPENAI_LEGACY = True
 
 try:
-    from openai.openai_object import OpenAIObject
     import openai.error
-    ERRORS = (openai.error.RateLimitError, openai.error.ServiceUnavailableError, openai.error.APIError)
+    from openai.openai_object import OpenAIObject
+
+    ERRORS = (
+        openai.error.RateLimitError,
+        openai.error.ServiceUnavailableError,
+        openai.error.APIError,
+    )
 except Exception:
     ERRORS = (openai.RateLimitError, openai.APIError)
     OpenAIObject = dict
@@ -40,8 +42,11 @@ def backoff_hdlr(details):
     print(
         "Backing off {wait:0.1f} seconds after {tries} tries "
         "calling function {target} with kwargs "
-        "{kwargs}".format(**details)
+        "{kwargs}".format(**details),
     )
+
+
+client = None
 
 
 class GPT3(LM):
@@ -66,8 +71,40 @@ class GPT3(LM):
     ):
         super().__init__(model)
         self.provider = "openai"
-        openai.api_type = api_provider
 
+        # Assert that all variables are available
+        assert api_key is not None, "Must provide api_key for OpenAI/Azure API"
+        if api_provider == "azure":
+            assert (
+                "engine" in kwargs or "deployment_id" in kwargs
+            ), "Must specify engine or deployment_id for Azure API instead of model."
+            assert "api_version" in kwargs, "Must specify api_version for Azure API"
+            assert api_base is not None, "Must specify api_base for Azure API"
+
+        # Define Client
+        if OPENAI_LEGACY:
+            openai.api_base = api_base
+            openai.api_key = api_key
+            openai.api_type = api_provider
+
+            if api_provider == "azure":
+                openai.api_version = kwargs["api_version"]
+
+            self.client = None
+
+        else:
+            if api_provider == "azure":
+                client = openai.AzureOpenAI(
+                    azure_endpoint=api_base,
+                    api_key=api_key,
+                    api_version=kwargs["api_version"],
+                )
+            else:
+                client = openai.OpenAI(api_key=api_key, base_url=api_base)
+
+            self.client = client
+
+        # Define model type
         default_model_type = (
             "chat"
             if ("gpt-3.5" in model or "turbo" in model or "gpt-4" in model)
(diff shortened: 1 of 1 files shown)


dsp/modules/gpt3.py
```diff
@@ -1,35 +1,37 @@
...
 import backoff
 import openai
 
+import dsp
 from dsp.modules.cache_utils import CacheMemory, NotebookCacheMemory, cache_turn_on
 from dsp.modules.lm import LM
 
 try:
-    OPENAI_LEGACY = int(openai.version.__version__[0]) == 0
+    OPENAI_LEGACY = int(openai.__version__[0]) == 0
```
okhat, line 22: What's the goal of this change? Does it affect behavior in any way?

dsp/modules/gpt3.py
```diff
@@ -66,8 +68,40 @@ def __init__(
     ):
         super().__init__(model)
         self.provider = "openai"
-        openai.api_type = api_provider
 
+        # Assert that all variables are available
+        assert api_key is not None, "Must provide api_key for OpenAI/Azure API"
```
okhat, line 73: The key is not generally necessary, it can be obtained from the ENV variables

dsp/modules/gpt3.py
```diff
@@ -222,57 +250,67 @@ def __call__(
...
 
-@CacheMemory.cache
-def v1_cached_gpt3_request_v2(**kwargs):
-    return openai.completions.create(**kwargs)
 
-@functools.lru_cache(maxsize=None if cache_turn_on else 0)
-@NotebookCacheMemory.cache
-def v1_cached_gpt3_request_v2_wrapped(**kwargs):
-    return v1_cached_gpt3_request_v2(**kwargs)
+def v1_chat_request(client, **kwargs):
```
okhat, line 277: I think I get what you're changing but unfortunately changing this function will break pre-existing caches which we really don't want to do

okhat: Are these changes here necessary?

okhat:
Thank you so much @KCaverly ! I really appreciate your contribution. I left some comments but the overall summary is: can we actually just have an AzureOpenAI client that's separate from the current "GPT3" class? That way there are no breaking changes for non-Azure users.

dsp/modules/gpt3.py
younes-io: it depends on the authentication type used; Azure offers two ways; one using the env variable, and another one which uses rotating keys (using MS Azure services)

younes-io:
> Thank you so much @KCaverly ! I really appreciate your contribution. I left some comments but the overall summary is: can we actually just have an AzureOpenAI client that's separate from the current "GPT3" class? That way there are no breaking changes for non-Azure users.

I agree, we need to separate Azure from the rest; it's safer and also more maintainable

KCaverly:
A separate AzureOpenAI class is much cleaner. 
What I can do is revert the GPT3 changes, pull azure out so GPT3 is exclusively for OpenAI and introduce a new AzureOpenAI LM.

dsp/modules/gpt3.py
KCaverly: Without a global client, these decorator changes may be necessary, as the client itself can not be hashed, but must be used to make the specific call. As we are splitting Azure out, we can just fall back on using the global client and revert these.

younes-io:
> A separate AzureOpenAI class is much cleaner. What I can do is revert the GPT3 changes, pull azure out so GPT3 is exclusively for OpenAI and introduce a new AzureOpenAI LM.

I would say it's better to delegate these LM abstractions to other libraries (that take care of legacy versions as well, like LangChain or LlamaIndex); also, this removes the burden of maintenance from DSPy contributors who should focus on improving DSPy concepts and features, not the "accessories/devices" around it.

KCaverly:
I split out the Azure functionality into its own class.
I agree that it makes more sense to abstract over LMs to some sort of external package.
In the meantime, the concerns above should be resolved with these changes.

The caching methods have not been changed, and this should not be breaking at all for existing non-Azure users.
The only difference is api_provider as an argument is no longer necessary, however, I have left it in, so we can redirect users who may have Azure api calls to the new class.

dsp/modules/azure_openai.py
```diff
@@ -0,0 +1,301 @@
+import logging
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(message)s",
+    handlers=[logging.FileHandler("openai_usage.log")],
```
insop, line 7: Would you keep the file the same as `openai_usage.log` or `azure_openai_usage.log`?

dsp/modules/azure_openai.py
```diff
@@ -0,0 +1,301 @@
...
+        self.history: list[dict[str, Any]] = []
+
+    def _openai_client(self):
+        if OPENAI_LEGACY:
+            return openai
+
+        return self.client
+
+    def log_usage(self, response):
+        """Log the total tokens from the OpenAI API response."""
```
insop, line 134: s/OpenAI API/Azure OpenAI/

insop:
Overall, LGTM. I made minor comments on the PR.

Two questions about this PR and one general question:

1. I am using Azure APIs. If you think it is useful, I'd be happy to test the PR.
2. Would you add the Azure OpenAI entry to this file as well to help users? `docs/language_models_client.md`.

3. We have seen more people support OpenAI compatible APIs, such that OpenAI APIs can be used instead of using different client modules. I am posing a question: are there enough differences between OpenAI and Azure APIs that we need to have a separate module? I have seen a couple of other software libraries have separate modules for Azure APIs and OpenAI APIs, but I am still asking questions.

dsp/modules/azure_openai.py
KCaverly: Updated for azure_openai_usage.log

dsp/modules/azure_openai.py
KCaverly: Updated

KCaverly:
Thanks @insop for the review.

I have updated the docs/and a few of the comments above to ensure consistency.
If you have access to the AzureOpenAI I would appreciate you doing a quick test.

On 3: We walked through why we broke Azure OpenAI out above, but at a high level. Given the new v1 OpenAI api, I don't think Azure can be integrated into the current OpenAI class, without invalidating all existing caches, which is something we definitely do not want to do.

dsp/modules/azure_openai.py
```diff
@@ -0,0 +1,302 @@
...
+
+def v1_chat_request(client, **kwargs):
+    @functools.lru_cache(maxsize=None if cache_turn_on else 0)
+    @NotebookCacheMemory.cache
+    def v1_cached_gpt3_turbo_request_v2_wrapped(**kwargs):
+        @CacheMemory.cache
+        def v1_cached_gpt3_turbo_request_v2(**kwargs):
+            if "stringify_request" in kwargs:
+                kwargs = json.loads(kwargs["stringify_request"])
+            return client.chat.completion.create(**kwargs)
```
bclavie, line 271: Typo: should be chat.completion**s**

bclavie:
Hey, just gave this a quick try using the latest OpenAI azure version. Great work! There are two small issues:

- Had to fix the typo highlighted above (otherwise got an error about `Chat` having no `completion` attribute) but otherwise it went smoothly.
- It's not safe to infer `model_type` based on the model name for azure, since everyone will have different deployment names. Given that the wide majority of use nowadays is for chat model (few people use 3.5-instruct, and `davinci-003` isn't available on Azure), I think it'd make sense to change:

```python
       default_model_type = (
            "chat"
            if ("gpt-3.5" in model or "turbo" in model or "gpt-4" in model)
            and ("instruct" not in model)
            else "text"
        )
```
to just having `model_type` assume the default value of `chat`, with a note in the documentation to manually change it to `text` if using an instruct model?

KCaverly:
Thanks @bclavie for giving it a test, adjusted both the model_type and typo fix.

insop:
@KCaverly 
Thank you for the update, I have not given a try, but since @bclavie has done the test. I will skip my test.

@okhat 
DSPy is used for [this course](https://github.com/cgpotts/cs224u/blob/628628bd330eae3c5d72b103e314ae2c5b837a8d/requirements.txt#L19), I think it would be great if we could consider followings:

- Adding a _git tag_ the repo, so that the repo that uses older OpenAI API can be checked out. So that the course notebook will work on colab environment simply clone the repo with the tag
- Identifying a specific _DSPy version that works with openai older API_ so that it can be used for the pip install? so that the course notebook and the grader still works?

okhat:
Lots of cool stuff here - thank you all. Just to be clear @insop old openai is still fully supported!

See intro.ipynb — it installs openai 0.28.1 and it works with very old caches

okhat:
The main test I need to merge this is to ensure that existing usage of openai is completely unaffected. I am only worried about the azure file using all the same names for functions and then importing with * so I worry about polluting the namespace with the wrong functions. The cache functions are named the same

insop:
> Lots of cool stuff here - thank you all. Just to be clear @insop old openai is still fully supported!
> 
> See intro.ipynb — it installs openai 0.28.1 and it works with very old caches

Sounds good, I just want to make sure to avoid any disturbing changes for the existing users.

That said, for adding a git tag and a description with version number for non-trivial changes are good things for the current users.

KCaverly:
Hey all, is there anything outstanding on this being merged?

okhat:
Hey @KCaverly nothing outstanding, merged :D Thanks a ton!

younes-io:
Thank you for this :D

PR #406, opened 2024-02-18
Added saving/loading instructions to Predict and Chain_Of_Thought

XenonMolecule:
After the optimizers change the instruction we need to be able to save/load the optimized instructions.  Currently the original (unoptimized) instructions are loaded and only the demos are saved.

This is a quick fix that adds this change for Predict and Chain of Thought.  If more modules are added for instruction optimization then this fix will need to be added to those modules as well.

Thanks to @klopsahlong for help with testing/validating the changes!


insop:
LGTM

Issue #397, opened 2024-02-15
Optimization dry runs


okhat:
This may need to be done on a per-optimizer basis but it may be good to think of a dev UX that involves showing an estimation upfront of the number of calls / tokens in an optimization run, and possibly asking for confirmation if `dspy.settings.confirm_first` (this doesn't exist yet) is True or something to that effect.

buzypi:
This would be a wonderful feature. I have 2 things in mind (not sure about how well this can be implemented):

1. A dry run at a module level: `predict(input, dry_run=True)`. We need to have a convention of how we can separate the args from these settings. But it will be good if we don't have to change settings at a global level but rather on a per prediction level, `predict` in `dry_run` mode and then run it without `dry_run` in case we are happy. This is great in a Notebook/REPL environment.
2. A stepper/debugger at a teleprompter level: `BootstrapFewShot(metric=..., step=True)`, which shows progress after each loop and asks if we want to continue. If we don't intend to continue, we can still work with the optimisations that have been obtained until the current cycle.

smwitkowski:
I've started to work on some of these - https://github.com/stanfordnlp/dspy/pull/408

Not going in the same order, but here's what I have so far:

>2. A stepper/debugger at a teleprompter level: BootstrapFewShot(metric=...).compile(..., step=True), which shows progress after each loop and asks if we want to continue. If we don't intend to continue, we can still work with the optimisations that have been obtained until the current cycle.

`dspy/teleprompt/bootstrap.py`
```diff
-   def compile(self, student, *, teacher=None, trainset, valset=None):
+   def compile(self, student, *, teacher=None, trainset, valset=None, step=False):
        self.trainset = trainset
        self.valset = valset
```

```diff
-   def _bootstrap(self, *, max_bootstraps=None):
+   def _bootstrap(self, *, max_bootstraps=None, step=False):
        max_bootstraps = max_bootstraps or self.max_bootstrapped_demos

        bootstrapped = {}
        self.name2traces = {name: [] for name in self.name2predictor}
        for round_idx in range(self.max_rounds):
            for example_idx, example in enumerate(tqdm.tqdm(self.trainset)):
                if len(bootstrapped) >= max_bootstraps:
                    break
                if example_idx not in bootstrapped:
                    success = self._bootstrap_one_example(example, round_idx)

                    if success:
                        bootstrapped[example_idx] = True
+           if step:
+               user_input = input("Continue bootstrapping? (Y/n): ")
[...]

PR #408, opened 2024-02-19
Optimization dry runs [WIP]

smwitkowski:
Adding items discussed in #397

### DIFF
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -63,7 +63,7 @@ def _generate(template: Template, **kwargs) -> Callable:
     generator = dsp.settings.lm
 
     def do_generate(
-        example: Example, stage: str, max_depth: int = 2, original_example=None
+        example: Example, stage: str, max_depth: int = 2, original_example=None, dry_run=False
     ):
         if not dsp.settings.lm:
             raise AssertionError("No LM is loaded.")
@@ -75,6 +75,11 @@ def _generate(template: Template, **kwargs) -> Callable:
 
         # Generate and extract the fields.
         prompt = template(example)
+
+        if dry_run:
+            # In dry run mode, return the prompt instead of calling the LM
+            return prompt, []
+
         completions: list[dict[str, Any]] = generator(prompt, **kwargs)
         completions: list[Example] = [template.extract(example, p) for p in completions]
 
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -1,4 +1,5 @@
 from collections import Counter
+from contextlib import contextmanager
 from typing import Callable, Any, Optional
 
 import dsp
@@ -63,7 +64,7 @@ def _generate(template: Template, **kwargs) -> Callable:
     generator = dsp.settings.lm
 
     def do_generate(
-        example: Example, stage: str, max_depth: int = 2, original_example=None, dry_run=False
+        example: Example, stage: str, max_depth: int = 2, original_example=None
     ):
         if not dsp.settings.lm:
             raise AssertionError("No LM is loaded.")
@@ -75,10 +76,6 @@ def _generate(template: Template, **kwargs) -> Callable:
 
         # Generate and extract the fields.
         prompt = template(example)
-
-        if dry_run:
-            # In dry run mode, return the prompt instead of calling the LM
-            return prompt, []
 
         completions: list[dict[str, Any]] = generator(prompt, **kwargs)
         completions: list[Example] = [template.extract(example, p) for p in completions]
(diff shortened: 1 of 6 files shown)


smwitkowski:
Sounds good, @okhat - there's definitely still work that needs to be done on this, but I wanted to get a start on shaping this.

okhat:
I guess the current dryrun plan seems to be mainly changing the `Predict` module.

But it's unclear to me how that will connect to the bigger picture of optimizers.

Maybe we can have .dryrun(.) method in each optimizer? Or is it better to do .estimate_cost(.) ?

smwitkowski:
@okhat Sounds good - I agree there are a lot of moving pieces, and the link between optimizers and the Predict module could be fleshed out a bit more. 

I'll revisit this and add some pseudocode for us to review.

smwitkowski:
@okhat - Taking a stab at a plan for the optimizer (I'm leaving out `step` here).

TLDR: Create a function or context manager that simulates the LLM output and optimization step to remove any early stopping on a per-optimizer basis. Count the input and output tokens and estimate a "max" number of tokens used. 

### Objective

The aim is to simulate the number of tokens used during optimization without executing actual calls to the Language Model (LM). To do this, we'll: 

1. Mock the output of an LM to bypass actual LM calls, allowing us to estimate the token count without calling any APIs.
2. Prevent early stopping in our optimization routines to ensure we know the most number of tokens to be used, enabling a "max cost" estimate. 

### Implementation Strategy

To achieve this, we'll implement a mock functionality encapsulated within a context manager or function within `teleprompt.py.` This will enable us to:

-   **Mock LM Outputs**: Create a placeholder for LM responses, facilitating the reuse of the optimizers as if they were interacting with a live and functional LM.
-   **Modularize Early Stopping Prevention**: Initially apply a mock code to a set of an optimizer's functions to bypass their early stopping mechanisms.

### Example Workflow

#### Defining Modules and Signatures

First, let's define a module `RAG` that uses a mocked version of `dspy.Module` for demonstration:

```python
class GenerateAnswer(dspy.Signature):
    """Generates short factoid answers to questions based on provided context."""
    context = dspy.InputField(desc="May contain relevant facts")
    question = dspy.InputField()
    answer = dspy.OutputField(desc="Typically between 1 and 5 words")

class RAG(dspy.Module):
    def __init__(self, num_passages=3):
        super().__init__()
        self.retrieve = dspy.Retrieve(k=num_passages)
        self.generate_answer = dspy.ChainOfThought(GenerateAnswer)
    
    def forward(self, question):
        context = self.retrieve(question).passages
        prediction = self.generate_answer(context=context, question=question)
        return dspy.Prediction(context=context, answer=prediction.answer)` 
```
#### Simulating Dry Runs

To simulate a dry run, we'll utilize a context manager `dspy.dry_run()` that mocks the LM interactions:

```python
dataset = HotPotQA(train_seed=1, train_size=20, eval_seed=2023, dev_size=50, test_size=0)

trainset = [x.with_inputs('question') for x in dataset.train]
devset = [x.with_inputs('question') for x in dataset.dev]

def validate_context_and_answer(example, pred, trace=None):
    # Validation logic for context and answer accuracy
    answer_EM = dspy.evaluate.answer_exact_match(example, pred)
    answer_PM = dspy.evaluate.answer_passage_match(example, pred)
    return answer_EM and answer_PM

teleprompter = BootstrapFewShot(metric=validate_context_and_answer)
with dspy.dry_run():
    compile_dryrun = teleprompter.compile(RAG(), trainset=trainset)` 
```

This approach ensures that we mock the LM's responses and bypass the early stopping mechanisms within the optimization routines.

Alternatively, `dry_run` can be created as a class in `teleprompt.py`, and it can be called simple as a method of `teleprompter`, as shown below.

```python
compiled_dryrun = teleprompter.dry_run(RAG(), trainset=trainset)
```

#### Context Manager for Dry Runs

A context manager `dry_run` will facilitate the dry run. This can either (A) be used as a global context manager, or (B) it can be added to `teleprompt.py,` and it would be called a function of the optimizer class.

```python

 @contextmanager
    def dry_run(self):
        original_lm = self.lm  # Save the original LM
        
        try:
            self.lm = mock_llm(original_lm.provider)  # Use a mocked LM
            if self.optimizer_type == 'bootstrap':
                self._mock_bootstrap()
            # Additional conditions for other optimizers
            yield
        finally:
            self.lm = original_lm  # Restore the original LM

```
### Mocking Logic for Optimizers

Each optimizer will have its specific mocking logic to simulate its optimization process without actual LM calls or early stopping. This simulation will estimate the maximum number of tokens and calls expected during the optimization.

```python
	def _mock_bootstrap():
		pass
	# Mocking methods for different optimizers...` 
```
These will need to be built specifically for each optimizer.


### Mocking Logic for LMs

Each LM will have a function that takes an input and returns some output without actually calling an LM.
```python
def mock_llm(llm_provider)
    # Returns a mocked LM based on the provider
    if llm_provider == 'open_ai':
        return mock_open_ai_llm
    # Additional logic for different LLM providers...` 
```

Many details are omitted here, but it gives a high-level idea of how we could approach this.

PR #410, opened 2024-02-19
Add CohereVectorizer for Embed 3

aidangomez:
Adds a new vectorizer for [Cohere's Embed models](https://docs.cohere.com/reference/embed) analogous to OpenAIVectorizer. Details on the performance of the Embed 3 model can be found [here](https://txt.cohere.com/introducing-embed-v3/)

Of note, the Vectorizer interface doesn't specify whether a query versus a document is being embedded, but Cohere's embedding models produce different embeddings depending on the content you want to embed. The different options are in the docs, but taking this into account can product a fairly dramatic boost to performance.

### DIFF
--- a/dsp/modules/sentence_vectorizer.py
+++ b/dsp/modules/sentence_vectorizer.py
@@ -1,7 +1,6 @@
 import abc
 from typing import List, Optional
 
-import cohere
 import numpy as np
 import openai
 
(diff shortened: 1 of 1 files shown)


insop:
LGTM

okhat (commented):
Move `import cohere` inside the constructor to avoid creating a global dependency

okhat:
Hey @aidangomez , thanks so much for this PR! Looks great. I just moved import cohere inside the constructor to avoid breaking when it's not installed

PR #411, opened 2024-02-19
Enable passing a non-default profile to dspy.Bedrock

jabalazs:
For workflows using aws profiles other than the default one.


insop:
LGTM

arnavsinghvi11:
Thanks @jabalazs ! updated the merge conflicts but running into a ruff fix error which should be handled if you run - `ruff check . --fix-only`

jabalazs:
Thank you for handling the merge conflicts @arnavsinghvi11! I just reformatted the code with `ruff` as you suggested.

PR #412, opened 2024-02-19
Add `metric_threshold` argument to BootstrapFewShot

CShorten:
## How it currently works

In line 151 of `BootstrapFewShot` we currently check examples with:

`success = (self.metric is None) or self.metric(example, prediction, trace)`

If success == True, we then wrap the predictor, inputs, and outputs in an Example object and append it to `name2traces`.

This works great if you have a boolean metric, but maybe we want to extend this with a `metric_threshold` argument for float or int metric return types.

## Metric Threshold

- `metric_threshold` is added as default `None`

Then line 151 is replaced with:

```
if self.metric:
  metric_val = self.metric(example, prediction, trace)
  if self.metric_threshold:
    success = metric_val > self.metric_threshold
  else:
    success = metric_val
else:
  success = True
```

We can now retain the float metrics like LLM rating metrics or what have you, but also add a little more strength to find high quality examples in `BootstrapFewShot`.

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -149,8 +149,12 @@ class BootstrapFewShot(Teleprompter):
                     for name, predictor in teacher.named_predictors():
                         predictor.demos = predictor_cache[name]
                 
-                if self.metric and self.metric_threshold:
-                    success = self.metric(example, prediction, trace) > self.metric_threshold
+                if self.metric:
+                    metric_val = self.metric(example, prediction, trace)
+                    if self.metric_threshold:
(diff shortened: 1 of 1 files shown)


insop:
I think this good idea for LM metric.
Is there a way to add to add description as a docstring or related markdown file?

Otherwise, LGTM.

okhat:
Thanks so much @CShorten ! And thanks for the review @insop !

Just converted this from `> threshold` to `>= threshold`... I think that's more intuitive? Unclear but let me know what you think!

CShorten:
Awesome! Oops, yes much better!

PR #418, opened 2024-02-20
Propagate device_map to hf model

stevelaskaridis:
## Description

Currently, when using the HF model backend, the `hf_device_map` parameter may not propagate to the `AutoModel*`. This PR solves that, by passing the `hf_device_map` to the underlying `AutoModel*`.

Ultimately, this allows for splitting models across multiple GPUs/accelerators.


insop:
Thank you, looks good to me.

PR #419, opened 2024-02-20
fix: update imports to allow global dspy.Google

KCaverly:
Currently, the new Google LM class, does not have consistent imports with the remainder of the LMs.

ie. dspy.Cohere works, but dspy.Google does not.

This PR makes small import changes to ensure consistency.

### DIFF
--- a/dsp/modules/__init__.py
+++ b/dsp/modules/__init__.py
@@ -9,6 +9,7 @@ from .pyserini import *
 from .ollama import *
 from .clarifai import *
 from .bedrock import *
+from .google import *
 
 
 from .hf_client import HFClientTGI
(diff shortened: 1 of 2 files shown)


okhat:
LGTM, running checks

dsp/modules/google.py
```diff
@@ -8,7 +8,8 @@
     import google.generativeai as genai
 except ImportError:
     google_api_error = Exception
-    print("Not loading Google because it is not installed.")
+    # print("Not loading Google because it is not installed.")
```
insop, line 11: Why commenting this out?

insop:
One minor question, and looks good to me.
Thank you,

dsp/modules/google.py
KCaverly: If it stays, we print this every time anybody imports dspy, regardless of whether they are importing Google specifically. It's similar to how imports are managed on the Cohere side in: dsp/modules/cohere.py.

I imagine there may be a cleaner way throughout the project to manage optional dependencies.

okhat:
Running some basic checks then merging

okhat:
I wish there was a way to say "please merge but only if the checks pass" instead of just waiting for the checks. Maybe there is.

Issue #390, opened 2024-02-14
[WIP] Major refactor roadmap 


okhat:
DSPy has a small number (maybe 5-6) of extremely powerful concepts that have grown organically over the past year as open source.

Internally, it's time for a major refactor that will simplify things greatly and make sure everything works more smoothly. I have received a lot of interest from the community to contribute to this, so we just need to make sure the goals and milestones are clearly defined.

Potential leaders in this space include @CyrusOfEden @stalkermustang (and from the internal DSPy side possibly @krypticmouse and @arnavsinghvi11 but I haven't checked with them) and there has been a lot of in-depth shots at this from @thomasahle so I'm curious if he's also interested broadly in this.

Starting this issue just to collect the necessary tasks and prioritize them in the right dependency order.

Off the top of my head, I think we have to have:

1. Cleaner LM abstraction that requires a lot less work to maintain and is clearer about the boundaries. The amazing @CyrusOfEden has already defined this on paper. This will include cleaner "Backend" abstraction, which is a bit of a wrapper around LMs that does template adaptation and other things that Cyrus has in mind.

2. Cleaner Signature abstraction. I think the push by @thomasahle here is perfectly on the right track. Types, and immutable signatures. We just need to make more decisions about how far down Pydnatic/Types we go, and how far down, say, SGLang we go, or having our own logic, etc. I do like outsourcing parsing logic and type logic but we need to make sure it doesn't break existing features.

3. Cleaner Modules. This is actually easy but needs to be streamlined. Predict, CoT need to be Modules. And they need to store instructions (not leave it to the signature). They need to handle multiple outputs more consistently. This can be any of us really, esp @arnavsinghvi11, me, @krypticmouse , @thomasahle if any of the folks is interested.

4. Cleaner Optimizers. Well, we can leave this to me and it's a final easy step once signatures and modules are good.

5. More guidance on common patterns in the docs. We now finally have docs for all the key components and we have plenty of individual examples. But we do not have enough guidance on the common e2e workflows. This also partly includes clear guidance on what people should do with local LMs: ollama for cpu, tgi or sglang or vllm(?) for gpus? What about quantization etc.

Thoughts? Other things we need?

1. Production concerns? Streaming. Async.
2. Docstrings.
3. Various individual issues.

Dumping list of stuff:

1. Assertions now has an open issue to collect things needing improvement
2. Don't fail silently if forgot kwarg passed to Predict or pass incorrect name

okhat:
[...]

PR #421, opened 2024-02-20
WIP - refactor LM backend

isaacbmiller:
Start to refactor backend as mentioned in #390

### DIFF
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -1,18 +1,17 @@
-from dsp.modules.hf_client import ChatModuleClient
-from dsp.modules.hf_client import HFServerTGI, HFClientVLLM, HFClientSGLang
-from .signatures import *
+import dsp
+from dsp.modules.hf_client import HFClientVLLM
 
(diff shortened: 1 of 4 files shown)


KCaverly:
If we are going to be abstracting away implementation details for individual providers, do we even need a OpenAI vs Cohere class etc.

Would something like this work, 

```python

class LiteLLM(BaseLM):

    def __init__(self, model: str, **kwargs):
        self.model = model
        self.kwargs = kwargs
    
    def request(self, prompt: str, **kwargs):
        return completion(model=self.model, messages=[{"content": prompt, "role": "user"}], **self.kwargs)

```

It looks like [LiteLLM abstracts all input params](https://docs.litellm.ai/docs/completion/input#translated-openai-params) so all models are basically just a Model ID, and completion params. Would make maintenance pretty easy for the DSPy team, if no provider classes needed maintenance.

isaacbmiller:
See #424 seems like @CyrusOfEden took a better stab at this

PR #424, opened 2024-02-21
Backend Refactor [WIP]

CyrusOfEden:


KCaverly:
Where does the responsibility sit for Parsing the Language Model output to the Signature compliant Structured Output?

It is my understanding that the `BaseBackend` would be compatible for the following:

Given the Signature, Demos, and the Signature Inputs. Generate the Examples, Templates before passing to the Language Model and returning Signature Compliant output.

However, as any given language model can return the following:
- Tool Use/JSON Format
- Raw Completion Strings

Would we expect the `BaseLM` to have individual functions for JSON/String outputs, which would then be called/parsed as necessary by the `BaseBackend`? Or would we expect the responsibility for returning Signature Compliant Signature Output sits entirely at the `BaseLM` level?

CyrusOfEden:
> Where does the responsibility sit for Parsing the Language Model output to the Signature compliant Structured Output?
> 
> It is my understanding that the `BaseBackend` would be compatible for the following:
> 
> Given the Signature, Demos, and the Signature Inputs. Generate the Examples, Templates before passing to the Language Model and returning Signature Compliant output.
> 
> However, as any given language model can return the following:
> 
> * Tool Use/JSON Format
> * Raw Completion Strings
> 
> Would we expect the `BaseLM` to have individual functions for JSON/String outputs, which would then be called/parsed as necessary by the `BaseBackend`? Or would we expect the responsibility for returning Signature Compliant Signature Output sits entirely at the `BaseLM` level?

Not really, in essence —

```python
class ToolBackend(BaseBackend):
  def __call__(...) -> list[dict[str, Any]]:
    return self.lm(..., tools=[...], tool_choice=...)

class JSONBackend(BaseBackend):
  def __call__(...):
    return self.lm(..., response_format={"mode": "json"})

class TemplateBackend(BaseBackend):
  def __call__(...):
    template = ...
    response = self.lm(...)
    results = [template.extract(...) for choice in response.choices]
    return results
```

KCaverly:
One thing we may have to tackle, is Template variations on a potentially per backend basis.

Ie. For the `JSONBackend`, the guidelines has to be augmented to accommodate for json specific instructions.

One option, would be to abstract away Template, have a `PromptTemplate`, which just leverages Text generation, and a `JSONTemplate` which accomodates for JSON specific instructions, and knows how to extract specifically for JSON output.

Currently, I've just got one `Template`, with an argument for `is_json`, but I could see some value for introducing more Templates, would also allow the users to override all template functionality if they want to. You would simply pass a Template to the `Backend.

ammirsm:
I just merged the main into the branch but seems like lots of conflicts, I resolved most of them based on my own knowledge but please take another look at the them.

I have fixed the lint issues but pass the test fixing to you, there should be some new tests that may fail.

KCaverly:
Okay - This guy is good to merge.

### Note on Breaking Changes:

This should be entirely backwards compatible with main. I've run a series of the example notebooks to ensure compatibility, along with passing our entire test suite. All of your existing DSPy programs should continue to operate in the same way. I would test this the same way we have as opposed to building directly with the new backends.

### How the New Backends Work:

The external API for the Backends, looks like the following:

```
class BaseBackend:
    def prepare_request(self, signature: Signature, example: Example, ...) -> dict:
        ...
        
    def generate_request(self, **kwargs) -> t.Any:
        ...
        
    def process_response(self, signature: Signature, example: Example, response: t.Any, ...) -> Completions:
        ...
```

Providing this interface to the user, allows the user to fully override all arguments going to the LM, and allows for dynamic extraction based on the prompt etc. We went through a lot of iterations trying to change this to be more general. Unfortunately the extraction process (generated content -> arguments) is still highly tied to the prompt provided, so we've started with a general API and hopefully something more efficient may emerge as the space evolves. To start, we've provided two backends, a `JSONBackend` and a `TextBackend`. The `TextBackend` operates the same as the current LMs, with a generation prompt and templated extraction. The `JSONBackend` is net new, leveraging the JSON mode inherent in LiteLLM models to provide JSON directly.

The Backend internals also allow for "healing" ie, retrying with partially complete generations, along with caching. 100% of all new backend calls should be cached, without any additional maintenance needed from the team.

### Where should we focus our review:

1. Predict Module
- The way this is built, if you have configured a Backend, Predict will leverage the backend for generation, otherwise it will fall back on the LM.
- When using the LM, the functionality should be no different.
- Additionally, when using the TextBackend and the LM, performance should be basically the same, as the prompt format and generation methods are basically identical.

2. Prediction/Completions Primitives
- While building this out, we found a series of inconsistencies with how Completion objects were constructed. Without diving too much in detail, it was possible for two fields in the same generation to be provided in two seperate "generated" responses.
- All the external APIs should be fairly similar (and have been updated) with tests passing, this was primarily meant to tighten up the internal consistency.
- For optimizers in which the internal Completions state is manipulated (COPRO/MIPRO, Boostrap) we may want an extra pair of eyes. But all my tests look great.

### Overall

This PR is getting Mega, and we've just been spending cycles catching up with main, so I would propose we review/merge, and then continue to harden and tweak once this guy is in.

Thanks all for the help on this - please let me know if we have any questions/concerns.

CyrusOfEden:
I’ve been in the loop with this through dev and it looks good to me. I think the broad brush strokes are there and agree with you in that we should merge to main and then do future PRs to refine. 

My proposal is to operate optimistically here — unless there are any objections from @okhat or @thomasahle within the next 48h then we merge. 

Do we have a version bump / process to publish a new version? @okhat are you the one that manually publishes new versions?

dspy/modeling/backends/json.py
```diff
@@ -0,0 +1,109 @@
...
+    def _guidelines(self, signature: Signature, example: Example) -> str:
+        included_fields = []
+        missing_fields = []
+        for name, field in signature.fields.items():
+            if name in example:
+                included_fields.append((name, field))
+            else:
+                missing_fields.append((name, field))
+
+        result = "Provided the following:"
```
arnavsinghvi11, line 33: this is a bit awkward. Can we rephrase it to what we have as "Given the following fields:"

dspy/modeling/backends/json.py
```diff
@@ -0,0 +1,109 @@
...
+        for _, field in included_fields:
+            result += f"\n{field.json_schema_extra['prefix']} {field.json_schema_extra['desc']}"
+
+        result += "\n\nPlease return the following fields:"
+        for _, field in missing_fields:
+            result += f"\n{field.json_schema_extra['prefix']} {field.json_schema_extra['desc']}"
+
+        result += "\n\nAccording to the following JSON schema:"
```
arnavsinghvi11, line 41: I like the[ typed instruction for JSON objects](https://github.com/stanfordnlp/dspy/blob/2ad84c2cc9fa34e69402bf775e37e049e90ebcd0/dspy/functional/functional.py#L260) better:  `(". Respond with a single JSON object. JSON Schema: " + schema`

dspy/modeling/backends/text.py
```diff
@@ -0,0 +1,226 @@
...
+        "num_retries": 3,
+    }
+
+    model: str
+    default_params: dict[str, t.Any] = Field(default_factory=dict)
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if _missing_litellm:
+            raise ImportError(
```
arnavsinghvi11, line 77: i think we can add this to the try/except block like down for other LMs?

dspy/predict/predict.py
```diff
@@ -16,13 +17,20 @@ def __init__(self, signature, **config):
         self.reset()
 
     def reset(self):
-        self.lm = None
+        if dspy.settings.get("backend", None) is not None:
+            self.backend = None
```
arnavsinghvi11, line 21: is `self.backend` used anywhere?

dspy/primitives/prediction.py
```diff
@@ -1,4 +1,132 @@
...
+
+from dsp.utils import normalize_text
 from dspy.primitives.example import Example
+from dspy.signatures.signature import Signature, SignatureMeta
+
+default_normalize = lambda s: normalize_text(s) or None
+
+
+class Completions(BaseModel):
+    # For some reason Signature is leading to an error here
```
arnavsinghvi11, line 14: what's the error?

dspy/teleprompt/bootstrap.py
```diff
@@ -183,10 +191,18 @@ def _bootstrap_one_example(self, example, round_idx=0):
...
+                    demo = Example(
+                        augmented=True,
+                        dspy_uuid=example.dspy_uuid,
+                        **inputs,
+                        **outputs,
+                    )
                 else:
                     # TODO: FIXME: This is a hack. RandomSearch will complain for now in this edge case.
-                    demo = Example(augmented=True, **inputs, **outputs)
+                    # The predictor, now adds the input into the output Prediction directly
```
arnavsinghvi11, line 202: a bit unclear on this. an example?

pyproject.toml
```diff
@@ -118,7 +118,8 @@ rich = "^13.7.1"
 psycopg2 = { version = "^2.9.9", optional = true }
 pgvector = { version = "^0.2.5", optional = true }
 structlog = "^24.1.0"
-
+snoop = "^0.4.3"
+litellm = "^1.26.3"
```
arnavsinghvi11, line 122: should these be [optional dependencies](https://github.com/stanfordnlp/dspy/blob/2ad84c2cc9fa34e69402bf775e37e049e90ebcd0/pyproject.toml#L42)?

dspy/modeling/backends/text.py
arnavsinghvi11: and mention `pip install dspy-ai[litellm]` since it's an extra

CyrusOfEden: @arnavsinghvi11 @KCaverly we should make litellm a required dep probably

KCaverly:
@okhat @arnavsinghvi11 are we good to merge this one?

dsp/modules/anthropic.py
```diff
@@ -38,12 +38,13 @@ class Claude(LM):
 
     def __init__(
         self,
-        model: str = "claude-3-opus-20240229",
+        model: str = "claude-instant-1.2",
```
okhat, line 41: seems like a bad idea

okhat: though opus is also a bad idea... we should probably not have a default model

dsp/modules/cohere.py
```diff
@@ -108,7 +110,36 @@ def request(self, prompt: str, **kwargs):
     def __call__(
         self,
         prompt: str,
+        only_completed: bool = True,
+        return_sorted: bool = False,
         **kwargs,
     ):
-        response = self.request(prompt, **kwargs)
-        return [response.text]
+        assert only_completed, "for now"
```
okhat, line 117: are these changes part of this refactor? or outdated things? can't tell

dspy/predict/aggregation.py
```diff
@@ -1,54 +1,15 @@
...
-    for completion in completions:
-        if normalize(completion[field]) == majority_value:
-            break
-    
-    # if input_type == Prediction:
-    return Prediction.from_completions([completion], signature=signature)
-
-
+    assert isinstance(prediction, Prediction)
+    return prediction.get_majority(field=field, normalize=normalize)
```
okhat, line 15: Hmm why should the prediction object itself know how to find its own majority? This made sense as a utility.

okhat:
Hey everyone, just catching up on this. Let's prioritize merging very soon (let's say within 2 days).

Biggest problem right now is that I gather (but not sure) that 90% of the files changed are just formatting. Is this accurate? It makes it harder to review this PR. Do we need these formatting changes? Isn't ruff supposed to take that out of the story? Basically I'd love to review a version of this PR that's mostly or entirely actual updates so nothing slips.

KCaverly:
That is correct, we did our best but its messy. Ive outlined the three areas to review above. It should just be the modeling/predict and primitives/prediction modules. Ive extensively tested this and it should be fine. 

I would love to see this merged within 48 hours. Its not perfect, and there are additional issues we can fix on follow up, but existing functionality remains unchanged.

dspy/evaluate/loss.py
```diff
@@ -0,0 +1,60 @@
...
+        if self.is_complete:
+            raise ValueError("Trace is complete and cannot be modified")
+        self.log.append(line)
+
+
+@dataclass
+class Loss:
+    """Expected usage:
+
+    loss_fn = dspy.Loss(...)
```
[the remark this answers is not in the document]
okhat: why is the file called Loss?

dspy/modeling/backends/text.py
okhat: what does litellm involve/add? how heavy?

CyrusOfEden: @okhat it's the backend that lets us have access to all the models from OpenAI, Anthropic, Huggingface, Cohere, etc. without having to build individual adapters for each one

dsp/modules/databricks.py
```diff
@@ -52,23 +54,26 @@ def __init__(
             **kwargs,
         )
 
-        self.kwargs.pop('frequency_penalty', None)
-        self.kwargs.pop('presence_penalty', None)    
+        self.kwargs.pop("frequency_penalty", None)
+        self.kwargs.pop("presence_penalty", None)
```
mikeedjones, line 58: is it worth having a seperate "run ruff on everything" PR? The pre-commit hooks only look at changed files so source which predates ruff is still unformatted.

samiur:
Hi @okhat any update on when this can be merged? We use Instructor for a few things, and I'd like to create an InstructorBackend for ourselves instead of creating a new `TypedInstructorPredictor` and `TypedInstructorChainOfThought`

PR #435, opened 2024-02-23
[Readme Update] Alphabetically sort Vector DBs and add Weaviate to `pip install dspy-ai[weaviate]`

CShorten:
Adds Weaviate to the optional dependencies, and alphabetically sorts Vector DB integrations in the readme and pyproject.toml.


insop:
Markdown update looks good.
Do we manually update `poetry.lock` file? I don't use `poetry` so I am asking.

PR #439, opened 2024-02-23
feat(Evaluate): callable objects can be used as metric for Evaluate

lzjever:
Minor changes to enable the evaluator to accept any callable objects as metrics.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -142,7 +142,7 @@ class Evaluate:
         # Truncate every cell in the DataFrame
         df = df.applymap(truncate_cell)
 
-        # Rename the 'correct' column to the name of the metric function
+        # Rename the 'correct' column to the name of the metric object
         assert(callable(metric))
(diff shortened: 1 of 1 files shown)


dspy/evaluate/evaluate.py
```diff
@@ -141,8 +142,9 @@ def wrapped_program(example_idx, example):
         # Truncate every cell in the DataFrame
         df = df.applymap(truncate_cell)
 
-        # Rename the 'correct' column to the name of the metric function
-        metric_name = metric.__name__
+        # Rename the 'correct' column to the name of the metric object
+        assert(callable(metric))
```
insop, line 146: Is the current `metric` already `callable`?

lzjever: Getting metric.__name__ only works on function. It fails when the metric is an object with __call__ method

lzjever: No. It won't break the existing code, since all current metrics are functions.

okhat:
Nice, this was needed! Thank you @lzjever and thanks for the review @insop !

PR #450, opened 2024-02-25
Save kwargs in HFClientVLLM

vikhyat:
It's not possible to override max_tokens without doing this.


okhat:
Thanks so much @vikhyat  ! I don't understand your comment: It's not possible to override max_tokens without doing this.

Why not?

PR #451, opened 2024-02-25
Updated Pydantic Branch

thomasahle:
This PR contains the new Signature type based on Pydantic.
It also contains 112 unit tests.
Finally it has the module dspy/functional which include the TypedPredictor class, that lets you generate constrained on the types in the signature.


dsp/modules/aws_lm.py
```diff
@@ -28,7 +28,6 @@ def __init__(
         max_new_tokens: int,
```
okhat, line 28: Should this file change? We can remove it from the PR I think?

thomasahle: Must have gotten mixed up in the rebase. Done.

dspy/teleprompt/signature_opt.py
```diff
@@ -48,6 +48,8 @@ class GenerateInstructionGivenAttempts(dspy.Signature):
 
 class SignatureOptimizer(Teleprompter):
     def __init__(self, prompt_model=None, metric=None, breadth=10, depth=3, init_temperature=1.4, verbose=False, track_stats=False):
+        if breadth <= 1:
```
okhat, line 51: It's fine with one, right? Just has to be positive

dspy/teleprompt/signature_opt.py
```diff
@@ -103,12 +107,13 @@ def compile(self, student, *, devset, eval_kwargs):
         for predictor in module.predictors():
             basic_instruction = None
             basic_prefix = None
+            *_, last_key = predictor.extended_signature.fields.keys()
             if (hasattr(predictor, 'extended_signature')):
                 basic_instruction = predictor.extended_signature.instructions
-                basic_prefix = predictor.extended_signature.fields[-1].name
+                basic_prefix = predictor.extended_signature.fields[last_key].json_schema_extra['prefix']
```
okhat, line 113: I'll just record here that I'm still confused about how .name and "prefix" are the same thing and I hope it's not an artifact of tests that are a bit too simple (i.e., tests that don't pass `dspy.OutputField(prefix="...")`)

examples/longformqa/DSPy_LongFormQA_Cache
okhat, line 1: Do we need this btw

okhat: (probably not, I think it's coming from running the notebook for assertions and it downloads this from HuggingFace)

okhat:
Hey @thomasahle , thanks so much for this! Looks ready to merge for me except for the comments above. The comment about prefix is probably ~safe to ignore. The others are probably easy to address. Let me know what you think. I think we can merge right after these, assuming the notebooks don't break.

dspy/teleprompt/signature_opt.py
thomasahle: No, it has to be at least 2. That's what was causing me problems originally, because I set breadth=1 and got weird error messages.

dspy/teleprompt/signature_opt.py
thomasahle: I'm confused about this too. Signatures don't have a "name" though, so I'm not sure what else it should be. If the only difference is a colon, I'm wondering if the LMs or some other code is masking over the difference.

examples/longformqa/DSPy_LongFormQA_Cache
thomasahle: I'm having some trouble deleting this, because it's apparently a submodule rather than a file. Will figure it out.

dspy/teleprompt/signature_opt.py
okhat: I guess last_key is the (proper) name. But actually in the old TemplateV2 world .name is equivalent to .prefix in practice (when separator is empty).

examples/longformqa/DSPy_LongFormQA_Cache
thomasahle: OK, it's gone

PR #455, opened 2024-02-25
Further Update YouRM to return an Iterable of Predictions

mgbvox:
This brings it in line with the behavior of other retriever implementations in `/retrieve`.

Related - we probably want to make dspy.Retrieve a Protocol or AbstractMethod so Pycharm and co can highlight/catch these errors ahead of time.


okhat:
Thanks so much @mgbvox ! I agree we need an abstract interface. #424 is probably a good example (but for LMs)

PR #461, opened 2024-02-26
Improve google LLM support

nbqu:
The previous script is basically borrowed from `cohere.py`,  and I found that the previous script doesn't work properly and should be fixed.
* The Google Python API requires different arguments, so I updated it by referring official document.
* Empirically, some requests may be blocked by safety reasons even if the given prompt is not harmful. So I propose to add default safety options with `BLOCK_ONLY_HIGH`.
* For now Google API seems to disallow multiple generations, so I implemented a loop to generate candidates iteratively.
*  Add Google module to DSP and DSPy package so we can access directly through `dspy.Google`.

Please let me know if you have any concerns or feedback.

Best,


dsp/modules/google.py
```diff
@@ -101,4 +148,52 @@ def __call__(
...
+        n = kwargs.pop("n", 1)
+
+        completions = []
+        for i in range(n):
+            response = self.request(prompt, **kwargs)
+            completions.append(response.parts[0].text)
+
+        return completions
+
+    def inspect_history(self, n: int = 1, skip: int = 0):
```
insop, line 163: Do we need this function instead of the existing one in `lm.py` file?

nbqu: I can see that in `lm.py` the response is handling response based on the `provider`. For now DSPy supports few models but in my opinion this is not scalable when new models are added since each model has its own response format. Modifying `lm.py` to support Google response schema would work fine too, so it can be another options.

nbqu:
https://github.com/stanfordnlp/dspy/pull/419#issue-2144782979 Now I checked my PR is somewhat outdated, should I close this PR?

insop:
I think it'd be great if we reduce duplicated code if possible.

insop (changes requested):
I think it'd be great if we reduce duplicated code if possible.

CShorten:
@nbqu could you please take a look at this - https://github.com/stanfordnlp/dspy/pull/472?

There is a new `.send_message()` syntax in the generativeai API.

CShorten:
Awesome! Worked for me!

CShorten:
@nbqu could you please take a look at this one - https://github.com/stanfordnlp/dspy/issues/462#issuecomment-1966518996?

CShorten:
Thanks @nbqu! https://github.com/stanfordnlp/dspy/issues/462#issuecomment-1967876420. I've increased the max_tries backoff a bit, hopefully something that resolves itself independent of DSPy.

Maybe an idea to add particular handling of network errors, it looks like we only have rate limiting, would be good to add overflow tokens as another example maybe.

Issue #432, opened 2024-02-21
Assert transform module doesn't work?


chengyjonathan:
```
class LongFormQAWithAssertions(dspy.Module):
    def __init__(self, passages_per_hop=3, max_hops=2):
        super().__init__()
        self.generate_query = [dspy.ChainOfThought(QASignatures.GenerateSearchQuery) for _ in range(max_hops)]
        self.retrieve = dspy.Retrieve(k=passages_per_hop)
        self.generate_cited_paragraph = dspy.ChainOfThought(QASignatures.GenerateCitedParagraph)
        self.max_hops = max_hops
    
    def forward(self, question):
        context = []
        for hop in range(self.max_hops):
            query = self.generate_query[hop](context=context, question=question).query
            passages = self.retrieve(query).passages
            context = deduplicate(context + passages)
        pred = self.generate_cited_paragraph(context=context, question=question)
        pred = dspy.Prediction(context=context, paragraph=pred.paragraph)
        dspy.Suggest(citations.citations_check(pred.paragraph), f"Make sure every 1-2 sentences has citations. If any 1-2 sentences lack citations, add them in 'text... [x].' format.", target_module=QASignatures.GenerateCitedParagraph)
        _, unfaithful_outputs = citations.citation_faithfulness(None, pred, None)
        if unfaithful_outputs:
            unfaithful_pairs = [(output['text'], output['context']) for output in unfaithful_outputs]
            for _, context in unfaithful_pairs:
                dspy.Suggest(len(unfaithful_pairs) == 0, f"Make sure your output is based on the following context: '{context}'.", target_module=QASignatures.GenerateCitedParagraph)
        else:
            return pred
        return pred

uncompiled_cited_rag = assertions.assert_transform_module(
        rags.LongFormQAWithAssertions().map_named_predictors(Retry), 
        assertions.backtrack_handler
        )
```

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/joncheng/Desktop/repos/dspy_dnd/scripts/chroma_2_dspyCitedRAG.py", line 61, in <module>
    main()
  File "/Users/joncheng/Desktop/repos/dspy_dnd/scripts/chroma_2_dspyCitedRAG.py", line 34, in main
    pred = uncompiled_cited_rag(question=question)
[...]

PR #467, opened 2024-02-26
Teleprompt Base Class -> ABC, Some Static Typing, and Ruff Fixes/Updates

mgbvox:
Only clear dsp.setting.trace in assertions.py if a trace object exists. 

Adapt YouRM API call signature to match the documentation on https://api.you.com/api-key; update return data to respect self.k (as opposed to returning all snippets from the first K hits).

### DIFF
--- a/dsp/modules/cache_utils.py
+++ b/dsp/modules/cache_utils.py
@@ -6,7 +6,7 @@ from joblib import Memory
 
 from dsp.utils import dotdict
 
-cache_turn_on = not os.environ.get('DSP_CACHEBOOL', 'True').lower() == 'false'
+cache_turn_on = os.environ.get('DSP_CACHEBOOL', 'True').lower() != 'false'
 
 
 def noop_decorator(arg=None, *noop_args, **noop_kwargs):


insop:
It's be great if you add a you.com entry to this document? that will help others who want to use you.com
If you prefer, a separate PR would work for sure.

mgbvox:
@insop do you mean to the documentation? Sure, I can add that. Will get to it this evening.

dspy/primitives/assertions.py
```diff
@@ -237,7 +237,9 @@ def wrapper(*args, **kwargs):
                     break
                 else:
                     try:
-                        dsp.settings.trace.clear()
+                        # only try to clear the trace if one exists
```
okhat, line 240: Overall this makes sense but I worry it's hiding a different issue under the hood?

cc @arnavsinghvi11 @manishshettym @Shangyint

mgbvox:
Had less bandwidth than I anticipated, I plan to sit down and contribute a bunch come early March.

okhat:
Tagging @arnavsinghvi11 for the issue above it comes up a lot: dsp.settings.trace being None by default causes issues. Is setting it globally to [] correct? Probably not? It affects optimization somewhere? Is there any code checking for dspy.settings.trace that assumes it None and [] are different?

arnavsinghvi11:
> dsp.settings.trace

Yeah I believe we should set it to [] going forward since there are [checks in assertion.py](https://github.com/stanfordnlp/dspy/issues/432). It seems that the optimizations logic also set [trace = [] ](https://github.com/stanfordnlp/dspy/blob/b2816c4a35e3144a06752423de1afb8e68e1005f/dspy/teleprompt/bootstrap.py#L133) internally so I believe we can make this change globally in settings. Let me know if that aligns and I can push a PR for it.

PR #458, opened 2024-02-26
ci(dspy): Add tests to CI

isaacbmiller:
1. Create a new action to run the tests that @thomasahle implemented. Fix #456 
2. Fix the pre-commit checks workflow
3. Adjusted the pyproject.toml to have pytest as a dependency.

Note: databricks integration does not pass tests. Commented out as that is the only integration to do so. To be added back at a later date

PR #470, opened 2024-02-26
migrated latest OpenAI version import

arnavsinghvi11:
-moved OpenAI dependencies within local functions to avoid OpenAI version incompatibilities in build: `from openai import OpenAI` is compatible with >1.11.1


isaacbmiller:
This will pass CI once #458 is merged (tested on another branch). LGTM

Issue #463, opened 2024-02-26
Integrate vertexai


Demontego:
Do you plan to integrate Google's VertexAI?
If not, I can add little by little based on this https://github.com/langchain-ai/langchain/blob/master/libs/partners/google-vertexai/langchain_google_vertexai/llms.py

arnavsinghvi11:
@Demontego VertexAI would be great. Feel free to open a PR for it! You'll find this documentation useful for [integrating an LM within DSPy](https://dspy-docs.vercel.app/docs/deep-dive/language_model_clients/custom-lm-client) or following the format of existing LM client integrations.

CShorten:
Hey both, would it be an idea to name this googlevertexai.py? And the LM `GoogleVertexAI`? Inspired by the azurecognitivesearch.py file, but I also see we have hf_client / hf_server.

arnavsinghvi11:
> Hey both, would it be an idea to name this googlevertexai.py? And the LM `GoogleVertexAI`? Inspired by the azurecognitivesearch.py file, but I also see we have hf_client / hf_server.

Yeah that sounds great!

Demontego:
Ok, I will start writing the integration bit by bit in the evenings during my free time.

chaitanya-parsana:
This would be super helpful. @Demontego happy to help test the integration when you have a PR.

Demontego:
https://github.com/stanfordnlp/dspy/pull/616 check please

Issue #462, opened 2024-02-26
'Google' object has no attribute 'co'  and  'AttributeError' object has no attribute 'message'


aparnajore:
After resolving the issue #417, I tried to use Google LM but facing some new issues like
AttributeError: 'Google' object has no attribute 'co'
AttributeError: 'AttributeError' object has no attribute 'message'

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/backoff/_sync.py](https://localhost:8080/#) in retry(*args, **kwargs)
    104             try:
--> 105                 ret = target(*args, **kwargs)
    106             except exception as e:

8 frames
[/usr/local/lib/python3.10/dist-packages/dsp/modules/google.py](https://localhost:8080/#) in request(self, prompt, **kwargs)
     94         """Handles retrieval of completions from Google whilst handling API errors"""
---> 95         return self.basic_request(prompt, **kwargs)
     96 

[/usr/local/lib/python3.10/dist-packages/dsp/modules/google.py](https://localhost:8080/#) in basic_request(self, prompt, **kwargs)
     73         }
---> 74         response = self.co.generate(**kwargs)
     75 

AttributeError: 'Google' object has no attribute 'co'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
[<ipython-input-13-740f17f5850a>](https://localhost:8080/#) in <cell line: 18>()
     16 # Call the predictor on a particular input.
     17 question='What is the color of the sky?'
---> 18 pred = generate_answer(question=question)
     19 
     20 print(f"Question: {question}")

[/usr/local/lib/python3.10/dist-packages/dspy/predict/predict.py](https://localhost:8080/#) in __call__(self, **kwargs)
     47 
     48     def __call__(self, **kwargs):
---> 49         return self.forward(**kwargs)
     50 
     51     def forward(self, **kwargs):

[/usr/local/lib/python3.10/dist-packages/dspy/predict/predict.py](https://localhost:8080/#) in forward(self, **kwargs)
     88 
     89         if self.lm is None:
---> 90             x, C = dsp.generate(template, **config)(x, stage=self.stage)
     91         else:
     92             # Note: query_only=True means the instructions and examples are not included.

[/usr/local/lib/python3.10/dist-packages/dsp/primitives/predict.py](https://localhost:8080/#) in do_generate(example, stage, max_depth, original_example)
     76         # Generate and extract the fields.
[...]

PR #472, opened 2024-02-27
Fix and rename google.py to googlegenerativeai.py

CShorten:
- Solves issue #462 

- Re issue #463 I'm not sure if the vertex API will be more reliable than google.generativeai, so I think it makes sense to rename this one to `googlegenerativeai` rather than just `google`.

- The google.generativeai API has changed significantly since last week. There is a new conv.send_message syntax. I have tested that this works.


CShorten:
Close in favor of https://github.com/stanfordnlp/dspy/pull/461.

PR #478, opened 2024-02-27
fix: default collection embedding function attribute in ChromadbRM constructor

ragul-kachiappan:
The recent PR #449 opened by @smwitkowski  refactored ChromadbRM constructor to use collection default embedding function with the syntax "self._chromadb_collection.embedding_function". But ChromaDB collection object has no such attribute and raises attribute error if no embedding function is passed to the constructor. 

This PR modifies the constructor to use Chromadb collection's protected attribute "self._chromadb_collection._embedding_function" and the module now works as expected.

I could be wrong in my assumptions as I am relatively new to dspy and am open to possible better resolution.

### DIFF
--- a/dspy/retrieve/chromadb_rm.py
+++ b/dspy/retrieve/chromadb_rm.py
@@ -70,11 +70,11 @@ class ChromadbRM(dspy.Retrieve):
         persist_directory: str,
         embedding_function: Optional[
             EmbeddingFunction[Embeddable]
-        ] = None,
+        ] = ef.DefaultEmbeddingFunction(),
         k: int = 7,
     ):
         self._init_chromadb(collection_name, persist_directory)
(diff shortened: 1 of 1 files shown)


smwitkowski:
You're correct, @ragul-kachiappan; I made a typo, so this would throw an error.

Plus, if a collection is being read in from disk then `_embedding_function` is unavailable and the default embedding model. 

I'd recommend rolling back to the version introduced in #400. 

@ragul-kachiappan since you've already opened the PR, could you make this change?

ragul-kachiappan:
As per your suggestion, @smwitkowski , I have rollback the default to generic embedding function introduced in #400.

arnavsinghvi11:
Thanks @ragul-kachiappan @smwitkowski !

Issue #166, opened 2023-10-14
long_text attribute for retriever passages


andreapiso:
Hi, I am creating a custom retriever, inheriting from `dspy.Retrieve` and overloading the `__init__` and `forward` method. 

Right now I am receiving an error on "long_text" not being present in the passages that my retriever is generating:

```
File ~/miniconda3/lib/python3.10/site-packages/dsp/primitives/search.py:10, in <listcomp>(.0)
      8     raise AssertionError("No RM is loaded.")
      9 passages = dsp.settings.rm(query, k=k, **kwargs)
---> 10 passages = [psg.long_text for psg in passages]
     12 if dsp.settings.reranker:
     13     passages_cs_scores = dsp.settings.reranker(query, passages)

AttributeError: 'str' object has no attribute 'long_text'
```

I am confused because I assumed that passages would be a list of strings, but it does not look like it's the case. However, when I look at the pinecone retriever, which is what I am using as a reference to implement mine, It does not look like it is using the "long_text"  field either. 

https://github.com/stanfordnlp/dspy/blob/main/dspy/retrieve/pinecone_rm.py

What am I missing?

wicusverhoef:
I am encountering the same.

jamesliu:
Same here

cyyeh:
@okhat Could you help answer this question? Thanks a lot!

vaaale:
Just had the same issue. This fixed it:

wiki_abstracts = dspy.ColBERTv2(url='http://20.102.90.50:2017/wiki17_abstracts')
dspy.settings.configure(lm=ollama_model, rm=wiki_abstracts)

ac-sagarmathpal:
Facing the same issue

koshyviv:
I was able to resolve by adding 
```
if hasattr(passages, 'passages'):
    passages = passages.passages
```
after this line https://github.com/stanfordnlp/dspy/blob/42a5943379d28d1673dc8fe332a3d596efdfc7a3/dsp/primitives/search.py#L12

It seems that we are getting a prediction object as the return of `passages = dsp.settings.rm(query, k=k, **kwargs)`

jettro:
I think there are a few parts important for this to work:
- dsp.search.retrieve: Calls the RetrieveModel and expects a list of objects with the key 'long_text'
- dspy.retrieve.Retrieve.forward: The return type is Prediction, which contains a key 'passages'
[...]

PR #486, opened 2024-02-28
fix: AttributeError for Non-Prediction Objects in Retriever

koshyviv:
Suggestion/Fix for this issue - https://github.com/stanfordnlp/dspy/issues/166

The retrieve function, seems to get the passages (as a Prediction object) during the use of certain retrievers - this breaks the downstream flow.

I have added a check to determine if its a Prediction object (currently I'm doing it by checking the attribute, but i feel there must be a better way - will check!)

Please let me know if there's something I should consider or have missed out.


arnavsinghvi11:
Hi @koshyviv , thanks for raising this. I think we'll need to account for a better way to handle this in the WIP refactor so feel free to raise this in #390 as needed! closing for now

PR #488, opened 2024-02-28
WIP - Testing on Notebooks - Refactor COT - Remove inheritance from predict and convert it to module.

ammirsm:
What is the change?
- I have converted the COT to use the Predict instead of inheritance of predict and make it as module.

Why the change is important?
- We are in refactor phase that we wanna clean up stuffs and one of the part is to make sure anything is not inheriting from predict.

How to test?
- Run tests in the tests/ folder with pytest.
- Run notebooks to make sure they are fine.


thomasahle:
LGTM

okhat:
Thanks so much @ammirsm   — will this break saved checkpoints

okhat:
Btw for the record, for this change I think the output of the following pieces of code may change, e.g. in ScoNe:


https://github.com/stanfordnlp/dspy/blob/main/examples/nli/scone/scone.ipynb


     cot_fewshot.save("scone-cot_fewshot-turbo-gpt4-demos.json")


    cot_fewshot = ScoNeCoT()
    cot_fewshot.load("scone-cot_fewshot-turbo-gpt4-demos.json")

Nothing special about scone, just basic .save() and .load() after a BootstrapFewShot run of any kind

PR #490, opened 2024-02-28
Cleaned up functional.py

thomasahle:
With new TypedChainOfThought function


tests/functional/test_functional.py
```diff
@@ -215,9 +220,9 @@ def simple_metric(example, prediction, trace=None):
     lm.inspect_history(n=2)
 
     # Check that the compiled student has the correct demos
-    assert len(compiled_student.output.predictor.demos) == 1
-    assert compiled_student.output.predictor.demos[0].input == trainset[0].input
-    assert compiled_student.output.predictor.demos[0].output == trainset[0].output
+    assert len(compiled_student.output.predictor.predictor.demos) == 1
```
okhat, line 223: Thank you so much Thomas! This .predictor.predictor is interesting. Do we maybe want to have .demos be a getter? feels like there's something to simplify here or a breaking change being made?

thomasahle: This just happened because I wrapped the functional style predictor/cot's inside another module. (They only have one output, so I strip the Prediction object.)
I think this is natural in DSPy to have many nested levels of Modules, no?
Probably the confusion is that I'm calling everything predictor instead of `inner` or `inner_module` or something like that.

thomasahle: It's just this `_StripOutput` class that's adding an extra layer of indirection.
And my unit tests that are "hacky" by trying to access the demos directly instead of calling `.predictors()` or something like that.

```python
def predictor(func):
    signature = _func_to_signature(func)
    *_, output_key = signature.output_fields.keys()
    return _StripOutput(TypedPredictor(signature), output_key)

def cot(func):
    signature = _func_to_signature(func)
    *_, output_key = signature.output_fields.keys()
    return _StripOutput(TypedChainOfThought(signature), output_key)

class _StripOutput(dspy.Module):
    def __init__(self, predictor, output_key):
        super().__init__()
        self.predictor = predictor
        self.output_key = output_key
    
    def copy(self):
        return _StripOutput(self.predictor.copy(), self.output_key)

    def forward(self, **kwargs):
        prediction = self.predictor(**kwargs)
        return prediction[self.output_key]
```

okhat: Ah got it, so this test doesn't actually affect other parts of the library, I get it now.

OK then all good. But down the line this feels like a pattern we want to clean up? Or maybe it's an irreducible complexity

okhat: Basically feel free to merge but I think the problem is going to pypi may cause some problem according to Amir's tests on main <-> notebook cache

thomasahle:
Now the test is just
```python
    demos = compiled_student.predictors()[0].demos
    assert len(demos) == 1
    assert demos[0].input == trainset[0].input
    assert demos[0].output == trainset[0].output
```

tests/functional/test_functional.py
thomasahle: I cleaned up the test, see below.
What does "going to pypi" mean?

okhat:
Basically releasing to pypi may be challenging for unclear reasons, assuming the report by Amir is not a false negative. He is getting cache errors on intro notebook. But if your notebook test runs fine then the issue is probably on his end. Maybe he has the wrong openai version. I think that’s it with 85% confidence

thomasahle:
I don't think this should affect any of what Amir is doing. It's mostly just cleaning up `functional.py` and fixing a few functions in signature where it was throwing away the types of things.

okhat:
There are other changes in main that presumably (but I doubt it) will cause issues when all of main is published on pypi — this is independent of this merge, but is kind of a prerequisite for pypi release

okhat:
But one thing at a time: happy to merge when you’re ready (or feel free to do it when ready). Thanks a lot Thomas!

Issue #428, opened 2024-02-21
After saving and loading, the program optimized by BayesianSignatureOptimizer encounters an error when calling forward


lzjever:
code as follows:
```
def train():
    teleprompter = BayesianSignatureOptimizer(prompt_model=sql_llm, task_model=sql_llm, metric=assess_api_selector, n=10, init_temperature=0.9, verbose=True,track_stats=True)
    evaluate_api_selector = Evaluate(devset=dataset_api_calling_gt_sample[10:20], metric=assess_api_selector, num_threads=1, display_progress=True, display_table=0)


    kwargs = dict(num_threads=1, display_progress=True, display_table=0)
    uncompiled_prompt = APISelector()
    #evaluate_api_selector(uncompiled_prompt)
    compiled_prompt_opt = teleprompter.compile(uncompiled_prompt, devset=dataset_api_calling_gt_sample[:10], optuna_trials_num=40, max_bootstrapped_demos=3, max_labeled_demos=5, eval_kwargs=kwargs)
    compiled_prompt_opt.save('0221_2_api_selector_expert_trainer.json')

def test():
    dataset_api_calling_gt_sample = [dspy.Example(question=x["question"],selected_api=x["api"]).with_inputs('question') for ind, x in result.iterrows()]
    compiled_prompt_opt = APISelector()
    compiled_prompt_opt.load('0221_2_api_selector_expert_trainer.json')
    print(compiled_prompt_opt)
    #compiled_prompt_opt.forward('What's the time?')
    evaluate_api_selector = Evaluate(devset=dataset_api_calling_gt_sample[20:30] , metric=assess_api_selector, num_threads=1, display_progress=True, display_table=0)
    evaluate_api_selector(compiled_prompt_opt)


train()
test()
```

The error occurs here:
```
Traceback (most recent call last):
  File "/home/percy/works/hongyou/walagi-offline/api_selector_expert_trainer.py", line 134, in <module>
    test()
  File "/home/percy/works/hongyou/walagi-offline/api_selector_expert_trainer.py", line 126, in test
    compiled_prompt_opt.forward('What's the time?')
  File "/home/percy/works/hongyou/walagi-offline/api_selector_expert_trainer.py", line 78, in forward
    prediction = self.generate_answer(question=question,api_inventory = self.api_inventory_text)
[...]

PR #500, opened 2024-02-29
Update template_v2.py to fix issue #428

drawal1:
Fix crash that occurs when 'augmented' key does not exist. The error is AttributeError: 'dict' object has no attribute 'augmented'


drawal1:
superseded by pull request #501

PR #503, opened 2024-02-29
Fix: Add logging path modification for OpenAI usage

neil-sola:
This is to ensure that extraneous files are created in specified folder.

Refer to #211

### DIFF
--- a/dsp/modules/azure_openai.py
+++ b/dsp/modules/azure_openai.py
@@ -1,12 +1,5 @@
 import logging
 
-# Configure logging
-logging.basicConfig(
-    level=logging.INFO,
-    format="%(message)s",
-    handlers=[logging.FileHandler("azure_openai_usage.log")],
-)
-
 import functools
 import json
 from typing import Any, Literal, Optional, cast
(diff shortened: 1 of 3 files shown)


neil-sola:
Hmm, I see. What are your opinions of using an env_var here: either to prefix the logging files to a specific folder, or to enable/disable?

I was thinking maybe something like `DSPY_LOGGING_DIR` to prefix a folder to save logs, or `DSPY_LOGGING` to enable/disable logging.

neil-sola:
@okhat I updated the PR to focus on just giving the user the ability to specify a folder to save logs to — the mechanism to allow enabling/disabling of logging seems more complicated, but this should provide some level on configurability on where logs are stored.

KCaverly:
I've added an initial global logger, with configuration here: #684.
Does the API offered in that PR, cover all the uses outlined here?

neild0:
> I've added an initial global logger, with configuration here: #684. Does the API offered in that PR, cover all the uses outlined here?

Yup, looks perfect — thanks so much!

PR #515, opened 2024-03-01
Reformatted all the code with ruff

thomasahle:
As per our discussions, I ran `ruff . --fix` on the entire code base.
I had to change a few things to get the tests to still run:
```
[tool.ruff.lint.per-file-ignores]
"**/{tests,docs}/*" = ["ALL"]
"**__init__.py" = ["F401"]
```
The first line just disables linting/error fixing on the tests.
This was necessary, because some tests do stuff like importing `List` rather than using `list`, because it wants to be backwards compatible.
The line `"**__init__.py" = ["F401"]` prevents `ruff` from removing unused imports in `__init__.py` files, which are used to make those modules publicly available.

### DIFF
--- a/dsp/evaluation/utils.py
+++ b/dsp/evaluation/utils.py
@@ -7,7 +7,7 @@ try:
     from IPython.display import display as ipython_display
 except ImportError:
     ipython_display = print
-from dsp.utils import EM, F1, HotPotF1
+from dsp.utils import EM
 
 
 def evaluateRetrieval(fn, dev, metric=None):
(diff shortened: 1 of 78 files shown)


okhat:
Happy to merge this. Just wanna do:

1. Push current to pypi first so there's a safe version up there.
2. Merge this. Then push again to pypi

thomasahle:
Ok, I've checked all the notebooks as discussed in the Discord.
Merge at your leisure.

PR #529, opened 2024-03-02
Feature data generation

fsndzomga:
This commit introduces the `synthetic_data_generation` function in the utils. The function dynamically generates synthetic data based on a given Pydantic schema class and sample size. The function constructs a custom dspy.Signature class, leverages dspy.Predict for data generation, and organizes the generated data into examples using dspy.Example. This enhancement streamlines the process of creating synthetic datasets for testing and development purposes.

Usage:
```python

# Generating synthetic data via a pydantic model
generator = SyntheticDataGenerator(schema_class=SyntheticFacts)
examples = generator.generate(sample_size=6)

# # Generating synthetic data via existing examples
generator = SyntheticDataGenerator(examples=existing_examples)
examples = generator.generate(sample_size=5)

```


okhat:
What is the main difference between your approach @fsndzomga and the link?

okhat:
We can probably merge the two efforts in some way!

fsndzomga:
I just finished reviewing the branch you sent to me.
krypticmouse approach is to generate synthetic data based on sample input data , so you still need some initial input data
my approach is different. You don't need any data upfront, you just define a pydantic schema and synthetic data generation is handled for you
Maybe the too approaches can coexist ?

fsndzomga:
I see the synthetic_data_generation more like a utils function. But I can move it to the datasets module if necessary.

fsndzomga:
I merged the two approaches @okhat and @krypticmouse ! Let me know what you think. I also created an experimental module as discussed.

krypticmouse:
@fsndzomga The implementation looks good! Thanks for contributing!

PR #530, opened 2024-03-02
Synthetic Data Generation

krypticmouse:
`Synthesizer` is a synthetic data generation utility that generates synthetic data based on sample input data you provide to it. All you need to pass is a few examples and `num_data` i.e. no. of synthetic datapoints you want to generate and it'll take care of it.

**Usage:** 
```python
from dspy.datasets import DataLoader, Synthesizer

dl = DataLoader()

data = dl.from_huggingface(
    "gsm8k", "main",
    fields=("question", "answer"),
    input_keys=("question",),
    split="train[:3]"
)

synthesizer = Synthesizer()

syn_data = synthesizer.generate(examples=data, num_data=10)

synthesizer.export(data=syn_data, path="syn_data_gpt_4.csv")
```


kostum123:
Can we specify the language of the data generated? My examples are in English, yet I want to generate data in another language. I think this is possible, but I am new to DSPY, as I discovered it today. How can we do this?

krypticmouse:
@kostum123 good idea I can add a language attribute to the  generator pretty sure it'll do the job.

kostum123:
> @kostum123 good idea I can add a language attribute to the generator pretty sure it'll do the job.

Awesome!

PR #532, opened 2024-03-02
Update README.md

thomasahle:
What do you think about having a types section in the examples like this?
I can expand it more, or make it shorter.


okhat:
Nice! Thanks so much. Merged.

PR #538, opened 2024-03-03
ci(dspy): Add Ruff linting workflow

isaacbmiller:
Makes CI better:
1. Caches poetry installation
2. Caches python install dependencies
3. Auto commit style fixes (That commit wont be tested, but the tests get run post-lint, so if it breaks something the tests will fail)


isaacbmiller:
@CyrusOfEden

.github/workflows/run_tests.yml
```diff
@@ -10,38 +10,78 @@ env:
   POETRY_VERSION: "1.6.1"
 
 jobs:
+  lint:
+    name: Lint
```
CyrusOfEden, line 14: Is this linting or because of --fix-only fixing? If it's fix, we should name it fix.

isaacbmiller: Ideally, it would Lint + auto fix and error if there are any lint errors, but right now there are 1600 ruff errors that arent auto-fixable. 

Maybe can add a check to only error on modified files, so the community will gradually refactor the modules in order to get their PR merged?

CyrusOfEden: Ideally yeah, let's call it 'Fix' for now and then circle back on the error check on modified files

CyrusOfEden (approved):

PR #544, opened 2024-03-04
Add `wandb` logging to `BootstrapFewShot`'s `metric_val`

CShorten:
Weights & Biases (wandb) are one of the industry leaders in experiment tracking software for Machine Learning.

I think `runs` and `sweeps` both offer significant value to monitoring DSPy optimization runs.

This PR will add wandb logging to `BootstrapFewShot` to log the values of `metric_val` as the teacher model is bootstrapping traces through the program.

Intended use:

```
# connect to wandb project
wandb.init(
  project: "my-dspy-program",
  config = {
    "program": "Perspective-Guided-STORM",
    "teacherLM": "GPT4"
  }
)

# wandb logging is internal to the `teleprompter.compile` call
# ^ you just add values to the project with wandb.log(...)

compiled_progarm = teleprompter.compile(PG_STORM, trainset=trainset, ...)
```

I think this will be especially effective when say looping through program configurations (in terms of multi-model design or the composition of modules) or teacherLMs.

As a result, DSPy users will have a nice visualization of how effectively the teacher LM is able to bootstrap traces through the program that surpass the `metric_threshold`.

I think this is a nice building block, but generally the wandb integration will be more effective with higher-level optimizers such as `BootstrapFewShotRandomSearch`, `BootstrapFewShotWithOptuna`, or the `BayesianSignatureOptimizer`.

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -49,7 +49,7 @@ class BootstrapFewShot(Teleprompter):
 
         self._prepare_student_and_teacher(student, teacher)
         self._prepare_predictor_mappings()
-        self._bootstrap(wandb_enabled)
+        self._bootstrap(wandb_enabled=wandb_enabled)
 
         self.student = self._train()
         self.student._compiled = True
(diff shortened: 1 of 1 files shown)


isaacbmiller:
Maybe better to implement this in config so we don't have to pass it as an individual parameter to each class and child?

CShorten:
Just in case anyone had seen this --> moved here: https://github.com/stanfordnlp/dspy/pull/849

PR #552, opened 2024-03-04
Introduce dspy.Template

KCaverly:
The original implementation of `TemplateBackend` relied upon the `signature_to_template` function. This would take in a `dspy.Signature` and return a `dsp.Template`. However, we would pass `dspy.Example` into the `dsp.Template` and the extract function would return `dsp.Example` a different type. Leaving us with two different types of `Example` within a single prediction. To combat this, a new `Template` primitive was introduced into `dspy` to mirror and replace the original `dsp.Template`.

I do not expect this current PR to achieve feature parity with the original class, but initial tests have been introduced to mock and mirror the current prompt structure used in `dsp`.

Let me know what you think @CyrusOfEden, the last piece I believe to start using the `TemplateBackend` is to fully explore the `do_generate` function dealing with incomplete or inaccurate completions.

### DIFF
--- a/dspy/backends/template.py
+++ b/dspy/backends/template.py
@@ -40,9 +40,6 @@ class TemplateBackend(BaseBackend):
         pred = self.lm(template(example), **kwargs)
 
         # This returns a list of Examples
-        for prediction in pred:
-            print(prediction.message.content)
-
         extracted = [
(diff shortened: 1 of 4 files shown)


CyrusOfEden:
YES man this is 🔥

PR #553, opened 2024-03-05
docs(dspy): add toy invocation of model to minimal-example

nattaylor:
The new docs are awesome, and this is a 1 line change that I propose to improve `minimal-example`


arnavsinghvi11:
Thanks @nattaylor !

PR #559, opened 2024-03-05
add 'k' as argument to FaissRM.forward()

software-artisan:
add 'k' as argument to FaissRM.forward()

### DIFF
--- a/dspy/retrieve/faiss_rm.py
+++ b/dspy/retrieve/faiss_rm.py
@@ -3,7 +3,7 @@ Author: Jagane Sundar: https://github.com/jagane.
 """
 
 import logging
-from typing import Union
+from typing import Union, Optional
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
LGTM! tagging @isaacbmiller as it seems like the ruff fix workflow is erroring out. Is [this](https://github.com/stanfordnlp/dspy/actions/runs/8156255381/job/22307837624?pr=559#step:5:124) something on the workflow end? or does it need to be handled within the PR?

isaacbmiller:
If you run `ruff check --fix-only`, then commit, it will pass CI.

isaacbmiller:
LGTM after that

arnavsinghvi11:
> If you run `ruff check --fix-only`, then commit, it will pass CI.

@software-artisan

PR #560, opened 2024-03-05
Fix `long_text` key problem with RMs

CShorten:
Draft, started with changing

`dsp.retrieveEnsemble` to `dsp.settings.rm`

in calls to dspy.Retrieve

=== TODO ==

I now need to remove the long_text key from all the RMs

=== RELATED PROJECT TODO (interface metadata with retrieval) ===

I then need to see how the [1], [2] is formatted in the Signatures when passing in the output from Retrieval. I think this should be self-contained in the Signatures.

### DIFF
--- a/dspy/retrieve/retrieve.py
+++ b/dspy/retrieve/retrieve.py
@@ -36,7 +36,8 @@ class Retrieve(Parameter):
         # print(queries)
         # TODO: Consider removing any quote-like markers that surround the query too.
         k = k if k is not None else self.k
-        passages = dsp.retrieveEnsemble(queries, k=k)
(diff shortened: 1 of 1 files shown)


koshyviv:
We're seeing that issue, even with RMs - Issue https://github.com/stanfordnlp/dspy/issues/166

If it helps - the current workaround i am using is https://github.com/stanfordnlp/dspy/pull/486 - but I'm not sure if its the best way forward.

Athe-kunal:
@CShorten 
Sorry, I was not able to comprehend the issue here. Can you please elaborate more on it?

PR #564, opened 2024-03-05
bug(dspy): Fix CI checking out wrong repo

isaacbmiller:


dspy/evaluate/evaluate.py
```diff
@@ -158,25 +183,25 @@ def wrapped_program(example_idx, example):
...
                     margin: 10px 0;'>
                     ... {truncated_rows} more rows not displayed ...
                 </div>
                 """
                 ipython_display(HTML(message))
-                
+
         if return_all_scores and return_outputs:
-            return round(100 * ncorrect / ntotal, 2), results
+            return round(100 * ncorrect / ntotal, 2), results, [score for *_, score in reordered_devset]
```
isaacbmiller, line 204: This is the actual change

arnavsinghvi11:
Thanks @isaacbmiller ! Not sure if I've missed this but is there some documentation on how to run these checks prior to opening PRs? It would be great to have this as it seems like some of the open PRs are failing these checks post-push. Thanks!

isaacbmiller:
> Thanks @isaacbmiller ! Not sure if I've missed this but is there some documentation on how to run these checks prior to opening PRs? It would be great to have this as it seems like some of the open PRs are failing these checks post-push. Thanks!

Rather than try to get the  pre-commit to work, I think I am just going to get a bot to comment that says to run `ruff check . --fix`

okhat:
Thanks Isaac. Since these CI pull requests were merged, the checks have strange names and they generally fail and/or take longer than usual. Is everything correct?

isaacbmiller:
In main, merging from non-forks should work correctly still. I was trying to get the PR comment bot to work correctly, but that seems harder than is worth it, and I am going to make a minimal fix

PR #565, opened 2024-03-05
ci(dspy): Fix failing tests coming from forks

isaacbmiller:
This PR does two main things:
1. It fixes an issue where the code would try to checkout code from a repo it did not have access to. The way I wrote it to checkout ```{{ github.head_ref}}``` was incorrect. This should resolve the issues that CI tests coming from forks face.
2. Starts to add a workflow to upload an artifact to add a comment to run the style checks


isaacbmiller:
@okhat This should address everyones failing tests. The first runs may be longer as the cache rebuilds

PR #567, opened 2024-03-06
DSPy Only Predict

KCaverly:
This PR is intended to get a "DSPy Only Predict" running, ie. No `dsp` code running. 
This is ultimately the point at which the Backend refactor becomes breaking, 

### 1. Move over `dsp`'s predict primitives:
- [x] Greedy & Partial decoding for Templates, allowing the `dspy.Template` to extract each field regardless of whether all fields were generated successfully.
- [x] Recursive "healing" for Template based generation calls. Ie. Some generate calls, often do not include all fields. 
- [x] Tests for BackendTemplates with partially completed calls.
- [x] Ensure parameters for `max_tokens`, `temperature` and recursive workflow `max`_retries` are respected and exposed appropriately.

### 2. Introduce new `DummyLanguageModel` for use with the `BaseLM`.
- [x] Build backend agnostic Dummy Language Model.
- [x] Add tests for `TemplateBackend` generate calls without recovery.
- [x] Add tests for `TemplateBackend` generate calls with recovery.

### 3. Update `Predict`
- [x] Remove all stale attributes (lm)
- [x] Transition away from `dsp.Example` and `dsp.Template` to `dspy.Example` and `dspy.Template`.
- [x] Generate new predictions for Backend infrastructure.
- [x] Expand `Predict` tests to ensure functionality remains consistent.

#### 1. Do we expect the 'non-Template' backends to use recursive healing functionality.
 I imagine, non-Templates would not need this functionality. I have currently included it within the `BaseBackend` but it could be pulled out and included within the `TemplateBackend` only.

#### 2. Load/Dump State for the current `dspy.Predict` class will be broken with this change.
Ultimately, starting this integration across the different module classes is going to be breaking. Are there specific concerns we have regarding this, how do we want to manage this? Migration details etc. Also is it worth merging the backend implementations earlier than transitioning the Module classes?

Let me know what you think? @CyrusOfEden @okhat.

### DIFF
--- a/dspy/backends/base.py
+++ b/dspy/backends/base.py
@@ -29,14 +29,10 @@ class BaseBackend(BaseModel, ABC):
             completions = self.generate(signature, **kwargs)
 
             # If 1 or more complete generations exist, simple return all complete
-            if len(completions) > 0:
+            if completions.has_complete_example():
                 break
 
-            # Try again
-            max_example = completions[0].example
-            for completion in completions.completions:
-                if len(max_example) < len(completion.example):
-                    max_example = completion.example
+            max_example = completions.get_farthest_example()
 
             for field in signature.fields:
                 if field in max_example:
@@ -52,6 +48,7 @@ class BaseBackend(BaseModel, ABC):
             i += 1
 
         assert completions is not None
+        completions.remove_incomplete()
         if len(completions) == 0:
             raise Exception(
                 "Generation failed, recursively attempts to complete did not succeed."
--- a/dspy/backends/base.py
+++ b/dspy/backends/base.py
@@ -10,13 +10,17 @@ class BaseBackend(BaseModel, ABC):
     """A backend takes a signature, its params, and returns a list of structured predictions."""
 
     history: list[Completions] = Field(default_factory=list)
+    attempts: int = Field(default=1)
 
     def __call__(
         self,
         signature: Signature,
-        attempts: int = 5,
+        attempts: int = 1,
         **kwargs,
     ) -> Completions:
+        # Allow overriding the attempts at the Backend Initialization Step
+        attempts = max(attempts, self.attempts)
+
         # Recursively complete generation, until at least one complete completion is available.
         signature = ensure_signature(signature)
 
--- a/dspy/backends/base.py
+++ b/dspy/backends/base.py
@@ -9,13 +9,12 @@ from dspy.primitives.prediction import Completions
 class BaseBackend(BaseModel, ABC):
     """A backend takes a signature, its params, and returns a list of structured predictions."""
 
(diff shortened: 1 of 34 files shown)


dspy/backends/template.py
```diff
@@ -12,12 +13,16 @@ class TemplateBackend(BaseBackend):
 
     lm: BaseLM
 
-    def __call__(
+    @property
+    def lm_history(self):
+        return self.lm.history
+
```
CyrusOfEden, line 19: Seems a little extra

dspy/predict/predict.py
```diff
@@ -50,66 +53,34 @@ def __call__(self, **kwargs):
...
 
-        for c in C:
-            completions.append({})
-            for field in template.fields:
-                if field.output_variable not in kwargs.keys():
-                    completions[-1][field.output_variable] = getattr(
-                        c, field.output_variable
-                    )
+        # TODO: What purpose does stage play here?
+        # assert self.stage in x, "The generated (input, output) example was not stored"
```
CyrusOfEden, line 78: @okhat can you comment here?

CyrusOfEden (approved):
This is fantastic man! Really appreciate you. Looks like you went through quite a concept maze to get here and the result shows it.

Pinging @okhat to explain 'stage' in your TODO, had 1 nit. lmk what you think

dspy/backends/template.py
KCaverly: I agree. I was using it for testing. Will likely remove all lm based history tracking and just keep it at the backend level, unless we find we need it somewhere else.

KCaverly:
I decided to sneak in a fleshed out JSONBackend, will add tests such that the Predict works with either backend.

dspy/backends/base.py
```diff
@@ -1,76 +1,42 @@
...
-
-BackendEvent = t.TypeVar("BackendEvent", bound=dict[str, t.Any])
+from dspy.primitives.prediction import Completions
 
 
 class BaseBackend(BaseModel, ABC):
     """A backend takes a signature, its params, and returns a list of structured predictions."""
 
-    _history: t.List[BackendEvent] = []
+    history: list[Completions] = Field(default=[])
```
CyrusOfEden, line 12: Little bug here — you wanna do 

`default_factory=list`

CyrusOfEden (changes requested):
Small bug but overall looks great!

KCaverly:
This is done, we have feature parity with DSPy from about a week or two ago, prior to COPRO/MIPRO.
I think we should merge this with the backend-refactor branch, pull main in, and do final tests that way.

We should only be failing on tests for the signature_opt, but that has been replaced by COPRO.

PR #574, opened 2024-03-06
New Typed Signature Optimizer

thomasahle:
Also fixed a bug with list inputs, and another bug on predictors with no input fields.


thomasahle:
I decided to not subclass `Teleprompter` and just make it a function.
Mostly because I didn't have any class state, and I couldn't see a reason for which arguments would go into the class constructor and which would go into `compile`.
Let me know if this is a problem.

thomasahle:
I called it "Typed Signature Optimizer" because it assumes TypedPredictors rather than Predicts. Unfortunately the other signature optimizers don't work when using TypedPredictors, as the signatures get manipulated too much.

examples/nli/scone/ScoNe
okhat, line 1: oh no a submodule haha. I hope it's not too hard to remove

thomasahle: Oops, thanks for noticing!

Issue #543, opened 2024-03-04
error when using azure api 


amirhossein-yousefi:
Completions.create() got an unexpected keyword argument 'api_base'

qianzheng12:
I think they shouldn't include api_key, api_base and api_version in the kwargs in azure_opanai.py.
Until it's being resolved, you can manually remove them at line 110 in azure_openai.py, which will get things working for the time being

insop:
This code is working for me the latest code (I am using 244a81705f1bb84eb3ccff24c2d18aa539fa1aa1)

```
        model_arg = {
            "engine": model_name,
            "deployment_id": model_name,
            "api_version": openai.api_version,
            "api_base": openai.api_base,
        }

        # provider_name = "azure"
        lm = dspy.AzureOpenAI(
            model=model_name,
            api_key=openai.api_key,
            **model_arg,
        )
```

gauthamk28:
> This code is working for me the latest code (I am using [244a817](https://github.com/stanfordnlp/dspy/commit/244a81705f1bb84eb3ccff24c2d18aa539fa1aa1))
> 
> ```
>         model_arg = {
>             "engine": model_name,
>             "deployment_id": model_name,
>             "api_version": openai.api_version,
>             "api_base": openai.api_base,
>         }
> 
>         # provider_name = "azure"
>         lm = dspy.AzureOpenAI(
>             model=model_name,
>             api_key=openai.api_key,
>             **model_arg,
>         )
> ```

Still I am facing the same issue :(
dspy-ai 2.3.6

ClaartjeBarkhofTNO:
Running into the same issue here, could it be due to this OPENAI_LEGACY global variable? 

[It is use in this line](https://github.com/stanfordnlp/dspy/blob/ee646535c232c5ced9e1e948754843e5af5f0d46/dsp/modules/azure_openai.py#L74)

JLKaretis:
Same issue for me as well with dspy-ai 2.3.6

I've tried Insop's code with no success. It seems Qianzheng12 is right, api_base is only useful to initialize the client, then it shouldn't be kept in kwargs after that

insop:
I am using openai `0.28.1`.
Are you using new API?

ClaartjeBarkhofTNO:
I am using a newer version `openai==1.13.3`

Issue #521, opened 2024-03-02
BootstrapFewShot failing due to lm.copy for AzureOpenAI


insop:
### Summary

BootstrapFewShot failing due to lm.copy for AzureOpenAI, likely due to positional argument issue from recent changes.

### Environment
- code: 98304a2eb9d1dddaaa846e30258cd5d8fc6b5d8d
- model azure openai

### Error message
- during `BootstrapFewShot` run, code errored out due to missing arguments, `api_version`

```
>[dspy/dsp/modules/lm.py](https://file+.vscode-resource.vscode-cdn.net/Users/insop/Projects/Github/NLP/dspy/dsp/modules/lm.py)(106)copy()
    102         model = kwargs.pop('model')
    105 
--> 106         return self.__class__(model, **kwargs)
```

### Analysis

AzureOpenAI expects positional arguments, so it is failing.
I will add my fixes and see if that will impact other models.

```
class AzureOpenAI(LM):
...

    def __init__(
        self,
        api_base: str,
        api_version: str,
        model: str = "gpt-3.5-turbo-instruct",
        api_key: Optional[str] = None,
        model_type: Literal["chat", "text"] = "chat",
        **kwargs,
    ):

okhat:
Resolved? Closing but lmk @insop , thanks a lot btw

insop:
> Resolved? Closing but lmk @insop , thanks a lot btw

Thank you @okhat.

insop:
Update.

Related to the issue (https://github.com/stanfordnlp/dspy/issues/543) that caused by the PR (https://github.com/stanfordnlp/dspy/pull/522) from this issue:


I tried to reproduce the issue I ran into earlier, https://github.com/stanfordnlp/dspy/issues/521, but I am not able to reproduce it.

The strange thing is that the code gets to this part as part of `BootstrapFewShot`, but it doesn't seem like `copy()` is called. It seems as if something to do with cache, but I am not able to find it.

https://github.com/stanfordnlp/dspy/blob/669ecd7e04431a0e890ca61c60afafcae1544517/dsp/modules/lm.py#L104

PR #584, opened 2024-03-06
Remove unused kwargs passed to requests sent through `AzureOpenAI`

Yummy2016:
This commit is to remove unused kwargs that we include in the requests sent through `AzureOpenAI` class.

This seems to fix this issue: https://github.com/stanfordnlp/dspy/issues/543

### DIFF
--- a/dsp/modules/azure_openai.py
+++ b/dsp/modules/azure_openai.py
@@ -107,9 +107,6 @@ class AzureOpenAI(LM):
             kwargs["model"] = model
 
         self.kwargs = {
-            "api_base": api_base,
-            "api_version": api_version,
-            "api_key": api_key,
             "temperature": 0.0,
             "max_tokens": 150,
             "top_p": 1,


insop:
@Yummy2016 thank you for the PR.
Let me check the change, first.

This code was updated recently due to this issue, https://github.com/stanfordnlp/dspy/issues/521.
I am using legacy openai.

insop:
Thank you for the change. Let's merge this as many are hitting the issue, @okhat. 


TL;DR

I tried to reproduce the issue I ran into earlier, https://github.com/stanfordnlp/dspy/issues/521, but I am not able to reproduce it.

The strange thing is that the code gets to this part as part of `BootstrapFewShot`, but it doesn't seem like `copy()` is called. It seems as if something to do with cache, but I am not able to find it.

https://github.com/stanfordnlp/dspy/blob/669ecd7e04431a0e890ca61c60afafcae1544517/dsp/modules/lm.py#L104

Since the `copy()` function does not appear to be called, I don't reproduce this, https://github.com/stanfordnlp/dspy/issues/521.

PR #588, opened 2024-03-07
Adding metric_threshold to BootstrapFewShotWithRandomSearch

insop:
*Change:*

- add `metric_threshold` parameter to `BootstrapFewShotWithRandomSearch` with default being `None`

*Reason:*

- when using a custom metric that are not 0 or 1, we need `metric_threshold` in `BootstrapFewShot`; however, when using `BootstrapFewShotWithRandomSearch` we don't have `metric_threshold` that can be passed to  `BootstrapFewShot`

@okhat

### DIFF
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -24,7 +24,7 @@ from .vanilla import LabeledFewShot
 
 
 class BootstrapFewShotWithRandomSearch(Teleprompter):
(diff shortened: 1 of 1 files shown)


insop:
@okhat 

This might not be related to my change (it could be however), any thought? or please tag who might give me pointers to dig more?

Screenshot 2024-03-06 at 9 44 35 PM

PR #602, opened 2024-03-07
Fix code issues in signature-optimizer.mdx

swairshah:
1. use the HotPotQA dataset instead of gsm8k
2. result.rationale instead of result.reason
3. process devset to include input_keys


krypticmouse:
Looks good to me!

PR #604, opened 2024-03-07
[DOCS] Typed Predictor Docs and API References

krypticmouse:


thomasahle:
Thanks for writing these!

thomasahle:
One thing: Maybe we shouldn't put the "_private_methods" in the documentation, as people aren't supposed to use them, and they may change or disappear in the future.

krypticmouse:
Yea I've removed those will push them along with deep dive in next PR :)

PR #606, opened 2024-03-08
WIP fix: cache for the baleen test and add notebook cache to everywhere.

ammirsm:
What is change?
- making the notebook cache available everywhere.

Why the change is important?
- we want to in baleen_test be able to use notebook cache

Why not putting it in actual test or decorator?
- because cached are used globally so it needs a huge change if we want to do that.

How to test?
- Running unit tests.
- Running the notebooks.

### DIFF
--- a/tests/examples/test_baleen.py
+++ b/tests/examples/test_baleen.py
@@ -1,4 +1,6 @@
 import pytest
+import os
+from dsp.modules.cache_utils import *
 from dsp.utils import deduplicate
 import dspy.evaluate
 import dspy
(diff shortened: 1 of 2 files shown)


tests/examples/test_baleen.py
```diff
@@ -106,7 +108,7 @@ def gold_passages_retrieved(example, pred, trace=None):
 
 # @pytest.mark.slow_test
 # TODO: Find a way to make this test run without the slow hotpotqa dataset
-def _test_compiled_baleen():
+def test_compiled_baleen():
```
thomasahle, line 111: Isn't it going to break things to have a unittest using OpenAI like this?

ammirsm: With the other change that I added in another file this should run based on the cache of the notebook.

PR #608, opened 2024-03-08
Feature mistralai api

fsndzomga:
I added support for Mistral AI's API. Thanks to this new feature, users will be able to play with the following models easily: mistral-small-latest, mistral-medium-latest, and mistral-large-latest (a GPT-4 level model).

Here is an example of usage:

```python

import dspy

lm = dspy.Mistral(api_key="your-mistralai-api-key")

mod = dspy.Predict("question -> answer")

print(mod(question="Who is Emmanuel Macron?"))

```


insop:
It'd be great if you could also add md file to this folder.
https://github.com/stanfordnlp/dspy/tree/main/docs/api/language_model_clients

fsndzomga:
Sure I will do that ! Thanks for the feedback !

fsndzomga:
Done @insop !

insop:
Thank you @fsndzomga .
Python code looks good to me; however I am not so sure about other files.

krypticmouse:
@fsndzomga can you change the link in docs to `api/language_model_clients/Mistral`

fsndzomga:
@krypticmouse , I am not sure I understand. Isn't it already the case ?
Screenshot 2024-03-08 at 23 22 07

krypticmouse:
@fsndzomga yes but you just need to specify the path and remove the base url from the link. So you can remove the `https://dspy-docs.vercel.app`

fsndzomga:
Okay, I get it. Done ! Thanks for the feedback.

dsp/modules/mistral.py
```diff
@@ -0,0 +1,124 @@
+from typing import Any, Optional
+
+import backoff
+
+from dsp.modules.lm import LM
+
+try:
```
arnavsinghvi11, line 7: add all related imports here

pyproject.toml
```diff
@@ -101,6 +102,7 @@ sphinx_rtd_theme = { version = "*", optional = true }
 autodoc_pydantic = { version = "*", optional = true }
 sphinx-reredirects = { version = "^0.1.2", optional = true }
 sphinx-automodapi = { version = "0.16.0", optional = true }
+mistralai = "^0.1.3"
```
arnavsinghvi11, line 105: remove from global dependencies

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "datasets~=2.14.6,<3.0.0",
     "requests",
     "optuna",
-    "pydantic==2.5.0",
+    "pydantic~=2.5.0,<2.5.3",
```
arnavsinghvi11, line 33: ideally, we don't want to update the dependencies if not needed, but tagging @okhat for context here

requirements.txt
```diff
@@ -8,5 +8,6 @@ tqdm
 datasets
 requests
 optuna
-pydantic==2.5.0
-# To be deleted
\ No newline at end of file
+pydantic==2.5.0,<2.5.3
+mistralai
```
arnavsinghvi11, line 12: remove mistral and refer to comment above on pydantic for updates

arnavsinghvi11:
Hi @fsndzomga , thanks for the addition!

We want to keep all external imports/dependencies local to the file requirements.

dsp/modules/mistral.py
fsndzomga: not sure I understand, I just followed the same canvas as for other models (google.py, cohere.py etc)

pyproject.toml
fsndzomga: thanks for the proposal. I wanted to do that initially. but when I did, the PR failed some tests in the CI/CD because of all the imports from mistralai client didn't work. Can you recommend a workaround ?

pyproject.toml
fsndzomga: okay, waiting for  @okhat comment, since it has a lot of implications about my current implementation. I was wondering why a strict requirement on pydantic if overall moving from 2.5.0 to 2.5.3 doesn't affect the package that much.

dsp/modules/mistral.py
arnavsinghvi11: Imports should be included locally - [import cohere](https://github.com/stanfordnlp/dspy/blob/a51aad7dcbe257e01d677025cad96acba7c13e90/dsp/modules/cohere.py#L9), [import google](https://github.com/stanfordnlp/dspy/blob/a51aad7dcbe257e01d677025cad96acba7c13e90/dsp/modules/google.py)

pyproject.toml
arnavsinghvi11: I believe the local imports from above should fix this!

dsp/modules/mistral.py
fsndzomga: Got it, working on it now.

requirements.txt
fsndzomga: I removed it

pyproject.toml
fsndzomga: No need, I removed the pydantic version change

fsndzomga:
@arnavsinghvi11 @okhat , I took into account all the feedbacks. It should be all good now.

dsp/modules/mistral.py
okhat: Thanks so much @fsndzomga ! I'm happy to merge pretty much once the imports do not affect people who don't need this module.

Basically: no errors/warnings ("Not loading Mistral AI..." is a confusing error if I don't need Mistral) if I don't try to use this class

fsndzomga: Yes I did that already, moved to local imports as per @arnavsinghvi11 advice.

okhat: Hmm I don't think it's local imports now. Isn't importing outside any methods? Also it's going to print "Not loading Mistral AI..."  for everyone?

fsndzomga: Ok got it, you want me to do it exactly like the cohere implementation and comment or remove the print statement.
On it !
Screenshot 2024-03-19 at 19 25 50

fsndzomga: @okhat done ! I now raise an import error inside the init of the Mistral class. 
Also added some docstrings to synthetic_data_generator since I was already working on it and plan to add tests later this week.
Screenshot 2024-03-19 at 20 05 46

fsndzomga: Yes sure !

requirements.txt
```diff
@@ -9,4 +9,4 @@ datasets
 requests
 optuna
 pydantic==2.5.0
-# To be deleted
\ No newline at end of file
+# To be deleted
```
thomasahle, line 12: Remove this?

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+anthropic = ["anthropic~=0.18.0"]
```
thomasahle, line 37: Why does a mistral patch change the anthropic requirements?

poetry.lock
thomasahle, line 1: Do we need to change this file?

thomasahle (commented):
I'm good with this, modulo the changes to poetry.lock, pyproject.toml and requirements.txt

axelpey:
Tested, I'm not sure if it really works. Mistral API models are chat completion models, and so they respond with repeating the task when using Predict (but not Chain of Thought which gives good results)

```python
sentence = "it's a charming and often affecting journey."  # example from the SST-2 dataset.

# 1) Declare with a signature.
classify = dspy.Predict('sentence -> sentiment')

# 2) Call with input argument(s). 
response = classify(sentence=sentence)

# 3) Access the output.
print(response.sentiment)
```

"Sentence: it's a charming and often affecting journey.\nSentiment: Positive\n\nNote: The sentiment is inferred as positive based on the use of words such as 'charming' and 'affecting' which convey a positive emotion or experience. However, please note that sentiment analysis is not always 100% accurate and can be subjective."

fsndzomga:
Thanks for the feedback @axelpey. To be clear, what you are saying is that it works, but the format of the response is sometimes not the one you expect. Right? Mistral models have a tendency to change their behavior, as you have mentioned. For instance, what you mentioned has not been happening over the past few weeks when I was testing. It depends on the model you use; even the same model can provide responses in different formats. I tested your code and got this: "Sentiment: Positive"  which is more concise.

Screenshot 2024-03-24 at 17 02 36

Since DSPy does not handle response formats per se, on way to achieve your desired goal is to use TypedPredictors or simply use a mistral-large for example.

A global solution would be to add a custom parser specifically for Mistral AI, and I am not sure if it is in line with DSPy approach.

axelpey:
Hmm which one did you use? I used medium. And I think this problem is not limited to mistral, it's for all models that are ChatCompletion models and not just standard Completion. I think it's still very relevant to have Mistral API avail in dspy, just we need to be mindful of it. @okhat I think you've seen these issues before with other ChatCompletion too right?

fsndzomga:
Maybe I am missing something, but we do use the chat completions API from OpenAI right ? Below a screenshot of the gpt3.py file. 
Screenshot 2024-03-24 at 20 00 20

Anyway, if you guys have a better idea how to handle it, let me know, or we can just drop this feature.

arnavsinghvi11 (approved):

arnavsinghvi11:
@fsndzomga @axelpey Thanks for the discussion on the limitation with ChatCompletion models. This is indeed something we've seen come across with running chat models with DSPy. There is some more discussion in #662 and #420 as we have some ongoing work to handle this, but this PR is good to merge for now! Thanks @fsndzomga for the contributions!

fsndzomga:
Oh so even the OpenAI integration has the same problem... I really wasn't aware of that. Thanks for sharing the links @arnavsinghvi11 and thanks for merging the PR !

PR #609, opened 2024-03-08
feature(dspy): return metadata along long_text for ChromaDB retrieval

### LINKED ISSUE #364: Retrieve should enable return of "things" other than passages
The retrieveEnsemble method focusses only on the long_text key. It ignores any other keys provided along with long_text key. I have a custom retriever that returns some metadata along with long_text. retrieveEnsemble should "pass through" this metadata all the way till self.retrieve(question). I mean self.retrieve(question) should return an object that has both passages and metadata keys. While passages can be used to provide the context to the LLM, metadata can be used by the consumer in whichever way he wants to. 

Is there a way to achieve this other than overriding the retrieveEnsemble method?

domci:
Returning metadata along long_text for ChromaDB.
Fixes #364 for ChromaDB

### DIFF
--- a/docs/api/retrieval_model_clients/ChromadbRM.md
+++ b/docs/api/retrieval_model_clients/ChromadbRM.md
@@ -34,7 +34,7 @@ Search the chromadb collection for the top `k` passages matching the given query
 - `k` (_Optional[int]_, _optional_): The number of results to retrieve. If not specified, defaults to the value set during initialization.
 
 **Returns:**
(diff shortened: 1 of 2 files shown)


insop:
Thank you.
It'd be great if you update the document with this change `docs/api/retrieval_model_clients/ChromadbRM.md`

domci:
> docs/api/ret

done! thanks @insop for pointing this out.

insop:
Looks great, thank you.

arnavsinghvi11:
Thanks @domci !

PR #611, opened 2024-03-08
hf models can use auth tokens now

mgbvox:
Access to gated HF models previously was not supported by dspy.HFModels. This is now supported - just pass in token=<your hf token> or load it into the env var HF_TOKEN (e.g. via dotenv.load_dotenv()).


isaacbmiller:
I was getting annoyed by this earlier. Glad someone made the change. LGTM

PR #614, opened 2024-03-09
feature(dspy): Support more embedding models and return all fields

Joshmantova:
The PGVector RM currently only supports a single OpenAI embedding model and although it allows users to input a list of fields to select from PGVector, it doesn't use those additional fields in the `dspy.Example`s output. This modification allows users to specify:
1. Any OpenAI embedding model to be used. Still will default to the current model so there's no regression.
2. Return data within the specified fields in the `dspy.Example`s. It will now use the column names in the examples rather than the hardcode fields the current RM specifies.

This PR still needs to be tested. I'll create a local PGVector DB and test once a maintainer has notified me that the goal of this change is desired.


arnavsinghvi11:
Thanks @Joshmantova ! These are great additions. Feel free to test (and add this to [tests/](https://github.com/stanfordnlp/dspy/tree/main/tests)) and lmk once it's ready to merge (post-merge conflict resolution).

Joshmantova:
> Thanks @Joshmantova ! These are great additions. Feel free to test (and add this to [tests/](https://github.com/stanfordnlp/dspy/tree/main/tests)) and lmk once it's ready to merge (post-merge conflict resolution).

Hi @arnavsinghvi11, it's tested and ready to go. I don't think unit tests are very useful here because I'd have to mock over nearly everything. It also looks like most if not all the tests are unit tests. Given that we can't run necessary tests without setting up postgres, do you think it's worth it to add tests to the repo for this?

arnavsinghvi11:
Hi @Joshmantova , looks great! yeah I think you can add some base integration tests and leave them out from the CI for now. Also, are the changes to poetry.lock needed? seems like they are not relevant to the changes but just want to confirm if you can remove that from the next commit. Thanks!

Joshmantova:
> Hi @Joshmantova , looks great! yeah I think you can add some base integration tests and leave them out from the CI for now. Also, are the changes to poetry.lock needed? seems like they are not relevant to the changes but just want to confirm if you can remove that from the next commit. Thanks!

I'll add those integration tests and ping you when they're in. It'll likely be today. The changes to the poetry lock file were due to me adding the requirements to integrate PGVector. This RM depends on some packages that were not in the list of dependencies so I added them. After that, I re-compiled the lock file. I'm on a M1 mac so I'm not sure if that caused the lock file to change more than expected. Do you want me to remove the added dependencies?

Joshmantova:
> Hi @Joshmantova , looks great! yeah I think you can add some base integration tests and leave them out from the CI for now. Also, are the changes to poetry.lock needed? seems like they are not relevant to the changes but just want to confirm if you can remove that from the next commit. Thanks!

@arnavsinghvi11 - woops! Looks like I accidentally upgraded the dependencies. I'll get that fixed up today as well, thanks for catching that.

Joshmantova:
> Hi @Joshmantova , looks great! yeah I think you can add some base integration tests and leave them out from the CI for now. Also, are the changes to poetry.lock needed? seems like they are not relevant to the changes but just want to confirm if you can remove that from the next commit. Thanks!

@arnavsinghvi11 - I added a few integration tests including some instructions on how to set up PGVector. They're all passing at the moment. It turns out that I did not upgrade any dependencies. Installing the postgres related extras is causing the diff in the poetry lock file. I think it's wise to keep these dependency extras in the PR. Do you agree? If so, it's good to merge.

arnavsinghvi11:
Hi @Joshmantova , thanks for adding the tests! 

If you could resolve the merge conflict with the dependencies in poetry.lock and pyproject.toml files, and also run `ruff check . --fix-only`, the PR is good to go!

Joshmantova:
@arnavsinghvi11 - Btw, [my other open PR](https://github.com/stanfordnlp/dspy/pull/704) will implement a solution that allows developers to easily install a valid dev environment including installation of all git hooks so we won't have to worry about these manual linting fixes moving forward. If you're able to take a look at that one as well, I'd appreciate it! There will be some merge conflicts that I'll resolve shortly after this PR goes in.

arnavsinghvi11:
Thanks @Joshmantova !

PR #618, opened 2024-03-09
Add system prompt to GPT

xdite:
## Usage

```
system_prompt = "You are a language model based on Wikipedia, answering users' questions using a formal and serious tone."
lm = dspy.OpenAI(model='gpt-4-1106-preview', api_key=openai_key, system_prompt=system_prompt, max_tokens=4000)
```

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -62,11 +62,14 @@ class GPT3(LM):
         api_provider: Literal["openai"] = "openai",
         api_base: Optional[str] = None,
         model_type: Literal["chat", "text"] = None,
+        system_prompt: Optional[str] = None,
         **kwargs,
     ):
         super().__init__(model)
         self.provider = "openai"
         openai.api_type = api_provider
+
+        self.system_prompt = system_prompt
 
         assert (
             api_provider != "azure"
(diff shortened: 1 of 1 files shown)


okhat:
I like this, very simple and will help a bunch of people who asked for it. Thanks a lot. Checking and merging.

Markovian99:
I believe there is a minor bug here in lines 124-127 of gpt3.py where the system prompt is getting overwritten in line 127:
```
            messages = [{"role": "user", "content": prompt}]
            if self.system_prompt:
                messages.insert(0, {"role": "system", "content": self.system_prompt})
            kwargs["messages"] = [{"role": "user", "content": prompt}]
```
I think would work as expected if replaced 127 with:
`kwargs["messages"] = messages`

rawwerks:
just to clarify - if i add the system prompt like above, should i assume that dspy is also optimizing the system prompt, or only the user prompt?

samiur:
> just to clarify - if i add the system prompt like above, should i assume that dspy is also optimizing the system prompt, or only the user prompt?

+1 to this question

Markovian99:
It is only optimizes the user prompt - the system prompt is save as part of the LLM object and the implementation is just to allow for this basic use of a system prompt. (vs not having one)

coderfengyun:
I want to ask the same question as @Markovian99 , does DSPy support optimizing System Prompt?

coderfengyun:
> just to clarify - if i add the system prompt like above, should i assume that dspy is also optimizing the system prompt, or only the user prompt?

+1 to this question

yubozhao:
@xdite Why is the system prompt defined at LLM init level? Shouldn't this be on the `dspy.Module` level?

yubozhao:
it is a bad design decision to define at the global level. You also add more responsibility to the configure level. If the configuration is doing lower level (aka model) level work, by adding system prompt to it, you are adding additional responsibility and overstepping boundaries

yubozhao:
System prompt (as usual practice) is part of the prompts work. That's part of the module, also system prompt should be optimize as well.

PR #622, opened 2024-03-10
fix: return value in the inspect_history beside printing.

ammirsm:
What is the change?
- in the inspect_history function on top of printing adding the return 

Why the change is important?
- to when we are trying to get the last prompt be able to save it to variable beside printing.

How to test?
- GIVEN you run the compile
- WHEN you use the inspect_history
- THEN be able to save it in variable and it print it too.


CyrusOfEden:
Hey Amir! Great initiative, can you please address the following?

1. The return isn't doing what you think it does, it's returning within the loop
2. This method should both print (to maintain compatibility) AND return the string

ammirsm:
> Hey Amir! Great initiative, can you please address the following?
> 
> 1. The return isn't doing what you think it does, it's returning within the loop
> 2. This method should both print (to maintain compatibility) AND return the string

Addressed.

dsp/modules/lm.py
```diff
@@ -69,28 +69,33 @@ def inspect_history(self, n: int = 1, skip: int = 0):
...
+            printing_value += prompt
+
             text = ""
             if provider == "cohere":
                 text = choices[0].text
             elif provider == "openai" or provider == "ollama":
                 text = ' ' + self._get_choice_text(choices[0]).strip()
             elif provider == "clarifai":
                 text=choices
-            elif provider == "google":
```
arnavsinghvi11, line 85: is there a reason to remove this?

arnavsinghvi11:
Thanks @ammirsm ! left a comment but otherwise LGTM.

PR #555, opened 2024-03-05
Fix return_all_scores in evaluate

quajak:
Addresses https://github.com/stanfordnlp/dspy/issues/477

PR #625, opened 2024-03-10
Fix return_all_scores in evaluate

quajak:
Remade version of https://github.com/stanfordnlp/dspy/pull/555 as requested by @isaacbmiller.


isaacbmiller:
Thanks for remaking!

PR #634, opened 2024-03-12
Add an option to choose embedding model in pgvector

### LINKED ISSUE #613: pgvector_rm is supporting only OpenAI as the embedding model
Currently, pgvector_rm supports only the OpenAI embedding model. However, it would be beneficial to make it more versatile by allowing any embedding model to be used. For reference, pinecone_rm offers a local embedding model option.

nbqu:
* What is this PR for?
  * This PR resolves #613, adding an option to choose whether use openai to get embeddings, or other embedding models like cloud embedding services(google, cohere, nomic, ...) or on-premise embedding models(huggingface).

Features:
* Now installing `openai` is not required. Move it into optional import.
* Now either `openai_client` or `embedding_func` should be specified, otherwise raises AssertionError.

This PR will not change the current behavior as it still accepts `openai_client` for openai embeddings.

### DIFF
--- a/dspy/retrieve/pgvector_rm.py
+++ b/dspy/retrieve/pgvector_rm.py
@@ -1,5 +1,6 @@
-from typing import List, Optional, Callable
 import warnings
+from typing import Callable, Optional
+
 import dspy
 
 try:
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @nbqu !

Issue #554, opened 2024-03-05
Multiple api bases


umarbutler:
Hi,
I'd like to see functionality added to make it possible to provide multiple api bases that are later round-robined when consecutive requests are made. This would allow me to host the same LLM on two servers and cut the time it takes to receive responses in half. I have already implemented a custom version of the `HFClientVLLM` class that does this but I'd like to see this implemented as part of the library's core functionality. The custom class is provided below:
```python
import dspy

from dsp.modules.hf_client import HFModel, send_hfvllm_request_v00
from config import openai_api_key, openai_api_bases, model_name

# Monkey patch the HFClientVLLM class to round robin api bases instead of using a single one.
class HFClientVLLM(HFModel):
    def __init__(self, model, urls, **kwargs):
        super().__init__(model=model, is_client=True)
        
        self.urls = [url.replace('/v1', '') for url in urls]
        self.headers = {"Content-Type": "application/json"}

    def _generate(self, prompt, **kwargs):
        kwargs = {**self.kwargs, **kwargs}

        payload = {
            "model": kwargs["model"],
            "prompt": prompt,
            "max_tokens": kwargs["max_tokens"],
            "temperature": kwargs["temperature"],
        }

        # Round robin the urls.
        url = self.urls.pop(0)
        self.urls.append(url)

        response = send_hfvllm_request_v00(
            f"{url}/v1/completions",
            json=payload,
            headers=self.headers,
        )

        try:
            json_response = response.json()
            completions = json_response["choices"]
            response = {
                "prompt": prompt,
                "choices": [{"text": c["text"]} for c in completions],
            }
            return response

        except Exception:
            print("Failed to parse JSON response:", response.text)
            raise Exception("Received invalid JSON response from server")
```

arnavsinghvi11:
Hi @umarbutler , this is great! Currently, HFClientTGI supports this [port routing](https://github.com/stanfordnlp/dspy/blob/226a0d94672f03a5d33215c5efce473a1a3419be/dsp/modules/hf_client.py#L71), but it would be great to support this for VLLM and other LM clients as well. Feel free to push a PR!

PR #638, opened 2024-03-13
Added support for multiple API endpoints to `HFClientVLLM`

umarbutler:
This PR adds support for multiple API endpoints being passed to `HFClientVLLM` via the `url` argument, in a backwards-compatible manner, as discussed in [this](https://github.com/stanfordnlp/dspy/issues/554) issue.

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -116,7 +116,16 @@ def send_hftgi_request_v00(arg, **kwargs):
 class HFClientVLLM(HFModel):
     def __init__(self, model, port, url="http://localhost", **kwargs):
         super().__init__(model=model, is_client=True)
-        self.url = f"{url}:{port}"
+        
+        if type(url) == list:
+            self.urls = ['/v1'.join(url_.split('/v1')[:-1]) for url_ in url]
+        
+        elif type(url) == str:
+            self.urls = [f'{url}:{port}']
+        
+        else:
+            raise ValueError(f"The url provided to `HFClientVLLM` is neither a string nor a list of strings. It is of type {type(url)}.")
+        
         self.headers = {"Content-Type": "application/json"}
 
     def _generate(self, prompt, **kwargs):
@@ -128,9 +137,13 @@ class HFClientVLLM(HFModel):
             "max_tokens": kwargs["max_tokens"],
             "temperature": kwargs["temperature"],
         }
+        
+        # Round robin the urls.
+        url = self.urls.pop(0)
+        self.urls.append(url)
 
         response = send_hfvllm_request_v00(
-            f"{self.url}/v1/completions",
+            f"{url}/v1/completions",
             json=payload,
             headers=self.headers,
         )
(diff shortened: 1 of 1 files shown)


dsp/modules/hf_client.py
```diff
@@ -116,7 +116,16 @@ def send_hftgi_request_v00(arg, **kwargs):
 class HFClientVLLM(HFModel):
     def __init__(self, model, port, url="http://localhost", **kwargs):
         super().__init__(model=model, is_client=True)
-        self.url = f"{url}:{port}"
+        
+        if type(url) == list:
+            self.urls = ['/v1'.join(url_.split('/v1')[:-1]) for url_ in url]
```
arnavsinghvi11, line 121: this seems a bit extra - would it be better to specify some base formatting in the documentation and having a condition check on that instead? feel free to add it!

arnavsinghvi11:
Thanks @umarbutler ! left a comment, but the rest looks good to me!

okhat:
This is pretty cool, I think @XenonMolecule @klopsahlong actually needed this right now, can you check and merge if all looks good @arnavsinghvi11 or @XenonMolecule ?

dsp/modules/hf_client.py
XenonMolecule: I agree with this. It's unclear why we need to do this splitting and joining.  I feel this might lead to unintuitive behavior, for example, if I don't include /v1/completions in my URLs now it drops the URL completely:

```
>>> urls = ["localhost:8000/v1/completions", "server.com:7410/v1/completions", "mycluster:4000", "oldservice.co:3000/v1/service/v1/completions"]
>>> new_urls = ['/v1'.join(url_.split('/v1')[:-1]) for url_ in urls]
>>> new_urls
['localhost:8000', 'server.com:7410', '', 'oldservice.co:3000/v1/service']
```

This would break the existing way people are inputting URLs if they just convert to list form.

dsp/modules/hf_client.py
```diff
@@ -116,7 +116,16 @@ def send_hftgi_request_v00(arg, **kwargs):
...
+        
+        elif type(url) == str:
+            self.urls = [f'{url}:{port}']
+
```
XenonMolecule, line 125: It is not necessary to add in this PR, but a related idea that I think it might be nice to add is the same way that HFClientTGI handles multiple ports.

`self.ports = port if isinstance(port, list) else [port]`

In this case if the user provides 1 URL and a list of ports we can round robin over the ports instead (by constructing self.urls over the URL + all ports)

If they specify both as a list we can just raise an exception and say if you have multiple ports and multiple URLs they need to all be specified in the URL list.

dsp/modules/hf_client.py
XenonMolecule: Also we might want to add a warning if the user passes a port and list of URLs because the port is now ignored.

XenonMolecule (commented):
Left a few comments as well.  Looks good!  The main one that I'd want addressed is the same thing Arnav commented on

umarbutler:
@arnavsinghvi11 @XenonMolecule I've removed the unnecessary dropping of `v1/*` from urls ending in `v1/*`. I added it because when I first used `HFClientVLLM`, I passed the full endpoint, which raised an error, yet there was no indication that I shouldn't have done so. It looks like there was a typo in my implementation anyways. It should have been `self.urls = [('/v1'.join(url_.split('/v1')[:-1]) if 'v1/' in url_ else url_) for url_ in url]`.

PR #616, opened 2024-03-09
 Vertex AI init commit

Demontego:
This is the first phase of implementing google vertex ai in dspy. Please take a look. Write where and how best to correct, what else should be added.

PR #644, opened 2024-03-13
feature(dspy): added support for Vertex AI Gemini API

marshmellow77:
Added support for Vertex AI Gemini API, see also https://github.com/stanfordnlp/dspy/discussions/640


nbqu:
#616 Addresses vertex AI support. Is there any differences?

marshmellow77:
Nope - I just wished the author of #616 would have spelled Vertex correctly so their PR would have popped up when I was searching for it, would have saved me some time 😂

PR #646, opened 2024-03-14
Fix do_generate looking for max_tokens when using Google LM

mitchellgordon95:
The google LM module has "max_output_tokens" in its kwargs dict, rather than "max_tokens". This causes `do_generate` to sometimes fail when trying to adjust the max tokens in the retry logic.

I considered updating the Google kwargs dict to also contain "max_tokens", but then the generation config object complains about extra keys, so it gets kind of annoying to keep track of as you switch back and forth.

It probably won't matter soon anyway, since I think the plan is to migrate to LiteLLM.


sandangel:
Thank you very much, I just got this issue yesterday.

arnavsinghvi11:
Thanks @mitchellgordon95 !

PR #647, opened 2024-03-14
Fix Google API generate_content usage

mitchellgordon95:
Per [the docs](https://ai.google.dev/api/rest/v1/GenerateContentResponse) GenerateContentResponse doesn't seem to have a `parts` param any more.

Completions are found under `candidates[i].content.parts[i].text`.

### DIFF
--- a/dsp/modules/google.py
+++ b/dsp/modules/google.py
@@ -157,6 +157,6 @@ class Google(LM):
         completions = []
         for i in range(n):
             response = self.request(prompt, **kwargs)
-            completions.append(response.parts[0].text)
+            completions.append(response.candidates[0].content.parts[0].text)
 
         return completions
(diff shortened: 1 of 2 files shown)


nbqu:
I found that `parts` are the quick accessor as you can check [here](https://github.com/google/generative-ai-python/blob/3704fa8b1859c2ac8135cdd36df73429a7b27acc/google/generativeai/types/generation_types.py#L321), so this may work equally.

mitchellgordon95:
ah, yes, you are right.

i think my issue was that `parts` was empty sometimes, and i fooled myself into thinking that this fixed it.

will close for now, but we should keep in mind that the quick accessor only works if you have a single candidate. (multiple candidates seem like they are [supported](https://ai.google.dev/api/rest/v1/GenerationConfig) now, so maybe I will take a pass at that later.)

mitchellgordon95:
For the posterity, the reason I was getting this error:

```
IndexError: list index (0) out of range
```

was because I was hitting the max token length, so the API was not returning any results.

https://www.reddit.com/r/GeminiAI/comments/1bbf739/list_index_out_of_range_calling_gemini_api/

it might make sense to handle this case gracefully somehow, but I'm not sure the best way to do it.

PR #656, opened 2024-03-15
ci(dspy): Add coverage to CI

isaacbmiller:
Add a coverage file and comment on each PR.

Coverage isn't something that it seems like we really care about currently, but it is a good feature to have.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -35,7 +35,7 @@ jobs:
           exit 1
 
   test:
-    name: Run Tests
+    name: Run Tests and Coverage
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


thomasnormal:
I'm not the type of person who thinks we necessarily need 100% coverage on everything.
But this table definitely points out some places we lack tests, like `dspy/retrieve`, `synthesizer` and `finetuning`.
I'm not sure who'd be best at providing those. Maybe @CShorten, @fsndzomga and @isaacbmiller?

isaacbmiller:
I absolutely agree. Definitely not 100%, but we should fill in the gaps. I can work on some of these tests

fsndzomga:
will add some test on the synthetic_data_generator too later this week !

PR #657, opened 2024-03-15
fix(ruff) | Ruff accidentally has been deleted from pre-commit

ammirsm:
What is the change?
- ruff came back to the repo pre-commit

Why the change is important?
- it will bring lots of merge conflicts later on if the IDE doesn't have the same config.

How to test?
- test it with your IDE

Note:
This has been removed accidentally in https://github.com/stanfordnlp/dspy/pull/611 which I just reverted the file.


thomasnormal:
Great commit messages :-D

PR #668, opened 2024-03-18
fix(colbertv2): Add error handler for server reports error

wronkiew:
When the ColBERT server returns an error in colbertv2_get_request_v2 the error message is not relayed to the calling function and the line that fails is not helpful for diagnosing the problem.
```
  File "dsp/modules/colbertv2.py", line 50, in colbertv2_get_request_v2
    topk = res.json()["topk"][:k]
           ~~~~~~~~~~^^^^^^^^
KeyError: 'topk'
```
Check for an error flag in the response and raise an exception with the server message instead.
```
  File "dsp/modules/colbertv2.py", line 49, in colbertv2_get_request_v2
    raise(Exception("Server error: " + res["message"]))
Exception: Server error: Exception occurred when connecting to server on port 2172: Cannot connect to host localhost:2172 ssl:default [Connect call failed ('127.0.0.1', 2172)]
```

### DIFF
--- a/dsp/modules/colbertv2.py
+++ b/dsp/modules/colbertv2.py
@@ -43,8 +43,11 @@ def colbertv2_get_request_v2(url: str, query: str, k: int):
 
     payload = {"query": query, "k": k}
     res = requests.get(url, params=payload, timeout=10)
+    res = res.json()
+    if "error" in res and res["error"]:
+        raise(Exception("Server error: " + res["message"]))
 
-    topk = res.json()["topk"][:k]
+    topk = res["topk"][:k]
     topk = [{**d, "long_text": d["text"]} for d in topk]
     return topk[:k]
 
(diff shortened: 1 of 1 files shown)


insop:
Thank you for the PR.
It looks good to me.

okhat:
Unfortunately we can't merge this at the moment. It will change the cached function `colbertv2_get_request_v2` and so it will invalidate/lose all existing caches, which affects a lot of runs and a lot of the notebooks that rely on caching.

In the long run this kind of change makes sense. In the short term, is there a way to get the same behavior by changing the *caller* of `colbertv2_get_request_v2`?

wronkiew:
Yes, thanks for pointing that out. Good to know. I can catch the KeyError in `ColBERTv2.__call__` and raise an exception with a more helpful error message, though the server error message will be lost.

wronkiew:
New error message:
```
  File "dsp/modules/colbertv2.py", line 34, in __call__
    raise KeyError(f"Key not found in ColBERTv2 server response: {e}") from e
KeyError: "Key not found in ColBERTv2 server response: 'topk'"
```

arnavsinghvi11:
Hi @wronkiew , thanks for the PR! This is compatible with existing caches so just checking if its ready to merge from your end.

PR #670, opened 2024-03-18
misc(dspy): Move skycamp into example directiory

isaacbmiller:
Just cleaning up home directory.

I need to adjust the first block with relative imports


okhat:
Makes sense. Btw there are links to this from the README too

okhat:
Also at this point we should just "complete" the uncompleted notebook. People always get confused

PR #672, opened 2024-03-18
Syntax and Typing

mgbvox:
Added some syntax and typing annotations, and applied black/ruff pre-commit passes to code so pre-commit doesn't fail on machines where it is installed.


pyproject.toml
```diff
@@ -109,15 +109,20 @@ autodoc_pydantic = { version = "*", optional = true }
...
+matplotlib = "^3.8.3"
 
 
 [tool.poetry.group.dev.dependencies]
+black = "^24.2.0"
+mypy = "^1.8.0"
+pre-commit = "^3.6.2"
 pytest = "^6.2.5"
-transformers = "^4.38.2"
-torch = "^2.2.1"
```
thomasahle, line 117: Do you need to change this?

pyproject.toml
```diff
@@ -265,6 +270,14 @@ ignore = [
...
     # Sometimes we need List and Tuple
     "UP006",
+    # Asserts sometimes needed (e.g. assert X is not None)
+    "S101",
+    # Prints are currently being used for logging
+    "T201",
+    # .compile shadows a python builtin, and this is fine
+    "A003",
+    # Commented out code allowable for the time being
+    "ERA001",
```
thomasahle, line 280: I don't think we should have comment out code in the repo.

poetry.lock
thomasahle, line 1: Do we need to change this?

thomasahle (commented):
Remove the poetry lock file, unless it's really needed?
Maybe make sure the versioning stuff in pyproject is correct.
Other that that this looks good!

okhat:
Will this affect other optimizers, like MIPRO?

PR #677, opened 2024-03-19
Fix save and load state of program with typed predictors

Neoxelox:
This simple PR aims to fix the `Pydantic model is not JSON serializable` error when saving the state of a compiled program that uses `TypedPredictors`.

However, this fix assumes that the demos of the predictors don't need to keep the same types after saving and loading their states. In older versions of DSPy the type was preserved, but after this commit it changed: https://github.com/stanfordnlp/dspy/commit/b9516c45dbcdedbed6b71dab7cd4a6c1024e0dcf

If that is not expected... then I don't how we would load the state to the pipeline. The JSON cannot know about the Pydantic models. A solution would be to pickle everything. Unfortunately I wasn't able to pickle the pipeline either, because it crashes on the dynamically created Pydantic models of the augmented TypedCoTs signatures.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -4,6 +4,7 @@ import dsp
 from dspy.predict.parameter import Parameter
 from dspy.primitives.prediction import Prediction
 from dspy.signatures.signature import ensure_signature, signature_to_template
+from pydantic import BaseModel
 
 
 class Predict(Parameter):
@@ -20,8 +21,18 @@ class Predict(Parameter):
         self.demos = []
 
     def dump_state(self):
-        state_keys = ["lm", "traces", "train", "demos"]
+        state_keys = ["lm", "traces", "train"]
         state = {k: getattr(self, k) for k in state_keys}
+
+        state["demos"] = []
+        for demo in self.demos:
+            demo = demo.copy()
+
+            for field in demo:
+                if isinstance(demo[field], BaseModel):
+                    demo[field] = demo[field].model_dump_json()
+
+            state["demos"].append(demo)
 
         # Cache the signature instructions and the last field's name.
         state["signature_instructions"] = self.signature.instructions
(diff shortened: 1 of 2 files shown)


dspy/predict/predict.py
```diff
@@ -20,9 +21,19 @@ def reset(self):
...
+        state_keys = ["lm", "traces", "train"]
         state = {k: getattr(self, k) for k in state_keys}
 
+        state["demos"] = []
+        for demo in self.demos:
+            demo = demo.copy()
+
+            for field in demo:
+                if isinstance(demo[field], BaseModel):
+                    demo[field] = demo[field].model_dump_json()
```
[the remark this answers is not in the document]
Neoxelox: That's the thing, I'm assuming I don't have to, because the demos of the predictors don't need to keep the same types after saving and loading their states, at least after this commit https://github.com/stanfordnlp/dspy/commit/b9516c45dbcdedbed6b71dab7cd4a6c1024e0dcf .

When the state is loaded, typed demos are converted to a dict of str fields, which are injected correctly when the pipeline is runned.

tests/predict/test_predict.py
```diff
@@ -53,14 +55,88 @@ def test_call_method():
...
+
+    class Input(pydantic.BaseModel):
+        content: str
+        language: str
+
+    class Output(pydantic.BaseModel):
+        translation: str
+
+    input: Input = dspy.InputField()
+    output: Output = dspy.OutputField()
```
thomasahle, line 109: I think it's better to define the Signature classes inside the `test_` functions, since in the future we might want to move the tests around, and then it's easier that they contain all the signatures they need.

thomasahle:
Sorry, I don't have any insight into loading/saving.
Regarding the ruff error, it's usually a single missing comma, and fixable by calling `ruff check . --fix-only`

tests/predict/test_predict.py
Neoxelox: Done

PR #684, opened 2024-03-20
Add initial Logger

KCaverly:
I decided to go ahead and add a small default logger for DSPy.
The implementation is fairly simple to use at the library level, offers rich output, has minimal dependencies, and allows alot of flexibility for default formatting should we choose to set some specific structure up.

At a high level, its implemented using [structlog](https://www.structlog.org/en/stable/index.html).
Structlog is a lightweight library (no dependencies) which wraps logging for improved ergonomics.

It is set up to be used with two methods exposed at the `dspy` level.

## How to Log Anything

```python
import dspy

# To log a basic info message
dspy.logger.info("hello world!")

# Similar methods exist for all other log levels (INFO, DEBUG, WARNING, CRITICAL, ERROR):
dspy.logger.debug("hello world from debug!")
````

## On Log Levels

The visual log level defaults to INFO, but can be overrided in two simple ways:

**Via Environment Variable:**

```bash
LOG_LEVEL=info python dspy_program.py
````

**Via Code:**

```python
import dspy

# Since the default level is set to INFO, this wont print to the console
dspy.logger.debug("this won't print to console")

# Update the log level to show debug items
dspy.set_log_level("debug")

dspy.logger.debug("this WILL print to console")
````

**Log Output**

A few utilities have been added to either print the log to console, or file, and whether to output it as a string, or json.

````python
import dspy

# This will print to console as a str
dspy.logger.info("this will show in console str formatted")

# Change output from str to json
dspy.set_log_output(output_type="json")

# This will print to console as json
dspy.logger.info("this will show in console as a json object")

# Change output method from console to file
dspy.set_log_output(method="file", file_name="log.txt")

# This will append to log.txt as a json
dspy.logger.info("this will show in console as a json object")

````


thomasahle:
Amazing! 
Is `structlog` better than the default `logging ` module?
I think the standard practice is to use `logger = logging.
getLogger(__name__) ` at the top of each module.
That gives you a logger hierarchy, and you can toggle e.g. `dspy.teleprompt` separately from `dspy.primitive`, or whichever you are debugging.

KCaverly:
I believe structlog allows you to be more dynamic with your outputs, versus just capturing text data. 

I do like the dunder method for hierarchical file names in the log though. Would be nice to see if we can get hierarchical file info without the getLogger calls everywhere.

KCaverly:
For example, we could tag which module the call is being called from which may be particularly helpful for debugging with DSPy but is likely advanced functionality beyond what we want right now.

isaacbmiller:
This looks awesome. Thanks for adding it!

KCaverly:
Changed a few items. This change will allow for hierarchical filtering fully compatible with `logging`, without having to initialize boilerplate on each file. Ie. We track call site details/module, which allow for hierarchical `__name__` based module filtering. 

The other benefit with `structlog`, is we can add any key/value information to any logging call, and export the logs directly as a JSON file. I have a feeling there are a few areas, in optimization/compilation where this could be used to get really strong log details.

With this change I removed `logging.basicConfig` from a few `dsp.modules` files. Setting the basicConfig for logging in numerous places in the codebase, leads to inconsistent logging handling. If we want to enable Streaming logs to a file as opposed to the console, we can enable this globally to ensure logging is consistent throughout the package.

I think this should be good to merge, and I would encourage people to transition to `dspy.logger` over `logging`, as they encounter logging.

KCaverly:
I also added small utilities for rendering the logs as a json object, or saving to a file versus streaming to the stdout.

hasalams:
This looks good. Thanks for the contribution.

okhat:
Do we merge? This seems great, thank you @KCaverly. I see some recent changes. Happy to merge when ready

okhat:
Might need to update setup.py

also I don’t understand the changes to azure or gpt3 — is this going to break the usage logs that were there, which saved to files

KCaverly:
It will break the logs to azure/gpt3, but I dont think its a concern. The config conflicts anyways based on import order, (ie. both cant be applied) and I am skeptical that logging output config should be handled at all at the library level. 

For example, if someone wants to output the gpt logs to a file, in a similar way they could just run the below in their code to do it. 

dspy.set_log_output(file=“gpt.log”, method=“file”)

KCaverly:
I agree we will need to update requirements.txt, before merge.

KCaverly:
@okhat I've added structlog to requirements.txt, and this should be good to merge, if you are comfortable with the changes to gpt3 and the logging defaults outlined above. Let me know if there are any other questions/concerns. Thanks!

okhat:
Thank you so much!!! Happy to merge if code that wasn’t using the gpt3 logs will feel completely unaffected. I think that’s the case but triple checking

fullstackwebdev:
how to disable this? It breaks the logs for GPT3

PR #693, opened 2024-03-21
Fix Retry for TypedPredictors

Neoxelox:
This simple PR aims to fix the errors and signature inconsistencies when using Assertions and Suggestions with modules with TypedPredictors.

### DIFF
--- a/dspy/predict/retry.py
+++ b/dspy/predict/retry.py
@@ -32,6 +32,12 @@ class Retry(Predict):
         return signature
 
     def forward(self, *, past_outputs, **kwargs):
+        # Take into account the possible new signature, as in TypedPredictor
+        new_signature = kwargs.pop("new_signature", None)
+        if new_signature:
+            self.original_signature = new_signature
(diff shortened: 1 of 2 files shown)


dspy/predict/retry.py
```diff
@@ -59,7 +65,7 @@ def __call__(self, **kwargs):
         # now pop multiple reserved keys
         # NOTE(shangyin) past_outputs seems not useful to include in demos,
         # therefore dropped
-        for key in ["_trace", "demos", "signature", "config", "lm", "past_outputs"]:
+        for key in ["_trace", "demos", "signature", "new_signature", "config", "lm", "past_outputs"]:
```
Neoxelox, line 68: I don't know whether this is necessary, but I added it because there is "signature" already in there.

thomasnormal:
Very nice!
Can you create a unit test to show how assertions work with TypedPredictor?

thomasahle:
Excellent. Thank you!

PR #700, opened 2024-03-23
(style) adding docstrings to the synthetic data generator class

fsndzomga:


dspy/experimental/synthetic_data.py
```diff
@@ -6,21 +7,33 @@
 import dspy
 
 
-class descriptionSignature(dspy.Signature):
-  field_name = dspy.InputField(desc="name of a field")
-  example = dspy.InputField(desc="an example value for the field")
-  description = dspy.OutputField(desc="a short text only description of what the field contains")
+class DescriptionSignature(dspy.Signature):
+    """Custom signature to generate descriptions."""
```
okhat, line 11: Docstrings are instructions when using Signatures. Notice here it’s not really an instruction.

arnavsinghvi11:
Thanks @fsndzomga ! just removed the signature instruction since it is not entirely separate from the descriptions present in the input and output fields already and would be handled by the default dspy.Predict template.

fsndzomga:
That's perfect ! Thank you @arnavsinghvi11 !

PR #701, opened 2024-03-23
feat: Add support for additional Anthropic models

Arbaaz-Mahmood:
This commit updates the `Claude` class to support additional models provided by the Anthropic API:

- claude-2.0
- claude-2.1
- claude-3-opus-20240229
- claude-3-haiku-20240307
- claude-3-sonnet-20240229

The `__init__` method of the `Claude` class has been modified to include a list of supported models. When creating an instance of the `Claude` class, the `model` parameter can be set to any of the supported model names. If an unsupported model is provided, a `ValueError` is raised with a message indicating the supported models.

This change allows users to utilize the newly added models for generating completions using the Anthropic API.

### DIFF
--- a/dsp/modules/anthropic.py
+++ b/dsp/modules/anthropic.py
@@ -45,15 +45,26 @@ class Claude(LM):
     ):
         super().__init__(model)
 
+
         try:
             from anthropic import Anthropic
         except ImportError as err:
             raise ImportError("Claude requires `pip install anthropic`.") from err
+        supported_models = [
+        "claude-instant-1.2",
+        "claude-2.0",
+        "claude-2.1",
+        "claude-3-opus-20240229",
+        "claude-3-haiku-20240307",
+        "claude-3-sonnet-20240229",
+    ]
+    
+        if model not in supported_models:
+            raise ValueError(f"Unsupported model: {model}. Supported models are: {', '.join(supported_models)}")
 
         self.provider = "anthropic"
         self.api_key = api_key = os.environ.get("ANTHROPIC_API_KEY") if api_key is None else api_key
         self.api_base = BASE_URL if api_base is None else api_base
-
         self.kwargs = {
             "temperature": kwargs.get("temperature", 0.0),
             "max_tokens": min(kwargs.get("max_tokens", 4096), 4096),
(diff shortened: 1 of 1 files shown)


okhat:
Thank you. Why do we need to check for the supported names

Arbaaz-Mahmood:
Checking for supported model names ensures compatibility and prevents errors when using the Anthropic API. It provides a clear error message to users if they attempt to use an unsupported model, guiding them to use the correct model names. This validation step improves the user experience and reduces potential issues caused by using incompatible or incorrect model names, saving time and effort in debugging and troubleshooting. We check for supported models on other API providers like OpenAI too which has conditions like if name in gpt-4 then it routes to chat if instruct it routes to completion models otherwise raises an error which is not obvious at all --- you end up troubleshooting twenty other things which have nothing to do with the root cause which is wrong model name. @okhat

Arbaaz-Mahmood:
If there are no issues with this? Could you please merge this? Checking for supported model names by Anthropic only makes it more maintainable and improves the UX for the developers. So instead of going down wrong rabbit holes you are provided with correct error messages. We can always expand this list... The existing LM class only supported one model claude-2 instant and it's not obvious at all if you run it with other models when it breaks what the problem is to people who are not very familiar with the dspy env.

okhat:
But this means whenever Anthropic adds new models we’ll have to update our class. Why not let the Anthropic API just raise its own exception

arnavsinghvi11:
Agreed on having Anthropic API raise its own errors when triggered (and users can reference this [documentation](https://docs.anthropic.com/claude/reference/errors) as needed). Let us know if you have any other concerns with this approach @Arbaaz-Mahmood !

Arbaaz-Mahmood:
I agree @arnavsinghvi11 @okhat. I have defaulted the model to claude-3-opus and allowed for arbitrary selection of models instead of raising value errors from my side. One thing I did notice however is that, the main repo's functionality for claude and other recent additions is not available when installed via pip. Someone should look into that.

Arbaaz-Mahmood:
@okhat ruff his has been applied. Kindly re-run

Arbaaz-Mahmood:
Hi! @okhat I think this is safe for merge. Could you merge?

PR #483, opened 2024-02-27
feat: add from_parquet to dataloader

JamesHWade:
A parquet file loader would be a convenient addition to the dataloader class. I particularly like that parquet files preserve types better than a csv file.

The primary change is the addition of the from_parquet() function:

```python
def from_parquet(self, file_path: str, fields: list[str] = None, input_keys: tuple[str] = ()) -> list[dspy.Example]:
        dataset = load_dataset("parquet", data_files=file_path)["train"]

        if not fields:
            fields = list(dataset.features)

        return [dspy.Example({field: row[field] for field in fields}).with_inputs(input_keys) for row in dataset]
```

The rest of the changes were caused by the ruff precommit hook.

PR #709, opened 2024-03-25
[feat] add from_parquet to dataloader

JamesHWade:
A parquet file loader would be a convenient addition to the dataloader class. I particularly like that parquet files preserve types better than a csv file.

Here's the change:

```python
def from_parquet(self, file_path: str, fields: list[str] = None, input_keys: tuple[str] = ()) -> list[dspy.Example]:
        dataset = load_dataset("parquet", data_files=file_path)["train"]

        if not fields:
            fields = list(dataset.features)

        return [dspy.Example({field: row[field] for field in fields}).with_inputs(input_keys) for row in dataset]
```

Note, this PR was originally made in #483, but something with git hooks prevented test from being able to run.

### DIFF
--- a/dspy/datasets/dataloader.py
+++ b/dspy/datasets/dataloader.py
@@ -62,6 +62,13 @@ class DataLoader(Dataset):
         
         return [dspy.Example({field:row[field] for field in fields}).with_inputs(*input_keys) for row in dataset]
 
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @JamesHWade !

PR #665, opened 2024-03-17
feat(program_of_thought): added support for custom import whitelist and added dynamic input and output field names for PoT 

Arbaaz-Mahmood:
While, the PythonInterpreter and CodePrompt modules allowed for importing a white_list, this functionality hadn't been made available directly inside ProgramofThought. This was causing repeated hinderance when language models would try to use any math packages that were not allowed like sympy etc, which makes for a bad dev UX.

PR #710, opened 2024-03-25
feat(ProgramofThought): added support for custom whitelists and dynamic input and output field naming (continuation of #665)

Arbaaz-Mahmood:
While, the PythonInterpreter and CodePrompt modules allowed for importing a white_list, this functionality hadn't been made available directly inside ProgramofThought. This was causing repeated hinderance when language models would try to use any math packages that were not allowed like sympy etc, which makes for a bad dev UX.
This also updates the ProgramOfThought class to support dynamic input and output field names, improving user experience and flexibility. The changes include:
Modifying the __init__ method to store input and output fields from the provided signature
Updating the generation of input and output field names in the instruction strings
Refactoring the _generate_signature and _generate_instruction methods to handle dynamic field names
Modifying the forward method to use dynamic input field names instead of hardcoded names
These changes allow users to define their own input and output field names in the signature, making the ProgramOfThought class more adaptable to different use cases.

### DIFF
--- a/dspy/predict/program_of_thought.py
+++ b/dspy/predict/program_of_thought.py
@@ -8,20 +8,22 @@ from ..primitives.python_interpreter import CodePrompt, PythonInterpreter
 
 
 class ProgramOfThought(Module):
-    def __init__(self, signature, max_iters=3):
+    def __init__(self, signature, max_iters=3, import_white_list=None):
         super().__init__()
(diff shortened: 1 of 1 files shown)


Arbaaz-Mahmood:
@arnavsinghvi11 here.

arnavsinghvi11:
Thanks @Arbaaz-Mahmood !

PR #712, opened 2024-03-25
fix(dsp/modules/gpt3): fix system_prompt overwrite, as discussed in PR #618 comments

mlederbauer:
The system_prompt is currently overwritten when setting kwargs[“messages”]. This is fixed by setting kwargs[“messages”] = messages, where messages contains the system prompt if self.system_prompt is given.
Full credit to @Markovian99 for spotting this bug in PR #618 comments.


arnavsinghvi11:
Thanks @mlederbauer !

PR #716, opened 2024-03-25
Update Cohere LM to use  the `/chat` API and add support for Command-R

CShorten:
Cohere has deprecated the /generate API in favor of /chat. More info can be found here: https://docs.cohere.com/reference/chat.

Implementation follows the example of `dsp/modules/gpt3.py`.

[ToDo in the future, not in this PR]
- Would be interesting to look into the `connectors` and `documents` APIs.
- Also would be good to add the same logging in OpenAI to Cohere, `log_usage`


beatobongco:
Love this. Would want to use command-r w/ dspy!

However seems like predict expects `num_generations`.

```
  File "/opt/venv/lib/python3.9/site-packages/dspy/predict/predict.py", line 82, in forward
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Trying to add that when instantiating the `Cohere` class then passes that kwarg to `co.chat` which is an unexpected keyword :( 

`TypeError: chat() got an unexpected keyword argument 'num_generations'`

CShorten:
Ah great catch!!... hm, ok looking into this -- it looks like the `Chat` API doesn't support multiple generations. Let me see how `Predict` calls it if I can access which LM it's calling.

CShorten:
Ok, I think we can put `num_generations` in `self.kwargs` to access it in `Predict` and then pop it out of `kwargs` without modifying the `self.kwargs` state before the call is made.

CShorten:
Ok I tested this works with:

```
import dspy
from dspy import Cohere, Predict, Signature

class test(Signature):
    """say hello to the given name."""

    name = dspy.InputField()
    greeting = dspy.OutputField()

command_r = Cohere(model="command-r", api_key="cohere_key")

print(command_r("say hello"))

dspy.settings.configure(lm=command_r)

greeting_module = dspy.Predict(test)
print(greeting_module(name="Connor").greeting)
print(dspy.settings.lm.inspect_history(n=1))

beatobongco:
@CShorten in an attempt to get it to work, I forked this last night and did this to only pass kwargs expected by `chat`: https://github.com/beatobongco/dspy/blob/beato/update-cohere-chat-api/dsp/modules/cohere.py#L86-L98

I forgot the exact error I got but generally its about `chat` complaining about kwargs. I think it didn't like `n` this time.

Context: I am using it in a project where we have a few modules and signatures.

I got it to work, but somehow running  Modules would result in single character outputs e.g. `H` 

Switching to `dspy.OpenAI` does not have this problem. 

I think I need to dive deeper to understand what is exactly happening. Doing `inspect_history` I get the prompt mostly right... but not exactly. I think there may have been a wrong input somewhere.

CShorten:
A supported keys type check could be interesting -- this is also quite related to the design of the RMs and optional search APIs in kwargs.

Yes -- I saw that as well. It is taking a `[0]` somewhere, fixed on line 55 by wrapping the response text in a list, `return [response.text]`.

Haha indeed, inspect_history helped enormously for figuring this out.

I have tested and I think this is now good to merge, I will take one more look tomorrow.

Issue #662, opened 2024-03-16
Support Chat Mode


thomasahle:
Hopefully the new LM backend will allow us to make better use of models that are trained for "Chat".
Below is a good example of how even good models like GPT-3.5 are currently have trouble understanding the basic DSPy format:
<img width="649" alt="Screenshot 2024-03-15 at 6 46 18 PM" src="https://github.com/stanfordnlp/dspy/assets/946355/4fd177f1-cc5b-424e-8182-047cbc002f30">

Right now we use chat mode as if it was completion mode.
We send:
```
messages: [
{"from": "user", "message": "guidance, input0, output0, input1, output1, input2"}
}
```
And expect the agent to reply with
```
{"from": "agent", "message": "output2"}
```
A better use of the Chat APIs would be to send
```
messages: [
{"from": "system", "message": guidance},
{"from": "user", "message": input0},
{"from": "agent", "message": output0},
{"from": "user", "message": input1},
{"from": "agent", "message": output2},
{"from": "user", "message": input2},
]
```
That is, we simulate a previous chat, where the agent always replied with the output in the format we expect.
This teaches the agent to not start it's message with "OK! Let me get to it!" or repeating the template as in the gpt-3.5 screenshot above.

Also, using the system message for the guidance should help avoid prompt injection attacks.

okhat:
Totally agreed but I'd love for this to be more data-driven.

Either: (a) meta prompt engineering for popular models + easy addition of new adapters for new LMs if needed, or (b) automatic exploration of a new LM on standard tasks to automatically establish the patterns that "work" for that LM.

Do you have a way that fixes the sql_query example you had?

okhat:
Also I wonder to what extent this behavior you saw is because of "Follow the following format." is not explicitly saying "Complete the unfilled fields in accordance with the following format." Basically the instruction is slightly misleading for chat models.

CyrusOfEden:
My sense is that interleaving inputs/outputs as a default would be a footgun because I would assume all outputs depend on all inputs, and the LLM doesn’t have access to this. 

Right now our focus is using LiteLLM for broad support + moving over all dsp code into DSPy. 

I’d love to tackle something like this when we look at the current Template usage and how that’s currently responsible for going from Example => prompt, and offering users some more flexibility with how an LLM gets called with an example.

thomasahle:
I'm not sure what you mean by "interleaving inputs/outputs". This is already how DSPy works, no?

[...]

PR #717, opened 2024-03-25
Final API Changes for Backend LM Refactor

KCaverly:
While working through some examples for the JSONBackend, I realized a few small changes were needed to the Template.

As such, I've abstracted the Templates, and allowed them to be passed directly during backend generalization/generation.
This PR will remain a Draft until, I am confident that there are no changes in functionality to existing examples, and new Backend examples are available. This PR is being drafted early for visibility.


thomasnormal:
This is great!

What about having a `Content` class instead of content just being a string?
The content class can contain a list of texts or images like in the GPT Vision API: https://platform.openai.com/docs/guides/vision/multiple-image-inputs

And `extract` in `BaseTemplate` can take a `Content` instance instead of a `str` in `raw_pred`.

dspy/primitives/prompt.py
```diff
@@ -0,0 +1,22 @@
...
+    def is_valid(self):
+        return self.content is not None or self.messages is not None
+
+    def get_messages(self):
+
+        if not self.is_valid:
+            raise ValueError("Prompt has not been initialized with a content or messages")
+
+        if self.messages is not None:
+            return self.messages
```
thomasnormal, line 20: Would it be simpler to have `Prompt` only have `messages` and then make a subclass `SimplePrompt` which just has `content` and wraps it as `[{"role": "user", "content": self.content}]`?

dspy/modeling/templates/json.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Generate spans for all the demos
+        for demo in example.demos:
+            prompt_spans.append(self._example_span(signature, demo))
+
+        # Generate span for the active example
+        prompt_spans.append(self._example_span(signature, example))
+
+        content = "\n\n--\n\n".join(prompt_spans)
+        return Prompt(content=content, messages=[{"role": "user", "content": content}])
```
thomasnormal, line 29: Following https://github.com/stanfordnlp/dspy/issues/662 this could also be
```
messages: [
{"from": "system", "message": guidance},
{"from": "user", "message": input0},
{"from": "agent", "message": output0},
{"from": "user", "message": input1},
{"from": "agent", "message": output2},
{"from": "user", "message": input2},
]
```

But maybe we should just make a separate JSONChatTemplate?

thomasnormal:
Also, regarding json, how can se set `response_format={ "type": "json_object" }` when using the `JSONTemplate`?
Maybe the template class needs to support giving extra "kwargs" to the LM? Or maybe the LMs just need to detect the JSONTemplate and enable it themselves?

dspy/primitives/prompt.py
KCaverly: As some models presumably would take in just a string, I imagine we want a structure in which we can provide either a text prompt, or ChatML messages. I was thinking of managing this with simpler constructors, ie. something like this:

`Prompt.from_str(content: str)`

And then having the LM identify what it needs from the Prompt.
Im pretty easy on whatever this looks like, just want to prefer keeping less classes, over having one class for each variant.

KCaverly:
> Also, regarding json, how can se set `response_format={ "type": "json_object" }` when using the `JSONTemplate`? Maybe the template class needs to support giving extra "kwargs" to the LM? Or maybe the LMs just need to detect the JSONTemplate and enable it themselves?

Extra LM args, are managed directly by the `JSONBackend`, but if we manage both prompts and extraction via Templates, there is limited difference between JSON/Template Backends.

CyrusOfEden:
Let's hop on a Discord call tomorrow or Friday to go through this :)

KCaverly:
@thomasahle I've incorporated your feedback from the above, simplifying the Prompt. I've also ensured that we have full type checked JSON deserialization for messages from Pydantic, and updated this format to accomodate for both basic text messages, along with GPT4-V compatible images.

KCaverly:
I've tested this across the example notebooks, and added a new "COT With Backend" example for GSM8K, which uses the JSONBackend along with a custom Template. This should be ready to merge to backend-refactor, and ultimately review and merge with main.

@CyrusOfEden

dspy/modeling/templates/json.py
```diff
@@ -0,0 +1,87 @@
...
+        example = example.copy()
+
+        try:
+            pred = json.loads(raw_pred)
+            for k, v in pred.items():
+                k = k.lower()
+                if k in signature.fields:
+                    example[k] = v
+        except:
+            pass
```
thomasahle, line 42: What error are we hiding here?
Json doesn't parse?
Should we at least give a warning or something?

thomasahle:
Super excited about this!

I didn't realize the difference between Backends and Templates.
Can you use any combination from the cartesian product of `templates x backends x [Predict, TypedPredictor]`?

dspy/modeling/templates/json.py
KCaverly: There are scenarios in which json does not parse. I think you are right in that we should provide a warning. I was planning on going through once this hits main, and updating all the print/todos surrounding logging at once.

KCaverly:
I am still slightly unclear on the full boundaries between these objects, as they are currently tightly coupled. But as I think about it:

`Signature`: Establishes problem schema/intention
`Example`: Provide examples and new problems to complete.
`Template`: Converts the Signature/Example into a format which can be leveraged by Language Models, and converts back to Complete/Incomplete Examples.
`BaseLM`: Completes given a Prompt.
`TemplateBackend`: Co-ordinates the entire generation piece. Takes in a Signature/Example/Template and Language Model and returns complete generations. This manages for orchestration and retries/healing.

Currently we basically only have a `TemplateBackend` which manages for structured generation via `Template` extraction, but this opens us up to have something like a `InstructorBackend` , or an `OutlinesBackend` which may manage for Prompt/Structured output in slightly different ways. Obviously this has impacts across the Module suite as well as with something like an `InstructorBackend` , all `Predict` objects become `TypedPredictor` objects automatically etc.

Currently, we set the backend during config, in a similar way to historically managing LMs:

```python
from dspy.modeling import LiteLM, TemplateBackend

lm = LiteLM(model="gpt-3.5-turbo", default_params={"max_tokens": 500})

# lm, is any language model which implements BaseLM
template_backend = TemplateBackend(lm=lm, attempts=3)

dspy.settings.configure(backend=template_backend)
````

Currently, each Backend has a default Template, which can be overridden during generation. But we could easily allow this to be set at the class level as well.

````python

# The below would generate Completions for the Signature/Example/Template provided.
template_backend(signature=..., example=..., template=JSONTemplate())
````

**Review by Josephrp — COMMENTED** [review]
interesting open questions and improved examples

**Comment by thomasahle** [comment]
Thanks, this is helpful. 

> this has impacts across the Module suite as well as with something like an `InstructorBackend` , all `Predict` objects become `TypedPredictor` objects automatically etc.

I feel a need to mention TypedPredictor has many tricks not in Instructor ;-), such as
- Explaining errors (Reflexion paper)
- Generating demo instances to help LMs understand schema.

Some other thoughts:

Could the extract method take a Content object instead of a raw string? Then we can also support image generation modules(!)

Would it make sense to set the backend on the LM, or in some other way allow the LM to set a default template?

I didn't realize `Example` had a `demos` attribute. I've always just used a `list[Example]`.

**Comment by KCaverly** [comment]
@CyrusOfEden and I took a look at this yesterday, and we tried to simplify it a little bit. We were a bit concerned about Class sprawl generally, and wanted to ensure we kept it flexible without needing 100 different objects.

Essentally we combined `Backend` and `Template`, and dropped the concept of a `Prompt`.

Ie. A Backend now looks like the following:

````python
class BaseBackend:
     ...
     
     def prepare_request(self, signature: Signature, example: Example, config: dict, **kwargs) -> dict:
         """Given a Signature, Example and LM Config, generate a dictionary of kwargs for the LM"""
         ...
         
     def process_response(self, signature: Signature, example: Example, output: LMOutput, **kwargs) -> Completions:
         """Given a Signature, Example and LMOutput, extract the appropriate field information and create Completions"""
         ...
````

This serves a similar way to the prior `Template`, essentially providing Generate/Extract functionality. the BaseBackend has default methods to generate content given an LM and these two methods. So for most use cases we can just Subclass `BaseBackend` with these two methods overridden and it will work like the current method.

With this we still remain super flexible and ensure the ability to almost replay all LM functionality, there is one primary drawback I can see.

This results in Limited API Boundary between `BaseLM` and `BaseBackend`.
- As the `BaseBackend.prepare_request` function, generates the entire kwargs object for the LM, the responsibility for augmenting the Request params for the `BaseLM` primarily sits with the `BaseBackend`.
- The `BaseLM` implementation, could of course sanitize this, but we no longer have any fixed "prompt" argument at the `BaseLM` level.

This works fine with the LiteLLM Model, but I am unsure how this would scale as we introduce more Language Model providers. The next step for me, would be to implement an additional LM Provider, and see how well this API provides for both. If anyone has any thoughts on this, I would be keen to hear.

**Comment by CyrusOfEden** [comment]
Love it! Go for it

**Review by Josephrp — APPROVED** [review]
really neat solution to a tough problem 🚀

**Comment by KCaverly** [comment]
@CyrusOfEden and I worked through a few other edge cases, and we have temporarily settled on the following:

Instead of having a seperate Template/BaseLM and Prompt objects, we will just have a single `BaseBackend` abstraction.
The backend, will consist of `prepare_request` and `process_response` methods. These methods will be overridable for each prompt variant people are interested in.

Ie. We now have a `TextBackend`, which uses LiteLLM behind the scenes, and mocks the existing `LM` solution in `dsp.modules`. We have also provided a `JSONBackend` which overrides the `prepare_request` and `process_response` methods for the `TextBackend` to use JSON relevant kwargs and prompts.

Moving forward, if you want to use a custom prompt format, you can subclass a suitable backend, and augment the `prepare_request` and `process_response` methods to get whatever behaviour you would like.

Please let us know what you think of something like this. We moved to this method for the following reasons:
1. We likely dont need any other vanilla Language Model backend, that is not LiteLLM. Thus it makes no sense to abstract LiteLLM out individually.
2. Prompt generation, and extraction are heavily related. Given the current vanilla LM backend, there is no way to augment the prompt without also augmenting the extraction methods individually. While we could abstract this to a `Template` or something, the templates are likely still backend specific. Ie. we could not have a single Prompt abstraction that would work for both LiteLLM/Instructor for example.

The old method would have looked something like this:

```python
# Provide a JSON compatible prompt template
backend = LiteLLMBackend(model="gpt-3.5-turbo", mixin=JSONMixin())
```

Whereas now, it would look something like this:

```python
# Provide a JSON compatible prompt format directly in the Backend.
backend = JSONBackend(model="gpt-3.5-turbo")
```

As we introduce a Backend for Instructor, this may make prompting more modular. But for now, we wanted to keep it simple to start, and build on pain points as we go. If there are any thoughts on this from the community, I am all ears.

isaacbmiller:
> We likely dont need any other vanilla Language Model backend, that is not LiteLLM. Thus it makes no sense to abstract LiteLLM out individually.
I agree with this. We can mostly operate with just LiteLLM, but I would imagine that we want to keep the door open for other backends. #722 is requesting Portkey integration. Maybe there's a world where we only officially support LiteLLM?

CyrusOfEden:
@isaacbmiller we are gonna have 1 standard way for everything that works for most people (LiteLLM) and a interface for others to extend/implement for other providers. Using the prepare_request / process_response paradigm and considering how similar Portkey is to LiteLLM it should be straightforward for others to subclass this in their own projects.

We can explore a .contrib directory for these things, or a separate contrib repo, but for now let's be a little more opinionated on this WHILE giving people the power to extend our opinions.

RahulBalan:
[KCaverly]
Can you please provide an example of how to use chat history with DSPy

A simple example of how we used to work without DSPy till now is.
---
client = OpenAI( base_url ="http://localhost:11434",api_key='ollama')
messages: [
    {"role": "system", "content": "You are a ai assistant"},
    {"role": "user", "content":"Hi am Rahul"},
    {"role": "assistant", "content": "What can i do for you rahul"},
    {"role": "user", "message": "whats the capital of France"}
]
response = client.chat.completions.create(
                    model="mistral,
                    messages=messages
                    )

------
Using dspy , this code can be changed to 

turbo = dspy.OpenAI(api_base='http://localhost:11434/v1/', api_key='ollama', model='mistral', model_type='chat',max_tokens=10000)
dspy.settings.configure(lm=turbo)
classify = dspy.Predict('question -> answer')
classify(question="whats the capital of france").answer.

In this DSPy example, how can we pass the other messages also , like i did in my previous example. 

(What i understood from DSPy is, it generates a single message in which the question, examples etc every thing is included. History also can be included in that single message. 
but am not asking example of that. 
in previous example, we could sent a list of messages. is that possible via DSPy also ?

CyrusOfEden:
I don’t feel like DSPy is suited for you trying to do a chat app, but I could be wrong

thomasahle:
@RahulBalan I'm afraid I agree with @CyrusOfEden on this.
The idea of DSPy is to include multiple few-shot examples of other conversations at the top, so the model can use in-context learning. To do this you'll need to send multiple existing conversations before the current one you are in.
But that could also confuse the model, and it would start using information from the examples that you wouldn't want it to. You could try to tell it not to, but I'm worried it would be hard.

PR #725, opened 2024-03-27
docs: fix cheatsheet no RM loaded error in ReAct

hanxiao:


okhat:
Actually just pass tools=[] ?

Issue #730, opened 2024-03-28
Opaque output truncation as `max_tokens` is not explicitally included in the __init__ for GPT3 (OpenAI) 


harrysalmon:
Currently the output tokens for Openai is set to 75, half of the value of the max tokens specified here in gpt3.py 

```
...
self.kwargs = {
    "temperature": 0.0,
    "max_tokens": 150,
    "top_p": 1,
    "frequency_penalty": 0,
    "presence_penalty": 0,
    "n": 1,
    **kwargs,
}  # TODO: add kwargs above for </s>
...
```

As this is only setable through kwargs, there's no indication to the user in the OpenAI class why the output is being truncated or how to modify it. 

I might be missing a reason for this, but I'll open a PR to modify it and see what folks think.

okhat:
Thank you @harrysalmon ! Left some quick comments on the PR. It's a bit more complex than it seems.

PR #734, opened 2024-03-28
enh(dspy): Move max_tokens explicitally into `GPT3` constructor

harrysalmon:
Makes the reason for output truncation clearer as per #730

Also adds a comment to the predict file where the max tokens are modified before the call is made.

### DIFF
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -111,6 +111,7 @@ def _generate(template: Template, **kwargs) -> Callable:
             max_tokens_key = "max_tokens" if "max_tokens" in keys else "max_output_tokens"
             new_kwargs = {
                 **kwargs,
+                # Set the required max tokens key to the new value.
                 max_tokens_key: max_tokens,
                 "n": 1,
                 "temperature": 0.0,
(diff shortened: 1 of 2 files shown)


dsp/primitives/predict.py
```diff
@@ -111,6 +111,7 @@ def do_generate(
             max_tokens_key = "max_tokens" if "max_tokens" in keys else "max_output_tokens"
             new_kwargs = {
                 **kwargs,
```
harrysalmon, line 113: How the value set by the user effects the maximum tokens passed and returned from the model is quite unclear here, would anyone be able to clarify how 101:118 relate to the eventual values passed to different models, I'll have a crack at clearing this up. Might it be possible to do the `75` checking upstream of here?

okhat:
This is mainly during retries etc. It's not very straightforward to change.

okhat:
Basically, the user's max_tokens is used at first, then max_tokens / 2 if we need more fields, but 75 is sort of like a min_max_tokens. Setting a large enough max_tokens could be all that's required here?

harrysalmon:
Thanks for taking a look @okhat 

Yeah the main thing for me is putting it into the init so it's clear to the user that they need to set it. What do you think of this part? 

In terms of the retries, do you mean if `max_tokens` doesn't leave the LLM enough tokens to pass the full input tokens, it'll retry with `max_tokens / 2` and then with 75?

arnavsinghvi11:
@harrysalmon just following up on this. I think we can merge the change but leave the existing retry behavior. lmk if it's ready to merge.

arnavsinghvi11:
Closing for now to avoid breaking cache errors.

PR #742, opened 2024-03-30
docs: examples for saving and loading modules after optimizing

nsobadzhiev:
## 📝 Added information about saving and loading modules in the documentation

* A new section in the Optimizers Building Blocks page about saving and loading an optimized program
* A new section in the Cheat Sheet about saving and loading modules


Josephrp (approved):
i think these additions really help the audience do something interesting quicker , thanks for the attention to this

PR #750, opened 2024-04-01
Add Custom Rationale Type option in TypedChainOfThought Module

krypticmouse:
Add an argument for custom rationale type like:

```python
        rationale_type = dsp.Type(
            prefix="Take a deep breath and let's think step by step in order...",
            desc="Based on the context...",
        )

        self.task = dspy.TypedChainOfThought(SomeTask, rationale_type=rationale_type)
```

### DIFF
--- a/dspy/functional/functional.py
+++ b/dspy/functional/functional.py
@@ -62,7 +62,7 @@ def TypedChainOfThought(signature, instructions=None, rationale_type=None, *, ma
         prefix="Rationale: Let's think step by step in order to",
         desc="${produce the " + output_keys + "}. We ...",
     )
-    rationale_type = DEFAULT_RATIONALE or rationale_type
+    rationale_type = rationale_type or DEFAULT_RATIONALE
 
     return TypedPredictor(
         signature.prepend(
--- a/dspy/functional/functional.py
+++ b/dspy/functional/functional.py
@@ -59,7 +59,7 @@ def TypedChainOfThought(signature, instructions=None, rationale_type=None, *, ma
     output_keys = ", ".join(signature.output_fields.keys())
 
     DEFAULT_RATIONALE = dspy.OutputField(
-        prefix="Rationale: Let's think step by step in order to",
+        prefix="Reasoning: Let's think step by step in order to",
         desc="${produce the " + output_keys + "}. We ...",
     )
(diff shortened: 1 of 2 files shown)


Josephrp (commented):
is this change required ?

Josephrp:
> Thought it might come handy 🤔

yeah absolutely handy, had not noted my previous review of the main changes , so that's why i made that comment - with all my appologies  ! i'm actually testing it out , because i'm in the same situation , trying out stuff and building demos :-)

krypticmouse:
No worries at all!! Thanks for the comment!!!

thomasahle:
Technical question: Can you prepend a dsp.Type to dspy.Signature? I'm pretty sure you can only prepend dspy.Fields, no?

Practical question: The TypedChainOfThought function is already doing very little work (just one call to signature.prepend). Why not just add the alternative field to the signature yourself, and then use TypedPredictor?

krypticmouse:
> Can you prepend a dsp.Type to dspy.Signature? I'm pretty sure you can only prepend dspy.Fields, no?
 
 You can't prepend dsp.Type ig. But the rationale_type being dsp.Type is deprecated now and is replace with dspy.OutputField overall.
 
 > The TypedChainOfThought function is already doing very little work (just one call to signature.prepend). Why not just add the alternative field to the signature yourself, and then use TypedPredictor?
 
 Good point. I think that'll reduce the friction of migrating away from `ChainOfThought` to `TypedChainOfThought`. I agree there is not much that is happening but I think this sort of functionality being native will reduce the burden on user "to do something" on their end.

arnavsinghvi11:
Thanks @krypticmouse ! Is this good to merge? LGTM, but checking if anything was left

thomasahle:
>  think that'll reduce the friction of migrating away from ChainOfThought to TypedChainOfThought.  

Ok, fair enough.

But can we at least rename it to `rationale_field`, since it's no longer a dsp.Type?

Oe maybe `reasoning_field`, since we are no longer calling the field rationale. So if people want to get the output value, they need to check output reasoning.

Or maybe just `reasoning` since we don't normally add the type of a variable to the name?

krypticmouse:
Yes we can! Wrote that just to keep the convention same, lemme edit it out and merge!

Issue #739, opened 2024-03-29
pytest failing in dev container


drawal1:
ImportError while importing test module '/workspaces/dspy/tests/teleprompt/test_mipro_optimizer.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/teleprompt/test_mipro_optimizer.py:4: in <module>
    import dspy
dspy/__init__.py:1: in <module>
    import dsp
dsp/__init__.py:1: in <module>
    from .modules import *
dsp/modules/__init__.py:1: in <module>
    from .anthropic import Claude
dsp/modules/anthropic.py:5: in <module>
    import backoff
E   ModuleNotFoundError: No module named 'backoff'

drawal1:
After doing a 'pip install .', I get No module named 'structlog'

drawal1:
Optional dependencies for dspy.Bedrock are not specified anywhere. Should we setup a group in toml for each module?

drawal1:
> Optional dependencies for dspy.Bedrock are not specified anywhere. Should we setup a group in toml for each module?

This module should probably be removed once LiteLLM is integrated

okhat:
Thank you @drawal1 ! Can we tag the relevant people here? The dev container was just merged last week

drawal1:
@Joshmantova - can you take a look at build fail due to backoff and structlog? ignore the posts re. bedrock

josh-mantovani:
> @Joshmantova - can you take a look at build fail due to backoff and structlog? ignore the posts re. bedrock

Yup, I'll get a fix up today. This also pushes the priority of getting dev container tests in CI. For now, just install the package manually within your container but I'll fix it shortly. Thanks for bringing it to my attention!

Joshmantova:
@drawal1 - this actually isn't a bug. When a Python interpreter is selected in VSCode for a given workspace, it will remember the interpreter you chose until you choose a different one. It doesn't look like there's a way to override the interpreter chosen if one has previously been selected for the workspace. To resolve the issue, you need to be using the Poetry virtual env python interpreter rather than whichever one you currently have selected.

[...]

PR #753, opened 2024-04-02
Increase chances that VSCode dev container will use Poetry venv

Joshmantova:
## 📝 Changes Description

* Builds the dev container Poetry venv within the workspace
* Includes VSCode setting to choose the poetry Python interpreter by default
* Includes additional warning about interpreter at the end of the dev container build to provide instructions on how to choose the correct interpreter if it's not already selected
* Updates contributing markdown file with these additional instructions

We got reports that people were not able to run the unit tests after building the dev container. After further investigation, I found that this is not due to the dev container itself but instead in the Python interpreter selected by the user. If the user has the OS default Python interpreter selected, that interpreter will not have access to the packages installed by Poetry. Because poetry creates it's own venv, we need to use the Poetry interpreter rather than any of the other interpreters installed in the dev container. By default, the dev container should find and use this Poetry interpreter but if a user has previously selected a different interpreter for the workspace, it will not attempt to find the Poetry interpreter. 

To resolve this, we're adding additional instructions in the logs of the dev container build as well as in the `contributing.md` file. Additionally, the Poetry venv will now be built within the VSCode workspace rather than outside of it. Because it's inside the workspace, we're now able to add the Poetry interpreter as the default interpreter in the VSCode settings. If a user has selected a different interpreter, this setting will not override that selection but it will make it easier to select the correct interpreter as this makes the Poetry interpreter not only the recommended interpreter but also the default. The instructions to ensure developers are using the correct interpreter should now be hard to miss and the issue should be less likely to occur in the first place. Unfortunately, there's no way to force VSCode to use the Poetry interpreter. The best we can do is set it explicitly as the default.

## ✅ Contributor Checklist

## ⚠️ Warnings

### DIFF
--- a/.devcontainer/interpreter_warning.py
+++ b/.devcontainer/interpreter_warning.py
@@ -1,0 +1,25 @@
+from rich.console import Console
+from rich.text import Text
+
+text = Text()
+console = Console()
+
+text.append(
+    "Be sure to check that you have the Poetry Python interpreter selected in VSCode otherwise"
+    " your terminal will not include the necessary packages to develop. If you have the correct interpreter"
+    " configured, you'll see an interpreter that looks like this: ",
+    style="bold yellow",
+)
+text.append("3.9.18 ('.venv': Poetry)", style="code")
+text.append(
+    ". If you have previously selected another Python interpreter for the VSCode workspace, you may need to select"
+    " the Poetry Python interpreter.\n\nTo do so, follow the instructions at the following link and choose the Poetry"
+    " virtual environment interpreter.",
+    style="bold yellow",
+)
+
+console.print(text)
+console.print(
+    "VSCode Python Interpreter Documentation\n",
+    style="link https://code.visualstudio.com/docs/python/environments#_working-with-python-interpreters",
+)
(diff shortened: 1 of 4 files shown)


Joshmantova:
https://github.com/stanfordnlp/dspy/issues/739

.devcontainer/interpreter_warning.py
Joshmantova, line 1: Warning looks like this:

Screenshot 2024-04-01 at 9 13 59 PM

.devcontainer/post_create.sh
```diff
@@ -6,6 +6,7 @@ git config --global --add safe.directory /workspaces/dspy
 
 pip install poetry==1.7.1
 poetry config installer.max-workers 4
+poetry config virtualenvs.in-project true
```
Joshmantova, line 9: This just makes it so the poetry venv is created within the workspace. This, combined with the additional setting specified in `devcontainer.json`, makes it more likely that VSCode picks it up as the correct interpreter. If someone has previously selected a different interpreter for this workspace, this will not override that selection but it will make changing it to the correct interpreter a bit easier.

Joshmantova:
@okhat - This clears up a bit of the confusion in the linked issue and provides additional instructions so others don't have similar issues. Small changes here, it's tested and good to go.

PR #760, opened 2024-04-03
Remove top_p default value

xdite:
Pratically, we should not use top_p and temperture at the same time, however, we may override  temperture in some cases, but since DSPy predefine `top_p = 1`, it's impossible to leave `top_p` as blank.

this weired behavior is due to there are very diffrent results from direct openai call V.S. DSPy.

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -87,7 +87,6 @@ class GPT3(LM):
         self.kwargs = {
             "temperature": 0.0,
             "max_tokens": 150,
-            "top_p": 1,
             "frequency_penalty": 0,
             "presence_penalty": 0,
             "n": 1,


arnavsinghvi11:
Thanks @xdite . Yes ideally, we can remove these definitions and have users define all the model kwargs directly (some models are not compatible with frequency_penalty, presence_penalty, etc.) but tagging @okhat to verify this won't break any existing caches with GPT3?

okhat:
This will break existing caches. Maybe this can somehow be marked as "for next breaking release" and then merged in the future.

But I don't understand the need for this. Can't the user just pass None? Why can't we specify top_p and temperature at once? What's the effect of this PR?

ammirsm:
hey folks,
just a follow up on this one, 
should we close this pull request?

Issue #591, opened 2024-03-07
BootstrapFewShot failing due to lm.copy for AzureOpenAI


insop:
Since https://github.com/stanfordnlp/dspy/issues/521 is closed state, creating a new issue.

Now, the initial PR https://github.com/stanfordnlp/dspy/pull/522 for the issue was reverted by this PR, https://github.com/stanfordnlp/dspy/pull/584 recently, I think we need to keep this issue.

----

BootstrapFewShot failing due to lm.copy for AzureOpenAI, likely due to positional argument issue from recent changes.

- Issue: https://github.com/stanfordnlp/dspy/issues/521

### Environment
- code: 98304a2eb9d1dddaaa846e30258cd5d8fc6b5d8d
- model azure openai

### Error message
- during `BootstrapFewShot` run, code errored out due to missing arguments, `api_version`

```
>[dspy/dsp/modules/lm.py](https://file+.vscode-resource.vscode-cdn.net/Users/insop/Projects/Github/NLP/dspy/dsp/modules/lm.py)(106)copy()
    102         model = kwargs.pop('model')
    105
--> 106         return self.__class__(model, **kwargs)
```

### Changes

AzureOpenAI expects positional arguments, so it is failing. I am putting my change for review if that will impact other models.

```
class AzureOpenAI(LM):
...

    def __init__(
        self,
        api_base: str,
        api_version: str,
        model: str = "gpt-3.5-turbo-instruct",
        api_key: Optional[str] = None,
        model_type: Literal["chat", "text"] = "chat",
        **kwargs,
    ):

arnavsinghvi11:
thanks for circling back on this @insop . What were the proposed changes for handling this?

insop:
Thank you @arnavsinghvi11 .

I catch up the latest main and will propose soon.

snimu:
I've run into the same issue, and it looks to me like the problem isn't8 that `api_base` needs to be renamed to `azure_endpoint`, but that `dspy.AzureOpenAI` hasn't redefined the `LM.copy`-method, so it falls back on that. For that method to work, `self.kwargs` in `AzureOpenAI` would have to include `api_base` and `api_version`, but they don't. So either, `AzureOpenAI.copy` should be implemented (preferred), or `api_base` and `api_version` have to be added to `AzureOpenAI.kwargs`.

PR #769, opened 2024-04-04
fix(dspy): add `copy` to AzureOpenAI to propagate `api_(key|version|...)`

snimu:
Fixes issue #591 without re-creating issue #543

The problem was that `AzureOpenAI.copy` used `LM.copy`, but because `api_key` etc. aren't in the `AzureOpenAI.kwargs`, and `LM.copy` per default only propagates `self.__class__.kwargs`.

I now gave `AzureOpenAI` its own `copy`-method. This isn't really beautiful, but it seems to work pretty well.


arnavsinghvi11:
Thanks @snimu !

PR #774, opened 2024-04-04
Fix the bug of lm in predict

minfawang:
In l71 of predict.py, there is a logic to use the lm from the kwargs: `lm = kwargs.pop("lm", self.lm) or dsp.settings.lm`, but the `generate` call does not use it. This PR fixes the bug by using the lm as the context.

Also, the previous comment stated that it's unclear why `query_only` should ever be used, and this is the only place the `query_only` option is used. So it sounds like a good choice to just get rid of this option and simplify the code logic.

Reference:

1. [self.lm in generate](https://github.com/stanfordnlp/dspy/commit/bc23a70f8f051903a9034b41a2a7837536cc3837) was introduced on 08/24/2023
2. [lm kwargs](https://github.com/stanfordnlp/dspy/commit/9cf152317f53db76c879ba9a6870479cf78f440c) was introduced on 10/11/2023

### DIFF
--- a/dsp/templates/template_v2.py
+++ b/dsp/templates/template_v2.py
@@ -199,9 +199,6 @@ class TemplateV2:
     def __call__(self, example, show_guidelines=True) -> str:
         example = dsp.Example(example)
 
-        if hasattr(dsp.settings, "query_only") and dsp.settings.query_only:
-            return self.query(example)
-
         # The training data should not contain the output variable
         if self.fields[-1].input_variable in example:
             del example[self.fields[-1].input_variable]
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -99,7 +99,8 @@ class Predict(Parameter):
         # Switch to legacy format for dsp.generate
         template = signature_to_template(signature)
 
-        x, C = dsp.generate(template, **config)(x, stage=self.stage)
+        with dsp.settings.context(lm=lm):
+            x, C = dsp.generate(template, **config)(x, stage=self.stage)
         assert self.stage in x, "The generated (input, output) example was not stored"
 
         completions = []
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -99,14 +99,7 @@ class Predict(Parameter):
         # Switch to legacy format for dsp.generate
         template = signature_to_template(signature)
 
-        if self.lm is None:
-            x, C = dsp.generate(template, **config)(x, stage=self.stage)
-        else:
-            # Note: query_only=True means the instructions and examples are not included.
-            # I'm not really sure why we'd want to do that, but it's there.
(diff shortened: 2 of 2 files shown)


arnavsinghvi11:
Hi @minfawang , thanks for checking up on this. the lm is actually used as the line `if self.lm is None` defaults to using the [`dspy.settings.lm` ](https://github.com/stanfordnlp/dspy/blob/dba1285fb8acdfd0154a1e7fdc0448e3eadd7dab/dsp/primitives/predict.py#L62)but internally when `self.lm` is set, we simply configure `dspy.settings.lm` using the context manager. 

As for the `query_only` behavior, it should be fine to leave as is in case there are use cases that arise to only have the query!

Feel free to comment otherwise if you think this PR is still needed. Thanks!

arnavsinghvi11:
hmm I'm still a bit confused. Could you share an example perhaps that illustrates this issue? To my understanding, the current logic ensures the correct LM is loaded in, and we don't need to touch `query_only` for now.

arnavsinghvi11:
@minfawang This will actually work using the DSPy context manager. Feel free to reference [this documentation on using multiple LMs together](https://dspy-docs.vercel.app/docs/building-blocks/language_models#using-multiple-lms-at-once).

If you run `with dspy.context(lm= lm2):` instead of specifying it in the generate_answer config, this will resolve the issue. lmk if that helps.

minfawang:
Thanks. I understand that I could use `dspy.context` to achieve the desirable effect. But `generate_answer(question=question, lm=lm2)` is legal, yet it results in inconsistency: the function will invoke lm1, using the `temperature` and `num_generations` from lm2 (see l76-l80).

I think we should:
1. either support the lm kwarg `generate_answer(..., lm=lm2)`, and adopt a fix like this
2. or completely ignore lm kwarg, and fix l71, l76-l80, making `temperature` and `num_generations` independent on lm kwarg

Are you more in favor of 2)?

arnavsinghvi11:
I think the change you've proposed (1) is making more sense to me now because we don't want to ignore the lm kwarg if passed in, but I'm realizing `self.lm` actually isn't really updated anywhere except [BootstrapFinetune](https://github.com/stanfordnlp/dspy/blob/c9468abf310a367a05c98ddda5a3479e268e457e/dspy/teleprompt/finetune.py#L186) where we assign the `predictor.lm` as the fine-tuned model which also sets `query_only` in the comment. 

I don't believe this is needed but I'll confirm. If not, we can merge this PR with the changes as is. Thanks for sharing the examples - really helpful to dive deeper to isolate the use cases!

minfawang:
Thanks for your patience! The plan sounds good.

okhat:
Please don't merge this yet. It's missing some context which I'll add asap

PR #783, opened 2024-04-08
allow passing of kwargs to typed predictor decorators

limnick:
This change allows you to pass key word arguments such as the max_retries argument to the typed predictors while using the decorator method to construct them. Example:
```py
class MyModule(dspy.Module):
    @dspy.cot(max_retries=5, explain_errors=True)
    def my_cool_function(self, input: Input) -> Output:
        """Does some incredibly cool stuff up to five times with an explanation for errors"""
        pass

    ... rest of module implementation
```

The arguments are optional, so `@dspy.cot` and `@dspy.cot(max_retries=1)` will both work under this patch.

### DIFF
--- a/dspy/functional/functional.py
+++ b/dspy/functional/functional.py
@@ -14,18 +14,32 @@ from dspy.primitives.prediction import Prediction
 from dspy.signatures.signature import ensure_signature, make_signature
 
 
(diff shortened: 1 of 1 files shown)


CholoTook:
Can you add some tests plz?

PR #794, opened 2024-04-09
docs(dspy): Fixed docstring for OpenAI having wrong default model

nsobadzhiev:
## 📝 Changes Description

I noticed that the docstring for `OpenAI` says that the default model value is davinci. However, the default value of the parameter in code is GT3 turbo instruct, which looks more reasonable.
I suspect the default changed a while ago and someone forgot to change the docstring.

* New docstring value for the model property of `OpenAI` and `AzureOpenAI` classes

## ⚠️ Warnings

Exactly because of these things (changing the parameter default value and forgetting to update the docs), I would consider not writing the default value in the docs at all. The IDE (I checked VS Code) anyway shows the available parameters and their default values, in addition to the docstring. So best case - it's written twice. Worst case - it's written twice AND one is incorrect.

### DIFF
--- a/dsp/modules/azure_openai.py
+++ b/dsp/modules/azure_openai.py
@@ -44,7 +44,7 @@ class AzureOpenAI(LM):
     Args:
         api_base (str): Azure URL endpoint for model calling, often called 'azure_endpoint'.
         api_version (str): Version identifier for API.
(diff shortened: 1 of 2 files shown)


isaacbmiller:
LGTM

isaacbmiller (approved):

PR #796, opened 2024-04-09
Cleanup Logging

KCaverly:
This PR makes a series of changes to clean up print logging, specifically in the evaluate/teleprompt folders.
This should not change any functionality, instead it is to remove individualized logging handling, to favour centralized logging.


KCaverly:
Additionally, there is one api change to Copro, in which I removed the 'verbose' argument on initialization.
To combat this, I moved all the verbose print calls to debug log messages. I believe for simplicity given this scenario, I have added the following DeprecationMessage. 

image

Would be interested in seeing how the team thinks about introducing messages such as this.

@okhat @CyrusOfEden @thomasahle @isaacbmiller

CyrusOfEden (approved):

arnavsinghvi11:
@KCaverly Thanks for cleaning the logging! The DeprecationWarning makes sense to me until we complete the refactor to remove the respective args. Quick note before merging: I noticed you left some [comments](https://github.com/stanfordnlp/dspy/pull/796/files#diff-9908d550b4fa5c66ce43b1d2bf058f92a66b6da37dc94c1326f5bc9dd067fe68R202) and just wanted to check if those were still in pending for this PR or to be addressed later on?

KCaverly:
> @KCaverly Thanks for cleaning the logging! The DeprecationWarning makes sense to me until we complete the refactor to remove the respective args. Quick note before merging: I noticed you left some [comments](https://github.com/stanfordnlp/dspy/pull/796/files#diff-9908d550b4fa5c66ce43b1d2bf058f92a66b6da37dc94c1326f5bc9dd067fe68R202) and just wanted to check if those were still in pending for this PR or to be addressed later on?

Thats a good catch, ill remove these comments before we merge.

KCaverly:
I've cleaned up the comments, should be good to merge.

arnavsinghvi11:
Thanks @KCaverly !

okhat:
Thanks for this PR! I tend to be more on the pragmatic side: Centralized logging is cool, but we must prioritize understanding the downstream effect of a PR before merging.

We've received a lot of complaints/questions on observability during optimization, and I only recently understood why they increased. This PR makes them all "info" level, so now optimization is practically fully opaque.

okhat:
Let's think of the best way to fix this:

1. Raise the default logging level so `info` things are always shown. What are the implications here?
2. Modify a whole lot of the `info` logs to be something louder.
3. Switch back to print statements for any important streams of information.

PR #800, opened 2024-04-10
Fix evaluation multithreaded results

liorfo:
This PR fixes the evaluation return results feature for multithreaded evaluation. previously it returned an unsorted results array and now it returns the desired ordered results


arnavsinghvi11:
Thanks @liorfo !

Issue #570, opened 2024-03-06
Upgrade pydantic to ^2.6


mwildehahn:
The project is pinned to pydantic 2.5.0. Any reason it can't be upgraded to the latest version ^2.6?

isaacbmiller:
There used to be tests with the pydantic version name in them, but those seem to have been removed. IMO good to upgrade but I would defer to @thomasahle

thomasahle:
Yes, I fixed those tests.
I'm happy to upgrade pydantic if the tests pass.

nlaffey:
👀 I'm also needing pydnatic at 2.6 so that I can generate an OpenApi 3.1 spec

thomasahle:
Yes, I don't see any reason to now upgrade. I don't think we are using any newer features though, so maybe we can just put the requirement `pydantic >= 2.0`?

PR #807, opened 2024-04-11
fix(dspy): Loosen pydantic version

isaacbmiller:
Loosen Pydantic version restriction to be >= 2.0

Reasoning is from #570


isaacbmiller:
@thomasahle Would you anticipate any issues with this change?

isaacbmiller:
I could also cap at 2.7

pyproject.toml
```diff
@@ -82,7 +82,7 @@ keywords = ["dspy", "ai", "language models", "llm", "openai"]
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-pydantic = ">=2.5.0,<=2.7"
```
paluchasz, line 85: You could use `pydantic="^2.0"` which also ensures that it would be `<3.0.0`. See https://python-poetry.org/docs/dependency-specification/

isaacbmiller: Good call. Thanks!

isaacbmiller:
Oops. Thanks for catching. will follow up soon

Issue #805, opened 2024-04-11
pypi.org version does not match


talalryz:
https://pypi.org/project/dspy-ai/2.4.5/ does not match or include the latest changes to the project. It also does not match the 2.4.3 here: https://github.com/stanfordnlp/dspy/releases/tag/v2.4.3

qhoxie:
I noticed the same earlier today.  It seems to be missing at least a couple of weeks worth of changes.  Something strange appears to have happened a few days ago with many yanked releases: https://pypi.org/project/dspy-ai/2.4.5/#history

okhat:
yeah working on this

okhat:
nothing major, we're just 2 weeks behind and it's challenging to publish some stuff to pypi, WIP!

vietvudanh:
And there is no tag to track `2.4.5`. 
I installed `2.4.5` and there are some documented features missing (e.q: GROQ intergration), went to code to check and confused what commit is related to this version.

arnavsinghvi11:
Resolved with latest updates!

Issue #578, opened 2024-03-06
How to remove and disable cache?


JPonsa:
Hi team.

I would like to:
1. remove existing change. I was recommended to remove the ~/cache_lib folder but I cannot find it.
2. disable the generation of cache. So I don't need to do step 1 in future iterations.  

Possible use case for disabling cache.  Get slightly different answer versions for the same question. Currently getting duplicates.

      def generate_qa(context: str, n_qa_pairs: int = 5) -> list:
          """Given a context generate n question and answer pairs"""
      
          query_generator = dspy.Predict("context -> question")
          answer_generator = dspy.Predict("context, query -> answer")
      
          qa_pairs = []
          for i in range(n_qa_pairs):
              dspy.settings.configure(lm=lm, temperature=0.7 + (0.1 * i))
              query = query_generator(context=context)
              answer = answer_generator(context=context, query=query.question)
              qa_pairs.append((query.question, answer.answer))
      
          return qa_pairs

okhat:
I believe you should pass `temperature=0.7 + (0.1 * i)` as follows:

```python
query = query_generator(context=context, config=dict(temperature=0.7 + (0.1 * i)))
```

meditans:
Could there be a flag `cache=False` that we can control from the single generation?

It seems to me that we have two solutions for now:
- manually editing the source code to change the cache flag
- doing the trick above slightly varying the temperature
 
If it was a per-generation flag we could write:
```
Predict(Signature)(input1="the input", dspy_cache=False)
```

meditans:
I have found a workaround I'm satisfied with: I defined
```
def t():
    return dict(temperature=0.7 + 0.0001 * random.uniform(-1, 1))
```
and I call:
```
Predict(MySignature, **t)(input1="the input")
```

mohamedScikitLearn:
if anyone still looking for a solution:

From v2.5, you can turn off the cache by setting cache parameter in dspy.LM to False:

`dspy.LM('openai/gpt-4o-mini',  cache=False)`

https://dspy.ai/faqs/

Issue #213, opened 2023-11-07
runtime cache


DmitriyG228:
Hey, I have noticed that in order to avoid redundant model calls cash is being stored. Hove can I control this behaviour?

okhat:
Sure what kind of control are you looking for?

DmitriyG228:
I would want to drop it

okhat:
Completely drop the cache?

okhat:
Just switch this line to false:

https://github.com/stanfordnlp/dspy/blob/87e58d943c93766f08ccfe3b4c35d689c32bd7d8/dsp/modules/cache_utils.py#L10C1-L11C1

DmitriyG228:
can I ignore cash while calling a pipeline?

okhat:
Yeah

UtkarshGarg-UG:
Hi, I am observing a behavior where even though my module has been updated using signature optimizer, the evaluator runs from the cache and gives me the same score. Is there a way to run the evaluator without it reusing the result from cache? It shouldn't in the first place because compiled_prompt_opt has an updated signature below using the MIPRO optimizer.

print(evaluator(compiled_prompt_opt, devset=devset, )

jasonmgeorge:
> I would want to drop it

I was able to locate my local cache directory (to delete it) with the following
`import dsp
from dsp.modules.cache_utils import CacheMemory
print(CacheMemory)`

Note that there are a few other memory locations in cache_utils that might be in use.

Required a reload of my local Jupyter Notebook to take effect.

PR #581, opened 2024-03-06
env var for cache_utils.py

JPonsa:
add environmental variable ''DPS_CACHEBOOL' to activate or deactivate cache. By default True.

PR #810, opened 2024-04-12
docs(faqs): update documentation for disabling the cache

majapavlo:
Documents the latest way to address these two issues: 
- https://github.com/stanfordnlp/dspy/issues/213
- https://github.com/stanfordnlp/dspy/issues/578

Please also note that the disable cache feature is not available on the latest version of the dspy package available on [pypi](https://pypi.org/project/dspy-ai/2.4.5/#history). **See**: https://github.com/stanfordnlp/dspy/issues/805

### DIFF
--- a/docs/docs/faqs.md
+++ b/docs/docs/faqs.md
@@ -70,7 +70,7 @@ Open source libraries such as [RAGautouille](https://github.com/bclavie/ragatoui
 
 - **How do I turn off the cache? How do I export the cache?**
 
-You can turn off the cache by setting the [`cache_turn_on` flag to `False`](https://github.com/stanfordnlp/dspy/blob/9d8a40c477b9dd6dcdc007647b5b9ddad2b5657a/dsp/modules/cache_utils.py#L10).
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Hi @majapavlo , does disabling the cache not work currently? Also just made a tweak to the phrasing since `DSP_CACHEBOOL` refers to the `cache_turn_on` flag.

majapavlo:
The change you added might confuse users, since as a consumer of this library the `cache_turn_on` variable is a hidden implementation detail. 
It should be possible to disable the cache without modifying the source code of the library, i.e. by setting an environment variable. This feature has already been [added to the code](https://github.com/stanfordnlp/dspy/pull/581), and is present on the main branch via the environment variable `DSP_CACHEBOOL`.

However, currently disabling the cache does **_not_** work, since the latest version of the package on pypi is missing this feature as the pypi releases are out of date. See this issue: https://github.com/stanfordnlp/dspy/issues/805.

arnavsinghvi11:
I see, thanks for raising the concern @majapavlo . We have updated the README to mention how to use the library locally to use `cache_turn_on` as well.

Issue #784, opened 2024-04-08
Inspect_history only shows the last history regardless of n/skip


noobHappylife:
While I'm using the inspect_history to check the output history, regardless of the n, and skip I chose. It always only return the last history. 

```python
        for idx, (prompt, choices) in enumerate((printed)):
            # skip the first `skip` prompts
            if (n - idx - 1) > skip:
                continue
            printing_value = ""
```

A changed to the inspect_history function as above would make it works to show the last n-th history after considering the skip.

arnavsinghvi11:
@noobHappylife could you elaborate on your test?

Specifying `n` should print the last `n` generations. Specifying `skip` will jump over the past `skip` generations and then print out the last `n` generations.

noobHappylife:
Hi, for your reference. here is the scenario i met. 

```
# lm response history
print([x['response'] for x in nous_lm.history])
```
[{'prompt': 'hi',
  'choices': [{'text': 'Hello! How can I help you today? If you have any questions or need assistance with a problem, feel free to ask.'}]},
 {'prompt': 'who are you?',
  'choices': [{'text': 'I am a super assistant designed to help you with a wide range of tasks, from answering questions to providing guidance and recommendations. My goal is to assist you in the most efficient and helpful way possible.'}]},
 {'prompt': 'who am i?',
  'choices': [{'text': "As a super assistant, I can't directly tell you who you are. However, I can help you discover more about yourself by providing self-reflection prompts or assisting with personal growth and self-discovery tasks. If you'd like, I can offer some resources or activities to help you better understand yourself."}]},
 {'prompt': 'what is 1 + 1?', 'choices': [{'text': '1 + 1 is equal to 2.'}]}]

- followings are inspect_history on different n, and skip.

```
# regardless of n, the inspect_history shows the same output
print(nous_lm.inspect_history(n=1))

what is 1 + 1?1 + 1 is equal to 2.

print(nous_lm.inspect_history(n=2))

what is 1 + 1?1 + 1 is equal to 2.

print(nous_lm.inspect_history(n=4))

what is 1 + 1?1 + 1 is equal to 2.

# including skip, results in None output
print(nous_lm.inspect_history(n=1, skip=1))
# empty output

print(nous_lm.inspect_history(n=3, skip=2))
# empty output
```

After making the changes mentioned earlier, the inspect_history can response with the n, skip correctly (at least in my testing). 

```
[x['response'] for x in nous_lm.history]
```
[...]

Issue #778, opened 2024-04-05
Support for vLLM chat completion endpoint


noobHappylife:
Currently, i noticed the module HFClientVLLM only calls the completion endpoints. As vLLM now support applying chat templates. Would it be possible to add support for chat completion endpoints? 

the following is a naive approach to add a chat completion support, edited the _generate method for HFClientVLLM
```python
    def _generate(self, prompt, **kwargs):
        kwargs = {**self.kwargs, **kwargs}

        model_type = kwargs.get("model_type",None)
        url = self.urls.pop(0)
        self.urls.append(url)

        system_prompt = kwargs.get("system_prompt",None)
        if model_type == "chat":
            messages = [{"role": "user", "content": prompt}]
            if system_prompt:
                messages.insert(0, {"role": "system", "content": system_prompt})
            payload = {
                "model": self.kwargs["model"],
                "messages": messages,
                **kwargs,
            }
            response = send_hfvllm_chat_request_v00(
                f"{url}/v1/chat/completions",
                json=payload,
                headers=self.headers,
            )

            try:
                json_response = response.json()
                completions = json_response["choices"]
                response = {
                    "prompt": prompt,
                    "choices": [{"text": c["message"]['content']} for c in completions],
                }
                return response

            except Exception:
                print("Failed to parse JSON response:", response.text)
                raise Exception("Received invalid JSON response from server")
        else:
            payload = {
                "model": self.kwargs["model"],
                "prompt": prompt,
                **kwargs,
            }

            response = send_hfvllm_request_v00(
                f"{url}/v1/completions",
                json=payload,
                headers=self.headers,
            )

            try:
                json_response = response.json()
                completions = json_response["choices"]
                response = {
                    "prompt": prompt,
                    "choices": [{"text": c["text"]} for c in completions],
                }
                return response

            except Exception:
                print("Failed to parse JSON response:", response.text)
                raise Exception("Received invalid JSON response from server")


@CacheMemory.cache
def send_hfvllm_request_v00(arg, **kwargs):
    return requests.post(arg, **kwargs)

@CacheMemory.cache
def send_hfvllm_chat_request_v00(arg, **kwargs):
    return requests.post(arg, **kwargs)
```

arnavsinghvi11:
Thanks @noobHappylife ! Happy to merge a PR for this change.

PR #811, opened 2024-04-13
Add vllm chat completion endpoints and fix inspect_history

noobHappylife:
Related to issue #778 #784 

- added vllm chat completion endpoints (support system_prompt)
- fixed an issue with .copy() with vllm client module. (need to include model, url, port into kwargs)
- fix inspect_histrory not respecting n/skip as mentioned in #784

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -3,6 +3,7 @@ import random
 import re
 import shutil
 import subprocess
+from typing import Literal
 
 # from dsp.modules.adapter import TurboAdapter, DavinciAdapter, LlamaAdapter
 import backoff
@@ -114,7 +115,7 @@ def send_hftgi_request_v00(arg, **kwargs):
 
 
 class HFClientVLLM(HFModel):
-    def __init__(self, model, port, url="http://localhost", **kwargs):
+    def __init__(self, model, port, model_type: Literal['chat', 'text'] = 'text', url="http://localhost", **kwargs):
         super().__init__(model=model, is_client=True)
 
         if isinstance(url, list):
@@ -126,27 +127,24 @@ class HFClientVLLM(HFModel):
         else:
             raise ValueError(f"The url provided to `HFClientVLLM` is neither a string nor a list of strings. It is of type {type(url)}.")
         
+        self.model_type = model_type
         self.headers = {"Content-Type": "application/json"}
         self.kwargs |= kwargs
         # kwargs needs to have model, port and url for the lm.copy() to work properly
         self.kwargs.update({
-            'model': model,
             'port': port,
-            'url': url
+            'url': url,
         })
 
 
     def _generate(self, prompt, **kwargs):
         kwargs = {**self.kwargs, **kwargs}
         
-        # get model_type
-        model_type = kwargs.get("model_type",None)
-
         # Round robin the urls.
         url = self.urls.pop(0)
         self.urls.append(url)
         
-        if model_type == "chat":
+        if self.model_type == "chat":
             system_prompt = kwargs.get("system_prompt",None)
             messages = [{"role": "user", "content": prompt}]
             if system_prompt:
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -128,44 +128,86 @@ class HFClientVLLM(HFModel):
         
         self.headers = {"Content-Type": "application/json"}
         self.kwargs |= kwargs
+        # kwargs needs to have model, port and url for the lm.copy() to work properly
+        self.kwargs.update({
+            'model': model,
+            'port': port,
+            'url': url
+        })
 
 
     def _generate(self, prompt, **kwargs):
         kwargs = {**self.kwargs, **kwargs}
-
-        payload = {
-            "model": self.kwargs["model"],
-            "prompt": prompt,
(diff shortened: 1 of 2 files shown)


dsp/modules/hf_client.py
```diff
@@ -128,47 +128,89 @@ def __init__(self, model, port, url="http://localhost", **kwargs):
         
         self.headers = {"Content-Type": "application/json"}
         self.kwargs |= kwargs
+        # kwargs needs to have model, port and url for the lm.copy() to work properly
+        self.kwargs.update({
+            'model': model,
```
arnavsinghvi11, line 133: model is in fact included in the kwargs already, inherited through [HFModel-> LM](https://github.com/stanfordnlp/dspy/blob/0b7289fca05988b2863cc46a3564e8edbc9b7021/dsp/modules/lm.py#L7) so we can remove this.

is there a use case for including the port and URL in the LM kwargs? seems like they are only used for setting the URL(s) and not in the model payload? feel free to remove if not.

dsp/modules/hf_client.py
```diff
@@ -128,47 +128,89 @@ def __init__(self, model, port, url="http://localhost", **kwargs):
...
-
-        payload = {
-            "model": self.kwargs["model"],
-            "prompt": prompt,
-            **kwargs,
-        }
-
         
+        # get model_type
+        model_type = kwargs.get("model_type",None)
```
arnavsinghvi11, line 143: let's actually move this to the initialization layer so the user can specify the model_type when configuring the vLLM client, not during generation time.

dsp/modules/lm.py
```diff
@@ -63,13 +63,11 @@ def inspect_history(self, n: int = 1, skip: int = 0):
             if len(printed) >= n:
                 break
 
-        for idx, (prompt, choices) in enumerate(reversed(printed)):
-            printing_value = ""
-
+        for idx, (prompt, choices) in enumerate(printed):
```
arnavsinghvi11, line 66: is there a reason for this change? seems to have the same effect originally?

arnavsinghvi11 (changes requested):

dsp/modules/lm.py
arnavsinghvi11: currently failing these [tests](https://github.com/stanfordnlp/dspy/actions/runs/8671183621/job/23806529670?pr=811)

arnavsinghvi11:
Thanks @noobHappylife  for the additions to vLLM. left a couple comments and should be good to merge following those comments and running `ruff check . --fix-only` .

dsp/modules/hf_client.py
noobHappylife: Noted on the model part. For url and port, basically is when calling lm.copy() (ex: while compiling a program with bootstrap), it is missing the url and port kwargs. As shown as below. Thus, requiring the port and url in the kwargs for the copy() to work properly. 

```
in BootstrapFewShot._bootstrap_one_example(self, example, round_idx)
    148 with dsp.settings.context(trace=[], **self.teacher_settings):
    149     lm = dsp.settings.lm
--> 150     lm = lm.copy(temperature=0.7 + 0.001 * round_idx) if round_idx > 0 else lm
    151     new_settings = dict(lm=lm) if round_idx > 0 else {}
    153     with dsp.settings.context(**new_settings):

File [~/micromamba/envs/yh_hf/lib/python3.10/site-packages/dsp/modules/lm.py:108](http://192.168.77.2:8892/lab/tree/~/micromamba/envs/yh_hf/lib/python3.10/site-packages/dsp/modules/lm.py#line=107), in LM.copy(self, **kwargs)
    105 kwargs = {**self.kwargs, **kwargs}
    106 model = kwargs.pop("model")
--> 108 return self.__class__(model=model, **kwargs)

TypeError: HFClientVLLM.__init__() missing 1 required positional argument: 'port'
```

dsp/modules/lm.py
noobHappylife: I've included some of my testing with the inspect_history in the issue #784. Can you take a look, or I'm misunderstanding how it should works?

arnavsinghvi11: Ah thanks for highlighting. Just realized that this is a bug where the `printing_value` is getting reset everytime in the loop. That should be all the error is and the skip condition is fine as is. Adding a commit to this file - lmk if it aligns with fixing your issue and will merge.

noobHappylife: Thank you. Let me know what else is needed to change. I've updated to the vllm initialization part, and fix the error, so the test should pass now.

arnavsinghvi11: yeah can you revert the changes made to `inspect_history` here and just move the [`printing_value`](https://github.com/stanfordnlp/dspy/pull/811/files#diff-6bdc17664251e0662d9b19a69a5ebd5acd2252e4ecde2bdb6d89c460b1d022a6L67) outside the for loop?

noobHappylife: ok, updated and tested it works. Thank you. Now inspect_history is able to view n-number of histories with skip.

ammirsm: good catch folks! 

thanks for it!

arnavsinghvi11:
Thanks @noobHappylife !

PR #822, opened 2024-04-15
feat(compatibility): Add compatibility with Weaviate client v4.

### LINKED ISSUE #787: Update weaviate_rm.py with v4 WeaviateClient
# Description

Actually the WeaviateRM class is using the deprecated v3 client. 

We may consider using the new class WeaviateClient that comes with weaviate v1.23.6 and higher. 

# From Weaviate Documentation 

```py
self.collections = _Collections(self._connection)
        """This namespace contains all the functionality to manage Weaviate data collections. It is your main entry point for all collection-related functionality.

        Use it to retrieve collection objects using `client.collections.get("MyCollection")` or to create new collections using `client.collections.create("MyCollection", ...)`.
        """
```

# Updates to be done 

In ReviateRM :
From : 

```py
def __init__(self, 
                 weaviate_collection_name: str, 
                 weaviate_client: weaviate.Client, 
                 k: int = 3,
                 weaviate_collection_text_key: Optional[str] = "content",
        ):
``` 
To :
```py
def __init__(self, 
                 weaviate_collection_name: str, 
                 weaviate_client: weaviate.WeaviateClient, 
                 k: int = 3,
                 weaviate_collection_text_key: Optional[str] = "content",
        ):
```

And we could  change the doc-string example with the now recommended way to connect to Weaviate :
```py
        weaviate_client = weaviate.Client("your-path-here")
``` 
to 
```py
        weaviate_client = weaviate.connect_to_[local,wcs,custom,embedded]("your-path-here")
```

Concerning the forward method, I think the correct approach would be something like :

```py
queries = [q for q in queries if q]
passages = []
for query in queries:
    results = self._weaviate_client.collections.get(self._weaviate_collection_name)\
        .query\
        .hybrid(query=query, limit=k)

    parsed_results = [result.properties[self._weaviate_collection_text_key] for result in results.objects]
    passages.extend(dotdict({"long_text": d}) for d in parsed_results)

return passages
```

I can't test it right now, but here is what I can think about the current usage.

Fiwex:
Maintain backward compatibility with v3 client.
Fix #787

### DIFF
--- a/dspy/retrieve/weaviate_rm.py
+++ b/dspy/retrieve/weaviate_rm.py
@@ -1,21 +1,19 @@
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 import dspy
 from dsp.utils import dotdict
+from dspy.primitives.prediction import Prediction
 
 try:
     import weaviate
-    import weaviate.classes as wvc
-    from weaviate.collections.classes.grpc import HybridFusion
-except ImportError:
+except ImportError as err:
     raise ImportError(
         "The 'weaviate' extra is required to use WeaviateRM. Install it with `pip install dspy-ai[weaviate]`",
-    )
+    ) from err
 
 
 class WeaviateRM(dspy.Retrieve):
-    """
-    A retrieval module that uses Weaviate to return the top passages for a given query.
+    """A retrieval module that uses Weaviate to return the top passages for a given query.
 
     Assumes that a Weaviate collection has been created and populated with the following payload:
         - content: The text of the passage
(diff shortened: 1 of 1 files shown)


CShorten:
Thank you so much, I can take a look this evening!

hsm207:
Why do we want to maintain compatibility with v3 client API?

Fiwex:
Thanks for the question! I decided to maintain backward compatibility with the v3 client API in this PR for smooth transition for users upgrading to the latest version.

Alternatively, I could keep v3 support for now but add deprecation warnings to encourage users to upgrade to v4. Or if you prefer, only keep the v4 compatibility.

Let me know your thoughts. I'm happy to modify the PR based on our project's priorities.

hsm207:
I'm not a member of the DSPy team (I work at Weaviate). My thoughts are:

We recommend [users use the v4 API whenever possible](https://weaviate.io/developers/weaviate/client-libraries/python#python-client-v3-api). Given that this integration is pretty recent and minor, I don't think it will cause users significant disruption to upgrade.

Fiwex:
@CShorten Just tel me what you prefer and I'll change the PR accordingly.

CShorten:
Thanks @Fiwex I think this is solely a positive change. Thank you for the thoughts @hsm207, I agree with the sentiment, but we are indeed seeing some issues arising and I can't see any drawbacks of this approach.

arnavsinghvi11:
Hi @Fiwex  @CShorten , just catching up on old PRs - is this ready to merge with the current approach?

betimd:
It would be awesome if this PR is merged ASAP. I'm having a ton of issues on my solutions, especially when I have to integrate other libs, that are specialized on specific tasks and they are depended on current version of package. It's conflicting with 3.* version hugely. Ex. I'm using DSPy and LlamaIndex import data functionality, while latest use updated dep package, because of former one I have to downgrade and have a ton of errors.

CShorten:
Hey @betimd! Thank you so much for telling us more about this problem, really happy to hear about use cases with DSPy, LlamaIndex, and Weaviate! I have reviewed the PR and will merge it to add support for using DSPy with Weaviate's v3 and v4 Python clients. 

P.S. @hsm207, arguments like `hybrid_alpha` and `hybrid_fusion_type` are supported with `**kwargs`. I really like this approach because of how we can add reranking, filters, and things like this.

PR #825, opened 2024-04-15
docs(rag): fix broken link

thliang01:


arnavsinghvi11:
Thanks for raising this @thliang01! turns out we wanted to show the home page for language model clients so just making that edit before merging.

arnavsinghvi11:
Thanks @thliang01!

PR #828, opened 2024-04-15
fixing issue in DatabricksRM preventing indexed text from being returned

lmoros-DB:
- There were two missing code lines that prevented the "text" column from being returned when the "froward" method was executed


arnavsinghvi11:
Thanks @lmoros-DB !

PR #829, opened 2024-04-15
Setting the default value of the DatabricksRM.forward  "query_type" 

lmoros-DB:
Setting the default value of the DatabricksRM.forward  "query_type" param to "text"

It is recommended to use the fully managed approach for Vector Search which covers getting the embeddings from the of the indexed column during the ingestion and the embeddings of the text query during inference


arnavsinghvi11:
Thanks @lmoros-DB !

PR #830, opened 2024-04-15
Add missing gold_passages_retrieved function

tom-doerr:
The function gold_passages_retrieved is currently missing, without it the tutorial doesn't work. 
I copied the function from https://github.com/stanfordnlp/dspy/blob/eab1324d45aa1a723e187b2523e4d20bfc75356d/tests/examples/test_baleen.py#L98

I tested the code with this new function and I get similar numbers as in the tutorial


arnavsinghvi11:
Thanks @tom-doerr !

PR #831, opened 2024-04-16
feat: FastEmbedVectorizer

Anush008:
## Description

This PR adds an implementation of `BaseSentenceVectorizer` using [FastEmbed](https://qdrant.github.io/fastembed/) - A lightweight, CPU-first library to generate embeddings locally.

This PR will be followed by https://github.com/Anush008/dspy/pull/1, which updates `QdrantRM` to support any `BaseSentenceVectorizer`. The default being the new `FastEmbedVectorizer`.


pyproject.toml
```diff
@@ -62,6 +62,7 @@ docs = [
     "sphinx-automodapi==0.16.0",
 ]
 dev = ["pytest>=6.2.5"]
+fastembed = ["fastembed>=0.1.0"]
```
ammirsm, line 65: it should be added in the [tool.poetry.extras] too.

dsp/modules/sentence_vectorizer.py
```diff
@@ -24,28 +25,29 @@ def __call__(self, inp_examples: List["Example"]) -> np.ndarray:
...
     def _extract_text_from_examples(self, inp_examples: List) -> List[str]:
         if isinstance(inp_examples[0], str):
-            return inp_examples 
+            return inp_examples
         return [" ".join([example[key] for key in example._input_keys]) for example in inp_examples]
 
 
 class SentenceTransformersVectorizer(BaseSentenceVectorizer):
-    '''
+    """
```
ammirsm, line 33: can I ask for reverting these changes in the other classes?

we have a plan to fix them all in one PR to keep the git blame cleaner.

ammirsm (commented):
All in all LGTM but I just added few small notes on your PR.

We don't have any testing for these Vectorizer so I would recommend adding some tests if it is possible for you with mocking OR adding some testing instruction in your PR description.

If it is going to be tested in the following PR I would recommend targeting that branch as a target branch and merge it after the testing completed meanwhile while it is not being used anywhere in the code it would be fine to merge with the master too.

LMK which one would works better for you.

tests/modules/vectorizer/integration_test_fastembed.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Skip the test if the 'fastembed' package is not installed
+pytest.importorskip("fastembed", reason="'fastembed' is not installed. Use `pip install fastembed` to install it.")
+
+
+@pytest.mark.parametrize(
+    "n_dims,model_name", [(384, "BAAI/bge-small-en-v1.5"), (512, "jinaai/jina-embeddings-v2-small-en")]
+)
+def test_fastembed_with_examples(n_dims, model_name):
+    vectorizer = FastEmbedVectorizer(model_name)
```
ammirsm, line 14: what do you think about mocking the `embed` function and keep it in the unit test?

the other question is does it require downloading of the models? if yeah I would say that would be good to be able to have it mocked and run it in the test in the CI without downloading.

ammirsm (commented):
Thanks for the tests and contribution!!
It really looks good. 

Just a quick question about that should we mock or not which I mentioned in the code for you.

Thanks again

tests/modules/vectorizer/integration_test_fastembed.py
Anush008: I planned tests. However, since the implementation is straightforward, the risk of failure comes from changes in the FastEmbed library, so these integration tests(Skipped if FastEmbed is not installed).

ammirsm: So I think you don't want to include them in CI, am I right?

Anush008: I've added them to the CI in https://github.com/stanfordnlp/dspy/pull/831/commits/9b04c68f8eed9da0af6d7ac19b2a5e657f2f4d26.

ammirsm: sorry for the delay in reply! 

I didn't meant to add them to the CI as they would take so long and I think they are not built for there to download each time we run the CI, just wanted to double check them for there with you! 

sorry if I made a confusion for you.

Anush008: Alright. I've removed it from the CI.

ammirsm (approved):
Thanks for contribution and sorry for all back and forth!

LGTM, just @arnavsinghvi11 can you take a final look and merge this one please if you think is good too?

arnavsinghvi11:
Thanks for the PR and edits @Anush008 @ammirsm ! Good to merge pending resolving the merge conflicts! (likely just rebasing to main should work)

Anush008:
Hi @arnavsinghvi11. It should be good now.

PR #839, opened 2024-04-16
fix(dsp): prevent cache instantiation when cache is disabled

stevenhoelscher:
The environment variable `DSP_CACHEBOOL` does not completely disable the filesystem cache. 

As a result, `joblib.Memory` is still instantiated, which will try to create a directory on the filesystem. 

In an environment where a user does not have write access to the `DSP_CACHEDIR`, this will throw a surprising exception despite `DSP_CACHEBOOL` being set to `false`.

```
File "/var/lang/lib/python3.11/site-packages/dsp/modules/cache_utils.py", line 27, in <module>
    CacheMemory = Memory(location=cachedir, verbose=0)
  File "/var/lang/lib/python3.11/site-packages/joblib/memory.py", line 1020, in __init__
    self.store_backend = _store_backend_factory(
  File "/var/lang/lib/python3.11/site-packages/joblib/memory.py", line 132, in _store_backend_factory
    obj.configure(location, verbose=verbose,
  File "/var/lang/lib/python3.11/site-packages/joblib/_store_backends.py", line 452, in configure
    mkdirp(self.location)
  File "/var/lang/lib/python3.11/site-packages/joblib/disk.py", line 61, in mkdirp
    os.makedirs(d)
 ```

### DIFF
--- a/dsp/modules/cache_utils.py
+++ b/dsp/modules/cache_utils.py
@@ -24,7 +24,12 @@ def noop_decorator(arg=None, *noop_args, **noop_kwargs):
 
 
 cachedir = os.environ.get('DSP_CACHEDIR') or os.path.join(Path.home(), 'cachedir_joblib')
-CacheMemory = Memory(location=cachedir, verbose=0)
+
+CacheMemory = dotdict()
+CacheMemory.cache = noop_decorator
+
+if cache_turn_on:
+    CacheMemory = Memory(location=cachedir, verbose=0)
 
 cachedir2 = os.environ.get('DSP_NOTEBOOK_CACHEDIR')
 NotebookCacheMemory = dotdict()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: dsp/modules/cache_utils.py [hunk 1]
```diff
@@ -24,19 +24,16 @@ def wrapper(*args, **kwargs):
...
+CacheMemory.cache = noop_decorator
+
+if cache_turn_on:
+    CacheMemory = Memory(location=cachedir, verbose=0)
 
 cachedir2 = os.environ.get('DSP_NOTEBOOK_CACHEDIR')
 NotebookCacheMemory = dotdict()
 NotebookCacheMemory.cache = noop_decorator
 
 if cachedir2:
```
stevenhoelscher, line 38: Looks like this could be updated to

```
if cache_turn_on and cachedir2:
```

stevenhoelscher: It's not obvious what `DSP_NOTEBOOK_CACHEDIR` denotes. Because it seems to assign the `NotebookCacheMemory` a fs cache but then walks that back if `DSP_CACHEBOOL` is true.

arnavsinghvi11:
Thanks @stevenhoelscher for the PR! This looks good and maintains the cache correctly. Is the PR ready to merge or pending [the change you've mentioned here](https://github.com/stanfordnlp/dspy/pull/839#discussion_r1567678653)?

PR #843, opened 2024-04-16
Fixed issue #894 and #858 (aws_models issues)

drawal1:
EnsembledProgram can now be loaded/saved
~~Inference of programs (forward function) is now concurrent~~
Added checks for size = 0 or size > number of programs

### DIFF
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -28,7 +28,7 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
     def __init__(
         self,
         metric,
-        teacher_settings=None,
+        teacher_settings={},
         max_bootstrapped_demos=4,
         max_labeled_demos=16,
         max_rounds=1,
@@ -39,12 +39,7 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
         metric_threshold=None,
     ):
         self.metric = metric
-
-        if teacher_settings is None:
-            self.teacher_settings = {}
-        else:
-            self.teacher_settings = teacher_settings
-
+        self.teacher_settings = teacher_settings
         self.max_rounds = max_rounds
 
         self.num_threads = num_threads
@@ -72,12 +67,6 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
         scores = []
         all_subscores = []
         score_data = []
-
-        if teacher:
-            if hasattr(teacher, "candidate_programs"):
-                teacher.candidate_programs = teacher.candidate_programs[:1]
-            elif hasattr(teacher, "programs"):
-                teacher = teacher.programs[0]
 
         for seed in range(-3, self.num_candidate_sets):
             if (restrict is not None) and (seed not in restrict):
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -28,7 +28,7 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
     def __init__(
         self,
         metric,
-        teacher_settings={},
+        teacher_settings=None,
         max_bootstrapped_demos=4,
         max_labeled_demos=16,
         max_rounds=1,
@@ -39,7 +39,12 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
         metric_threshold=None,
     ):
         self.metric = metric
-        self.teacher_settings = teacher_settings
+
+        if teacher_settings is None:
+            self.teacher_settings = {}
+        else:
+            self.teacher_settings = teacher_settings
+
         self.max_rounds = max_rounds
 
         self.num_threads = num_threads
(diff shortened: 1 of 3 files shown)


okhat:
This is so cool.... I need to check the parallelization, it's a bit tricky in DSPy due to dspy.settings.

In dspy.evaluate.Evaluate we do things in a bit of a special way, which has to be applied when there's threading

drawal1:
@okhat - any progress on this review? I am continuing to work on intelligent ensemble routing and don't won't to muddy this PR with commits related to that work

drawal1:
> This is so cool.... I need to check the parallelization, it's a bit tricky in DSPy due to dspy.settings.
> 
> In dspy.evaluate.Evaluate we do things in a bit of a special way, which has to be applied when there's threading

I think I understand. I see this code in evaluate.py:
```
        def wrapped_program(example_idx, example):
            # NOTE: TODO: Won't work if threads create threads!
            thread_stacks = dspy.settings.stack_by_thread
```
So I need to test and verify the parallelization NOT in inference but as part of another optimization pipeline

drawal1:
Fixed #894 in dsp/modules/aws_models.py, in class AWSMeta(AWSModel) by moving below functionality from __init__ function:
`self.kwargs["max_gen_len"] = self.kwargs.pop("max_tokens")`
to the __create_body__ function

drawal1:
@okhat - I have reverted the parallelization in EnsembledProgram.forward() based on your review and feedback. 

The remaining fixes/improvements are targeted and should not have side-effects

dspy/teleprompt/random_search.py
```diff
@@ -49,27 +55,42 @@ def __init__(
...
+            elif hasattr(teacher, "programs"):
+                teacher = teacher.programs[0]
+            if has_programs:
+                # if we are working with a teacher that already has candidate programs
+                # lets skip the zero shot and LabeledFewShot cases since they don't use teachers
+                restrict = [-1]
+
         for seed in range(-3, self.num_candidate_sets):
-            if (restrict is not None) and (seed not in restrict):
+            if seed < 0 and (restrict is not None) and (seed not in restrict):
```
arnavsinghvi11, line 93: why are we skipping over the zero-shot, labeled_fewshot and unshuffled_fewshot cases for all scenarios? I think even in the conditions above when a teacher is set, it is fine to run through these preliminary candidates and have it weighted in the scores.

dspy/teleprompt/random_search.py
```diff
@@ -96,7 +117,8 @@ def compile(self, student, *, teacher=None, trainset, valset=None, restrict=None
                 program2 = program.compile(student, teacher=teacher, trainset=trainset2)
 
             else:
-                assert seed >= 0, seed
+                if seed < 0:
```
arnavsinghvi11, line 120: can remove this based on decision above

pyproject.toml
```diff
@@ -278,7 +284,14 @@ ignore = [
     "E731",
     # Sometimes we need List and Tuple
     "UP006",
+    # optimzer are using the 'compile' method which is a built-in
```
arnavsinghvi11, line 287: spelling typo

pyproject.toml
```diff
@@ -278,7 +284,14 @@ ignore = [
...
+    "A003",
+    # commented-out code should be allowed
+    "ERA001",
+    # allow pickle
+    "S301",
```
arnavsinghvi11, line 292: what are the purpose of these changes? I think the ruff has been working correctly so far but just curious

arnavsinghvi11:
Thanks @drawal1 for the PR! left a few comments. The changes to Ensemble look correct to me, but I think the ones for RandomSearch are not needed? or can be handled separately without impacting existing behavior. lmk what you think.

There's also a merge conflict to resolve and feel free to remove extraneous comments from the changes that do not help with following the code. Thanks!

tagging @okhat for reference on Ensemble/RandomSearch again (but I believe the parallelization changes are removed so should be good there).

pyproject.toml
drawal1: ERA001 because Ruff complains about commented out print() statements. I often find these useful for debugging

ensemble.py load_folder/save_folder use pickle for serialization. Ruff flags pickle as unsafe for load because it could load arbitrary objects. I suppose we could switch to https://jsonpickle.github.io/. If jsonpickle is acceptable, I can make the change and remove the S301.

dspy/teleprompt/random_search.py
drawal1: assert fails Ruff check (S101 Use of `assert` detected). Lets keep the raise exception here. its good practice

pyproject.toml
drawal1: fixed

pyproject.toml
drawal1: @arnavsinghvi11 - what's your opinion on jsonpickle? see my comment above...

drawal1:
@arnavsinghvi11 - I have reverted ALL the complicated changes. Hopefully now this PR is acceptable

arnavsinghvi11:
Hi @drawal1 , thanks for the changes. can you also revert the changes to [BootstrapWithRandomSearch](https://github.com/stanfordnlp/dspy/pull/843/files#diff-cfdd85de1987474649224c297047325dc2f0b8b890f7a45810556be440ecf84a) since those changes are not relevant to patching #858 .

arnavsinghvi11:
Thanks @drawal1 !

PR #849, opened 2024-04-17
Add Weights & Biases logging to BootstrapFewShot

CShorten:
# `wandb_enabled` to `BootstrapFewShot`

Here is a minimal first step to adding Weights & Biases optimization logging to DSPy.

This can be used for monitoring runs to `BootstrapFewShot` in order to see the `metric_val` returned for each bootstrapped example. To motivate the use case, you may have a rating on a scale of 1 to 5 for answers and you only want to use examples that achieve a 5 in your prompt.

There are definitely more directions to take with DSPy <> Weights & Biases logging, hopefully this is a nice first step. I've found this to be pretty interesting particularly when you are testing multiple teacher models for bootstrapping.

### Docs

Note, this contains a quick note in `6-optimizers.md`. I think it would be a premature refactoring, but I think it's likely we see a lot of the "callback" style logging of PyTorch / Keras carry over.

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -1,7 +1,9 @@
 import random
+random.seed(42) # meaning of life reproducibility.
 import threading
 
 import tqdm
+import wandb # Maybe an idea to refactor into a config to isolate the dependency, but it doesn't add too many LOC to customize the observability to the compiler.
 
 import dsp
 import dspy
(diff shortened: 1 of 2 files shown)


dspy/teleprompt/bootstrap.py
```diff
@@ -1,7 +1,10 @@
 import random
+
+random.seed(42) # meaning of life reproducibility.
```
arnavsinghvi11, line 3: we can remove the comment (although i agree haha)

dspy/teleprompt/bootstrap.py
```diff
@@ -1,7 +1,10 @@
...
 import threading
 
 import tqdm
+import wandb  # Maybe an idea to refactor into a config to isolate the dependency, but it doesn't add too many LOC to customize the observability to the compiler.
```
arnavsinghvi11, line 7: this can be wrapped in a try/except block similar to [handling import dependencies like here](https://github.com/stanfordnlp/dspy/blob/d7dace6b51e1c0899fd798c575262310939f6deb/dsp/modules/cohere.py#L8). This will also fix the failing test

also can remove comment

arnavsinghvi11 (changes requested):

arnavsinghvi11:
Hi @CShorten , thanks for the PR! left a few comments that will also address the failing test here by fixing the import with a try-except block. There is also a merge conflict to fix. Should be good to merge after that!

PR #857, opened 2024-04-17
Deprecate old logging

KCaverly:
Deprecate the 'log_openai_usage' configuration option.
As we are now moving to proper logging, we now log all usage by default, and allow the user to specify how they want to handle the logs themselves.

This removes the setting, and all mentions of this settings across gpt3/anthropic/azure_openai/groq.

### DIFF
--- a/dsp/modules/anthropic.py
+++ b/dsp/modules/anthropic.py
@@ -8,12 +8,14 @@ from dsp.modules.lm import LM
 
 try:
     import anthropic
+
     anthropic_rate_limit = anthropic.RateLimitError
 except ImportError:
     anthropic_rate_limit = Exception
 
 logger = logging.getLogger(__name__)
 BASE_URL = "https://api.anthropic.com/v1/messages"
+
 
 def backoff_hdlr(details):
     """Handler from https://pypi.org/project/backoff/."""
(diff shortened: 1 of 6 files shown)


okhat:
hey this sounds good (didn't check the code yet) but qq @KCaverly : is logging now expected to print out every HTTP request to OpenAI when using DSPy? that's a bit overkill

KCaverly:
The change is simply moving all the logs that were routing to the .log files before to logging regardless of the configuration setting. This may be overkill to log all the time, it may be a good idea to move these to debug, so they wouldnt clutter up most useful logs.

okhat:
Ah @KCaverly that's not what I meant to ask. I'm saying that right now for some reason logs print every HTTP request to OpenAI to console that's too much logging going on

KCaverly:
Thats odd - I dont think that should be because of this, let me take a look.

KCaverly:
I just took a look, we honestly arent logging that much from the open ai side at all. I just gave it a run, and for rate limiting, I see a ton of messages. This is not due to dspy though, the default logger for 'backoff' the package openai uses for retries, is defaulting to 'info'. 

As such, when we set the previous log level to info globally with the `logging.basicConfig` call, all those get filtered through. Kinda painful, but you can silence them with: 

```
logging.getLogger("backoff").setLevel(logging.WARNING)
```

KCaverly:
As this removes all default log handling from dspy, I've added a helper:

`dspy.show_logging()` which will set a default logger for people.

I made this helper silence the overeager backoff logger aswell.

okhat:
Hmm @KCaverly I merged this but I suspect it turned off most/all INFO logging by accident?

okhat:
When I run BootstrapFewShot I see no logs at all now iiuc

KCaverly:
@okhat that is correct. Its not best practice for a framework to define how applications should show their logs, we had a few issues posted about this. 

To help, I added a helper, ‘dspy.show_logging()’

PR #863, opened 2024-04-18
Fix retreive typo

tom-doerr:


krypticmouse:
@tom-doerr Can you try running `ruff check . --fix-only` to see if there is any issue?

tom-doerr:
Can't imagine that the error is related to my PR, I only switched to letters in the docs. I can try this latter anyway if you want

krypticmouse:
Thank You for the contribution!!

krypticmouse:
Thank You for the contribution!!

PR #869, opened 2024-04-19
feature(dsp) Add context to details

Vasilije1990:
Add context to details to make it easier to run evaluations.
We want want to make it easier to retrieve context outside of dspy retriever

### DIFF
--- a/dspy/datasets/hotpotqa.py
+++ b/dspy/datasets/hotpotqa.py
@@ -18,7 +18,7 @@ class HotPotQA(Dataset):
         for raw_example in hf_official_train:
             if raw_example['level'] == 'hard':
                 if keep_details is True:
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Hi @Vasilije1990 , thanks for the PR. This addition is fine to me, but it seems like in you want to use 'context' within evaluation, and my understanding is this is the 'gold context' from the data, and not actually retrieved context to be used for RAG evaluation. lmk if I'm misinterpreting.

Vasilije1990:
It's a correct estimate. What we are trying to do is to enrich and process gold context, in order to have an enrichment layer. This enrichment layer then evaluates vs qa pairs. 
In this way, we can see how good our enrichment does the job vs vanilla gold context.
Hope that makes sense.

For more details please check here: https://github.com/topoteretes/cognee

PR #873, opened 2024-04-19
Passing the "filters_json" param in the "forward" method

lmoros-DB:
- This change allows passing different filters to the same instance of a DatabricksRM object

### DIFF
--- a/dspy/retrieve/databricks_rm.py
+++ b/dspy/retrieve/databricks_rm.py
@@ -115,10 +115,9 @@ class DatabricksRM(dspy.Retrieve):
             payload["query_text"] = query
         else:
             raise ValueError("Invalid query type specified. Use 'vector' or 'text'.")
-        if filters_json:
-            payload["filters_json"] = filters_json 
-        elif self.filters_json:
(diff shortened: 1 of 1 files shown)


dspy/retrieve/databricks_rm.py
```diff
@@ -115,8 +115,10 @@ def forward(self, query: Union[str, List[float]], query_type: str = 'text') -> d
             payload["query_text"] = query
         else:
             raise ValueError("Invalid query type specified. Use 'vector' or 'text'.")
-        if self.filters_json:
-            payload["filters_json"] = self.filters_json     
+        if filters_json:
```
arnavsinghvi11, line 118: could this be simplified with `payload["filters_json"] = filters_json if filters_json else self.filters_json`

lmoros-DB: done

Tested the change and it works fine. 

Nitpicking: if `filters_json` and `self.filters_json` are None,  we send an extra element (`filters_json`) in the payload with value `null`.  The REST API handles it fine

Issue #893, opened 2024-04-23
DSPy is incompatible with cohere>=5.0.0


mikeldking:
The new cohere python SDK doesn't have the property `cohere.CohereAPIError`, causing a runtime exception on import.

At a minimum it might be best to just advise on downgrading the SDK?

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[<ipython-input-2-eae9ba8b34d4>](https://localhost:8080/#) in <cell line: 1>()
----> 1 import dspy

3 frames
[/usr/local/lib/python3.10/dist-packages/dsp/modules/cohere.py](https://localhost:8080/#) in <module>
      8 try:
      9     import cohere
---> 10     cohere_api_error = cohere.CohereAPIError
     11 except ImportError:
     12     cohere_api_error = Exception

AttributeError: module 'cohere' has no attribute 'CohereAPIError'
```
Minimum example: https://colab.research.google.com/drive/14ljK45-tM4_eNbjPv8QceBVj5xksyyEx?usp=sharing

mikeldking:
This is present in DSPy 2.4.5

PR #906, opened 2024-04-25
fix(cohere): Catch attribute error

isaacbmiller:
Adds another except statement to ensure that cohere can be imported correctly, regardless of version.

Missing:
Tests

Link to collab with repro/fix:
[Collab notebook](https://colab.research.google.com/drive/1yK0X0TEp_J2m9khQxbi8Ueh6nxfpSePo?usp=sharing)

Fixes: #893

### DIFF
--- a/dsp/modules/cohere.py
+++ b/dsp/modules/cohere.py
@@ -6,7 +6,6 @@ from dsp.modules.lm import LM
 
 try:
     import cohere
-
     cohere_api_error = cohere.errors.UnauthorizedError
 except ImportError:
     cohere_api_error = Exception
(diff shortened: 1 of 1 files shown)


dsp/modules/cohere.py
```diff
@@ -6,10 +6,14 @@
 
 try:
     import cohere
+
     cohere_api_error = cohere.errors.UnauthorizedError
 except ImportError:
     cohere_api_error = Exception
     # print("Not loading Cohere because it is not installed.")
+except AttributeError:
+    cohere_api_error = Exception
```
mikeldking, line 15: I'm guessing the location of the errors for cohere are different across 4.X.X and 5.X.X. Could have one more fallback for 4.X.X if supporting 4 fully is desired.

isaacbmiller: They actually aren't different, it was just wrong before from what I could dig up

PR #819, opened 2024-04-15
fix(dependencies): move builds and dependencies to poetry

ammirsm:
## 📝 Changes Description
This PR contains the following changes:

- migrated setuptools to poetry
- remove poetry main project stuffs and move them to the tool.dependencies
- export the poetry stuffs to requirements to if someone wanted to use it till June
- update PR structure message.

## 🛠 Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] This change will expire our current cache and need cache update

## 🔍 How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Added new unit/integration tests
- [x] Updated one notebook (that tests end-to-end)
- [ ] I stared at the code and made sure it makes sense

## How to test?

- the way that you can test it is to setup an environment and follow the steps for setting up your poetry and use that.

- we should run notebooks and make sure they works.

- another part that should be tested is the build of the package and push it to pypi, which should be done with poetry build 

- To publish a Python package to PyPI using Poetry, you need to follow these steps:

1. Poetry will create a distribution package that you can upload to PyPI. You can build the package with the following command:

```bash
poetry build
```

2. Before you can publish the package, you need to add the PyPI repository to Poetry if you haven't done so already. You can do this with the following command:

```bash
poetry config repositories.pypi https://upload.pypi.org/legacy/
```

4. Now you can publish your package to PyPI. You'll be prompted to enter your PyPI username and password:

```bash
poetry publish -r pypi
```

## Checklist

- [x] I have added dependencies to the `pyproject.toml` file
- [x] I have updated the `poetry.lock` file
- [x] I have updated the `requirements.txt` file (till 2024-06-01)
- [ ] I have updated the documentation accordingly
- [x] New and existing unit tests pass locally with my changes
- [ ] New and existing integration notebooks pass locally with my changes
- [x] Pre-Commit checks are passing (locally and remotely)
- [x] Title of your PR / MR corresponds to the required format
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
[...]

PR #911, opened 2024-04-26
Added support for LlamaIndex Retrievers as a DSPy Retriever Module

no-dice-io:
Given that Llama Index has some of the most wide and robust support for various retrieval abstractions and methods, adding an interface from LlamaIndex into DSPy would allow many to leverage the various retrieval classes open in LlamaIndex.

https://docs.llamaindex.ai/en/stable/api_reference/retrievers/


dspy/retrieve/llama_index_rm.py
```diff
@@ -0,0 +1,63 @@
...
+
+    retriever: BaseRetriever
+
+    def __init__(self, retriever: BaseRetriever):
+        self.retriever = retriever
+
+    @property
+    def similarity_top_k(self) -> int:
+        """Return similarity top k of retriever."""
+        return self.retriever.similarity_top_k
```
logan-markewich, line 35: not every retriever in llama-index will have this property 👀 It kind of depends. The `BaseRetriever` class is fairly generic, and just exposes `retrieve()`/`aretrieve()` methods

no-dice-io: Ooh thank you for pointing this out - I'll make an adjustment shortly!

pyproject.toml
```diff
@@ -118,6 +118,7 @@ rich = "^13.7.1"
 psycopg2 = { version = "^2.9.9", optional = true }
 pgvector = { version = "^0.2.5", optional = true }
 structlog = "^24.1.0"
+llama-index = "^0.10.30"
```
ammirsm, line 121: don't you think it should be optional?

tests/retrieve/test_llama_index_rm.py
```diff
@@ -0,0 +1,63 @@
...
+        "retriever": retriever,
+        "rm": rm,
+        "lm": DummyLM(answers=dummyset),
+        "trainset": trainset,
+        "devset": devset,
+    }
+
+
+def test_lirm_as_rm(rag_setup):
+    """Test the retriever as retriever method"""
```
ammirsm, line 48: I think this test will fail in the CI if we make that dependency optional...

pyproject.toml
ammirsm: I know it is causing lots of confusion but for the dependencies right now we use 3 different places... I have a PR to fix that in a short but if it is going to be merge before that you need to add your dependencies in requirements.txt (which use for building our package), here that you added it, and `dependencies` in the pyproject.toml ...

 more info:
https://github.com/stanfordnlp/dspy/pull/819

dspy/retrieve/llama_index_rm.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @similarity_top_k.setter
+    def similarity_top_k(self, k: int) -> None:
```
ammirsm, line 38: just to mention the `similarity_top_k` should be removed from here too.

ammirsm: and I should say these kind of configs will be handled during defining the retriever.

ammirsm (changes requested):
Thanks for the contribution and help!

The PR overal LGTM, couple small stuffs which needs attention and we can approve and merge it.

dspy/retrieve/llama_index_rm.py
ammirsm: Thanks @logan-markewich for looking into this!

pyproject.toml
no-dice-io: I'm open to making the dependency optional - but wouldn't my tests fail as you pointed out if we run that route?

dspy/retrieve/llama_index_rm.py
no-dice-io: I believe I've resolved this - please review!

dspy/retrieve/llama_index_rm.py
no-dice-io: I believe I've refactored to take this into account. Now you can specify anywhere really and it will also handle cases where the underlying LI retriever does not have the similarity_top_k property.

dspy/retrieve/llama_index_rm.py
logan-markewich: I think this looks good, but the output should be typed as `Optional[int]` though

Kind of curious what the motivation is to expose this on the class? Is this something that dspy can optimize later?

no-dice-io: I don't recall if it can be optimized, but its a parameter that is often used with their `forward` method so I was looking for a way to pass it to the underlying LI retriever.

pyproject.toml
ammirsm: I think the best way is to tag those tests to just be ran when the llama-index is installed and we fix the CI based on that.

no-dice-io: I'll go with that. I'll make the adjustment today or tomorrow. I'll only run those tests when llamaindex is present.

ammirsm:
There are some merge conflicts in pyproject and poetry and I think after that we should be good to go.

no-dice-io:
> There are some merge conflicts in pyproject and poetry and I think after that we should be good to go.

I believe I've resolved this now!

ammirsm (approved):
LGTM.

Issue #914, opened 2024-04-27
Number of generation doesn't give 'n' results on other LM clients (vllm, ollama) except OpenAI


AmoghM:
The param `n` determines the total number of generations but it only works with openAI and not with vLLM or Ollama. From the ollama logs, it appears that its`/api/generate` endpoint is hit 5 for the example below but the output is not concatenated.

```python
from typing import List
import dspy

vllm_llm = dspy.HFClientVLLM(model="meta-llama/meta-llama-3-8B-Instruct", port=8000, url="http://localhost", temperature=0.7)
dspy.settings.configure(lm=vllm_llm)

class SciGen(dspy.Signature):
    """context and answer on science questions"""
    question = dspy.InputField(desc="Input question")
    context = dspy.OutputField(desc="context") 
    answer = dspy.OutputField(desc="Answer using only the context")

context_generator = dspy.Predict(SciGen, n=5)
response = context_generator(question="Why do we have solar eclipse?")

print(len(response.completions)) # Outputs 5 for openai call but 1 for ollama

```

I can use the for loop to hit the endpoint multiple times as mentioned in the docs but I am assuming the batching of request would provide better performance.

```python
question = "Why do we have solar eclipse?"
for idx in range(5):
    response = context_generator(question=question, config=dict(temperature=0.4+0.0001*idx))
    print(f'{idx+1}.', response.completions)
```

arnavsinghvi11:
Hi @AmoghM , I believe this is related to a previously-surfaced [issue](https://github.com/stanfordnlp/dspy/issues/749) with [ollama only printing the first completion, regardless of the specified `n`](https://github.com/stanfordnlp/dspy/issues/749#issuecomment-2032789021).

It seems like ollama would need some batched request handling like done for [OpenAI](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/modules/gpt3.py#L189)

This likely applies for vLLM with respect to its corresponding batching configuration.

Feel free to open a PR to add support for batching!

mikeedjones:
AFAICT everything in modules just uses a for loop to iterate over `range(n)`? `ollama.py` has essentially the same logic as everything else in there.

mikeedjones:
Reprodued the bug with the below using OpenAI:

```python
vllm_llm = dspy.OpenAI(model='gpt-3.5-turbo')
dspy.settings.configure(lm=vllm_llm)

[...]

Issue #749, opened 2024-04-01
dspy.ChainOfThoughtWithHint() won't respect n=x


RamXX:
Traditional behavior with `dspy.ChainOfThought()`:

```
qa = dspy.ChainOfThought('question -> answer', n=3)
response = qa(question="How many floors are in the castle David Gregory inherited?")

print(len(response.completions.answer))
3
```

Current behavior with `dspy.ChainOfThoughtWithHint()`:

```
qa = dspy.ChainOfThoughtWithHint('question -> answer', n=3)
response = qa(question="How many floors are in the castle David Gregory inherited?", hint="we know it's at least 3 floors")

print(len(response.completions.answer))
1
```

LM Settings:

```
lm = dspy.OllamaLocal(model="openhermes:7b-mistral-v2.5-fp16", max_tokens=4000, timeout_s=240)
dspy.settings.configure(lm=lm)
```

okhat:
This probably only happens because the LM fails to produce 3 valid completions in parallel

RamXX:
Changing the settings to:

```
turbo = dspy.OpenAI(model='gpt-3.5-turbo-0125', max_tokens=2000)
dspy.settings.configure(lm=turbo)
```

produces the same results thou.

okhat:
Very interesting. Could you do turbo.inspect_history(n=2) ?

RamXX:
OK, I reseted the notebook and created a zero shot basic test based on the known examples like this:

```
import dspy

turbo = dspy.OpenAI(model='gpt-3.5-turbo-0125', max_tokens=2000)
ollama = dspy.OllamaLocal(model="openhermes:7b-mistral-v2.5-fp16", max_tokens=4000, timeout_s=240)

with dspy.context(lm=turbo):
    qa = dspy.ChainOfThoughtWithHint('question -> answer', n=3)
    response = qa(question="How many floors are in the castle David Gregory inherited?", hint="we know it's at least 3 floors.")
    print(f"GPT3.5 turbo answer length: {len(response.completions.answer)}")

with dspy.context(lm=ollama):
    qa = dspy.ChainOfThoughtWithHint('question -> answer', n=3)
    response = qa(question="How many floors are in the castle David Gregory inherited?", hint="we know it's at least 3 floors.")
    print(f"Ollama answer length: {len(response.completions.answer)}")
```

The output is:

```
GPT3.5 turbo answer length: 3
Ollama answer length: 1
```

[...]

PR #918, opened 2024-04-28
fix/remove n=1 in fallback generate logic when not all fields have been generated

mikeedjones:
Fixes https://github.com/stanfordnlp/dspy/issues/914 and I suspect https://github.com/stanfordnlp/dspy/issues/749

Had to disable pre-commit, linting, and formatting to keep this change atomised to a single line as this file pre-dates changes to the pre-commit settings.

Explaination:

When the required fields are not generated by in any of the completions, [fallback logic](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/primitives/predict.py#L95) is entered which `generate` is called recursively until the fields are created:

```python
        # If none of the completions is completed (i.e., none has the final field set).
        if last_field_idx < len(field_names):
            # Pick the first completion that has gone farthest.
            completion = completions[0]
```
However, in this logic, `n` - the number of completions - is set to 1 before `generate` is called again. 

```python
            new_kwargs = {
                **kwargs,
                max_tokens_key: max_tokens,
                "n": 1,
                "temperature": 0.0,
            }

            assert max_depth > 0
            return generate(template, **new_kwargs)(
                completion,
                stage=stage,
                max_depth=max_depth - 1,
                original_example=original_example,
            )
```

This PR removes `"n":1`.

These updates to the kwargs have been in since the initial commit so caution is advised! However, the signature of the function is not affected and based on my understanding of the logic, this should "just work". The temperature being 0 is another thing I would like a maintainer's view upon - as with this fix all the completions would be the same!

I do wonder if the "correct" way to fix this is to recursively continue the completion for all candidate completions as opposed to the one which has got the furthest when the user has set `n` - but that feels like a more extensive change to the logic which would need some additional thought and testing.

There are other parts of `do_generate` which [quietly strip completions](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/primitives/predict.py#L95) which could still result in behaviour like #914 even if this change to [`predict.py`]((https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/primitives/predict.py#L95)) is made. 

```python 

        last_field_idx = 0
        for field_idx, key in enumerate(field_names):
            completions_ = [
                c for c in completions if key in c.keys() and c[key] is not None
            ]

            # Filter out completions that are missing fields that are present in at least one completion.
            if len(completions_):
                completions = completions_
                last_field_idx = field_idx + 1
```

Another option is to update the documentation of `n` to make it clear that it is a limit on the number of returned completions per step in the graph and not a guarantee. For example [here](https://dspy-docs.vercel.app/docs/building-blocks/language_models#tips-and-tricks) in the docs it seems pretty clear that users are meant to treat `n` as a guarantee, not a limit.  

Cheers

### DIFF
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -112,7 +112,6 @@ def _generate(template: Template, **kwargs) -> Callable:
             new_kwargs = {
                 **kwargs,
                 max_tokens_key: max_tokens,
-                "n": 1,
                 "temperature": 0.0,
             }
 


AmoghM:
Thanks for providing the insight. 

I do think that we should use the same config and not set `temperature=0` etc.
> These updates to the kwargs have been in since the initial commit so caution is advised! However, the signature of the function is not affected and based on my understanding of the logic, this should "just work". The temperature being 0 is another thing I would like a maintainer's view upon - as with this fix all the completions would be the same!

We will need another parameter like `max_iter (max no. of times we can hit the lm API)` if we go to this route otherwise we may hit rate limits.
>I do wonder if the "correct" way to fix this is to recursively continue the completion for all candidate completions as opposed to the one which has got the furthest when the user has set n - but that feels like a more extensive change to the logic which would need some additional thought and testing.

Would love to know more thoughts from the maintainers.

okhat:
This is indeed a difficult problem. The most intuitive thing to do, is to complete each incomplete output in a big loop. This is expensive. Currently, there are no guarantees indeed.

mikeedjones:
That is currently set by max_depth - which i don't think is surfaced to the user.

mikeedjones:
> This is indeed a difficult problem. The most intuitive thing to do, is to complete each incomplete output in a big loop. This is expensive. Currently, there are no guarantees indeed.

I think then, at least, there should be warnings exposed when entering the fallback logic, when setting `n!=1`, and explicitly called out in the docs - [here](https://dspy-docs.vercel.app/docs/building-blocks/language_models#tips-and-tricks) for example.

PR #919, opened 2024-04-28
Automate pypi deployment

hmoazam:
WIP for automatically publishing package to pypi and creating github release. Testpypi automation in place for package at: https://test.pypi.org/project/dspy-ai-test/


build_utils/tests/intro.py
```diff
@@ -0,0 +1,223 @@
+###
+# Copy of intro notebook for the sake of testing new versions of the package.
```
ammirsm, line 2: I would say let's move this to a pytest and run that test case instead of running the actual intro.py file. what do you think?

setup.py
```diff
@@ -9,8 +9,8 @@
     requirements = f.read().splitlines()	
 
 setup(	
-    name="dspy-ai",	
-    version="2.4.9",	
+    name="{{PACKAGE_NAME_PLACEHOLDER}}",
```
ammirsm, line 12: are they actual place holder or they automatically will be filled up?

ammirsm: oh nvm, I think they are filled up with 
```      - name: Update package name in pyproject.toml
        run: sed -i "s/{{PACKAGE_NAME_PLACEHOLDER}}/dspy-ai-test/g" pyproject.toml```

File: .github/workflows/build_and_release.yml [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+    runs-on: ubuntu-latest
+    steps:  
+      - uses: actions/checkout@v2  
+      - name: Set up Python 3.9  
+        uses: actions/setup-python@v3
+        with:  
+          python-version: "3.9"  
+      - name: Install package from TestPyPI
+        run: |  
+          for i in {1..5}; do
```
ammirsm, line 69: for keeping this file clean do you wanna move this to the .sh file?

ammirsm (changes requested):
Great job! Thank you for your time and contribution!
Overall, this looks good to me, but here are a few suggestions:

1. It would be helpful to have a document explaining how we are building this. Including this document in the PR should be sufficient. While I have a general understanding of how it works currently, a short document connecting the dots would make it much easier to understand.
2. I've added a few comments which may help in cleaning up, but they aren't necessary.
3. Instead of copying the intro.py into the file, it might be better to move it to pytest and run it from there.

setup.py
hmoazam: Yes they are automatically filled in by the version as set by the tag (or with the version + an auto incremented identifier in case multiple pushes are made to test pypi)

.github/workflows/build_and_release.yml
hmoazam: Sure I can move this to an sh file

build_utils/tests/intro.py
hmoazam: Yes that's a good call, I'll do that. Shall I put it under tests/examples?

hmoazam:
> Great job! Thank you for your time and contribution! Overall, this looks good to me, but here are a few suggestions:
> 
> 1. It would be helpful to have a document explaining how we are building this. Including this document in the PR should be sufficient. While I have a general understanding of how it works currently, a short document connecting the dots would make it much easier to understand.

-> Shall I expand to what I've got [here](https://github.com/stanfordnlp/dspy/pull/919#issuecomment-2096652623) or is that enough detail? I also included a `release-checklist.md ` file which contains instructions for use and prereqs.

> 2. I've added a few comments which may help in cleaning up, but they aren't necessary.
> 3. Instead of copying the intro.py into the file, it might be better to move it to pytest and run it from there.

build_utils/tests/intro.py
ammirsm: I would say that would be a great place to put or we can move all of these tests to a new folder name `tests/integration` and keep them there, this way we can distinct it later to run them or not to run them where ever we want. 

I think it is good to separate these integration and maybe later we want to run them in their own CI step.

ammirsm:
> > Great job! Thank you for your time and contribution! Overall, this looks good to me, but here are a few suggestions:
> > 
> > 1. It would be helpful to have a document explaining how we are building this. Including this document in the PR should be sufficient. While I have a general understanding of how it works currently, a short document connecting the dots would make it much easier to understand.
> 
> -> Shall I expand what I've mentioned [here](https://github.com/stanfordnlp/dspy/pull/919#issuecomment-2096652623) or is that enough detail? I also included a `release-checklist.md ` file which contains instructions for use and prereqs.
> 
> > 2. I've added a few comments which may help in cleaning up, but they aren't necessary.
> > 3. Instead of copying the intro.py into the file, it might be better to move it to pytest and run it from there.

@hmoazam I think that's a good draft but i would say 
1. let's keep it somewhere in our repo as a document not a comment
2. let's add a little bit more detail like which part of the CI is doing what 

All in all, great job! 

Let's tweak it a little bit and merge it as soon as we can.

build_utils/tests/intro.py
ammirsm: I can help with setting up this one for you, it will be a little bit tricky, I will push it in another branch and add it to CI for you in couple days.

ammirsm: https://github.com/stanfordnlp/dspy/pull/1048

I added my initial commits here.

hmoazam:
This PR depends on #1192 and #1048.

Once #1048 is merged in we need to:
- Update the `test-intro-script` task to point to the pytest at `tests_integration/test_intro.py` instead of `build_utils/tests/intro.py`
- Delete `build_utils/tests/intro.py`

ammirsm (approved):
LGTM! Thanks for the hard work here @hmoazam 🚀🥳

PR #920, opened 2024-04-28
fix/Extend generation for all candidate completions

mikeedjones:
Following from https://github.com/stanfordnlp/dspy/pull/918 - This PR attempts a fuller implementation of the fix suggested there such that every candidate completion is extended until it contains valid content for every field in the template. 

Fixes https://github.com/stanfordnlp/dspy/issues/914 and I suspect https://github.com/stanfordnlp/dspy/issues/749 and some other raised issues.

The current, unpatched, behaviour also seems to be replicated in the [backend-refactor](https://github.com/stanfordnlp/dspy/tree/backend-refactor) branch.

This PR makes the following changes to the logic of `do_generate` in `dsp/primitives/predict.py`:
- Differentiates between `completions` and `finished_completions` where `finished_completions` have entries for every field in the template.
- Introduces the function`extend_generation` to `dsp/primitives/predict.py`, in the ns of `_generate` which ensures (up to two levels of recursion) that every template field has valid content for all `n` completions requested.
- Introduces `get_last_field` to acquire the last valid field of an Example.
- Maintains overwriting of temperature in extended completions.

Further suggestions:
- Should this method raise warnings suggesting that `max_tokens` should be increased? This implementation could slow down forward passes significantly.

### DIFF
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -63,6 +63,7 @@ def _generate(template: Template, **kwargs) -> Callable:
 
     def extend_generation(completion: Example, field_names: list[str], stage:str, max_depth: int, original_example:Example):
         """If the required fields are not present in the completion, extend the generation."""
+        assert max_depth > 0, "Max depth exceeded - failed to complete in one pass - increase max_tokens"
         # remove content of last field to avoid half-completed content
         for field_name in get_all_fields_following_missing_field(completion, field_names):
             completion.pop(field_name, None)
@@ -86,7 +87,6 @@ def _generate(template: Template, **kwargs) -> Callable:
             "temperature": 0.0,
         }
 
-        assert max_depth > 0, "Max depth exceeded - failed to complete in one pass - increase max_tokens"
         _, finished_completion = generate(template, **new_kwargs)(
             completion,
             stage=stage,
@@ -117,11 +117,7 @@ def _generate(template: Template, **kwargs) -> Callable:
 
         finished_completions = []
         for completion in completions:
-            all_keys_valid = all(
-                (completion.get(key, None) is not None) and 
-                (completion.get(key, None) !="") for key in field_names
-            )
-            if all_keys_valid:
+            if all((completion.get(key, "") != "") for key in field_names):
                 finished_completions.append(completion)
                 continue
             finished_completions.append(
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -63,13 +63,9 @@ def _generate(template: Template, **kwargs) -> Callable:
 
     def extend_generation(completion: Example, field_names: list[str], stage:str, max_depth: int, original_example:Example):
         """If the required fields are not present in the completion, extend the generation."""
+
         # remove content of last field to avoid half-completed content
-        for field_idx, key in enumerate(field_names):
-            if key not in completion:
-                break
-            if completion[key] is None or completion[key] == "":
-                break
-        completion[field_names[field_idx]] = ""
+        completion[get_last_field(completion, field_names)] = ""
 
         # Recurse with greedy decoding and a shorter length.
         max_tokens = (kwargs.get("max_tokens") or 
@@ -165,6 +161,12 @@ def _generate(template: Template, **kwargs) -> Callable:
 
     return do_generate
 
+def get_last_field(completion, field_names):
+    for key in field_names:
+        if key not in completion:
+            return key
+        if completion[key] is None or completion[key] == "":
+            return key
 
 def generate_sc(
     example, prompt, normalize=True, extract=None, prediction_field=None, **kwargs,
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -112,7 +112,6 @@ def _generate(template: Template, **kwargs) -> Callable:
             new_kwargs = {
                 **kwargs,
                 max_tokens_key: max_tokens,
-                "n": 1,
                 "temperature": 0.0,
             }
 
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -65,7 +65,7 @@ def _generate(template: Template, **kwargs) -> Callable:
         """If the required fields are not present in the completion, extend the generation."""
         # remove content of last field to avoid half-completed content
         for field_name in get_all_fields_following_missing_field(completion, field_names):
-            completion[field_name] = ""
+            completion.pop(field_name, None)
 
         # Recurse with greedy decoding and a shorter length.
         max_tokens = (kwargs.get("max_tokens") or 
--- a/dsp/primitives/predict.py
+++ b/dsp/primitives/predict.py
@@ -61,6 +61,45 @@ def _generate(template: Template, **kwargs) -> Callable:
 
     generator = dsp.settings.lm
 
+    def extend_generation(completion: Example, field_names: list[str], stage:str, max_depth: int, original_example:Example):
+        """If the required fields are not present in the completion, extend the generation."""
+        # remove content of last field to avoid half-completed content
+        for field_idx, key in enumerate(field_names):
+            if key not in completion:
+                break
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @mikeedjones for the PR! This makes sense, but just curious from #734 , whether setting a large number of tokens likely solves this issue? I feel like it's actually better to give the user control over request parameters and adjust accordingly than to excessively increase to more than needed (unless the recursion has some control over this)

>Should this method raise warnings suggesting that max_tokens might be increased?  

This should definitely be logged as any impact to request parameters are important to flag!

mikeedjones:
What do you mean by "solves the issue"? Increasing `max_tokens` would make it more likely that the fallback logic is not entered into, but very long or complicated signatures might still exceed even very high token-generation limits. For example, whilst Claude's context window is 200k tokens, the generation limit is 4096.

arnavsinghvi11:
I'm not sure I understand. If the generation limit is restricted, does setting `max_tokens = 4096` not capture what's done here? If the long signatures exceed the very high token-generation limits, it would not work anyways right? maybe I'm misinterpreting so feel free to correct with an example!

mikeedjones:
The current flow, implemented now on main, checks which template fields are in the n completions made in the first pass. If none of the completions contain all the fields, there is some [fallback logic](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/primitives/predict.py#L95) is entered, in which the LM (`generate` method) is called recursively until the fields are created:

```python
        # If none of the completions is completed (i.e., none has the final field set).
        if last_field_idx < len(field_names):
            # Pick the first completion that has gone farthest.
            completion = completions[0]
            ...

            new_kwargs = {
                **kwargs,
                max_tokens_key: max_tokens,
                "n": 1,
                "temperature": 0.0,
            }

            assert max_depth > 0
            return generate(template, **new_kwargs)(
                completion,
                stage=stage,
                max_depth=max_depth - 1,
                original_example=original_example,
            )
```

The fallback logic basically gets the "most complete" completion and makes a further call to the LM to generate an extra k tokens (k is chosen by some more logic in [primirtives/predict.py](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/primitives/predict.py#L95).

If `max_tokens` is increased then the likelihood the LM generates all the required fields goes up, but it is not certain. 

For example, with the below:

```python
import dspy
import os

llm = dspy.OpenAI(model='gpt-3.5-turbo', max_tokens=100)
dspy.settings.configure(lm=llm)

class SciGen(dspy.Signature):
    """context and answer on science questions"""
    question = dspy.InputField(desc="Input question")
    foo = dspy.OutputField(desc="foo_context") 
    bar = dspy.OutputField(desc="bar_context") 
    bazz = dspy.OutputField(desc="bazz_context") 
    buzz = dspy.OutputField(desc="buzz_context") 
    fuzz = dspy.OutputField(desc="fuzz_context") 
    fooz = dspy.OutputField(desc="fooz_context") 
    answer = dspy.OutputField(desc="Answer using only the contexts")

context_generator = dspy.Predict(SciGen, n=10)
response = context_generator(question="Why do we have solar eclipse?")

print(len(response.completions))
# 1
```

The first call to gpt-35 produces 10 completions, none of which contain the required fields`fooz` or `answer`. 

The fallback logic is entered, where the "most complete" completion is used (maybe the one which contains `fuzz`) as the `Example` for the next completion, with updated kwargs `{"n":1, "max_tokens": k, "temperature":0}. It is this logic which is causing only one completion to be returned in https://github.com/stanfordnlp/dspy/issues/914.

arnavsinghvi11:
Thanks @mikeedjones , this is really helpful! I do see the issue now lies more in the response parsing which triggers the fallback completion logic. 

With your code above and the proposed changes in the PR, there are indeed 10 outputted completions, but these are actually 10 "incomplete" completions due to output parsing errors (e.g. 

```
Prediction(
    foo='....',
    bar=''....',
    bazz=''...',
    buzz=''....',
    fuzz=''....\n\nFooz',
    fooz='', #is empty because of the previous parsing error in fuzz likely not producing the Fooz prefix as "Fooz:"
    answer="'.... \n\nAnswer: '...."
)
```

whereas with the existing logic, there are only 2 completions outputted, but they are "complete" with all fields parsed correctly (from the fallback logic). 

This occurs even when I remove `"n": 1" from the #918 . Based on this, I believe we need to tackle a deeper parsing issue rather than extending generation for all candidates, especially since it's better to have 2 "complete" completions instead of 10 - but ideally we want 10 "complete" completions! 

Let me know if that makes sense as this PR doesn't fully correct the existing issue (but potentially is on the right track!).

mikeedjones:
Good catch! The _last_ filled field was being used to determine if the completion should be updated - not the first _missed_ field. 

So the completion would pick up from `fuzz`, with an already empty `buzz` in the prompt.

The below should now pass:

```python
import dspy


llm = dspy.OpenAI(model='gpt-3.5-turbo', max_tokens=200)
dspy.settings.configure(lm=llm)
class SciGen(dspy.Signature):
    """context and answer on science questions"""
    question = dspy.InputField(desc="Input question")
    foo = dspy.OutputField(desc="foo_context") 
    bar = dspy.OutputField(desc="bar_context") 
    bazz = dspy.OutputField(desc="bazz_context") 
    buzz = dspy.OutputField(desc="buzz_context") 
    fuzz = dspy.OutputField(desc="fuzz_context") 
    fooz = dspy.OutputField(desc="fooz_context") 
    answer = dspy.OutputField(desc="Answer using only the contexts")

context_generator = dspy.Predict(SciGen, n=10)
response = context_generator(question="Why do we have solar eclipse?")

assert len(response.completions) == 10

for answer in response.completions:
    for key in ["foo", "bar", "bazz", "buzz", "fuzz", "fooz", "answer"]:
        assert key in answer
        assert answer[key] is not None
        assert answer[key] != ""
```

mikeedjones:
Good catch! 

Yes, looks like I was using the _last filled_ field to restart the completion as opposed to the _first missing_ field. 

Updated the test as the LM didn't reliably fill the nonsense fields - leading to inconsistent results.

```python
import dspy


llm = dspy.OpenAI(model='gpt-3.5-turbo', max_tokens=200)
dspy.settings.configure(lm=llm)
class SciGen(dspy.Signature):
    """context and answer on science questions"""
    question = dspy.InputField(desc="Input question")
    sun_context = dspy.OutputField(desc="sun_context") 
    moon_context = dspy.OutputField(desc="moon_context") 
    earth_context = dspy.OutputField(desc="earth_context") 
    relative_distances_context = dspy.OutputField(desc="relative_distances_context") 
    annual_cycle_context = dspy.OutputField(desc="annual_cycle_context") 
    distant_future_context = dspy.OutputField(desc="distant_future_context") 
    answer = dspy.OutputField(desc="Answer only when you have all the context fields.")

context_generator = dspy.Predict(SciGen, n=10)
response = context_generator(question="Why do we have solar eclipse?")

assert len(response.completions) == 10

for answer in response.completions:
    for key in ["sun_context",
                "moon_context",
                "earth_context",
                "relative_distances_context",
                "annual_cycle_context",
                "distant_future_context",
                "answer",
        ]:
        assert key in answer
        assert answer[key] is not None
        assert answer[key] != ""
```

arnavsinghvi11:
Thanks @mikeedjones . Could you run `ruff check . --fix-only` and push again? Ready to merge after that.

To confirm, this change is more comprehensive than #918 and that PR can be closed after this is merged?

mikeedjones:
@arnavsinghvi11 - is there anything outstanding for this PR? cheers! :)

arnavsinghvi11:
Thanks @mikeedjones ! Very useful PR that caught an elaborate issue!

okhat:
This should not have been merged.

mikeedjones:
There's an open issues on empty input fields which is being caused by this PR - or is the problem more serious?

Issue #887, opened 2024-04-22
Langchain Integration Broken on Change in dspy.signatures.field 


mkotlarz:
OutputField no longer has .finalize() method.

hack....
in dspy/predict/langchain.py

Replace:
from dspy.signatures.field import InputField, OutputField

With:
from dspy.signatures.field import OldInputField as InputField, OldOutputField as OutputField

Im sure this probably breaks something else....

arnavsinghvi11:
Hi @mkotlarz , thanks for raising this. `InputField` and `OutputField` were migrated to using Pydantic types, no longer using [`finalize` to set the prefix. ](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dspy/signatures/field.py#L39). It seems like we need to make this change to langchain.py as well. Does removing [those lines](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dspy/predict/langchain.py#L79) and keeping the original field imports resolve the issue for you?  If so, feel free to push a PR so we can merge this correction!

PR #921, opened 2024-04-28
Update langchain.py for old field types

mkotlarz:
in langchain.py import Old Fields as Fields

This is a short-term fix until langchain.py can be migrated to the new Pydantic fields

### DIFF
--- a/dspy/predict/langchain.py
+++ b/dspy/predict/langchain.py
@@ -9,7 +9,7 @@ import dspy
 from dspy.predict.parameter import Parameter
 from dspy.predict.predict import Predict
 from dspy.primitives.prediction import Prediction
-from dspy.signatures.field import InputField, OutputField
+from dspy.signatures.field import OldInputField as InputField, OldOutputField as OutputField
 from dspy.signatures.signature import infer_prefix
 
 # TODO: This class is currently hard to test, because it hardcodes gpt-4 usage:


arnavsinghvi11:
Thanks for the PR @mkotlarz ! 

From #887 , I'm curious if removing the specified finalized lines would correct the change instead of modifying the import? Ideally, we'd like to keep "new" versions over legacy, and handle code changes accordingly.

mkotlarz:
I agree with you, I dont like it either, but short of rewriting langchain pieces, which i may take a stab at later, its the only thing I got to work without the real forensic change.  Additionally langchain only supports pydantic 1.  I have alot of langchain code, and there is so much synergy between the two platforms, i may build some more bridges.  Agent tools is another example.. I honestly didnt want to waste time learning the old DSPY fields code, when this stuff is moving so fast.

arnavsinghvi11:
ah this is not rewriting langchain code, but rather the lines in [langchain.py](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dspy/predict/langchain.py#L79) that may not be needed given the new `dspy.InputField` and `dspy.OutputField`. Does removing those lines resolve the issue? 

```for k, v in inputs.items():
            v.finalize(k, infer_prefix(k))  # TODO: Generate from the template at dspy.Predict(Template2Signature)

        for k, v in outputs.items():
            output_field_key = k
            v.finalize(k, infer_prefix(k))
```
If so, it makes sense to remove those than to refactor the imports to old DSPy field types. Let me know if that makes sense!

mkotlarz:
No, it didnt work.

File /opt/homebrew/Caskroom/miniforge/base/envs/hugface/lib/python3.12/site-packages/dsp/templates/template_v3.py:37, in Template.__init__(self, instructions, **kwargs)
     30 self.format_handlers: dict[str, Callable] = {
     31     "context": passages2text,
     32     "passages": passages2text,
     33     "answers": format_answers,
     34 }
     36 for key, value in kwargs.items():
---> 37     if value.prefix:
     38       prefix: str = value.prefix
     39     else:

AttributeError: 'FieldInfo' object has no attribute 'prefix'

arnavsinghvi11:
Thanks for checking @mkotlarz . Could you run `ruff check . --fix-only` and push again? Should be good to merge after that!

PR #923, opened 2024-04-28
add support for multiple GPT3 clients

### LINKED ISSUE #915: I can't use two models with `dspy.OpenAI`
When I try to use two different models using `dspy.OpenAI`, it ends up using only the last declaration. As in the example below, even using `llama3` it ends up using `phi3`.
In this case I am using `dspy.OpenAI` but behind the scenes there are two instances of [llama.cpp](https://github.com/ggerganov/llama.cpp) (It does not appear to be a problem with [llama.cpp](https://github.com/ggerganov/llama.cpp), as using the API directly the problem does not recur).
When I try to use `dspy.OpenAI` for Llama-3 and `dspy.OllamaLocal` for Phi-3, everything works correctly.

In short: Even using two models, the responses are only generated by the last declared model.

```
import dspy

llama3 = dspy.OpenAI(api_base="http://localhost:8080/v1/", model_type="chat", api_key="no-key")
phi3 = dspy.OpenAI(api_base="http://localhost:4040/v1/", model_type="chat", api_key="no-key")
#phi3 = dspy.OllamaLocal(model="phi3")

#dspy.settings.configure(lm=llama3)

question = "Who are you? Who developed you?"

class QA(dspy.Signature):

with dspy.context(lm=llama3):
    a = dspy.TypedPredictor(QA)(question=question)

    print("\nLlama3: ")
    #print(llama3(question))
    print(a.answer)

with dspy.context(lm=phi3):
    b = dspy.TypedPredictor(QA)(question=question)

    print("\nPhi3: ")
    #print(phi3(question))
    print(b.answer)

#llama3.inspect_history(n=1)
#phi3.inspect_history(n=1)
```
Env:  `DSP_CACHEBOOL=false`

DSPy:    v2.4.5
OpenAI:    v1.23.4
Python:    v3.11.9
OS:      Fedora Linux 40 (KDE Plasma), Linux 6.8.7-300.fc40.x86_64
CPU:     AMD Ryzen 7 5700X 8-Core Processor
RAM:     32010 Mi

Su3h7aM:
In my tests, versions of [openai](https://github.com/openai/openai-python) above `1.0.0` are working correctly, however versions below `0.28.1` are crashing with the error:
```
choices = response["choices"]
               ~~~~~~~~^^^^^^^^^^^
KeyError: 'choices
```
but as this error also happens in `main`, I didn't look into what it could be. (I don't know how important it is for you to maintain compatibility with such an old version, but I'll take a look later anyway)

---

It was also necessary to make changes to [databricks.py](https://github.com/Su3h7aM/dspy/blob/main/dsp/modules/databricks.py) to adapt to the new behavior of the GPT3 class.

---

Pointing out that I did all the tests using the OpenAI-compatible API of [llama.cpp](https://github.com/ggerganov/llama.cpp), if anyone is willing to do tests with the OpenAI and Databricks APIs it would help to validate the PR.
Another important fact is that this PR breaks compatibility with existing caches, I couldn't think of a way to keep the caches compatible (Taking into account how the cache was implemented, I don't think it's possible).
The entire structure was based on [azure_openai.py](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/azure_openai.py), I tried to make it consistent with other parts of the code.

---

Fix #915  /  Alternative implementation to #744

### DIFF
--- a/dsp/modules/databricks.py
+++ b/dsp/modules/databricks.py
@@ -62,11 +62,11 @@ class Databricks(GPT3):
         if self.model_type == "chat":
             kwargs["messages"] = [{"role": "system", "content": "You are a helpful assistant."}, {"role": "user", "content": prompt}]
             kwargs = {"stringify_request": json.dumps(kwargs)}
-            response = custom_client_chat_request(**kwargs).json()
+            response = chat_request(self.client, **kwargs).json()
             response = json.loads(response)
         else:
             kwargs["prompt"] = prompt
-            response = custom_client_completions_request(**kwargs).json()    
+            response = completions_request(self.client, **kwargs).json()    
             response = json.loads(response)
         
         history = {
@@ -85,7 +85,7 @@ class Databricks(GPT3):
         kwargs.pop('max_tokens', None)
         kwargs.pop('top_p', None)
         kwargs.pop('n', None)
-        response = custom_client_embeddings_request(**kwargs).json()    
+        response = embeddings_request(self.client, **kwargs).json()    
         response = json.loads(response)
         embeddings = [cur_obj['embedding'] for cur_obj in response['data']][0]
         return embeddings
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
Hi @Su3h7aM , thanks for raising this PR!
These changes are similar to #744 as you've mentioned, and in both cases, this breaks existing caches, so we can't merge the PR. Tagging the comment from there:

> yeah ideally if we can add this support without modifying the existing behavior, this won't break the existing caches. Curious if you can only add the client creation as a separate flag, while keeping the rest of the cache function behavior the same. 

Also note that Databricks Model Serving is not compatible with openai 0.28.1 so the changes to `dspy.Databricks` can be reverted.

PR #925, opened 2024-04-29
Add the cohere client name to Cohere calls 

BeatrixCohere:


arnavsinghvi11:
Hi @BeatrixCohere , why is the `client_name=dspy` needed? I have not used Cohere myself, but curious where this client_name is getting used.

BeatrixCohere:
> Hi @BeatrixCohere , why is the `client_name=dspy` needed? I have not used Cohere myself, but curious where this client_name is getting used.

Hi, thanks for reviewing. It aids in debugging issues with the dspy integration with Cohere.

jalammar:
@arnavsinghvi11 This change would help prioritize DSPy at Cohere and contribute to how the models may work in the future to support its use cases.

arnavsinghvi11:
Merged. thanks @BeatrixCohere !

PR #929, opened 2024-04-29
Make Dataloader/Datasets Type Safe, Consistent and 100% Compatible with HF Datasets

KCaverly:
Hi Team, 

In an effort to increase transparency and type safety, the below PR does the following:

1. Make Dataloader/Dataset objects Pydantic & Type Safe
2. Make Dataloader/Dataset consistent. Some provided datasets were not identified as Dataset objects specifically.
3. Make Dataloaders 100% compatible with HuggingFace Datasets

As we are currently not using Datasets directly downstream (we only use list[Example]) this change should be fairly inconsequential, but opens up the opportunity for us to harden the interaction between Data/DSPy in the future.

@krypticmouse Let me know what you think.

### DIFF
--- a/dspy/datasets/colors.py
+++ b/dspy/datasets/colors.py
@@ -149,6 +149,10 @@ _all_colors = [
 class Colors(Dataset):
     sort_by_suffix: bool = Field(default=True)
 
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.load()
+
     def load(self) -> None:
         if self.sort_by_suffix:
             colors = self.sort_by_suffix(_all_colors)
--- a/dspy/datasets/colors.py
+++ b/dspy/datasets/colors.py
@@ -1,26 +1,164 @@
 import random
 
-from dspy.datasets.dataset import Dataset
+from .dataset import Dataset
+from dspy.primitives.example import Example
+from pydantic import Field
 
 ### A bunch of colors, originally from matplotlib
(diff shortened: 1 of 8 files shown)


dspy/datasets/colors.py
```diff
@@ -1,33 +1,176 @@
...
+    "olive drab",
+    "olive",
+    "dark sea green",
+]
+
 
 class Colors(Dataset):
-    def __init__(self, sort_by_suffix=True, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
+    sort_by_suffix: bool = Field(default=True)
```
arnavsinghvi11, line 151: why is this removed from the constructor?

dspy/datasets/colors.py
```diff
@@ -1,33 +1,176 @@
...
+        else:
+            colors = _all_colors
 
-        self._train = [dict(color=color) for color in train_colors]
-        self._dev = [dict(color=color) for color in dev_colors]
+        train_size = int(len(colors) * 0.6)
+        self.data["train"] = [Example(color=color) for color in colors[:train_size]]
+        self.data["dev"] = [Example(color=color) for color in colors[train_size:]]
 
-        random.Random(0).shuffle(self._train)
```
arnavsinghvi11, line 21: reason to remove randomization?

dspy/datasets/dataset.py
```diff
@@ -1,114 +1,90 @@
...
-        self.train_seed = train_seed
-        self.dev_size = dev_size
-        self.dev_seed = eval_seed
-        self.test_size = test_size
-        self.test_seed = eval_seed
-        self.do_shuffle = True
-
-        self.name = self.__class__.__name__
-
-    def reset_seeds(self, train_seed=None, train_size=None, eval_seed=None, dev_size=None, test_size=None):
```
arnavsinghvi11, line 20: are existing examples compatible with these changes?

arnavsinghvi11:
Hi @KCaverly , thanks for the PR! took a stab at it and left some comments. The changes look great and make sense to me but I just want to confirm that these are compatible with existing DSPy example notebooks (particularly intro.ipynb and any others using datasets ([GSM8K](https://github.com/stanfordnlp/dspy/blob/main/examples/math/gsm8k/CoT.ipynb)).

as an aside, I feel like these specific datasets (colors, gsm8k, hotpotqa) can likely be removed with some refactoring to ensure users can call the Dataloader abstraction to retrieve them. We could instead add them as examples/documentation (but that can be for a separate PR).

KCaverly:
Thanks @arnavsinghvi11, for the review. I can double check on the examples, and I agree with the comment on future refactors. 

I think two improvements for following up, would be:

1. Move examples out of the core library. 
2. Collapse Dataloader functionality into Dataset.
- I dont see the need for a separate Dataloader class, it feels to me like these should just be moved to factory functions on the Dataset object.

dspy/datasets/colors.py
KCaverly: Nice catch - I've added it back in.

krypticmouse:
@KCaverly Sorry about the delay, just catching up to this. Why did we remove sample and split functionalities :thinking:

KCaverly:
Hey @krypticmouse! 

Weve still got ‘split_existing_split’ and ‘sample_split’. Which offer functionality to split existing splits, and sample/shuffle. I tried to make the base functionality more deterministic and consistent, but all the existing functionality should be possible still. 

Not quite sure what your comment means on other formats. Does the two outline methods above have all the functionality we need? How can we improve this?

PR #935, opened 2024-04-30
Add docstrings and type hints to Evaluate class.

rpgoldman:
Add docstrings and type hints for the Evaluate class.

Needs review!

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -152,8 +152,9 @@ class Evaluate:
             num_threads: Optional[int] = ...,
             display_progress: Optional[bool] = ...,
             display_table: Union[bool, int, None] = ...,
-            return_all_scores: Literal[True] = ...,
-            return_outputs: Literal[True] = ...,
+            *,
+            return_all_scores: Literal[True],
+            return_outputs: Literal[True],
     ) -> Tuple[float, List, List[float]]:
         ...
 
@@ -166,8 +167,9 @@ class Evaluate:
             num_threads: Optional[int] = ...,
             display_progress: Optional[bool] = ...,
             display_table: Union[bool, int, None] = ...,
-            return_all_scores: Literal[True] = ...,
-            return_outputs: Literal[False] = ...,
+            *,
+            return_all_scores: Literal[True],
+            return_outputs: Literal[False],
     ) -> Tuple[float, List[float]]:
         ...
 
@@ -180,8 +182,9 @@ class Evaluate:
             num_threads: Optional[int] = ...,
             display_progress: Optional[bool] = ...,
             display_table: Union[bool, int, None] = ...,
-            return_all_scores: Literal[False] = ...,
-            return_outputs: Literal[True] = ...,
+            *,
+            return_all_scores: Literal[False],
+            return_outputs: Literal[True],
     ) -> Tuple[float, List[Tuple[Example, Prediction, Any]]]:
         ...
 
@@ -194,8 +197,9 @@ class Evaluate:
             num_threads: Optional[int] = ...,
             display_progress: Optional[bool] = ...,
             display_table: Union[bool, int, None] = ...,
-            return_all_scores: Literal[False] = ...,
-            return_outputs: Literal[False] = ...,
+            *,
+            return_all_scores: Literal[False],
+            return_outputs: Literal[False],
     ) -> float:
         ...
 
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -54,7 +54,7 @@ class Evaluate:
         If `True`, all rows will be displayed.
     max_errors: int, optional: default 5
     return_outputs: bool, optional
-        Whether or not to return the predictions.
+        Whether or not to return individual (per-`Example`) metrics.
 
     Notes
     =====
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -222,17 +222,21 @@ class Evaluate:
         """
         Returns
         =======
-        score: float, results: list of Prediction, all_scores: list of scores
-            if `return_all_scores` and `return_outputs` are both True.
+        score: float, results: list of `Prediction`, all_scores: list of scores
+            if `return_all_scores` and `return_outputs` are both `True`.
         score: float, all_scores: list of scores
-            if `return_all_scores` is True and `return_outputs` is True.
-        score: float, results: list of Prediction
-            if `return_all_scores` is False and `return_outputs` is True.
+            if `return_all_scores` is `True` and `return_outputs` is `False`.
+        score: float, results: list of `Prediction`
+            if `return_all_scores` is `False` and `return_outputs` is `True`.
         score: float
-            if both flags are false
+            `return_all_scores` and `return_outputs` are both `False`.
         """
         metric = metric if metric is not None else self.metric
+        if metric is None:
+            raise ValueError("metric is required for Evaluation")
         devset = devset if devset is not None else self.devset
+        if devset is None:
+            raise ValueError("devset is required for Evaluation")
         num_threads = num_threads if num_threads is not None else self.num_threads
         display_progress = (
             display_progress if display_progress is not None else self.display_progress
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -51,6 +51,7 @@ class Evaluate:
     display_table: Union[bool, int], optional
         This is either a boolean indicating whether or not to display a table of
         results or an integer indicating how many rows of results to display.
+        If `True`, all rows will be displayed.
     max_errors: int, optional
     return_outputs: bool, optional
         Whether or not to return the predictions.
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -1,12 +1,17 @@
 import sys
 import threading
 import types
+from collections.abc import Callable
+from typing import Literal, List, Tuple, Optional, Iterable, overload, Union
 
+# noinspection PyUnresolvedReferences
 import pandas as pd
+# noinspection PyUnresolvedReferences
 import tqdm
 from tqdm.contrib.logging import logging_redirect_tqdm
 
 import dspy
+from dspy import Example
 
 try:
     from IPython.display import HTML
@@ -14,9 +19,9 @@ try:
 except ImportError:
     ipython_display = print
 
+
     def HTML(x):
         return x
-
 
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
@@ -29,17 +34,39 @@ we print the number of failures, the first N examples that failed, and the first
 
 
 class Evaluate:
+    """
+    Reification of the process of evaluating a model.  Invoked using its
+    `__call__` method.
+
+    Parameters
+    ==========
+    devset: Iterable[Example]
+       An iterable of Examples
+    metric: Callable, optional
+       A function that takes an Example and a Prediction and keyword args as arguments
+       and returns a boolean
+    num_threads: int, optional
+       How many
+    display_progress: bool, optional
+    display_table: Union[bool, int], optional
+        This is either a boolean indicating whether or not to display a table of
+        results or an integer indicating how many rows of results to display.
+    max_errors: int, optional
+    return_outputs: bool, optional
+        Whether or not to return the predictions.
+    """
+
     def __init__(
-        self,
-        *,
-        devset,
-        metric=None,
-        num_threads=1,
-        display_progress=False,
-        display_table=False,
-        max_errors=5,
-        return_outputs=False,
-        **_kwargs,
+            self,
+            *,
+            devset: Iterable[Example],
+            metric: Optional[Callable] = None,
+            num_threads=1,
+            display_progress=False,
+            display_table: Union[bool, int] = False,
+            max_errors=5,
+            return_outputs=False,
+            **_kwargs,
     ):
         self.devset = devset
         self.metric = metric
@@ -53,7 +80,10 @@ class Evaluate:
 
         if "display" in _kwargs:
             dspy.logger.warning(
-                "DeprecationWarning: 'display' has been deprecated. To see all information for debugging, use 'dspy.set_log_level('debug')'. In the future this will raise an error.",
+                ("DeprecationWarning: 'display' has been deprecated. "
+                 "To see all information for debugging, use 'dspy.set_log_level('debug')'. "
+                 "In the future this will raise an error."
+                 )
             )
 
     def _execute_single_thread(self, wrapped_program, devset, display_progress):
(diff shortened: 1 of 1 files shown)


rpgoldman:
Still some typing errors:
```
dspy/evaluate/evaluate.py:17: error: Cannot find implementation or library stub for module named "IPython.display"  [import-not-found]
dspy/evaluate/evaluate.py:147: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
dspy/evaluate/evaluate.py:147: error: Overloaded function signatures 1 and 3 overlap with incompatible return types  [overload-overlap]
dspy/evaluate/evaluate.py:161: error: Overloaded function signatures 2 and 3 overlap with incompatible return types  [overload-overlap]
dspy/evaluate/evaluate.py:264: error: "type[tqdm[Any]]" has no attribute "_instances"  [attr-defined]
dspy/evaluate/evaluate.py:303: error: "Series[Any]" not callable  [operator]
```

Will check these as type permits.

rpgoldman:
With help from the mypy folks, the type errors are now fixed, and mypy likes the file.

dspy/evaluate/evaluate.py
```diff
@@ -1,23 +1,28 @@
 import sys
 import threading
 import types
+from collections.abc import Callable
+from typing import Literal, List, Tuple, Optional, Iterable, overload, Union, Any
 
+# noinspection PyUnresolvedReferences
```
arnavsinghvi11, line 7: are these comments needed? prefer to remove non-code related comments

dspy/evaluate/evaluate.py
```diff
@@ -1,23 +1,28 @@
...
 import tqdm
 from tqdm.contrib.logging import logging_redirect_tqdm
 
 import dspy
+from dspy import Example, Prediction
 
 try:
-    from IPython.display import HTML
-    from IPython.display import display as ipython_display
+    # noinspection PyPackageRequirements
```
arnavsinghvi11, line 17: here as well

dspy/evaluate/evaluate.py
```diff
@@ -1,23 +1,28 @@
...
+    from IPython.display import HTML  # type: ignore
+    # noinspection PyPackageRequirements
+    from IPython.display import display as ipython_display  # type: ignore
 except ImportError:
     ipython_display = print
-
-    def HTML(x):
+    def HTML(x): # noqa - linters dislike upper case function name
```
arnavsinghvi11, line 23: can remove this as well

dspy/evaluate/evaluate.py
```diff
@@ -29,17 +34,40 @@ def HTML(x):
 
 
 class Evaluate:
+    """
+    Reification of the process of evaluating a model.  Invoked using its
```
arnavsinghvi11, line 38: the evaulate class is for a DSPy program, not a model. Also, can the wording be simplified? :)

dspy/evaluate/evaluate.py
```diff
@@ -29,17 +34,40 @@ def HTML(x):
...
+
+    Parameters
+    ==========
+    devset: Iterable[Example]
+       An iterable of Examples
+    metric: Callable, optional
+       A function that takes an Example and a Prediction and keyword args as arguments
+       and returns a boolean
+    num_threads: int, optional
+       How many
```
arnavsinghvi11, line 49: comment left incomplete

dspy/evaluate/evaluate.py
```diff
@@ -29,17 +34,40 @@ def HTML(x):
...
+    display_progress: bool, optional
+    display_table: Union[bool, int], optional
+        This is either a boolean indicating whether or not to display a table of
+        results or an integer indicating how many rows of results to display.
+        If `True`, all rows will be displayed.
+    max_errors: int, optional
+    return_outputs: bool, optional
+        Whether or not to return the predictions.
```
arnavsinghvi11, line 57: return_outputs is for returning the evaluations of the program.

dspy/evaluate/evaluate.py
```diff
@@ -29,17 +34,40 @@ def HTML(x):
...
+    metric: Callable, optional
+       A function that takes an Example and a Prediction and keyword args as arguments
+       and returns a boolean
+    num_threads: int, optional
+       How many
+    display_progress: bool, optional
+    display_table: Union[bool, int], optional
+        This is either a boolean indicating whether or not to display a table of
+        results or an integer indicating how many rows of results to display.
+        If `True`, all rows will be displayed.
```
arnavsinghvi11, line 54: add comment for behavior when `int` type is specified

dspy/evaluate/evaluate.py
```diff
@@ -29,17 +34,40 @@ def HTML(x):
...
 class Evaluate:
+    """
+    Reification of the process of evaluating a model.  Invoked using its
+    `__call__` method.
+
+    Parameters
+    ==========
+    devset: Iterable[Example]
+       An iterable of Examples
+    metric: Callable, optional
```
arnavsinghvi11, line 45: change to match this [description](https://github.com/stanfordnlp/dspy/pull/951/files#diff-5a58ab8e720c8980fe710fbcacb3c8ce11ac0df0e6bcc14122156ac4f3e4b0baR53)

dspy/evaluate/evaluate.py
```diff
@@ -93,27 +135,110 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
-        devset=None,
-        num_threads=None,
-        display_progress=None,
-        display_table=None,
-        return_all_scores=False,
-        return_outputs=False,
+            self,
+            program,
+            metric: Optional[Callable] = None,
+            devset: Optional[Iterable] = None,  # Needs more specific type, if possible
```
arnavsinghvi11, line 211: remove comment

dspy/evaluate/evaluate.py
```diff
@@ -93,27 +135,110 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
+        Returns
+        =======
+        score: float, results: list of Prediction, all_scores: list of scores
+            if `return_all_scores` and `return_outputs` are both True.
+        score: float, all_scores: list of scores
+            if `return_all_scores` is True and `return_outputs` is True.
+        score: float, results: list of Prediction
+            if `return_all_scores` is False and `return_outputs` is True.
+        score: float
+            if both flags are false
```
arnavsinghvi11, line 228: keep consistent with phrasing in 222

dspy/evaluate/evaluate.py
```diff
@@ -132,8 +257,10 @@ def wrapped_program(example_idx, example):
 
                 # increment assert and suggest failures to program's attributes
                 if hasattr(program, "_assert_failures"):
+                    # noinspection PyProtectedMember
```
arnavsinghvi11, line 260: remove comments

dspy/evaluate/evaluate.py
```diff
@@ -151,11 +278,14 @@ def wrapped_program(example_idx, example):
                 if creating_new_thread:
                     del thread_stacks[threading.get_ident()]
 
-        devset = list(enumerate(devset))
-        tqdm.tqdm._instances.clear()
+        devset = list(enumerate(devset))  # This actually changes the type of `devset`
```
arnavsinghvi11, line 281: remove comments

dspy/evaluate/evaluate.py
```diff
@@ -164,27 +294,39 @@ def wrapped_program(example_idx, example):
...
 
         # Truncate every cell in the DataFrame
-        if hasattr(df, "map"):  # DataFrame.applymap was renamed to DataFrame.map in Pandas 2.1.0
+        if hasattr(
+                df, "map"
+        ):  # DataFrame.applymap was renamed to DataFrame.map in Pandas 2.1.0
             df = df.map(truncate_cell)
         else:
-            df = df.applymap(truncate_cell)
+            df = df.applymap(truncate_cell)  # type: ignore
```
arnavsinghvi11, line 322: remove comments

dspy/evaluate/evaluate.py
```diff
@@ -93,27 +135,110 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
-    def _update_progress(self, pbar, ncorrect, ntotal):
-        pbar.set_description(f"Average Metric: {ncorrect} / {ntotal}  ({round(100 * ncorrect / ntotal, 1)})")
+    @staticmethod
+    def _update_progress(pbar, ncorrect, ntotal):
+        pbar.set_description(
+            f"Average Metric: {ncorrect} / {ntotal}  ({round(100 * ncorrect / ntotal, 1)})"
+        )
         pbar.update()
 
+    ### Type stubs below for three alternative calling patterns, depending
```
arnavsinghvi11, line 145: Given that we have these overloading functions, does it make sense to add the return types here instead of all in the single __call__ function?
Also remove comments when resolved in L145, L152, L167, L182

arnavsinghvi11 (changes requested):

dspy/evaluate/evaluate.py
rpgoldman: Not sure what you mean by simplifying the wording.  I can't think of a synonym for "reification," if that's what you mean.

> Class representing the process of evaluating a DSpy program.

Would that be better?

dspy/evaluate/evaluate.py
rpgoldman: Whoops. Fixing. Thanks for the catch.

dspy/evaluate/evaluate.py
rpgoldman: These are used to quash linter complaints. I'd prefer to keep them so that I don't see the same linter complaints over and over.

dspy/evaluate/evaluate.py
rpgoldman: Thanks again.  Fixed.

dspy/evaluate/evaluate.py
rpgoldman: > or an integer indicating how many rows of results to display.
??

dspy/evaluate/evaluate.py
rpgoldman: I'm not sure I know what is meant here -- that link just takes me to a very long diff output.  If you can give me a line number, I will fix it.

dspy/evaluate/evaluate.py
rpgoldman: Should I refine these type hints?  I believe they should be:

``` python
metric: Optional[Callable[[Iterable[Example], Iterable[Prediction], MetricValue] = None
devset: Optional[Iterable[Example]] = None
```
with `MetricValue` defined as `Union[bool, float]`.

dspy/evaluate/evaluate.py
rpgoldman: Good catch.  Fixed this. Also one of the lines was wrong, and I added a check (to produce a more understandable error message when required arguments are missing).

dspy/evaluate/evaluate.py
rpgoldman: Fixed, but renamed.  See if you are ok with the change, please.

dspy/evaluate/evaluate.py
rpgoldman: Actually, I'm a little confused about what's going on here with respect to `return_scores`.  I have said that `__call__` returns a list of floats for `return_scores`, but metric functions can return boolean values, so maybe the right type for this return value is `Union[List[bool], List[float]]`

I'm not sure if booleans are coerced to floats at some point

cdowellmdb:
@arnavsinghvi11 @rpgoldman what is left for this to be merged?

rpgoldman:
@cdowellmdb I think it's good to go. @arnavsinghvi11 didn't like the comments I put in to muffle linters; I prefer to keep them. Is that a show-stopper?

If not, do you need me to merge in the changes since this was pushed?

dspy/evaluate/evaluate.py
arnavsinghvi11: we don't maintain these comments throughout the repo and it's actually fine to flag the linter and add in the changes within the PR. would prefer to remove them.

dspy/evaluate/evaluate.py
arnavsinghvi11: link was for L53 above, corresponding to the metric description

dspy/evaluate/evaluate.py
arnavsinghvi11: missed that, thanks!

dspy/evaluate/evaluate.py
arnavsinghvi11: I think the latter`Union[List[bool], List[float]]` makes sense for this!

dspy/evaluate/evaluate.py
```diff
@@ -132,8 +265,10 @@ def wrapped_program(example_idx, example):
 
                 # increment assert and suggest failures to program's attributes
                 if hasattr(program, "_assert_failures"):
+                    # noinspection PyProtectedMember
                     program._assert_failures += dspy.settings.get("assert_failures")
                 if hasattr(program, "_suggest_failures"):
+                    # noinspection PyProtectedMember
```
arnavsinghvi11, line 271: flagging the comments

dspy/evaluate/evaluate.py
```diff
@@ -151,40 +286,55 @@ def wrapped_program(example_idx, example):
                 if creating_new_thread:
                     del thread_stacks[threading.get_ident()]
 
-        devset = list(enumerate(devset))
-        tqdm.tqdm._instances.clear()
+        ndevset: List[int, Example] = list(enumerate(devset))
+        # noinspection PyProtectedMember
```
arnavsinghvi11, line 290: flagging the comments

arnavsinghvi11 (changes requested):

arnavsinghvi11:
Thanks for circling back to this @rpgoldman @cdowellmdb ! Left a few comments back on the review and it should be good to merge once the last parts are resolved! (There's also a merge conflict but I believe rebasing to main should fix this).

The in-line comments are definitely not a show-stopper, but I'd prefer to leave them out from committed code to ensure they catch any changes needed.

rpgoldman:
> The in-line comments are definitely not a show-stopper, but I'd prefer to leave them out from committed code to ensure they catch any changes needed.

Not sure I understand this last.  What's "they" here?  I assume you are saying that we should continue to see these linter warnings.  Here's an example of why I disagree:

E.g., I put in `# noinspection PyProtectedMember` because that piece of code uses a private member:

    program._assert_failures += dspy.settings.get("assert_failures")

One of 2 things should happen, IMO:
1. We add a method like `increment_assert_failures` to the sort of thing that counts assertion and suggestion failures and use that instead.
2. We decide we want to continue to use the protected member, and as I do, quash the warning.

I'm ok with either solution, but a third solution, where we just have linters continue shout at us, seems generally bad.  Bad because if we end up with lots of these linter shoutings then it's just human nature to start ignoring the linter warnings altogether. And we certainly cannot put "must build cleanly" into our testing process, if we do this.

I have been bitten many times by code repositories that emit lots of warnings, training their developers to ignore linter warnings.  Because sooner or later there's a warning that's really important, but that signal is lost in the noise.  I'm a strong believer in building clean. If there's a warning, either it's real and you fix it, or it's not real and you quash it.

PR #941, opened 2024-04-30
Removes unnecessary printing of the training_data_directory

Su3h7aM:


arnavsinghvi11:
Hi @Su3h7aM , thanks for the PR! is it better to instead refactor into logging instead of removing entirely?

arnavsinghvi11:
Hi @Su3h7aM , just following up on the comment above. Let me know if anything is unclear!

PR #942, opened 2024-05-01
feature(dspy): Copro optimizer prompts are now injectable

tobicoveo:
What:
Allow additional instruction to be added to the signatures used to generate the prompts.

Why:
Generally speaking, it allows the users more freedom regarding the instructions used to generate the prompts.

In our case, It unluck new possibilities to guide the generating prompt. Optimization based on metrics is already a powerful approach. However, being able to add details regarding how the prompt will be evaluated would help generate prompts that fulfill the requirements that must be filled out without relying only on the LLM to understand what is essential from the metrics score.

How:
- Simply added a field to append instructions to the base one in property methods. 
- The changes are backward compatible from the None initialization.

### DIFF
--- a/dspy/teleprompt/copro_optimizer.py
+++ b/dspy/teleprompt/copro_optimizer.py
@@ -29,6 +29,7 @@ Note that this teleprompter takes in the following parameters:
                     * results_latest: The min,max,avg,stddev of newest prompt scores for each predictor at each depth.
                     * total_calls: The total number of calls to the task metric.
                 These statistics will be returned as attributes of the best program.
+* additional_instructions: Instructions appended to the generation signatures. Can be used to provide explicit details on the optimization metric. 
 """
 
 
@@ -63,8 +64,7 @@ class COPRO(Teleprompter):
         depth=3,
         init_temperature=1.4,
         track_stats=False,
-        basic_generate_instruction=None,
-        generate_instruction_given_attempts=None,
+        additional_instructions=None,
         **_kwargs,
     ):
         if breadth <= 1:
@@ -75,11 +75,26 @@ class COPRO(Teleprompter):
         self.init_temperature = init_temperature
         self.prompt_model = prompt_model
         self.track_stats = track_stats
-        self.basic_generate_instruction = basic_generate_instruction or BasicGenerateInstruction
-        self.generate_instruction_given_attempts = generate_instruction_given_attempts or GenerateInstructionGivenAttempts
+
+        self.basic_generate_instruction = (
+            self._get_signature(BasicGenerateInstruction).with_instructions(
+                " ".join([BasicGenerateInstruction.instructions, additional_instructions])
+            )
+            if additional_instructions
+            else BasicGenerateInstruction
+        )
+        self.generate_instruction_given_attempts = (
+            self._get_signature(GenerateInstructionGivenAttempts).with_instructions(
+                " ".join([GenerateInstructionGivenAttempts.instructions, additional_instructions])
+            )
+            if additional_instructions
+            else GenerateInstructionGivenAttempts
+        )
 
         if "verbose" in _kwargs:
-            dspy.logger.warning("DeprecationWarning: 'verbose' has been deprecated. To see all information for debugging, use 'dspy.set_log_level('debug')'. In the future this will raise an error.")
+            dspy.logger.warning(
+                "DeprecationWarning: 'verbose' has been deprecated. To see all information for debugging, use 'dspy.set_log_level('debug')'. In the future this will raise an error."
+            )
 
     def _check_candidates_equal(self, candidate1, candidate2):
         for p1, p2 in zip(candidate1["program"].predictors(), candidate2["program"].predictors()):
@@ -315,7 +330,9 @@ class COPRO(Teleprompter):
                     )(attempted_instructions=attempts)
 
                 if self.prompt_model:
-                    dspy.logger.debug(f"(self.prompt_model.inspect_history(n=1)) {self.prompt_model.inspect_history(n=1)}")
+                    dspy.logger.debug(
+                        f"(self.prompt_model.inspect_history(n=1)) {self.prompt_model.inspect_history(n=1)}"
+                    )
                 # Get candidates for each predictor
                 new_candidates[id(p_base)] = instr.completions
                 all_candidates[id(p_base)].proposed_instruction.extend(instr.completions.proposed_instruction)
--- a/dspy/teleprompt/copro_optimizer.py
+++ b/dspy/teleprompt/copro_optimizer.py
@@ -75,26 +75,31 @@ class COPRO(Teleprompter):
         self.init_temperature = init_temperature
         self.prompt_model = prompt_model
         self.track_stats = track_stats
+        self._additional_instructions = additional_instructions
 
-        self.basic_generate_instruction = (
-            self._get_signature(BasicGenerateInstruction).with_instructions(
-                " ".join([BasicGenerateInstruction.instructions, additional_instructions])
-            )
-            if additional_instructions
-            else BasicGenerateInstruction
-        )
-        self.generate_instruction_given_attempts = (
-            self._get_signature(GenerateInstructionGivenAttempts).with_instructions(
-                " ".join([GenerateInstructionGivenAttempts.instructions, additional_instructions])
-            )
-            if additional_instructions
-            else GenerateInstructionGivenAttempts
-        )
 
         if "verbose" in _kwargs:
             dspy.logger.warning(
                 "DeprecationWarning: 'verbose' has been deprecated. To see all information for debugging, use 'dspy.set_log_level('debug')'. In the future this will raise an error."
             )
+
+    @property
+    def basic_generate_instruction(self):
+        return (self._get_signature(BasicGenerateInstruction).with_instructions(
+                " ".join([BasicGenerateInstruction.instructions, self._additional_instructions])
+            )
+            if self._additional_instructions
+            else BasicGenerateInstruction)
+
+    @property
+    def generate_instruction_given_attempts(self):
+        return (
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
Thanks for the PR @tobicoveo ! 
Small clarification that this ensures the DSPy *signature* is injectable, not the prompts exactly.

Could you please rebase and merge this branch with the latest version on main? Seems like there was a failing test on our end (nothing wrong with this PR) that just needs an update. 

Also, could you add a couple lines of documentation to the [Usage Suggestions](https://github.com/stanfordnlp/dspy/blob/d7dace6b51e1c0899fd798c575262310939f6deb/dspy/teleprompt/copro_optimizer.py#L10), specifying what the injectable signatures are defaulted to and what they are used for, now that they are parameters? Ready to merge after that! 

As an aside, seems like this can be extended to the MIPRO optimizer, if you'd like to tag that along with this PR! :)
(small note - that would require a bit refactoring in the [test](https://github.com/stanfordnlp/dspy/blob/d7dace6b51e1c0899fd798c575262310939f6deb/tests/teleprompt/test_mipro_optimizer.py#L6) as well)

dspy/teleprompt/copro_optimizer.py
```diff
@@ -153,13 +157,13 @@ def compile(self, student, *, trainset, eval_kwargs):
             if self.prompt_model:
                 with dspy.settings.context(lm=self.prompt_model):
                     instruct = dspy.Predict(
-                        BasicGenerateInstruction,
+                        self.basic_generate_instruction,
                         n=self.breadth - 1,
                         temperature=self.init_temperature,
                     )(basic_instruction=basic_instruction)
```
mikeedjones, line 178: These calls to Predict, and later attempts to access members of instruct place pretty strict requirements on the acceptable signatures - maybe more flexible to allow the partial-creation of the entire instruct object?

Think a protocol or the like which specifies the signature of the signature you pass in is probably a must to allow users to understand exactly what they can put in here.

tobicoveo:
Thanks @arnavsinghvi11! 

- I rebased and merged the main from dspy.
- I also limited the change to injecting instructions via string to avoid potential errors where the signature fields aren't the same as the base class.
- Added the changes to MIPRO.
- Added details about how to use it in the Usage Suggestions

dspy/teleprompt/copro_optimizer.py
tobicoveo: Hey @mikeedjones, yeah I agree it was a little bit too loose given the requirements of the signatures. Instead of making the signature injectable, I allowed the user to add instructions and create the signatures from the base class.

Let me know if you see an issue with this approach.

dspy/teleprompt/copro_optimizer.py
```diff
@@ -74,8 +76,25 @@ def __init__(
...
+            if additional_instructions
+            else BasicGenerateInstruction
+        )
+        self.generate_instruction_given_attempts = (
+            self._get_signature(GenerateInstructionGivenAttempts).with_instructions(
+                " ".join([GenerateInstructionGivenAttempts.instructions, additional_instructions])
+            )
+            if additional_instructions
+            else GenerateInstructionGivenAttempts
+        )
```
mikeedjones, line 92: use `dspy.Signature.with_updated_fields` to update `BasicGenerateInstruction` and `GenerateInstructionGivenAttempts`?

tobicoveo: I don't want to update the fields, but the instructions. Unless I am mistaken, the Signature class doesn't have a method for this.

mikeedjones: Yes - just realised the change in approach. Think it might be worth adding as a helper method? @arnavsinghvi11 ?

dspy/teleprompt/copro_optimizer.py
mikeedjones: Ah - that's what "with_instructions" does!

dspy/teleprompt/copro_optimizer.py
mikeedjones: sorry - that's what `with_instructions` does!

dspy/teleprompt/copro_optimizer.py
tobicoveo: But I want to preserve the initial instructions as well; it would be something more towards the line of `append_to_instructions.`

dspy/teleprompt/copro_optimizer.py
```diff
@@ -73,6 +75,8 @@ def __init__(
         self.init_temperature = init_temperature
         self.prompt_model = prompt_model
         self.track_stats = track_stats
+        self.basic_generate_instruction = basic_generate_instruction or BasicGenerateInstruction
+        self.generate_instruction_given_attempts = generate_instruction_given_attempts or GenerateInstructionGivenAttempts
```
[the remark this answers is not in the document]
tobicoveo: To avoid mutation of `BasicGenerateInstruction`,  I encapsulated them into properity methods.

Let me know if it is good with you!

dspy/teleprompt/copro_optimizer.py
```diff
@@ -73,9 +75,31 @@ def __init__(
...
+
+    @property
+    def generate_instruction_given_attempts(self):
+        return (
+            self._get_signature(GenerateInstructionGivenAttempts).with_instructions(
+                " ".join([GenerateInstructionGivenAttempts.instructions, self._additional_instructions])
+            )
+            if self._additional_instructions
+            else GenerateInstructionGivenAttempts
+        )
```
mikeedjones, line 102:
```suggestion

    def append_instructions(self, base_signature, additional_instructions):
        return self._get_signature(base_signature).with_instructions(
            " ".join([BasicGenerateInstruction.instructions, self._additional_instructions or ""])
        )

    @property
    def basic_generate_instruction(self):
        return append_instructions(BasicGenerateInstruction, self._additional_instructions)

    @property
    def generate_instruction_given_attempts(self):
        return append_instructions(GenerateInstructionGivenAttempts, self._additional_instructions)
```

dspy/teleprompt/copro_optimizer.py
mikeedjones: Do you think it would be more straightforward for users to specify a protocol which defines what the signature of an injectable `BasicGenerateInstruction` has to be and then leave it to the user to grab the original instructions from this file, if they want to append to the originals?

I think there's a wider discussion on how to manage the internal prompts of dspy and how to enable user access to them?

dspy/teleprompt/copro_optimizer.py
tobicoveo: Added the intuitions behind your recommendation. There were a few typos, but I fixed them.

dspy/teleprompt/copro_optimizer.py
tobicoveo: I think down the line, it would make sense to allow the user to change the whole instructions. One could even consider some kind of meta-learning approach on the instructions of the signature used to generate the other prompts (i.e., what signatures lead to the best prompt under a few-shot learning approach).

There should be a discussion about enabling users to access these signatures safely IMO. 

I am merely starting to get familiar with this library and wanted to keep the changes very scoped in the context of this PR. 😄

tobicoveo:
@arnavsinghvi11 and @mikeedjones, is there anything missing to close this PR?

dspy/teleprompt/copro_optimizer.py
mikeedjones: I guess the concern would be making a small change here actually alters the interface to Copro, which then has to be supported until 1.x is released. Maybe better to figure out how to more modify the templates more generally?

arnavsinghvi11:
tagging @XenonMolecule to further confirm the behavior of COPRO and MIPRO. LGTM.

XenonMolecule:
This looks good to me!  Cool idea to allow user-defined grounding for instruction proposal!

mikeedjones:
I'm really not sure about this PR. Making a small change here actually alters the interface to Copro, which then has to be supported until a breaking 3.x release. 

I think this functionality would be better implemented by figureing out how to more modify the internal prompts of dspy, and then apply that same logic here. Does that make sense?

XenonMolecule:
Sure, I see your point, Michael.  Are you suggesting some config or shared interface that lets you swap out the prompts for COPRO, MIPRO, and any other DSPy internal program that uses prompts?  This is a good design decision to come up with before we lock into a specific interface too soon.

One thing that we've been working on for some internal projects is refactoring MIPRO and COPRO to allow user-defined proposal programs that can include any number of details about your program.  This would mean taking into account more than just the Dataset Summary, but also things like the code in your program itself, various prompt engineering tips, etc.  Enabling users to plug-in from existing proposal programs or write their own might be a better interface to commit to.

mikeedjones:
Yes - there are ~8 `Signature`s in mipro, any one of which might be causing a user some headache. 

I think it would be possible to use something like a contextvar and a `replace` method associated with the `Signature` class which allows the user to replace or update subclasses by reference. The UX would be something like:

```python
import dspy

class MyBasicGenerateInstructionWithExamples(teleprompt.mipro_optimizer.MyBasicGenerateInstructionWithExamples):
    "you are foo"
    basic_instruction = dspy.InputField(desc="The initial instructions before bar")

with teleprompt.mipro_optimizer.BasicGenerateInstructionWithExamples.use(MyBasicGenerateInstructionWithExamples):
    ...
```

mikeedjones:
Sketched out in https://github.com/stanfordnlp/dspy/pull/1090

tobicoveo:
> ```python
> with mipro_optimizer.BasicGenerateInstruction.replace(MyBasicGenerateInstruction):
>     print(mipro_optimizer.BasicGenerateInstruction.__doc__)
> ```

I already checked it out, and it looks good to me! My understanding is that I could write the custom signature with additional instructions that I want and use the context manager to replace the base signature.

PR #943, opened 2024-05-01
fix(dspy): added `copy` to `OllamaLocal` to propagate `model`, `model_type`, `base_url` and `timeout_s`.

xyloid:
This problem is very similar to #769. `OllamaLocal.copy` uses `LM.copy`. However, `OllamaLocal`'s `kwargs` dict doesn't have `model` key, which means `model = kwargs.pop("model")` will create a `KeyError`. This fix added a `copy` method to `OllamaLocal` class so that it's own attributes like `model`, `model_type`, `base_url` and `timeout_s` can be copied.

### DIFF
--- a/dsp/modules/ollama.py
+++ b/dsp/modules/ollama.py
@@ -180,3 +180,15 @@ class OllamaLocal(LM):
         completions = [self._get_choice_text(c) for c in choices]
 
         return completions
+    
+    def copy(self, **kwargs):
+        """Returns a copy of the language model with the same parameters."""
+        kwargs = {**self.kwargs, **kwargs}
+
+        return self.__class__(
+            model=self.model_name,
+            model_type=self.model_type,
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Hi @xyloid , thanks for the contribution! Could you please rebase and merge to the latest version on main? Seems like there was a failing test on our end (nothing wrong with this PR) that just needs an update. Ready to merge after that!

xyloid:
> Hi @xyloid , thanks for the contribution! Could you please rebase and merge to the latest version on main? Seems like there was a failing test on our end (nothing wrong with this PR) that just needs an update. Ready to merge after that!

Hi @arnavsinghvi11 , I have rebased my branch on the latest main branch.

arnavsinghvi11:
Thanks @xyloid !

PR #945, opened 2024-05-01
Enhance README "Some other examples" section

rpgoldman:
Two modifications to make it easier to get to resources:

1. Modified so that "benchmark on a bunch of different tasks" sends the reader to a README.md file explaining the benchmarking process, instead of just dropping them into the directory on GitHub.
2. Added a link to ArXiV for "The Unreasonable Effectiveness..." article, so that now the interested reader can jump directly to ArXiV instead of having to go through the Register article. The link to the Register article *is* retained, though.


arnavsinghvi11:
Thanks @rpgoldman !

rpgoldman:
@arnavsinghvi11 Thanks for the review and corrections (and merging)!

PR #948, opened 2024-05-01
feat(dspy): add dspy.Claude helper

qhoxie:
dspy.Claude was a missing alias.  I checked through the others for consistency and this appears to be the only one left out.

Import move was a ruff fix.

### DIFF
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -5,10 +5,10 @@ from .predict import *
 from .primitives import *
 from .retrieve import *
 from .signatures import *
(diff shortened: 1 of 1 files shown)


dspy/__init__.py
```diff
@@ -5,10 +5,10 @@
 from .primitives import *
 from .retrieve import *
 from .signatures import *
+from .utils.logging import logger, set_log_output
```
qhoxie, line 8: ruff fix

arnavsinghvi11:
Thanks @qhoxie !

PR #950, opened 2024-05-01
fix(dspy): remove unreachable type condition in TypedPredictor

qhoxie:
`bool` was elevated into the top condition [here](https://github.com/stanfordnlp/dspy/commit/982db444322cbf6a545a2504ce162b15dafee975), so it can't be reached in the elif.


arnavsinghvi11:
Thanks @qhoxie !

Issue #962, opened 2024-05-03
Type hints for signature.py are inaccurate


rpgoldman:
`_parse_signature()` in [signature.py](https://github.com/stanfordnlp/dspy/blob/main/dspy/signatures/signature.py) is declared as returning a `Tuple` but it returns a dictionary.  I checked and mypy confirms this:

    Incompatible return value type (got "dict[Any, tuple[Any, Any]]", expected "tuple[type[Any], Field?]")  [return-value]

I'll make a Pull Request with edited type hints.

PR #964, opened 2024-05-03
Add type hints to signature.py

rpgoldman:
Address #962 

This is substantially complicated by the use of Pydantic.

/cc @thomasahle


rpgoldman:
I could definitely use help here. I got started just fixing issue #962 , but tried to add more type hints, and have been having issues because mypy does not seem to understand that `Signature()` is a class-constructor, not an instance-constructor, and also because it is confused by its signature (inherited from `BasicModel`, I believe).

PR #971, opened 2024-05-05
Loosen requirement on `datasets` package

krishnan-chandra:
Right now, the requirement on the `datasets` package is `"datasets~=2.14.6,<3.0.0"`. Based on [Poetry's specifications for tilde requirements](https://python-poetry.org/docs/dependency-specification/#tilde-requirements), this means that only 2.14 patch releases can be used for the datasets package (and the `<3.0.0` part of the requirement isn't doing anything).

But this is needlessly restrictive, as the datasets package is now up to version [2.19.0](https://github.com/huggingface/datasets/tags). Additionally, restricting datasets to such a low version causes issues with resolution alongside other packages that require `datasets` as well.

The primary usage of datasets is to call `load_dataset`, which shouldn't be affected by this version constraint.

**Solution**: Change the `~=2.14.6` to `>=2.14.6` instead, which makes version solving significantly easier.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -266,6 +266,23 @@ dev = ["cogapp", "pre-commit", "pytest", "wheel"]
 tests = ["pytest"]
 
 [[package]]
+name = "asgiref"
+version = "3.8.1"
+description = "ASGI specs, helper code, and adapters"
+optional = true
+python-versions = ">=3.8"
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
Thanks @krishnan-chandra !

PR #978, opened 2024-05-06
add missing kwargs when invoking retrieveRerankEnsemble

envytt:
When invoking `retrieveRerankEnsemble` in  https://github.com/stanfordnlp/dspy/blob/d7dace6b51e1c0899fd798c575262310939f6deb/dsp/primitives/search.py#L52, it's missing `kwargs`


arnavsinghvi11:
Thanks @envytt !

PR #985, opened 2024-05-07
Deep dive teleprompter rename

tom-doerr:
I keep getting confused when I search for information on optimizers and can't find it since some sections are still named teleprompters. I just went ahead and refactored the deep dive section. I hope you don't mind!


arnavsinghvi11:
Hi @tom-doerr , thanks for this change! This has been long overdue but we will also need to do this refactoring within the repository to align with Optimizers overall. @okhat is this a good time to make this transition?

tom-doerr:
If it's basically just renaming teleprompt to optimize and adding deprecation warnings if someone continues to use teleprompt I could try to add it to the PR

tom-doerr:
Adding deprecation warnings while keeping it functional seems straightforward. 
`dspy/teleprompt/teleprompt.py` would become:
```
import warnings
warnings.warn(
    "The 'teleprompt' module has been renamed to 'optimize'. Please update your imports.",
    DeprecationWarning
)
from dspy.optimize.optimize import *
```

Would be happy to help out with refactoring.

tom-doerr:
Ah okay, `LabeledFewShot` is probably also not an optimizer

PR #990, opened 2024-05-07
fix(evaluate): `Evaluate` should not error when `display_table=True`

christeefy:
Obtained an error when `display_table=True` is provided for `Evaluate`. 
See simple reproducible example below:

```py
from dspy.evaluate.evaluate import Evaluate
from dspy.evaluate.metrics import answer_exact_match
from dspy.predict import Predict

def new_example(question, answer):
    """Helper function to create a new example."""
    return dspy.Example(
        question=question,
        answer=answer,
    ).with_inputs("question")

devset = [new_example("What is 1+1?", "2")]

program = Predict("question -> answer")

ev = Evaluate(
    devset=devset,
    metric=answer_exact_match,
    display_table=True,
)
ev(program)  # TypeError: cannot do positional indexing on RangeIndex with these indexers [True] of type bool
```

This stems from `isinstance(True, int) == True`. 

This PR contains a proposed fix with updated unit tests. Feel free to re-structure the unit tests as you see fit.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -218,12 +218,12 @@ class Evaluate:
         result_df = result_df.rename(columns={"correct": metric_name})
 
         if display_table:
-            if isinstance(display_table, int):
-                df_to_display = result_df.head(display_table).copy()
-                truncated_rows = len(result_df) - display_table
-            else:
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
Thanks @christeefy ! great find!

PR #992, opened 2024-05-08
fix(react): `ReAct` doesn't use `instructions` from the given `Signature`

xyloid:
When I use `ReAct` with `Signature`, I find that the `instructions` field in the given signature is ignored by the `ReAct` class, this behavior is not consistent with `Predict`. This fix provides a workaround for the issue.


arnavsinghvi11:
Thanks for the PR @xyloid ! this makes sense.

PR #995, opened 2024-05-08
fix: assertion and chain-of-thought

Shangyint:
Previously, `ChainOfThought` does not work well with assertions due to an error in handling the `rationale` field of the CoT module. This pr fixes this issue. A future improvement would be implementing unit tests for CoT and Retry in `test_retry.py`.

### DIFF
--- a/dspy/predict/retry.py
+++ b/dspy/predict/retry.py
@@ -10,16 +10,17 @@ class Retry(Predict):
     def __init__(self, module):
         super().__init__(module.signature)
         self.module = module
-        self.original_signature = module.signature
+        self.original_signature = module.extended_signature if isinstance(module, dspy.ChainOfThought) else module.signature
         self.original_forward = module.forward
         self.new_signature = self._create_new_signature(self.original_signature)
 
     def _create_new_signature(self, signature):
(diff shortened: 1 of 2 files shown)


okhat:
Thank you @Shangyint !! Looks like we had tests that are tied to the old behavior. Do we trmove/efit them?

Shangyint:
Yes, I figured out the same thing. Pushed a fix!

chiragshah285:
hey all @Shangyint the assertion fix broke a few things. I'm unable to use dspy.Suggest(), here is a minimal working example of an error: `AttributeError: 'Predict' object has no attribute 'new_signature'. Did you mean: 'signature'?`


https://gist.github.com/chiragshah285/cc1756cde35b8cf14aa04282f3f744e1

line 260 in assertions.py, i've patched it with this but not sure if this is the right fix:
```
                            if hasattr(error_state[0], 'new_signature'):
                                output_fields = error_state[0].new_signature.output_fields
                            else:
                                output_fields = error_state[0].signature.output_fields
```

Shangyint:
Hey @chiragshah285, thank you! I submitted your patch as #1372.

PR #1017, opened 2024-05-12
Update FAQ to mention `DSP_CACHEDIR` for local cache.

tom-doerr:


okhat:
This is good but it will cause people to set both paths to the sane string….

arnavsinghvi11:
Hi @tom-doerr , was this resolved and ready to merge?

tom-doerr:
Yes I think it should be clearer now

arnavsinghvi11:
Thanks @tom-doerr !

PR #1026, opened 2024-05-14
support multiple corpora in vectara retriever

ofermend:
Minor update to support multiple Vectara corpora with the retriever.


arnavsinghvi11:
Thanks @ofermend !

PR #342, opened 2024-02-05
Fix bugs in pinecone integration, add support for using cohere embed with pinecone and create basic example notebook

DanielUH2019:
@CShorten , @okhat this should resolve #322. I  removed the _init_pinecone function because I believe it shouldn't be the responsibility of the PineconeRM class to create an index. Let me know what you think.

PR #791, opened 2024-04-09
feature(dspy): Add MyScale in Retrieve

usamajamil43:
Add MyScaleRM to support retrieval for MyScaleDB:

- MyScaleDB: https://github.com/myscale/myscaledb

PR #1027, opened 2024-05-15
fix(pinecone_rm): refactored to use cloud_embed and fix pinecone init

csellis:
source is from https://github.com/stanfordnlp/dspy/pull/342/files
- update had gotten stale. updates written by @DanielUH2019

### DIFF
--- a/dspy/retrieve/pinecone_rm.py
+++ b/dspy/retrieve/pinecone_rm.py
@@ -3,21 +3,23 @@ Retriever model for Pinecone
 Author: Dhar Rawal (@drawal1)
 """
 
-import os
-from dsp.utils import dotdict
-from typing import Optional, List, Union, Any
-import dspy
-import backoff
 from abc import ABC, abstractmethod
+from typing import List, Optional, Union
+
+import backoff
+
+import dspy
+from dsp.utils import dotdict
 
 try:
     import pinecone
 except ImportError:
     raise ImportError(
-        "The pinecone library is required to use PineconeRM. Install it with `pip install dspy-ai[pinecone]`"
+        "The pinecone library is required to use PineconeRM. Install it with `pip install dspy-ai[pinecone]`",
     )
 
 import openai
+
 try:
     OPENAI_LEGACY = int(openai.version.__version__[0]) == 0
 except Exception:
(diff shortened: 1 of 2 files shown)


dspy/retrieve/pinecone_rm.py
```diff
@@ -33,6 +29,64 @@
...
+        if self.org:
+            openai.organization = org
+
+    
+    @backoff.on_exception(
+        backoff.expo,
+        ERRORS,
+        max_time=15,
+    )
+    def get_embeddings(self, queries: List[str]) -> List[List[float]]:
```
arnavsinghvi11, line 57: could the caching functionality included [in this PR](https://github.com/stanfordnlp/dspy/pull/791/files#diff-c6ca43067290c41691d236de3c2a7648012ad0174adcb63f040e481a15f5b28bR171) be used here?

I realize we probably need an abstracted Retriever class to handle this but for now, feel free to check if this suffices

examples/integrations/pinecone/pinecone_wikipedia_example.ipynb
```diff
@@ -0,0 +1,625 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "[upserted_count: 100,\n",
```
arnavsinghvi11, line 153: maybe better to remove the output for this cell :)

examples/integrations/pinecone/pinecone_wikipedia_example.ipynb
```diff
@@ -0,0 +1,625 @@
...
+     "output_type": "stream",
+     "text": [
+      "100%|██████████| 20/20 [03:30<00:00, 10.54s/it]"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Bootstrapped 0 full traces after 20 examples in round 0.\n"
```
arnavsinghvi11, line 559: it seems like the bootstrapping in this case couldn't find any valid examples. I'm curious if it is because of the retriever and what passages are being retrieved - worth investigating if it is due to the changes in this PR. We typically expect this example to work with most LMs/RMs but it could just be these 20 were not valid outputs. Just worth checking IMO before merging this example if that is the case

arnavsinghvi11:
Hi @csellis , thanks for the PR! left a few comments to be addressed before merging. Also could you run `ruff check . --fix-only` and commit again to pass the failing test?

arnavsinghvi11:
Hi @csellis , I updated the ruff fix but accidentally merged this. Could you reopen this with the updated changes as per the comments? Thanks.

Issue #1025, opened 2024-05-14
dspy.HFClientVLLM broken since 2.4.9


Wolfsauge:
Using the dspy.HFClientVLLM method is broken since dsp version 2.4.9.

I have reproduced the error with the below script and the versions 2.4.5, 2.4.6 and 2.4.7 of dspy work as expected, while 2.4.8 doesn't seem to exist and 2.4.9 is broken.

I've been doing these tests with `Python 3.11.9`.

The erroneous run looks like this:
```shell
$ /home/ns/.pyenv/versions/3.11.9/bin/python /home/ns/GitHub/Playground/dspy_examples/repro.py
/home/ns/.pyenv/versions/3.11.9/lib/python3.11/site-packages/datasets/table.py:1421: FutureWarning: promote has been superseded by promote_options='default'.
  table = cls._concat_blocks(blocks, axis=0)

### Generate Response ###

Failed to parse JSON response: {"object":"error","message":"[{'type': 'extra_forbidden', 'loc': ('body', 'port'), 'msg': 'Extra inputs are not permitted', 'input': 8000}, {'type': 'extra_forbidden', 'loc': ('body', 'url'), 'msg': 'Extra inputs are not permitted', 'input': ['http://mnemosyne.local:8000']}]","type":"BadRequestError","param":null,"code":400}
Traceback (most recent call last):
  File "/home/ns/.pyenv/versions/3.11.9/lib/python3.11/site-packages/dsp/modules/hf_client.py", line 199, in _generate
    completions = json_response["choices"]
                  ~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'choices'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ns/GitHub/Playground/dspy_examples/repro.py", line 40, in <module>
    pred = generate_answer(question=example.question)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ns/.pyenv/versions/3.11.9/lib/python3.11/site-packages/dspy/predict/predict.py", line 61, in __call__
    return self.forward(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ns/.pyenv/versions/3.11.9/lib/python3.11/site-packages/dspy/predict/predict.py", line 103, in forward
    x, C = dsp.generate(template, **config)(x, stage=self.stage)
[...]

PR #1012, opened 2024-05-11
bug(dspy): removed extra params from kwargs sent to vllm pydantic model

omkar-sh:
Can you just checkout and suggest any changes?

PR #1029, opened 2024-05-16
bug(dspy): removed system_prompt from kwargs sent to vllm pydantic model 

LorenzoMinto:


Wolfsauge:
Hello @LorenzoMinto Thank you for your perseverance in trying to iron out this problem. I have run into the same issue and fixed it differently, see here:
https://github.com/stanfordnlp/dspy/issues/1025#issuecomment-2111249607

I then noticed the PR #1012 being in the works, despite having a different approach, thinking it would be sufficient.

Also, I am not thinking my "fix" is any better, but I decided to start commenting on this, because I feel there is something wrong that I have to create such a "fix" in the first place.

There are sampling parameters of vLLM stored in the same set of key words as some keywords used for only DSPy internal processing. This is asking for trouble down the road.

While I don't want to discourage anything about adding "system_prompt" to the elements which should be popped out here, I want to ask for discussion if that is not asking for further trouble down the road and if not a deeper fix is needed to separate DSPy internal parameters from sampling parameters, which will make vLLM choke. It is very strict with regard to the correctness of the requests it will process.

What do you think?

tom-doerr:
I merged this to my local dspy version and have the same issue, seems like another parameter vllm doesn't take

tom-doerr:
@Wolfsauge I went ahead and created a pull request using your code, hope you don't mind.

PR #1034, opened 2024-05-18
Update cheatsheet references

tom-doerr:
Does it make sense to remove the `dspy.` if a class is not actually accessible over it? In case that's wanted it might make sense to actually allow access with `dspy.`


arnavsinghvi11:
Thanks @tom-doerr !

Issue #879, opened 2024-04-20
AI feedback using logprobs


tom-doerr:
Whenever I use a LM to evaluate something I ask a Yes/No question and do `logprob_yes - logprob_no`. This worked very well for me in the past. How can I access logprobs in my metric function?

arnavsinghvi11:
Hi @tom-doerr , logprobs are indeed quite useful. Currently, DSPy doesn't support returning additional parameters besides the [response content](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/modules/gpt3.py#L188), but you can extract the logprobs from [the response](https://github.com/stanfordnlp/dspy/blob/d09d984ecaf17f7262294d50fe46fd8105fbf291/dsp/modules/gpt3.py#L178) (subject to different LM configurations). 

Feel free to push a PR that enables some flag for returning completions with logprobs while maintaining internal behavior!

arthurgreef:
This no longer works with the LiteLLM client on 2.5.13.

```python
output = gpt("Hello, world!", logprobs=True, top_logprobs=1)
print (output)
```

Expected: Logprobs after the text object.
Actual: Just the text: `['Hello! How can I assist you today?']`

tranhoangnguyen03:
```
from google.colab import userdata

lm = dspy.OpenAI(
    model='gpt-4o', 
    max_tokens=512, 
    api_key=userdata.get('openai_api_key')
)

dspy.settings.configure(lm=lm)

system_prompt = "answer in a single word.\n"
test_text = system_prompt + "The answer to the ultimate question to the meaning of life, the universe, and everything is ..."
test_output = lm(test_text, logprobs=True, top_logprobs=20)
print("test_output:", test_output)
```
The script above is working in terms of outputing the logprobs, but it triggers the deprecation warning.

```
from google.colab import userdata

lm = dspy.LM(
    'openai/gpt-4o', 
    max_tokens=512, 
    api_key=userdata.get('openai_api_key')
)
dspy.configure(lm=lm)

system_prompt = "answer in a single word.\n"
test_text = system_prompt + "The answer to the ultimate question to the meaning of life, the universe, and everything is ..."
test_output = lm(test_text, logprobs=True, top_logprobs=20)
print("test_output:", test_output)
```
This script doesn't produce the logprobs.

Are there plans to fix this? Is there a way for me to apply a patch locally on my machine to hotfix this?

[...]

PR #999, opened 2024-05-09
Add logprob support for OpenAI models

tom-doerr:
This is part of https://github.com/stanfordnlp/dspy/issues/879. 
Is this way of adding support for logprobs acceptable?
```
lm = dspy.OpenAI(model='gpt-3.5-turbo-instruct', max_tokens=6, api_key=config['openai']['secret_key'])
test_text = "This is a test article."
test_output = lm(test_text, logprobs=1)
print("test_output:", test_output)
```
```
test_output: [{'text': '\n\nLorem ipsum dolor sit amet', 'logprobs': {'text_offset': [23, 25, 30, 36, 42, 46], 'token_logprobs': [-0.26951057, -0.28366262, -0.0010195904, -5.4312077e-05, -0.0001788689, -0.00023739056], 'tokens': ['\n\n', 'Lorem', ' ipsum', ' dolor', ' sit', ' amet'], 'top_logprobs': [{'\n\n': -0.26951057, '\n': -2.4457762}, {'Lorem': -0.28366262, 'Test': -2.9440362}, {' ipsum': -0.0010195904, ' Ipsum': -6.9318686}, {' dolor': -5.4312077e-05, ' d': -11.2004175}, {' sit': -0.0001788689, ' amet': -9.340998}, {' amet': -0.00023739056, ' am': -8.778377}]}}]
```
If this is okay I would look into adding it to `dspy.Predict` and the vllm backend.

PR #1035, opened 2024-05-18
Add logprobs hfclientvllm

tom-doerr:
This is part of https://github.com/stanfordnlp/dspy/issues/879.
I tested it and it behaves the same as the other PR.
```
lm = dspy.HFClientVLLM(model="NurtureAI/Meta-Llama-3-8B-Instruct-32k", port=38242, url="http://localhost", max_tokens=4)
test_text = "This is a test article."
output_normal = lm(test_text)
print("output_normal:", output_normal)

output_with_logprobs = lm(test_text, logprobs=2)
print("output_with_logprobs:", output_with_logprobs)
```
```
output_normal: [' It is a test']
output_with_logprobs: [{'text': ' It is a test', 'logprobs': {'text_offset': [0, 3, 6, 8], 'token_logprobs': [-1.7945036888122559, -0.6691504716873169, -1.303508996963501, -0.7093929052352905], 'tokens': [' It', ' is', ' a', ' test'], 'top_logprobs': [{' It': -1.7945036888122559, ' This': -1.7945036888122559}, {' is': -0.6691504716873169, ' will': -2.0441503524780273}, {' a': -1.303508996963501, ' not': -1.803508996963501}, {' test': -0.7093929052352905, ' sample': -3.20939302444458}]}}]
```

### DIFF
--- a/dsp/modules/hf.py
+++ b/dsp/modules/hf.py
@@ -188,7 +188,7 @@ class HFModel(LM):
             kwargs["do_sample"] = True
 
         response = self.request(prompt, **kwargs)
-        return [c["text"] for c in response["choices"]]
+        return [{"text": c["text"], "logprobs": c["logprobs"]} if 'logprobs' in c else c["text"] for c in response["choices"]]
 
 
 # @functools.lru_cache(maxsize=None if cache_turn_on else 0)
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -206,7 +206,7 @@ class HFClientVLLM(HFModel):
                 completions = json_response["choices"]
                 response = {
                     "prompt": prompt,
-                    "choices": [{"text": c["text"]} for c in completions],
+                    "choices": [{"text": c["text"], "logprobs": c["logprobs"]} if c["logprobs"] else {"text": c["text"]} for c in completions],
                 }
                 return response
 


okhat:
Hmm this seems fine to me but it's ad-hoc.... it would be specific to this one client

tom-doerr:
It also works already for the OpenAI API: https://github.com/stanfordnlp/dspy/pull/999.
This feature wouldn't need to be documented until more clients are supported.
In my initial testing the logprob feedback was much better than binary feedback. In my case non of the dataset examples passed and BootstrapFewShot gave up. With logprob feedback it was able to consistently improve over multiple iterations.

okhat:
does it block you if we keep this open for a week or two? i need to think more about logprobs

tom-doerr:
No doesn't block me at all, I'm using my custom branch anyway.
If you are thinking about logprobs anyway, my personal wish list would be:
- Allows for access to prompt logprobs and not just completion logprobs
- Allows access to unnormalized logprobs. This can be especially useful when using logprobs to sort something

I know that not all backends can give us all that, users would have to choose a backend that supports their usecase.

arnavsinghvi11:
@tom-doerr just wanted to add some logprobs-related issues here. 

With the past PR merged for OpenAI logprobs support, I realize that this only outputs the logprobs for "direct" LM calls, and is not compatible with configuring the OpenAI LM within a DSPy program. 

For instance, the example provided outputs logprobs as intended:

```
lm = dspy.OpenAI(model='gpt-3.5-turbo-instruct', max_tokens=6, api_key=config['openai']['secret_key'])
test_text = "This is a test article."
test_output = lm(test_text, logprobs=1)
```

but if we had some DSPy program and did `dspy.settings.configure(lm=lm, ...)` to have it run with that LM, the [DSPy Completions logic ](https://github.com/stanfordnlp/dspy/blob/7a78cb5de9b922d2c7a700150528a1629be7afb1/dsp/primitives/predict.py#L113C3-L114C1)does not output the logprobs in the response. This signals that we potentially we need a greater refactor to integrate logprobs correctly. I might open a PR soon with some baseline code that handles this per client, but definitely good to think in the direction of non-client-specific functionality. @okhat

tom-doerr:
@arnavsinghvi11 Yes I know what you mean, I planned to add support to DSPy components after support for the clients is merged. Since it's a bigger refactor I would have suggested adding support for it bit by bit.
Is this something I could help out with? If so it might make sense to have a call

arnavsinghvi11:
Sounds good! Yes would love to connect @tom-doerr - truly appreciate the enthusiasm you've shown for DSPy :))

Issue #1018, opened 2024-05-13
How to convert a AutoModelForCausalLM object to a dspy model object?


pawanGithub10:
import dspy

llm = dspy.HFModel(model='model')

This method takes a string as input for the model if i have a quantized model object of the class AutoModelForCausalLM How i can convert the model object to dspy object?

direct assignment gives error on inference

llm = model #previously created as AutoModelForCausalLM class object

llm("Testing testing, is anyone out there?")

Error After Code Line 4 File /opt/conda/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py:623, in LlamaModel.forward(self, input_ids, attention_mask, position_ids, past_key_values, inputs_embeds, use_cache, output_attentions, output_hidden_states, return_dict) 621 raise ValueError("You cannot specify both decoder_input_ids and decoder_inputs_embeds at the same time") 622 elif input_ids is not None: --> 623 batch_size, seq_length = input_ids.shape 624 elif inputs_embeds is not None: 625 batch_size, seq_length, _ = inputs_embeds.shape

AttributeError: 'str' object has no attribute 'shape'

Anindyadeep:
I see, but internally hf module internally uses AutoModel module to instantiate weights. So can you explain me why we require to use already loaded model to dspy instead of giving the weight path?

pawanGithub10:
> I see, but internally hf module internally uses AutoModel module to instantiate weights. So can you explain me why we require to use already loaded model to dspy instead of giving the weight path?

Thanks for reply, The reason is I have a 4 bit quantized model and i want to use it directly.  I have tried to save it first to hugging face so that i can load it from weight path but then error comes that the Hugging face does not support saving 4 bit quantized model.

Anindyadeep:
can you please share the full code for the loading process and your approach? Would appreciate this.

pawanGithub10:
[dspy_4bitquantized_llama2_error.zip](https://github.com/stanfordnlp/dspy/files/15305279/dspy_4bitquantized_llama2_error.zip)
I have attached the file which can run on kaggle or may be on google colab. When I am converting the quantized model then it searches for the config.json as I am giving the AutoModel file please suggest some workaround or API call to use the quantized model.

Anindyadeep:
Hey @pawanGithub10 I have started to raise a PR by seeing the issue that you faced. Here are some of the cases of loading models would look like:

```python

from dsp.modules.hf_new import HFLocalModel
[...]

PR #1044, opened 2024-05-20
Migrating existing HuggingFace support to a more robust one

Anindyadeep:
Current status: WIP 

The PR motivation started from this issue: #1018 

### What this PR do?

We have an existing HuggingFace model integration on DsPY. However developers can not use this class with much flexibility. So the goal of the PR is to provide a balanced abstraction such that developers can

1. add their own model object
2. or add just different configs and arguments while instantiating, example: quantization_config, lora_config directly in the dspy.HF class
3. make it also directly compatible with the finetuning procedures (might not be applicable now)

The reason is, HuggingFace is very popular for sometimes writing out entire lifecycle. So writing entre pipelines with huggingface and dspy might need to have almost all the supported configs and HF's own addons (like bitsandbytes, peft etc), so that they can focus more on dspy instead of finding wayaround on how to add this or that with dspy huggingface.


Anindyadeep:
Hi would love to get your thoughts on this, so that according to the feedbacks I can move forward. Thanks 

cc: @arnavsinghvi11

Anindyadeep:
> Current status: WIP
> 
> The PR motivation started from this issue: #1018
> 
> ### What this PR do?
> We have an existing HuggingFace model integration on DsPY. However developers can not use this class with much flexibility. So the goal of the PR is to provide a balanced abstraction such that developers can
> 
> 1. add their own model object
> 2. or add just different configs and arguments while instantiating, example: quantization_config, lora_config directly in the dspy.HF class
> 3. make it also directly compatible with the finetuning procedures (might not be applicable now)
> 
> The reason is, HuggingFace is very popular for sometimes writing out entire lifecycle. So writing entre pipelines with huggingface and dspy might need to have almost all the supported configs and HF's own addons (like bitsandbytes, peft etc), so that they can focus more on dspy instead of finding wayaround on how to add this or that with dspy huggingface.
> 
> ### Todos
> * [x]  Initial working draft implementation
> * [x]  Testing the implementation passing HF models object
> * [x]  Testing the implementation passing HF models object as string
> * [x]  Testing the implementation passing HF models with quantization support
> * [ ]  Testing the implementation passing HF models with multi-gpu support (currently not have multi gpu settings)
> * [ ]  Refine the code (check once more, typos, check on some todos I wrote inside the code etc)
> * [ ]  Write good tutorial like documentation for this
> * [ ]  Fininshing (link etc)
> 
> cc: @okhat, let's discuss more about this approach and feel free to provide feedbacks :)

Hi just wanted to check on whether we can move forward with this PR or not, else we can close it. Thanks :)

dsp/modules/hf_new.py
```diff
@@ -0,0 +1,193 @@
...
+        bnb_config: Optional["transformers.BitsAndBytesConfig"]=None, 
+        **kwargs,
+    ):
+        try:
+            import torch
+            import transformers
+        except ImportError as exc:
+            raise ModuleNotFoundError(
+                "You need to install the following libraries: ",
+                "transformers >= 4.30.0, torch==2.2.2, accelerator"
```
arnavsinghvi11, line 44: would eventually be added to pyproject.toml and/or poetry?

arnavsinghvi11:
Hi @Anindyadeep , thanks for this PR refactor of HF! This looks great!

I anticipate this PR will eventually replace [`hf.py`](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/hf.py) so it would be great to either build on the existing file if possible. It would also be easier to understand what’s being retained and any additional differences. Following the leftover todos and testing, let me know if you have any comments/additions to review.

Looking forward to the next stage of this PR!

Anindyadeep:
> Hi @Anindyadeep , thanks for this PR refactor of HF! This looks great!
> 
> I anticipate this PR will eventually replace [`hf.py`](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/hf.py) so it would be great to either build on the existing file if possible. It would also be easier to understand what’s being retained and any additional differences. Following the leftover todos and testing, let me know if you have any comments/additions to review.
> 
> Looking forward to the next stage of this PR!

Hey thanks for the reply, on a second though I feel this PR is a very big addition so in that case, I could re-start from the PR implementation, where I can break into smaller parts (which could be easier for you all as well for review) and doing changes from `hf.py`. Does that sounds good to you?

Anindyadeep:
> > Hi @Anindyadeep , thanks for this PR refactor of HF! This looks great!
> > I anticipate this PR will eventually replace [`hf.py`](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/hf.py) so it would be great to either build on the existing file if possible. It would also be easier to understand what’s being retained and any additional differences. Following the leftover todos and testing, let me know if you have any comments/additions to review.
> > Looking forward to the next stage of this PR!
> 
> Hey thanks for the reply, on a second though I feel this PR is a very big addition so in that case, I could re-start from the PR implementation, where I can break into smaller parts (which could be easier for you all as well for review) and doing changes from `hf.py`. Does that sounds good to you?

Let me know your thoughts @arnavsinghvi11 whenever you find some bandwidth, thanks :)

arnavsinghvi11:
>  I could re-start from the PR implementation, where I can break into smaller parts (which could be easier for you all as well for review) and doing changes from hf.py. Does that sounds good to you?

Yes that would be great!

okhat:
Excellent effort here! Tagging @dilarasoylu and @isaacbmiller

okhat:
I wonder how much of this can we delegate to an opinionated tool like axotol or torchtune or something btw?

Anindyadeep:
> I wonder how much of this can we delegate to an opinionated tool like axotol or torchtune or something btw?

Well, it depends, I mean sure, we can go for axotol or torchtune, but then it would be less coupled with native hf scripts imo. However let me know what do you think on this and we can go accordingly.

okhat:
My only goal is to support most useful models & quantization settings & loss functions (DPO etc) with as little code as possible. Do you think HF is still the best path for that?

okhat:
If HF is still best, let's stick to it! Thanks for the amazing work here @Anindyadeep !

Anindyadeep:
> My only goal is to support most useful models & quantization settings & loss functions (DPO etc) with as little code as possible. Do you think HF is still the best path for that?

Ahh yes, in that case, we can go for HF, okay then let me complete this PR and I will ping you guys. (Putting it into draft now)

PR #1046, opened 2024-05-20
fix(dspy): fix save/load for predict method

GFarnon:
Allow the dump_state/load_state methods in dspy.Predict to include prefix and desc for all signature fields

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -38,8 +38,13 @@ class Predict(Parameter):
         # Cache the signature instructions and the last field's name.
         state["signature_instructions"] = self.signature.instructions
 
-        *_, last_key = self.signature.fields.keys()
-        state["signature_prefix"] = self.signature.fields[last_key].json_schema_extra["prefix"]
+        field_details = {}
+        for key, field in self.signature.fields.items():
+            field_details[key] = {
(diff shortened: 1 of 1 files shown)


okhat:
Thank you so much @GFarnon ! If I understand correctly, this will not be backwards compatible with existing saved files?

okhat:
Also looks like the ruff check is failing but that's more minor.

arnavsinghvi11:
Hi @GFarnon , thanks for the PR! the ruff fix will be fixed once you merge with main. 

@okhat I've tested the change with the intro.ipynb and it doesn't have any breaking changes - although it does not rely on save/load functionalities. Would the [CoT.ipynb ](https://github.com/stanfordnlp/dspy/blob/main/examples/math/gsm8k/CoT.ipynb) be better to test this (although that does not have cache currently so it wouldn't break any repo-wide changes). 

@GFarnon would this change impact [ChainOfThought](https://github.com/stanfordnlp/dspy/blob/41d0f0edf4ba934f36e59a9f3b7e8c1721abeaa8/dspy/predict/chain_of_thought.py#L62) as well, given that there are `extended_...` fields added to the state? For context, ChainOfThought inherits from Predict but reconstructs the signature within its internals.

Issue #1024, opened 2024-05-14
DSPy Typed Predict is not extracting the LLM output correctly.


felixgao:
I am trying to use the TypePredictor to extract information from some inputs.  The program failed due to exceeding number of retries but when inspecting the output of the LLM it seems to be in a parsable way. 

The LLM used is gemini-1.5-pro

Code

```python
import dspy
from pydantic import BaseModel, Field

class Field(BaseModel):
    key: str = Field(description="The key to be extracted from the context")
    value: str = Field(description="the value to be extracted from the context")
    truthful: bool = Field(
        description="is a tuple contains the extracted value and a boolean indicating if the information is truthful and correct."
    )
    confidence: float = Field(ge=0, le=1, description="The confidence score for the answer")


class Fields(BaseModel):
    fields: list[Field]

class QAExtractionSignature(dspy.Signature):
    """Your task is to extract the key-value pairs from the document that will follow the instructions on the field to extract.\nplease keep the content you extract truthful and correct based on the document text provided"""

    document: str = dspy.InputField()
    instruction: str = dspy.InputField(description="The instruction on which field to extact and how to extract it.")
    fields: Fields = dspy.OutputField(
        description="The list of fields extracted from the document text. IMPORTANT!! This must follow a semicolon separated list of values!"
    )

[...]

Issue #1001, opened 2024-05-09
[Feature Request] Better Typed Predictors for Reliable JSON Generation


StevenGuo42:
The [Typed Predictors](https://github.com/stanfordnlp/dspy/blob/7227e7081d8edc3d0ffc2729f7c97871aa61338b/dspy/functional/functional.py#L72) output JSON as plain text. They do not use function calling method/template and do not enforce the output to be a valid JSON. 
This makes generating structured output challenging (see my previous issue #957 ).

Some possible solutions: 
- Use built-in function calling method like OpenAI's [function calling](https://platform.openai.com/docs/guides/function-calling/function-calling). (There is currently a hacky solution for it #192 )
- Use model specific template for models fine-tuned for function calling, [Meta-Llama-3-8B-Instruct-function-calling](https://huggingface.co/Trelis/Meta-Llama-3-8B-Instruct-function-calling), [Qwen1.5-function-calling-chat-template](https://huggingface.co/Trelis/Qwen1.5-function-calling-chat-template), [glaive-function-calling-v1](https://huggingface.co/glaiveai/glaive-function-calling-v1). This may be achieved by allowing user provided template #996 , #228 .
- Integrate other Python libraries that help with structured output, such as [Jsonformer](https://github.com/1rgs/jsonformer/tree/main), [Instructor](https://github.com/jxnl/instructor), [lm-format-enforcer](https://github.com/noamgat/lm-format-enforcer)

StevenGuo42:
@AriMKatz Thanks for the suggestion. I added both to the issue.

juicesharp:
A half of responses failed validation actually contain a valid json object in it, but has a prefix like: "Here is my response: ```json VALID_JSON or "Written doc fotmmated as Json object {VALID JSON}. A simple ability to inject/customize a post-process of Prediction inside of Pipeline, for example use your own clean-up function, would have drastically mitigated the problem. 

One possible approach would be to allow subclasses of dspy.Module to override/exend part of post process pipeline. if Module has a function like def _post_prediction_process(self, pred : Prediction) -> Prediction: Even beyond the discussed issue a such hook has lot of sense.

juicesharp:
Another idea after the object was not deserialized correctly DSPY before doing a "rephrase" attempt could simple check the failed response with a help of regex something like: 
```python
def _extract_content_within_backticks(self, code):
        pattern = r"```[\w\s]*\n([\s\S]*?)\n```"
        match = search(pattern, code)
[...]

PR #1053, opened 2024-05-22
feat(dspy): try from json before applying formatting logic in typed predictor

mikeedjones:
Adds an attempt to use the function supplied to `from_json` before running string parsing. This allows users to modify the pydantic's `model_validate_json` to customise the deserialization of the generation.

Other changes were to satisfy ruff.

This PR gives users a route to inject code which can solve issues like https://github.com/stanfordnlp/dspy/issues/1024 and https://github.com/stanfordnlp/dspy/issues/1001

After this PR, the user could implement something like the following:

```python
class TaxField(BaseModel):
    key: str = Field()
    value: str = Field()
    truthful: bool = Field(
        description="is a tuple contains the extracted value and a boolean indicating if the inFieldsation is truthful and correct."
    )
    confidence: float = Field(ge=0, le=1, description="The confidence score for the answer")

    @classmethod
    def model_validate_json(
        cls, json_data: str, *, strict: bool | None = None, context: dict[str, Any] | None = None
    ) -> "list[ModelField]":
        try:
            __tracebackhide__ = True
            return cls.__pydantic_validator__.validate_json(
                json_data, strict=strict, context=context
            )
        except ValidationError:
            # custom parsing logic here
        raise ValueError("Could not find valid json") from last_exc


class QAExtractionSignature(Signature):
    """Your task is to extract the key-value pairs from the document that will follow the instructions on the field to extract.\nplease keep the content you extract truthful and correct based on the document text provided"""

    document: str = InputField()
    instruction: str = InputField(
        description="The instruction on which field to extact and how to extract it."
    )
    tax_from_fields: TaxField = OutputField(
        description="The list of fields extracted from the document text. IMPORTANT!! This must follow a semicolon separated list of values!"
    )


class TypedQAExtraction(dspy.Module):
    def __init__(self):
        self.qa_extraction = dspy.functional.TypedPredictor(QAExtractionSignature)

    def forward(self, document: str, instruction: str) -> TaxFromFields:
        return self.qa_extraction(document=document, instruction=instruction).tax_from_fields

```

### DIFF
--- a/dspy/functional/functional.py
+++ b/dspy/functional/functional.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import typing
+from contextlib import suppress
 from typing import Annotated, Callable, List, Tuple, Union  # noqa: UP035
 
 import pydantic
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
Thanks for adding this @mikeedjones ! Based on some of the referenced issues, does this PR suffice for nested JSON outputs?

mikeedjones:
Yes it ought to, but I'll add some tests to make sure :)

mikeedjones:
@arnavsinghvi11 test added for nested models :)

arnavsinghvi11:
Thanks @mikeedjones!

PR #1063, opened 2024-05-26
Allow to configure Anyscale API in constructor

Neoxelox:
This simple PR just allows to configure Anyscale API in it's constructor, following the pattern on the others LMs.

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -401,6 +401,7 @@ class Anyscale(HFModel):
         super().__init__(model=model, is_client=True)
         self.session = requests.Session()
         self.api_base = os.getenv("ANYSCALE_API_BASE") or api_base
+        assert not self.api_base.endswith("/"), "Anyscale base URL shouldn't end with /"
         self.token = os.getenv("ANYSCALE_API_KEY") or api_key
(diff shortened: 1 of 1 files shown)


krypticmouse:
This is pretty neat! Thanks for this contribution!!

Just to be clear to the users can you add an asset to make sure the URL doesn't end with `/`. Somthing like:

```
self.api_base = os.getenv("ANYSCALE_API_BASE") or api_base
self.token = os.getenv("ANYSCALE_API_KEY") or api_key

assert not self.api_base.endswith("/"), "Anyscale base URL shouldn't end with /"
```

Neoxelox:
> This is pretty neat! Thanks for this contribution!!
> 
> Just to be clear to the users can you add an asset to make sure the URL doesn't end with `/`. Somthing like:
> 
> ```
> self.api_base = os.getenv("ANYSCALE_API_BASE") or api_base
> self.token = os.getenv("ANYSCALE_API_KEY") or api_key
> 
> assert not self.api_base.endswith("/"), "Anyscale base URL shouldn't end with /"
> ```

Okay, done

krypticmouse:
Thanks for the contribution!

PR #1067, opened 2024-05-27
Small fix on logging on bootstrap.py

rogeriochaves:
Since >= is used above before the break, example_idx then is already the length of the example, so the logging was sligthly misleading adding one more example count


rogeriochaves:
actually it's more tricky, the +1 is correct when the bootstrapping goes over all the examples, but wrong when it finishes earlier due to reaching `max_bootstraps`. The way to make it work for both is moving the break at the end of the loop, I'll make this change

PR #1069, opened 2024-05-28
Fix pgvector work with the tutorial

behitek:
When working with the [Tutorial](https://dspy-docs.vercel.app/docs/tutorials/simplified-baleen), using OllamaLocal and PgVectorRM, I found some mismatch below and fixed it.

- Fix `long_text` KeyError by added `content_filed` attr
- Fix `psycopg2.errors.UndefinedFunction: operator does not exist: vector <=> numeric[]` when `include_similarity=True`
- Fix `TypeError: PgVectorRM.forward() got an unexpected keyword argument 'k'` in the forward method
- Code reformat with Ruff

### DIFF
--- a/dspy/retrieve/pgvector_rm.py
+++ b/dspy/retrieve/pgvector_rm.py
@@ -14,8 +14,10 @@ except ImportError:
 try:
     import openai
 except ImportError:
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @behitek ! made a minor typo correction and merging!

PR #1085, opened 2024-05-30
Fix misc assertion errors

Shangyint:


arnavsinghvi11:
Thanks @Shangyint ! will this change break any existing caches with the assertions notebooks in the examples folder?

Shangyint:
I’m afraid that they do. Maybe another practice we could do is to pointing
each notebook at different commit of the repo? ;)

arnavsinghvi11:
I see yes we can do that or we can remove the cache for now to avoid blocking modifications to assertions behavior since besides the intro notebook, the rest of the example notebooks are not fully cached yet

Shangyint:
> I see yes we can do that or we can remove the cache for now to avoid blocking modifications to assertions behavior since besides the intro notebook, the rest of the example notebooks are not fully cached yet

Agreed! Later would be good to develop a standardized way to stage all caches and automatically update them based on need.

PR #1100, opened 2024-06-03
feature(dspy): added retrieval module for Watson Discovery

andreswagner:
This feature allow to search and retrieve documents stored in Watson Discovery


arnavsinghvi11:
Thanks @andreswagner !

PR #1105, opened 2024-06-04
Add message for assert

### LINKED ISSUE #1103: MultiChainComparison M parameter
Do we really need the M parameter for MultiChainComparison? Took me a few minutes of debugging to figure out it only worked if I input three predictions and that I need to change M to fix it.
I would suggest moving the code that assembles the signature to the forward pass so it's no longer necessary to set M.
MultiChainComparison code:
https://github.com/stanfordnlp/dspy/blob/main/dspy/predict/multi_chain_comparison.py

tom-doerr:
Closes #1103
I didn't test this since it's just adding the message for the assert. 
Do I need to test it or is MultiChainComparison covered well enough by tests anyway?

### DIFF
--- a/dspy/predict/multi_chain_comparison.py
+++ b/dspy/predict/multi_chain_comparison.py
@@ -42,7 +42,7 @@ class MultiChainComparison(Module):
                 f"«I'm trying to {rationale} I'm not sure but my prediction is {answer}»",
             )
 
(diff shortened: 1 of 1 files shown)


tom-doerr:
Since it failed for the first commit it seems to be covered well enough by tests

PR #627, opened 2024-03-11
fix(dspy): together client response parsing bug + endpoint

adam-simple:
I opened up a related issue just now https://github.com/stanfordnlp/dspy/issues/626

Basically I was getting errors when using the Together client. If I set use_chat_api to true, the response did not have an "output" field. Also, if I set use_chat_api to false, I get a 404. These changes fix these issues.

I've only done a smoke test. probably worth at least a smoke test by whoever is reviewing this as well.

Issue #626, opened 2024-03-11
[Bug] Together AI completions bugs


adam-simple:
Hi, I believe there are a few bugs in the together API. When hitting the chat/completions endpoint, the response json I get doesn't contain an "output" field, but the Together client requires one. 

changing
`resp_json["output"].get("choices", [])[0] `
to 
`resp_json.get("choices", [])[0]`
fixes the issue. (and should be done both when the chat api is being used and when it isn't)

In addition, when not using the chat api, the url is by default set to just api_base, but it should probably be set to the completions endpoint like
`url = f"{self.api_base}/completions"`
so that when api_base is "https://api.together.xyz/v1", the natural choice, it would become "https://api.together.xyz/v1/completions", which is the correct endpoint.

isaacbmiller:
These issues are still present as of 5/13.

PR #1113, opened 2024-06-05
fix(dspy): together client response parsing

atmiguel:
This should fix the bug specified in https://github.com/stanfordnlp/dspy/issues/626. It looks like there was a potential fix with https://github.com/stanfordnlp/dspy/pull/627, but it was never merged in.

It also adds `api_base` and `api_key` to the constructor, matching the pattern for `Anyscale`.


arnavsinghvi11:
Thanks @atmiguel !

PR #1130, opened 2024-06-11
System prompt support for databricks module

blakekleinhans:
Found this necessary for guiding output length when Signature descriptions weren't impactful. Would be interesting to see system prompt integration in optimizers.

### DIFF
--- a/dsp/modules/databricks.py
+++ b/dsp/modules/databricks.py
@@ -41,6 +41,7 @@ class Databricks(GPT3):
         api_key: Optional[str] = None,
         api_base: Optional[str] = None,
         model_type: Literal["chat", "text", "embeddings"] = None,
+        system_prompt: Optional[str] = None,
         **kwargs,
     ):
         super().__init__(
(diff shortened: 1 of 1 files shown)


Shangyint (changes requested):
Looks good! Would it be possible to also add a fallback so that when no `system_prompt` is provided we can just use `"You are a helpful assistant."`?

blakekleinhans:
I wrote it to match what was in the [GPT3 module](https://github.com/stanfordnlp/dspy/blob/main/dsp/modules/gpt3.py#L106) and thought it would be nice to allow for the None case which uses the endpoint's own default system prompt [(Databricks - Docs)](https://docs.databricks.com/en/machine-learning/foundation-models/api-reference.html#chatmessage)? If you want to retain consistency with existing programs we could make that the default value for the `system_prompt` argument?

arnavsinghvi11:
I think it would actually be fine to leave the default system prompt as None (since changing it for GPT3 will break existing caches). but it would be great to support Databricks' default system prompts for dspy.Databricks with this PR!

PR #1143, opened 2024-06-13
Improvement: Prem Templates

Anindyadeep:
This PR adds an improvement to `dsp.modules.prem.PremAI` by introducing using Prem Prompt Templates throughout dspy calls.

### DIFF
--- a/dsp/modules/premai.py
+++ b/dsp/modules/premai.py
@@ -143,6 +143,8 @@ class PremAI(LM):
                     "params": params,
                 },
             ]
+            all_kwargs.pop("template_id")
+            all_kwargs.pop("params")
 
         response = self.client.chat.completions.create(
             project_id=self.project_id,
--- a/dsp/modules/premai.py
+++ b/dsp/modules/premai.py
@@ -143,6 +143,9 @@ class PremAI(LM):
                     "params": params,
                 },
             ]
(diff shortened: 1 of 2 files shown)


dsp/modules/premai.py
```diff
@@ -122,8 +120,31 @@ def _get_all_kwargs(self, **kwargs) -> Dict[str, Any]:
...
+                raise TypeError("params must be a dictionary")
+
+            messages = [
+                {
+                    "role": "user",
+                    "template_id": template_id,
+                    "params": params,
+                },
+            ]
+            all_kwargs.pop("template_id")
```
arnavsinghvi11, line 146: any reason to remove the params from the kwargs which are propagated in history? might be relevant to see how the DSPy prompt was changed in addition to the Prem template but I'll defer to you here.

arnavsinghvi11:
Thanks @Anindyadeep ! left a minor comment and revised some of the documentation - should be good to merge after that!

Anindyadeep:
> Thanks @Anindyadeep ! left a minor comment and revised some of the documentation - should be good to merge after that!

Yes you are right, so I added some latest changes addressing those. Additionally I combined the raw and all kwargs, just to remove redundancy.

arnavsinghvi11:
Thanks @Anindyadeep !

PR #1145, opened 2024-06-13
KNN few shot fixes

adam-simple:
- Don't pass valset to bootstrap fewshot
- forward_pass unpacks appropriately
- few shot bootstrap args to pass through during forward pass


adam-simple:
It might make sense to put a warning in the KNN call if **kwargs is empty. this can easily happen if someone calls .forward without keyword arguments, instead using args without keywords, like module.forward(input) or something

okhat:
Thanks! Probably remove `*args` altogether then?

okhat:
tagging @adam-simple

adam-simple:
done

arnavsinghvi11:
Thanks @adam-simple !

PR #1146, opened 2024-06-13
Adding in a rockset retriever class

danpechi:
- the input can be finicky; it works with something like: `str(openai_client.embeddings.create(model="text-embedding-ada-002", input=[question]).data[0].embedding)`

- this has some inflexibility wrt underlying schema, ie all results are coming from a singular vector db collection
- note: rockset doesn't have a 'retrieve' function, it just leverages cosine_sim
- this is mostly meant for singular questions, and not batch inference


dspy/retrieve/rockset_rm.py
```diff
@@ -0,0 +1,110 @@
...
+                    """
+
+            results = self._rockset_client.sql(query=sim_query)
+
+            if not pd.DataFrame(results["results"], dtype='str').empty:
+                data = pd.DataFrame(results["results"], dtype='str')[[self._rockset_collection_text_key,
+                                                                      self._rockset_collection_source_key]].values
+                passages.extend(data)
+
+        # Return type not changed, needs to be a Prediction object. But other code will break if we change it.
```
arnavsinghvi11, line 109: comment to be addressed? This can be a dspy.Prediction(passages=...).

arnavsinghvi11:
Thanks @danpechi  for the PR!

Couple of changes needed to merge:

1. To support the import message here, can you add rockset as a dependency to the pyproject.toml? Feel free to follow how it's [done for the other retrievers ](https://github.com/stanfordnlp/dspy/blob/bb97c72da6312600b8ac1a5228f79b07afc427b1/pyproject.toml#L51) and then add rockset as a supported RM in the README (under [Installation](https://github.com/stanfordnlp/dspy?tab=readme-ov-file#1-installation)).

2. Please run `ruff check . --fix-only` and push to pass the failing ruff test.

It would also be great if you could add documentation for the Rockset RM to provide context as done for the [RMs](https://github.com/stanfordnlp/dspy/tree/main/docs/api/retrieval_model_clients).

PR #1148, opened 2024-06-14
Improve CohereLM

behitek:
- CohereLM now supports AWS provider (tested locally with AWS bedrock)
- The retrieval support returns other attributes (current only returns relevant text). This is a quick modification to match my needs. Because I also want other attributes to do the evaluation, post-filtering, etc. So, I am not sure if this feature is good for DSPy or not. 

The second change affects the tutorial. Because the previous retrieval result was a list of strings, but now it is a list of objects. See the line starting with `*` in the code below:

```py
class SimplifiedBaleen(dspy.Module):
    def __init__(self, passages_per_hop=3, max_hops=5):
        super().__init__()

        self.generate_query = [
            dspy.ChainOfThought(GenerateSearchQuery) for _ in range(max_hops)
        ]
        self.retrieve = dspy.Retrieve(k=passages_per_hop)
        self.generate_answer = dspy.ChainOfThought(GenerateAnswer)
        self.max_hops = max_hops

    def forward(self, question):
        context = []
        for hop in range(self.max_hops):
            *context_contents = [c.long_text for c in context]
            query = self.generate_query[hop](
                context=context_contents, question=question
            ).query
            passages = self.retrieve(query).passages
            context = deduplicate(context + passages)
        *context_contents = [c.long_text for c in context]
        pred = self.generate_answer(context=context_contents, question=question)
        return dspy.Prediction(context=context, answer=pred.answer)
```

### DIFF
--- a/dsp/modules/cohere.py
+++ b/dsp/modules/cohere.py
@@ -1,11 +1,14 @@
-from typing import Any, Optional, Literal
-import backoff
 import datetime
 import hashlib
+from typing import Any, Literal, Optional
+
+import backoff
+
 from dsp.modules.lm import LM
 
 try:
     import cohere
+
     cohere_api_error = cohere.errors.UnauthorizedError
 except ImportError:
     cohere_api_error = Exception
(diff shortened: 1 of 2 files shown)


behitek:
Hmm ...

PR #1149, opened 2024-06-14
docs: Update minimal-example.mdx

yyq:
make the quick-start can start well


okhat:
thank you! but this seems wrong? instruct is not a chat model?

PR #1154, opened 2024-06-14
Fixing Cody Snider's PR for LlamaCPP local support (no server)

IMJONEZZ:
Ran ruff check . --fix-only

Added import llama_cpp to fix build test

Added install commands for CPU, CUDA, and Metal with link to pypi for other configurations

Added comment explaining model: str


arnavsinghvi11:
Thanks for adding to the LlamaCPP support @IMJONEZZ ! The changes still fail the tests as [`llama_cpp` is not imported ](https://github.com/stanfordnlp/dspy/actions/runs/9519667491/job/26269685187?pr=1154#step:8:26). To fix this, you can follow the try/except Exception structure [here](https://github.com/stanfordnlp/dspy/blob/bb97c72da6312600b8ac1a5228f79b07afc427b1/dsp/modules/cohere.py#L7)
and/or add llama_cpp to the optional dependencies in the [pyproject.toml](https://github.com/stanfordnlp/dspy/blob/23cd780897c908f61906069c810aae4ae0c89b9f/pyproject.toml#L123). 

It would also be great if you could add documentation for the LlamaCpp LM to provide context as done for the LMs in [our documentation here?](https://github.com/stanfordnlp/dspy/tree/main/docs/api/language_model_clients). maybe can pull from the example in @codysnider 's PR #862 too.

arnavsinghvi11:
Closed by #1347

PR #1168, opened 2024-06-18
fix(dspy): fixed bug in deeplake_rm retriever part

efenocchi:
This commit fixes a bug in the DeeplakeRM class, which is a retriever model that uses Deep Lake to return the top context for a given query. The following changes have been made:

- Updated the import of OpenAI errors.
- Updated the embedding function.
- Modified the vector store instance initialization: it was instantiated every time the user made a request, now it is instantiated only at the beginning.


arnavsinghvi11:
Thanks @efenocchi ! should be ready to merge after you run `ruff check . --fix-only` and push again.

efenocchi:
thanks @arnavsinghvi11, done

PR #1172, opened 2024-06-19
extend generation logic tests

mikeedjones:
Added tests to better define the behaviour of the extend generation logic in `dsp/primitives/predict.py`.

They currently don't pass with either version of the extend generation logic! I'm not sure what the intended behaviour should be - can @XenonMolecule, @okhat @arnavsinghvi11 please explain what these tests should look like?

Cheers!

### DIFF
--- a/tests/predict/test_predict.py
+++ b/tests/predict/test_predict.py
@@ -245,32 +245,182 @@ def test_extend_generation(SandwichIdea):
     lm = DummyLM(
         [
             " whole wheat\n\nProtein: turkey\n\nFat: avocado",
-            " tomato\n\nSauce: mustard",
+            # Incomplete generation leads to tomato field being assigned as an
+            # empty string ("") in dsp.primitives.predict l98 the generation
+            # therefores continues with the next field.
+            " tomato \n\nSauce: mustard",
         ]
     )
     dspy.settings.configure(lm=lm)
 
     prediction = Predict(SandwichIdea)(meal="lunch", dietary_requiements="N/A")
+
+    # The logged conversation:
+    # === DummyLM ===
+    # Based on the meal and dietary requirements, suggest a sandwich idea.
+
+    # ---
+
+    # Follow the following format.
+
+    # Meal: ${meal}
+
+    # Dietary Requiements: ${dietary_requiements}
+
+    # Bread: ${bread}
+
+    # Protein: ${protein}
+
+    # Fat: ${fat}
+
+    # Garnish: ${garnish}
+
+    # Sauce: ${sauce}
+
+    # ---
+
+    # Meal: lunch
+
+    # Dietary Requiements: N/A
+
+    # Bread: whole wheat
+
+    # Protein: turkey
+
+    # Fat: avocado
+    # ===
+    # === DummyLM ===
+    # Based on the meal and dietary requirements, suggest a sandwich idea.
+
+    # ---
+
+    # Follow the following format.
+
+    # Meal: ${meal}
+
+    # Dietary Requiements: ${dietary_requiements}
+
+    # Bread: ${bread}
+
+    # Protein: ${protein}
+
+    # Fat: ${fat}
+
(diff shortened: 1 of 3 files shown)


tests/predict/test_predict.py
```diff
@@ -209,14 +212,81 @@ class OutputOnlySignature(dspy.Signature):
...
+        ]
+    )
+    dspy.settings.configure(lm=lm)
+
+    prediction = Predict(SandwichIdea)(meal="lunch", dietary_requiements="N/A")
+    assert prediction.bread == "whole wheat"
+    assert prediction.protein == "turkey"
+    assert prediction.fat == "avocado"
+    assert prediction.garnish == "tomato"
+    assert prediction.sauce == "mustard"
```
mikeedjones, line 258: @okhat - I'm not sure about the other tests I've added, but I was definitely expecting this test to pass. [However](https://github.com/stanfordnlp/dspy/actions/runs/9577937593/job/26407127842?pr=1172#step:8:48):

```python
SandwichIdea = SandwichIdea(meal, dietary_requiements -> bread, protein, fat, garnish, sauce
    instructions='Based on the meal and ...notation=str required=True json_schema_extra={'__dspy_field_type': 'output', 'prefix': 'Sauce:', 'desc': '${sauce}'})
)

    def test_extend_generation(SandwichIdea):
        lm = DummyLM(
            [
                " whole wheat\n\nProtein: turkey\n\nFat: avocado",
                " tomato\n\nSauce: mustard",
            ]
        )
    
        prediction = Predict(SandwichIdea)(meal="lunch", dietary_requiements="N/A")
        assert prediction.bread == "whole wheat"
        assert prediction.protein == "turkey"
        assert prediction.fat == "avocado"
>       assert prediction.garnish == "tomato"
E       AssertionError: assert '' == 'tomato'
E         - tomato
```

I think when "" is added as the result of the field on https://github.com/stanfordnlp/dspy/blob/34725d08e337fef293b5a2126cf7bf8abb80e41a/dsp/primitives/predict.py#L98 it is interperated as the final value by the extract method of template and the actual value is not included correctly.

okhat:
Hey @mikeedjones, thanks so much for the deep dive! The changes do need to be reverted for now, because they break more fundamental things than the regressions you mentioned, although these are very important too.

In the longer run, I like the direction of this PR overall. Let's think of what the right long-term behavior is for parsing.

The original DSPy behavior, which strikes a very good compromise IMO but needs to be better documented, is: when you ask for n=1 completion, you'll always get it back. If you request n>1 completions, you get at least one. No guarantees. If you need guaranteed n > 1 behavior, create multiple modules with `temperature=0.7 + i*0.001`.

mikeedjones:
Hi @okhat - I think what I show above is that the reverted logic isn't currently working as I expect? At least according to the test cases.

The logic from #920 is also failing on 2/3 test cases.

What should the test cases for this logic look like? Given that #920 was merged - think it makes sense to revert and add some tests so someone else can't come and inadvertently break something further down the line!

tests/predict/test_predict.py
```diff
@@ -209,14 +212,81 @@ class OutputOnlySignature(dspy.Signature):
...
+        sauce: str = dspy.OutputField()
+
+    return SandwichIdea
+
+
+def test_extend_generation(SandwichIdea):
+    lm = DummyLM(
+        [
+            " whole wheat\n\nProtein: turkey\n\nFat: avocado",
+            " tomato\n\nSauce: mustard",
```
mikeedjones, line 248: @okhat So this generation should actually look like

```
" mustard"
```
?

mikeedjones:
Updated the tests and added a comment to each including the logged lm calls. This branch has been pointing at https://github.com/stanfordnlp/dspy/tree/mipro_v2 and the tests are passing for the reverted logic. I think the dummy LM generations are representative - but I'm not sure the behaviour demonstrated by the tests is desired?

PR #1193, opened 2024-06-22
Update evaluate.py-Fix ZeroDivisionError in metric calculation

leegang:
Fix ZeroDivisionError in metric calculation

- Added a check to handle cases where no examples are processed (ntotal == 0)
- Introduced a warning log when the devset is potentially empty
- Refactored the average metric calculation to avoid division by zero
- Updated return statements to use the safely calculated average_metric
- Improved robustness of the evaluation process for edge cases

This commit prevents crashes due to division by zero and provides meaningful  feedback when no examples are processed, enhancing the overall stability of  the evaluation functionality.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -197,12 +197,18 @@ class Evaluate:
                 display_progress,
             )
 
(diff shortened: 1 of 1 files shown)


XenonMolecule:
This was auto-closed by merging #1194. It could be useful to throw a more helpful exception than divide by zero here, though, just so it's more obvious what is going wrong. Perhaps we could make a change with a more descriptive exception but still throw an exception when running on an empty devset.

PR #1200, opened 2024-06-24
Bug - Dataset summary generation failing due to log file set to None

shubham-skr:
The dataset summary generation was failing when log_file is None, as at line 76 it does not check if the log_file is None resulting in an exception.

I fixed the problem by adding the code to check if log_file is None before write operation.


XenonMolecule:
Thanks for catching this!  Looks good to me, will merge if the checks all pass

PR #1202, opened 2024-06-24
Fixed Bug - MIPROv2 fails when program_aware_proposer parameter set to False

shubham-skr:
If program_aware_proposer parameter is set to False when compiling the program with MIPROv2 optimizer, it results in the following error - `AttributeError: 'MIPROv2' object has no attribute 'program_code_string'`. 

The error occurs on line no. 220 when program_code_string in GroundedProposer is assigned. It is due to program_code_string not set to None if program_aware_proposer is false. 

image


XenonMolecule:
This fix looks both good and minimal; thanks, @shubham-skr!

XenonMolecule (approved):

PR #1205, opened 2024-06-25
Fix: package name and version in setup.py and pyproject.toml

hmoazam:
Leave default values in project files for installation from source. Default values get replaced by automation by using comments as markers instead of placeholders.


arnavsinghvi11:
Thanks @hmoazam !

PR #1214, opened 2024-06-27
fix: Remove cache issue and retriever bugs

usamajamil43:
There were some issues related to the cache. Now, I have updated the code a little and tested it. The cache running as expected.
There were also some bugs in the retriever but Now I have resolved these and tested.


arnavsinghvi11:
Thanks @usamajamil43 !

PR #1217, opened 2024-06-28
fix(MiproV2): custom view_data_batch_size for dataset summary creation

shubham-skr:
**Scenario:** When compiling with MiproV2, my large trainset examples led to an error - `model context limit exceeded`. To address this, I set a custom view_data_batch_size. However, the error persisted as the parameter value was not reflected during dataset summary generation. 

**Problem:** The custom view_data_batch_size is not applied during the creation of the dataset summary in GroundedProposer. This occurs because the create_dataset_summary() function within the GroundedProposer defaults to a predefined batch size, ignoring the custom view_data_batch_size parameter value. 
image
image

**Solution:** The solution involves passing the view_data_batch_size parameter to GroundedProposer in MiproV2, ensuring it is utilized when create_dataset_summary() is called in GroundedProposer. 

**Testing:** Verified that the changes made in this PR successfully resolve this issue and the view_data_batch_size parameter set upon initialization of MIPROv2 optimizer is getting used for dataset summary generation. (Also, if not set by the user then it uses the default value of 10 defined in MIRPOv2 class). 

_Feel free to adjust the specifics as needed to better match the project's requirements or to add additional information._

### DIFF
--- a/dspy/propose/grounded_proposer.py
+++ b/dspy/propose/grounded_proposer.py
@@ -237,6 +237,7 @@ class GroundedProposer(Proposer):
         prompt_model,
         trainset,
         program_code_string=None,
+        view_data_batch_size=10,
         use_dataset_summary=True,
         program_aware=True,
         use_task_demos=True,
@@ -257,7 +258,7 @@ class GroundedProposer(Proposer):
         self.prompt_model = prompt_model
         self.program_code_string = program_code_string
         self.data_summary = create_dataset_summary(
-            trainset=trainset, view_data_batch_size=10, prompt_model=prompt_model,
+            trainset=trainset, view_data_batch_size=view_data_batch_size, prompt_model=prompt_model,
         )
         print(f"DATA SUMMARY: {self.data_summary}")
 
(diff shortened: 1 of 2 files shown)


XenonMolecule (approved):
This looks good to me!  It won't break any caches or change any downstream user program behavior since this defaults to the old behavior of always using a batch size of 10 when unspecified.

I noticed that the MIPROv2 dataset summarizer is also missing the max_iterations breakout condition that we added to MIPROv1:     [def _observe_data(self, trainset, max_iterations=10):](https://github.com/shubham-skr/dspy/blob/174b9b42db53cc59da3a3e9d2bd16cf5bd0de007/dspy/teleprompt/mipro_optimizer.py#L175)
(This will reduce the number of calls to the prompt LM when writing the dataset summary, especially for large datasets)

However, this would be a breaking change to user caches, so I'll add this fix for a larger MIPROv2 update separated by a DSPy version update.

Thanks Shubham!

shubham-skr:
Yeah, that would be great. Thanks @XenonMolecule.

arnavsinghvi11:
Thanks @shubham-skr !

PR #620, opened 2024-03-09
trace patch

arnavsinghvi11:
updated trace to now be configured as `[]` instead of `None` to avoid manual setting to enable assertions. 
Patch per comments from #434 #467. 

@okhat - feel free to confirm this does not break any existing caches/behavior (@Shangyint and I have checked to ensure this already!)

Issue #528, opened 2024-03-02
Bug: `AttributeError: 'NoneType' object has no attribute 'clear'`


younes-io:
I got this simple code:

```python
question = "What is the capital of Japan?"

rag_with_assertions(question=question)
```
which throws this error `AttributeError: 'NoneType' object has no attribute 'clear'` because of this line:

https://github.com/stanfordnlp/dspy/blob/98304a2eb9d1dddaaa846e30258cd5d8fc6b5d8d/dspy/primitives/assertions.py#L240

```bash
{
	"name": "AttributeError",
	"message": "'NoneType' object has no attribute 'clear'",
	"stack": "---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[13], line 3
      1 question = \"What is the capital of Japan?\"
----> 3 rag_with_assertions(question=question)

File /workspaces/aider_repos/.venv/lib/python3.12/site-packages/dspy/primitives/program.py:29, in Module.__call__(self, *args, **kwargs)
     28 def __call__(self, *args, **kwargs):
---> 29     return self.forward(*args, **kwargs)

File /workspaces/aider_repos/.venv/lib/python3.12/site-packages/dspy/primitives/assertions.py:317, in handle_assert_forward.<locals>.forward(self, *args, **kwargs)
    314     dspy.settings.configure(bypass_assert=args_to_vals[\"bypass_assert\"])
    316 wrapped_forward = assertion_handler(self._forward, **handler_args)
--> 317 return wrapped_forward(*args, **kwargs)

File /workspaces/aider_repos/.venv/lib/python3.12/site-packages/dspy/primitives/assertions.py:240, in backtrack_handler.<locals>.wrapper(*args, **kwargs)
    238 else:
    239     try:
--> 240         dsp.settings.trace.clear()
    241         result = func(*args, **kwargs)
    242         break

AttributeError: 'NoneType' object has no attribute 'clear'"
}
```

okhat:
I believe assertions currently ask you to add a line at the top of your file.

```dspy.configure(trace=[])```

or something like that?

sleeper:
I had the same problem, and initializing the traces to an empty array as advised by @okhat worked. This bug can probably be closed.

fireking77:
@okhat your sollution worked for me too Thanks! 
 ;), but may I ask where can I check this in doc or somewhere. Can u please point me into the right direction?

Issue #434, opened 2024-02-22
WIP: Assertions FAQ


okhat:
Gathering questions about assertions and suggestions from different users.

This will be most relevant for @Shangyint @manishshettym @arnavsinghvi11.

1. What's changing between different versions? What will be the final stable API for everything below?

2. Do users need to manually set `trace=[]`. Can this be automated?

3. Does `dspy.Assert` retry? If not, shouldn't it?

4. What happens if I don't set `target_module`? Does it directly re-run the last module? What's confusing people here currently.

5. Is it true that you can't really pass a module to the target_module , it must be it's signature?

6. I saved my compiled program with assertions. When I load it (like code below), I get `DSPySuggestionError`.

```python
loaded_compiled_module = MyModuleWithAssertions()
loaded_compiled_module.load(path="compiled_mods_assertions/BootstrapFewShotWithRandomSearch_With_Assertions.json")
# MyModuleWithAssertions embeds dspy.Suggest 
module_with_assertions = assert_transform_module(MyModuleWithAssertions().map_named_predictors(Retry), backtrack_handler) 
```

arnavsinghvi11:
1. Assertions API is currently consistent with the existing [documentation](https://github.com/stanfordnlp/dspy/blob/87ea3acce67ce18f283eb7f92b2aad9c5f109a93/docs/assertions.md), which will be updated if any other changes are made! 

2. Yes this should actually be configured by default, just modifying [`trace=[]`](https://github.com/stanfordnlp/dspy/blob/87ea3acce67ce18f283eb7f92b2aad9c5f109a93/dsp/utils/settings.py#L38) here which will resolve issues like https://github.com/stanfordnlp/dspy/issues/432. @okhat let me know if this change can be added without breaking other existing behavior!

3. `dspy.Assert` performs the same retrying attempts as `dspy.Suggest`, with the difference only lying in if the validation function is not passed even when the `max_backtracking_attempts` are reached,  `dspy.Assert` will trigger a `DSPyAssertionError` while `dspy.Suggestion` will log this error but continue program execution. 
 
One thing to note is once an DSPyAssertionError/DSPySuggestionError is triggered at any point, this is considered as an "attempt." This can have some implications with multiple Assertion statements as if a particular statement continually produces an invalid response, the result will error out on that corresponding statement(if using `dspy.Assert` statements) or continue to the next examples (if using `dspy.Suggest` statements) without having the output fully tested on the rest of the Assertion statements. 
[...]

PR #1219, opened 2024-06-28
fix(dspy): fix null reference in settings.trace

grant-d:
Using `dspy-ai===2.4.5`
When using the following code:

```python
__llm = dspy.Google(model="models/gemini-1.5-flash")
dspy.settings.configure(lm=__llm) # error below!
dspy.settings.configure(lm=__llm, trace=[]) # mitigates error (but NOT the fix)

# __llm = dspy.OpenAI(model="gpt-4o") # Works fine


class PlannerModule(dspy.Module):
  def forward(self, ...):
    # ... trivial implementation ...
    dspy.Suggest(...)

exec = PlannerModule().activate_assertions(max_backtracks=3)
prediction: dspy.Prediction = exec(input)
```

I get the following error:
```text
  File "...xyz.py", line 91, in <module>
    prediction: dspy.Prediction = exec(input)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/Data/repos/jigxie/.venv/lib/python3.12/site-packages/dspy/primitives/program.py", line 26, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/Data/repos/jigxie/.venv/lib/python3.12/site-packages/dspy/primitives/assertions.py", line 318, in forward
    return wrapped_forward(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/Data/repos/jigxie/.venv/lib/python3.12/site-packages/dspy/primitives/assertions.py", line 241, in wrapper
    dsp.settings.trace.clear()
    ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'clear'
```

Not sure why there is a difference in behavior between `OpenAI` and `Google`, however, via visual inspection the fix is both defensive & trivial:
- Added a null-check per similar code elsewhere in the codebase.

### DIFF
--- a/dspy/primitives/assertions.py
+++ b/dspy/primitives/assertions.py
@@ -211,12 +211,14 @@ def backtrack_handler(func, bypass_suggest=True, max_backtracks=2):
                 if i == max_backtracks:
                     if isinstance(current_error, DSPyAssertionError):
                         raise current_error
-                    dsp.settings.trace.clear()
+                    if dsp.settings.trace:
+                        dsp.settings.trace.clear()
                     result = bypass_suggest_handler(func)(*args, **kwargs) if bypass_suggest else None
                     break
                 else:
                     try:
-                        dsp.settings.trace.clear()
+                        if dsp.settings.trace:
+                            dsp.settings.trace.clear()
                         result = func(*args, **kwargs)
                         break
                     except (DSPySuggestionError, DSPyAssertionError) as e:


okhat:
Thank you @grant-d ! I'm not sure this is a correct fix. Why clear the trace altogether?

okhat:
i.e., I agree we may need a correction or something but clear seems not to be the fix?

grant-d:
@okhat it was the simplest fix, I didn't have time to figure out the root cause in the `Google` provider.

grant-d:
Oh wait. I think you misunderstood my fix. The `clear` statement was already there (see the diff), I just wrapped it in a null-check

arnavsinghvi11:
Hi @grant-d , this is a chance actually only needed when using DSPy.assertions. Please refer to #528, #434

grant-d:
I do agree that it only happens when using `assertions`. But in terms of defensive coding best-practice, why not add the null-check anyway?

arnavsinghvi11:
took a deeper look and it seems like this fix is not the intended behavior as #620 actually covered setting trace to be initialized to an empty list on each configuration. while the suggested change adds a null-check (which we can merge), ideally, we do not want to force the user to manually set `trace=[]` to activate assertions. `dspy.settings.configure(lm=__llm)` should work as is since trace should default to `[]` anyways.

seems like the bug is actually in `dspy.Google`. I think it could be because of the defined [`self.config` ](https://github.com/stanfordnlp/dspy/blob/d98859b0185de4912bddbf00bd49489da8cb5eeb/dsp/modules/google.py#L97) overriding `self.config` for settings and other use cases in DSPy. I don't have access to test this myself, but if you change that variable to another name and refactor `dspy.Google` accordingly, does this resolve the error and work with `dspy.settings.configure(lm=__llm)` as is.

PR #1226, opened 2024-07-01
Loosen Anthropic requirement

krishnan-chandra:
The latest version of the Anthropic library is `0.30.0`, but [DSPy currently enforces that the Anthropic library is pinned to the `0.18` minor version](https://python-poetry.org/docs/dependency-specification/#version-constraints). This PR changes the requirement to allow versions < 1.0.0.


okhat:
Thank you @krishnan-chandra ! Can we find who locked the version at 0.18 and tag them to review this?

Issue #1227, opened 2024-07-01
Missing import GSM8K dataset in __init__.py


aabhi02:
There is an import statement missing in the datasets init file for GSM8K dataset

PR #1228, opened 2024-07-01
added gsm8k import to init

aabhi02:
This PR solve the issue #1227

### DIFF
--- a/dspy/datasets/__init__.py
+++ b/dspy/datasets/__init__.py
@@ -2,3 +2,4 @@ from .colors import Colors
 from .dataloader import DataLoader
 from .dataset import Dataset
 from .hotpotqa import HotPotQA
+from .gsm8k import GSM8K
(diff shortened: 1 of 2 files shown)


okhat:
I don't understand the goal of this PR. What are the problems it's fixing?

aabhi02:
The main issue i was facing is as mentioned in #1227 , all the other datasets can be used by imported as `from dspy.datasets` but to use `gsm8k` we have to do `from dspy.datasets.gsm8k import GSM8K`.

and also i want to know what is the reason behind GSM8K not being the child class of `Dataset` class.

If i try to make the `GSM8K` class as the child class of `Dataset` class i am facing the error as shown in #1229

okhat:
Thank you @aabhi02 ! I would close this for now. GSM8K has a metric and other special functions so we want importing it to be a little different for now.

aabhi02:
okay got it. Thanks!

PR #1239, opened 2024-07-03
Add filter for QdrantRM

hawktang:
Add filter support for QdrantRM to support qdrant search filter like

https://qdrant.tech/documentation/concepts/filtering/

uses

image


arnavsinghvi11:
Thanks @hawktang !

PR #1240, opened 2024-07-03
Add MultiOpenAIVectorizer to allow general openai api format embeddings to be used for DSPY RM

hawktang:
With LiteLLM proxy, the embedding model that LiteLLM supported can be used in DSPY

### DIFF
--- a/dsp/modules/sentence_vectorizer.py
+++ b/dsp/modules/sentence_vectorizer.py
@@ -306,3 +306,50 @@ class PremAIVectorizer(BaseSentenceVectorizer):
 
         embeddings = np.array(embedding_list, dtype=np.float32)
         return embeddings
+
+    class MultiOpenAIVectorizer(BaseSentenceVectorizer):
+        '''
+        This vectorizer uses a general openai api format to convert texts to embeddings.
+        '''
+
+        def __init__(
+                self,
+                model: str = 'text-embedding-ada-002',
+                embed_batch_size: int = 1024,
+                base_url='http://0.0.0.0:4000',
+                api_key: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks for opening this PR @hawktang . just curious, where is the LiteLLMVectorizer being used? seems to me it is just using the openAI embeddings but just wanted to double-check here.

hawktang:
Sorry for the late relay, I am on travel now.

LiteLLMVectorizer is used to call embedding models which LiteLLM proxy supported.

LiteLLM proxy is used to  call different LLM APIs using the OpenAI API format. 

With LiteLLM adaptor DSPY can direct support all the cloud and local model LiteLLM supports.

Because it is using OpenAI API format, LiteLLMVectorizer I wrote is quite similar except the base_url

Directly add base_url as parameter for openAI embeddings can achieve the result. 

I will raise another PR if LiteLLMVectorizer class is redundant.

okhat:
See #1357 also, may overlap?

hawktang:
yes, totally overlap.

#1357 is what I expected to have.

Peter Ze TANG (汤赜)

hawktang:
I have change the name to MultiOpenAIVectorizer to follow the new MultiOpenAI api in dspy. Can we merge this PR for general embedding service to be use in dspy

hawktang:
> I have change the name to MultiOpenAIVectorizer to follow the new MultiOpenAI api in dspy. Can we merge this PR for general embedding service to be use in dspy RM.

This should be a quick solution for LM and RM to use LiteLLM before the roadmap finished.

Any feedback for the update of the  PR

Issue #1254, opened 2024-07-06
lm.copy() not working for Together API


Patchwork53:
# Minimum Code To Reproduce
```python
import dspy
lm = dspy.Together(model="meta-llama/Llama-3-8b-chat-hf")
lm.copy() 
```
# Output
```python
KeyError                                  Traceback (most recent call last)
Cell In[1], line 3
      1 import dspy
      2 lm = dspy.Together(model="meta-llama/Llama-3-8b-chat-hf")
----> 3 lm.copy() 

File /opt/conda/lib/python3.10/site-packages/dsp/modules/lm.py:135, in LM.copy(self, **kwargs)
    133 """Returns a copy of the language model with the same parameters."""
    134 kwargs = {**self.kwargs, **kwargs}
--> 135 model = kwargs.pop("model")
    137 return self.__class__(model=model, **kwargs)

KeyError: 'model'
```

# Expected Output
Works for OpenAI endpoints.
```python
import dspy
lm = dspy.OpenAI(model='gpt-3.5-turbo' )
lm.copy() 
```
```python
<dsp.modules.gpt3.GPT3 at 0x7fbb33f920e0>
```

## This error popped up in training with Together AI

I actually faced this error after training. Can't show the entire code but here's a snippet:
```python
compiled_program = teleprompter.compile(
    TaskMatcherProgram(),
    trainset=train_data
)
```
And the output:
```python
100%|██████████| 100/100 [03:13<00:00,  1.94s/it]
  4%|▍         | 4/100 [00:00<00:00, 2907.16it/s]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[16], line 12
      3 teleprompter = BootstrapFewShot(
      4     metric=task_matcher_metric,
      5     max_bootstrapped_demos=3,
      6     max_labeled_demos=3,
      7     max_rounds=2
      8 )
     10 print(help(teleprompter.compile))
---> 12 compiled_program = teleprompter.compile(
     13     TaskMatcherProgram(),
     14     trainset=train_data
     15 )

File /opt/conda/lib/python3.10/site-packages/dspy/teleprompt/bootstrap.py:84, in BootstrapFewShot.compile(self, student, teacher, trainset)
     82 self._prepare_student_and_teacher(student, teacher)
     83 self._prepare_predictor_mappings()
---> 84 self._bootstrap()
     86 self.student = self._train()
     87 self.student._compiled = True

[...]

PR #1257, opened 2024-07-07
hotfix(dspy): update Together client self.kwargs to include "model"

Patchwork53:
Solves issue: https://github.com/stanfordnlp/dspy/issues/1254


arnavsinghvi11:
Thanks @Patchwork53 !

PR #1268, opened 2024-07-10
added configurable backoff max time

arnavsinghvi11:
allows for user-configurable backoff times across LM/RM providers.

Note that if you'd like to set individual backoff times for specific providers, you can do through the dspy context manager 

```
with dspy.context(backoff_time = ..):
      dspy.OpenAI(...) # example

with dspy.context(backoff_time = ..):
      dspy.AzureOpenAI(...) # example
```

### DIFF
--- a/docs/docs/faqs.md
+++ b/docs/docs/faqs.md
@@ -137,7 +137,27 @@ Firstly, please refer to your LM/RM provider to ensure stable status or sufficie
 
 Additionally, try reducing the number of threads you are testing on as the corresponding servers may get overloaded with requests and trigger a backoff + retry mechanism.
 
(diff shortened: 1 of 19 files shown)


LukvonStrom:
Hi @arnavsinghvi11 would it be possible to ship a release with this change please? I am currently facing an issue where I sometimes overrun the context limit and am then sent into 16 minutes of continuous timeout. Reading the docs I was led to believe that the setting already takes effect, started my experimentation chain (takes >48hrs) and have to watch my code now doing needless retries all the time :(

arnavsinghvi11:
Hi @LukvonStrom missed responding back but this change is on the [latest DSPy release](https://github.com/stanfordnlp/dspy/releases/tag/2.4.13)!

PR #1272, opened 2024-07-12
Add the groq client version 0.9.0 to requirements*.txt

curious-cat-163:
Was tinkering with this library today and found that there is code that uses Groq client but it is not expressed as a requirement. 

Perhaps, it is intentional; but raising a PR anyway to clarify the stance on client libs because I saw the Open AI client in the requirements.txt.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,4 @@ pytest-mock==3.12.0
 ruff==0.3.0
 torch==2.2.1
 transformers==4.38.2
+groq==0.9.0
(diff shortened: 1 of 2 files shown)


isaacbmiller:
We have openai as a dep because so many things are based on it, all other extra clients/libs e.g. groq, anthropic, llama-index are marked as optional dependencies inside pyproject.toml

https://github.com/stanfordnlp/dspy/blob/b96e199b661d8eb3f3389e94689441f1e123ebcb/pyproject.toml#L122

Issue #1206, opened 2024-06-25
No module named 'mlc_chat'


wuyuzhe:
when i used this to load local module,it print No module named 'mlc_chat' in dsp/modules/hf_client.py   line 464
```
agent = dspy.ChatModuleClient(model=model, model_path=model_path)
```
how to resolve it

my versions:
python:  3.10
dspy-ai:  2.4.10

arnavsinghvi11:
Hi @wuyuzhe , please reference the ChatModuleClient [documentation](https://dspy-docs.vercel.app/api/local_language_model_clients/MLC) to install the dependencies. We can eventually add this within the file wrapped with a try/except block too.

wuyuzhe:
first,thank you for your answer.
i didn't find the mlc-chat-nightly package from https://mlc.ai , now i installed mlc-ai-nightly-cu121 and mlc-llm-nightly-cu121 on my machine. but it also show that error.
i find the shell  at https://dspy-docs.vercel.app/api/local_language_model_clients/MLC
```
pip install --no-deps --pre --force-reinstall mlc-ai-nightly-cu118 mlc-chat-nightly-cu118 -f https://mlc.ai/wheels
pip install transformers
git lfs install
```
and the command in mlc
```
python -m pip install --pre -U -f https://mlc.ai/wheels mlc-llm-nightly-cu121 mlc-ai-nightly-cu121
```
so, what wrong with my environment

wuyuzhe:
mlc_chat was deprecated, it turned to mlc_llm

okhat:
ah good catch, do you want to open a pr to correct this somehow @wuyuzhe or to note it in the right place?

zhaohan-dong:
The ChatModule from mlc is depreciated as well. Tried to make a pr (#1285) to mimic the behavior.

PR #1285, opened 2024-07-16
fix(mlc-llm): fix mlc-llm ChatModuleClient with MLCEngine

zhaohan-dong:
Fix on issue #1206.
Both mlc_chat and ChatModule are depreciated. Replaced with MLCEngine which will be suppoorted going forward.
A few considerations:
- Renaming the class to reflect the change from ChatModule
- Use EngineConfig when setting up the MLCEngine to have more control over configuration


arnavsinghvi11:
Thanks @zhaohan-dong !

PR #1287, opened 2024-07-17
add OpenAICompatible

SomeoneKong:
Add OpenAICompatible, likes GPT3, but each instance uses a separate cache.

Allow to use multi OpenAICompatible LLM service with different base_url

related #744, #915

### DIFF
--- a/docs/api/language_model_clients/MultiOpenAI.md
+++ b/docs/api/language_model_clients/MultiOpenAI.md
@@ -7,12 +7,19 @@ sidebar_position: 1
 ### Usage
 
 ```python
-lm = dspy.MultiOpenAI(model='gpt-4o-mini', 
-                      api_key='xxxx',
-                      api_provider='openrouter',
-                      api_base='https://openrouter.ai/api/v1/',
-                      model_type='chat',
-                      )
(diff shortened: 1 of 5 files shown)


arnavsinghvi11:
Hi @SomeoneKong , thanks for the PR! looks good and helps resolve the issues. Could we refactor this class to be `dspy.MultiOpenAI` to help distinguish from the existing `dspy.OpenAI` while highlighting the use case for supporting configuring multiple open ai clients while using this?

It would also be great if you could add documentation for this class as done for the LMs in [our documentation here?](https://github.com/stanfordnlp/dspy/tree/main/docs/api/language_model_clients), detailing a working example of configuring 2 OpenAI SDK clients using this. Thanks!

SomeoneKong:
@arnavsinghvi11 done

SomeoneKong:
Please note:
If it is used for multiple instances with different base_url, it only works with openai>=1.0 dependency.
When using only one base_url, this implementation is also compatible with openai <1.0.

Therefore, I previously named it OpenAICompatible.

arnavsinghvi11:
Thanks @SomeoneKong !

PR #1289, opened 2024-07-17
Add support for Langchain Tool use Optimization via Avatar

krypticmouse:
Adding support for Avatar Actor optimization for optimal tool usage using comparator based optimization that [Avatar](https://arxiv.org/abs/2406.11200) proposes. 

This also comes with thread-safe evaluator for module like Avatar that keep modifying their signature during execution. Using `dspy.Evaluate` doesn't work for such modules because it works on evaluating the pass argument and not the deepcopy of it. We do not want to change that so added a separate implementation as part of optimizer class.

Adding a step by step guide for Avatar Optimizer module for user to navigate with.

Evals for Arxiv QA:
image

Evals for Search QA:
image

### DIFF
--- a/dspy/predict/avatar/avatar.py
+++ b/dspy/predict/avatar/avatar.py
@@ -4,8 +4,8 @@ from copy import deepcopy
 from pydantic.fields import FieldInfo
 
 from dspy.predict.avatar.signatures import Actor
-from dspy.predict.avatar.models import Action, Tool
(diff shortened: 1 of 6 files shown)


CShorten:
👀

arnavsinghvi11:
Looks great!! We should eventually create a section in DSPy docs highlighting all these wonderful results 🚀 

Thanks @krypticmouse !

PR #1294, opened 2024-07-19
Fix conditional in Module.named_parameters() to prevent infinite recursion

theta-lin:
The problem is illustrated as follows:

For a module such as the following I added to the tests:
```python3
class DuplicateModule(Module):
    def __init__(self):
        super().__init__()
        self.p0 = dspy.Predict("question -> answer")
        self.p1 = self.p0
```
Running its `named_parameters()` method would result in infinite recursion. This is due to that `Predict` is both a `Module` and a `Parameter`. Thus, according to the original arrangement of the if statements, suppose that `p0` is already visited. Then, by having the `Predict` instance already a member of the `visited` set and that it is a `Module`, the second condition would still execute when `p1` is visited.

I originally discovered this problem when using `assert_transform_module()`, which calls `Module.named_parameters()` at some point.

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -23,10 +23,11 @@ class BaseModule:
         named_parameters = []
 
         def add_parameter(param_name, param_value):
-            if isinstance(param_value, Parameter) and id(param_value) not in visited:
-                visited.add(id(param_value))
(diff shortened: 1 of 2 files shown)


okhat:
Wonderful catch, thank you @theta-lin !!

Issue #1298, opened 2024-07-20
semver dependency error when setting up dev container


will-0:
Tests do not run when setting up devcontainer from clean repo clone, due to missing `semver` dependency.

## Steps to reproduce

As per instructions in [CONTRIBUTING.md](https://github.com/stanfordnlp/dspy/blob/main/CONTRIBUTING.md)
1. Clone repo
2. Select "Dev Containers: Rebuild and Reopen in container"
3. Ensure poetry interpreter selected
4. Run `pytest`

Output:

```
============================================================================================================================================== test session starts ===============================================================================================================================================
platform linux -- Python 3.9.19, pytest-6.2.5, py-1.11.0, pluggy-1.5.0
rootdir: /workspaces/dspy
plugins: anyio-4.4.0, mock-3.14.0
collected 165 items / 1 error / 1 skipped / 163 selected                                                                                                                                                                                                                                                         

===================================================================================================================================================== ERRORS =====================================================================================================================================================
__________________________________________________________________________________________________________________________________ ERROR collecting build_utils/test_version.py __________________________________________________________________________________________________________________________________
ImportError while importing test module '/workspaces/dspy/build_utils/test_version.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
build_utils/test_version.py:5: in <module>
    import semver
E   ModuleNotFoundError: No module named 'semver'
============================================================================================================================================ short test summary info =============================================================================================================================================
ERROR build_utils/test_version.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================================================================================================== 1 skipped, 1 error in 9.21s ===========================================================================================================================================
```

## Suggested fix

[...]

PR #1299, opened 2024-07-20
fix(dependency/semver): updated semver to allow tests to run

will-0:
Running `pytest` in a newly created dev container fails, due to a missing dependency on `semver`.

This PR adds `semver` as a dev dependency (with `poetry add semver --group dev`). Tests ran locally once included.


will-0:
Fixes #1298

PR #1301, opened 2024-07-21
Support nested objects in set_attribute_by_name() via magicattr

### LINKED ISSUE #882: Typed predictors initiated with a range make assert_transform_module crash
Hi there, first, thanks a lot for this very cool library!

Code to reproduce the issue:
```
import functools
from logging import Logger
from typing import List, Optional
from dspy import Module, Suggest, assert_transform_module, backtrack_handler, Predict
from dspy.functional import TypedChainOfThought

class Sandbox(Module):
    def __init__(self, logger: Logger = Logger("Sandbox")):
        super().__init__()
        self.logger = logger

    def forward(self, context: List[str] = []):
        transformed_module_with_array_nested_predictors = assert_transform_module(
            TransformedModule(),
            functools.partial(backtrack_handler, max_backtracks=2))
        transformed_module_with_array_nested_predictors([])


class TransformedModule(Module):
    def __init__(self, logger: Logger = Logger("TransformedModule")):
        super().__init__()
        self.logger = logger
        self.array_predictor = [TypedChainOfThought("question -> answer") for _ in range(3)]

    def forward(self, context: List[str] = []):
        self.array_predictor[0](question="What is the capital of the world?")
```

Trace:
```
module
    module.map_named_predictors(dspy.retry.Retry)
  File "/Users/X/Library/Caches/pypoetry/virtualenvs/rag-jAwKM_yQ-py3.11/lib/python3.11/site-packages/dspy/primitives/program.py", line 47, in map_named_predictors
    set_attribute_by_name(self, name, func(predictor))
  File "/Users/X/Library/Caches/pypoetry/virtualenvs/rag-jAwKM_yQ-py3.11/lib/python3.11/site-packages/dspy/primitives/program.py", line 88, in set_attribute_by_name
    sub_obj = getattr(obj, module_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'TransformedModule' object has no attribute 'array_predictor[0]'. Did you mean: 'array_predictor'?
```

It works as it should with non typed arrays of predictors or typed predictors alone

My dirty workaround:
```
# __init__
        [setattr(self, f"finalize_sub_section_{i}", TypedChainOfThought(GenerateFinalisedSubsection)) for i in range(max_subsections)]

# forward
for i, sub_section in enumerate(sub_sections[:self.max_subsections]):
            # p = self.finalize_sub_section[i](
            p = getattr(self, f"finalize_sub_section_{i}")(context=...
```

theta-lin:
## Fix

I propose using [magicattr](https://github.com/frmdstryr/magicattr) for implementing `set_attribute_by_name()`. While this does introduce an additional dependency, I think it is better than the alternatives of either reinventing the wheel or using `exec()` directly.

## Problem

Originally, `set_attribute_by_name()` sets an attribute specified by its string name via regex. However, for a name such as `"some_list[0].some_attrib"`, it would be matched by `module_pattern`, which would in turn interpret `"some_list[0]"` as the name of a submodule and an attribute of the parent object `obj`, and call `getattr(obj, "some_list[0]")` to retrieve it. However, `getattr()` does not support accessing the member of a list/dict attribute directly and would thus fail.

Fixes #882. While #882 concerns the issue of using both a list of predictors and that the predictors are typed, I think the use of typed predictors is not the root cause of the issue. The original regex implementation works just fine with non-nested list/dict attributes, but when using typed predictors, note that `TypedPredictor` wraps a `Predict` object in the `predictor` attribute. Therefore, the actual predictor/parameter would not be named as `"predictor_list[0]"`, but rather `"predictor_list[0].predictor"` and trigger the issue.

I encountered this issue without using typed predictors, but rather simply using `assert_transform_module()` on nested modules with lists and dicts of other modules.

### DIFF
--- a/dspy/primitives/program.py
+++ b/dspy/primitives/program.py
@@ -1,4 +1,5 @@
 import re
+import magicattr
 
 from dspy.primitives.assertions import *
 from dspy.primitives.module import BaseModule
@@ -74,37 +75,7 @@ class Module(BaseModule, metaclass=ProgramMeta):
     #     return new_copy
 
 
-# FIXME(Shangyint): This may cause some problems for nested patterns.
 def set_attribute_by_name(obj, name, value):
-    # Regular expressions for different patterns
-    module_pattern = re.compile(r"^([^.]+)\.(.+)$")
-    list_pattern = re.compile(r"^([^\[]+)\[([0-9]+)\]$")
-    dict_pattern = re.compile(r"^([^\[]+)\['([^']+)'\]$")
-
-    # Match for module.attribute pattern
-    module_match = module_pattern.match(name)
-    if module_match:
-        module_name, sub_name = module_match.groups()
-        sub_obj = getattr(obj, module_name)
-        set_attribute_by_name(sub_obj, sub_name, value)
-        return
-
-    # Match for list[index] pattern
-    list_match = list_pattern.match(name)
-    if list_match:
-        list_name, index = list_match.groups()
-        getattr(obj, list_name)[int(index)] = value
-        return
-
-    # Match for dict['key'] pattern
-    dict_match = dict_pattern.match(name)
-    if dict_match:
-        dict_name, key = dict_match.groups()
-        getattr(obj, dict_name)[key] = value
-        return
-
-    # Default case for simple attributes
-    setattr(obj, name, value)
-
+    magicattr.set(obj, name, value)
 
 Program = Module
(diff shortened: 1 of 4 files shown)


okhat:
Ah interesting, thanks @theta-lin ! 

@Shangyint thoughts?

Shangyint:
Thanks @theta-lin!

This is a great improvement over the existing error-prone code. LGTM to merge.

arnavsinghvi11:
Thanks @theta-lin !

PR #1305, opened 2024-07-22
fix (mipro): enable running mipro using only labeled demos

famimahmud:
Enable performing mipro prompt optimization while only using labeled demos. Here, bootstrapped demos are solely used for the meta-prompt to enhance the final prompt.

Before fix:  

- max_bootstrapped_demos = 0 
and 
- max_labeled_demos > 0  

mipro optimizer raises ValueError: "No examples found for the given predictor"

### DIFF
--- a/dspy/teleprompt/mipro_optimizer.py
+++ b/dspy/teleprompt/mipro_optimizer.py
@@ -419,12 +419,10 @@ class MIPRO(Teleprompter):
             module = student.deepcopy()
             evaluate = Evaluate(devset=trainset, metric=self.metric, **eval_kwargs)
 
-            # In the case where the bootstrapped and labeled demos are set to 0, we'll stil bootstrap examples to use in our meta prompt
-            if (
-                max_bootstrapped_demos == 0 and max_labeled_demos == 0
-            ):  # TODO: address case when max_bootstrapped alone is 0
+            # In the case where the bootstrapped demos are set to 0, we'll stil bootstrap examples to use in our meta prompt
+            if max_bootstrapped_demos == 0:
                 max_bootstrapped_demos_for_candidate_gen = 1
-                max_labeled_demos_for_candidate_gen = 1  # TODO: this might only need to be 0
+                max_labeled_demos_for_candidate_gen = max_labeled_demos
             else:
                 max_bootstrapped_demos_for_candidate_gen = max_bootstrapped_demos
                 max_labeled_demos_for_candidate_gen = max_labeled_demos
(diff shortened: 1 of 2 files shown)


okhat:
Thank you @famimahmud ! 

@XenonMolecule @klopsahlong thoughts?

dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -254,9 +254,9 @@ def compile(
             evaluate = Evaluate(devset=trainset, metric=self.metric, **eval_kwargs)
 
             # Determine the number of fewshot examples to use to generate demos for prompt
-            if max_bootstrapped_demos == 0 and max_labeled_demos == 0:
+            if max_bootstrapped_demos == 0:
                 max_bootstrapped_demos_for_candidate_gen = BOOTSTRAPPED_FEWSHOT_EXAMPLES_IN_CONTEXT
```
arnavsinghvi11, line 258: shouldn't this be set to 0 to ensure bootstrapped_demos in the few-shot generation is 0 as intended to only use labeled demos?

dspy/teleprompt/mipro_optimizer.py
```diff
@@ -419,12 +419,10 @@ def compile(
             module = student.deepcopy()
             evaluate = Evaluate(devset=trainset, metric=self.metric, **eval_kwargs)
 
-            # In the case where the bootstrapped and labeled demos are set to 0, we'll stil bootstrap examples to use in our meta prompt
-            if (
-                max_bootstrapped_demos == 0 and max_labeled_demos == 0
-            ):  # TODO: address case when max_bootstrapped alone is 0
+            # In the case where the bootstrapped demos are set to 0, we'll stil bootstrap examples to use in our meta prompt
+            if max_bootstrapped_demos == 0:
                 max_bootstrapped_demos_for_candidate_gen = 1
```
arnavsinghvi11, line 424: same point as below

dspy/teleprompt/mipro_optimizer_v2.py
famimahmud: In the current implementation of MIPRO and MIPROv2 at least one bootstrapped demo is required for the candidate generation of the meta-prompt. By setting it like this, bootstrapped demos are only used for the meta-prompt but not in the final prompt.

okhat:
Thanks a lot @famimahmud ! Deciding to close after discussion internally. Bootstrapping is good :D Let us know if you want to consider this more deeply.

PR #1313, opened 2024-07-24
databricks_rm: returning extra columns

lmoros-DB:
- Extension to return extra columns in the result of the Vector Search
- Added extra error checks:  REST API error, ID column not in index, TEXT column not in index
- Docs, Doc_ids and extra columns are sorted by the search "score" in descending order

### DIFF
--- a/dspy/retrieve/databricks_rm.py
+++ b/dspy/retrieve/databricks_rm.py
@@ -1,10 +1,8 @@
 import json
 import os
 from collections import defaultdict
-from typing import List, Union
-
+from typing import List, Union, Any, Dict
 import requests
-
 import dspy
 from dspy.primitives.prediction import Prediction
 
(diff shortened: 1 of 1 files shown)


dspy/retrieve/databricks_rm.py
```diff
@@ -125,23 +189,42 @@ def forward(self, query: Union[str, List[float]], query_type: str = 'text', filt
...
+        # Extracting the results
+        items = []
+        for idx, data_row in enumerate(results["result"]["data_array"]):
+            item = {}
+            for col_name, val in zip(col_names, data_row):
+                item[col_name] = val
+            items += [item]
+
+        # Sorting results by score in descending order
+        sorted_docs = sorted(items, key=lambda x: x["score"], reverse=True)
```
arnavsinghvi11, line 220: I believe this should slice up to top-k results - as done previously [here](https://github.com/stanfordnlp/dspy/pull/1313/files#diff-4f9ab9a7b9160924690ab31d1c9d59c6dd5280e53fcaef2f5420170d031bffcdL146)

maybe we can also wrap this issue into this PR as well, adding self.k to the forward pass

lmoros-DB: slicing added

arnavsinghvi11:
Thanks @lmoros-DB !

PR #1321, opened 2024-07-26
fix(dspy): added type check on evaluator result

jmadden12:
signature_opt_typed.py previously assumed that the result of evaluator() was necessarily a float, but if the user sets the flags return_all_scores or return_outputs, it returns a tuple, which is not sortable.


arnavsinghvi11:
Thanks @jmadden12 !

PR #1323, opened 2024-07-26
feat(dspy): add system and format options to ollama

LLukas22:
Adds the `format` and `system` options to allow the use of `JSON` mode and the definition of a system prompt when using ollama as defined [here](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion).

Example:
```python
llm = OllamaLocal(
    model="llama3.1",
    format="json",
    system="You are an sentiment analysis assistant. You should classify given texts into one of the following categories: `positive`, `negative`, `neutral`"
)
```

### DIFF
--- a/dsp/modules/ollama.py
+++ b/dsp/modules/ollama.py
@@ -1,6 +1,6 @@
 import datetime
 import hashlib
-from typing import Any, Literal
+from typing import Any, Literal, Optional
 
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @LLukas22 !

PR #1331, opened 2024-07-29
http client added to GPT3

mjaliz:
Set http_client on openai to support http_proxy


arnavsinghvi11:
Thanks @mjaliz ! (failing test is from an earlier commit)

satyaloka93:
Will this be supported in the new dspy.LM adapters? The gpt3.py is being deprecated, right?

PR #1339, opened 2024-07-31
Fixed some bugs in the ReAct module

xnmp:
- Fixes issue where model responds with literally `Finish[answer]`, instead of having the `answer` be the answer. 
- Fixes issue where the model generates more than one thought, action, or observation.


arnavsinghvi11:
Thanks @xnmp !

PR #1347, opened 2024-08-02
Llama CPP Support - Addressed some PR Comments - Extending 2 other PRs

AbdulDridi:
A continuation of these two:

https://github.com/stanfordnlp/dspy/pull/862

https://github.com/stanfordnlp/dspy/pull/1154

Addressing their pr comments

### DIFF
--- a/docs/api/local_language_model_clients/LlamaCpp.md
+++ b/docs/api/local_language_model_clients/LlamaCpp.md
@@ -39,4 +39,24 @@ After initializing the Llama model, you can interact with it using the `LlamaCpp
 import dspy
 
 llamalm = dspy.LlamaCpp(model="llama", llama_model=llm,  model_type="chat", temperature=0.4)
+dspy.settings.configure(lm=llamalm)
+
+
+#Define a simple signature for basic question answering
+class BasicQA(dspy.Signature):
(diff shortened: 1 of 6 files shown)


arnavsinghvi11:
Thanks for carrying the LlamaCPP PR forward @AbdulDridi !

It's failing some of the [tests](https://github.com/stanfordnlp/dspy/actions/runs/10207843430/job/28245014846?pr=1347) due to the import error.

Could you instead move the try/except block to [inside the LlamaCPP init method](https://github.com/stanfordnlp/dspy/pull/1347/files#diff-c0d52589d87f25668240d9a7eaa5afa50a2cdd73e4ab77b16aa2c5d6ee5d26e1R26) so it only tries to import the library when dspy.LlamaCpp is being used?
You can then remove the addition of this optional poetry dependency in the `pyproject.toml` as well.

AbdulDridi:
Updated

arnavsinghvi11:
Thanks @AbdulDridi !

shakfu:
@AbdulDridi 

Does this work with llama.cpp [server](https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md) REST interface or does one have to install llama-cpp-python?

PR #1352, opened 2024-08-05
docs(ChromadbRM): change reference to chromadb embeddings

Gwenn-LR:
# Description
Some links in documentation are broken, here is a small PR to update one of them.

# Content
An update of the reference to the page about ChromaDB embeddings.


arnavsinghvi11:
Thanks @Gwenn-LR !

Issue #293, opened 2024-01-23
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'


mneedham:
Hey,

When I try the [Getting Started Notebook](https://github.com/stanfordnlp/dspy/blob/main/intro.ipynb) with Ollama, the first time I run one of the examples it works fine. But the next time it's maybe cached and throws an error around token use?

```python
import dsp.modules.ollama
import dspy

llm = dsp.modules.ollama.OllamaLocal(
    model="mistral"
)

colbertv2_wiki17_abstracts = dspy.ColBERTv2(url='http://20.102.90.50:2017/wiki17_abstracts')

dspy.settings.configure(lm=llm, rm=colbertv2_wiki17_abstracts)

class BasicQA(dspy.Signature):
    """Answer questions with short factoid answers."""

    question = dspy.InputField()
    answer = dspy.OutputField(desc="often between 1 and 5 words")

generate_answer_with_chain_of_thought = dspy.ChainOfThought(BasicQA)
```

First time:

```python
In [31]: generate_answer_with_chain_of_thought(question="What is Python's pandas library?")
[...]

PR #1361, opened 2024-08-06
fix(dspy): prompt_eval_count occasionally is None and causes TypeError

371258:
References https://github.com/stanfordnlp/dspy/issues/293#issuecomment-2270834347

Approach provided and used  by @INF800   (https://github.com/stanfordnlp/dspy/issues/293#issuecomment-1921965485  , https://github.com/stanfordnlp/dspy/pull/325 )


arnavsinghvi11:
Thanks @371258 !

PR #1362, opened 2024-08-06
fix(dspy): save and restore `dsp.settings.trace` in `backtrack_handler()`

### LINKED ISSUE #1356: Bug: Assertion `backtrack_handler()` clears `dsp.settings.trace`, which deletes demo from e.g. `BootstrapFewShot`
Note these `dsp.settings.trace.clear()` calls in `backtrack_handler()`:
https://github.com/stanfordnlp/dspy/blob/6a316ff95e1fb8900c760b8f9ae4cb5056e11a70/dspy/primitives/assertions.py#L211-L221

Also note that `BootstrapFewShot` replies on `dsp.settings.trace` to generate the bootstrapped demos:
https://github.com/stanfordnlp/dspy/blob/6a316ff95e1fb8900c760b8f9ae4cb5056e11a70/dspy/teleprompt/bootstrap.py#L191
https://github.com/stanfordnlp/dspy/blob/6a316ff95e1fb8900c760b8f9ae4cb5056e11a70/dspy/teleprompt/bootstrap.py#L214-L216

Thus, using `backtrack_handler()` would erase the demos during bootstrapping such as in the following example:
```python
import dspy
from dspy.datasets.gsm8k import GSM8K
from dspy.teleprompt import BootstrapFewShot
from dspy.primitives.assertions import assert_transform_module, backtrack_handler
import functools

class TransformedModule(dspy.Module):
    def forward(self):
        pass


class CoT(dspy.Module):
    def __init__(self):
        super().__init__()
        self.prog = dspy.ChainOfThought("question -> answer")

    def forward(self, question):
        result = self.prog(question=question)
        t = assert_transform_module(
            TransformedModule(),
            functools.partial(backtrack_handler, max_backtracks=0),
        )
        # Uncomment the following line to trigger the issue:
        # t()
        return result


# Setup LM here

gsm8k = GSM8K()
gsm8k_trainset, gsm8k_devset = gsm8k.train[:1], gsm8k.dev[:1]
config = dict(max_bootstrapped_demos=4, max_labeled_demos=4)
teleprompter = BootstrapFewShot(**config)
teleprompter.compile(CoT(), trainset=gsm8k_trainset)
print("-" * 10)
print("traces:")
print(teleprompter.name2traces)
print("-" * 10)
```

With the indicated line commented, everything works fine:
```
Bootstrapped 1 full traces after 1 examples in round 0.
----------
traces:
{'prog': [Example({'augmented': True, 'question': "The result from the 40-item Statistics exam Marion and Ella took already came out. Ella got 4 incorrect answers while Marion got 6 more than half the score of Ella. What is Marion's score?", 'rationale': "Reasoning: Let's think step by step in order to find Ella's score first. Since Ella got 4 incorrect answers and there are 40 items in total, Ella's score is 40 - 4 = 36. \n\nNow, let's find Marion's score. Marion got 6 more than half the score of Ella. Half of Ella's score is 36 / 2 = 18. So, Marion got 18 + 6 = 24 more than half the score of Ella. Marion's score is 40 - 24 = 16 incorrect answers. Therefore, Marion's total score is 40 - 16 = 24.", 'answer': '24'}) (input_keys=None)]}
----------
```

However, if you uncomment the line, `BootstrapFewShot` would succeed with empty traces:
```
Bootstrapped 1 full traces after 1 examples in round 0.
----------
traces:
{'prog': []}
----------
```

I think saving `dsp.settings.trace` at the start of `wrapper()` of `backtrack_handler()`, and then prepend it at the end of `wrapper()` might be enough to fix this? I can fix this if it is just this simple, but I think this might not work for multi-threading?

Edit: As #1215 points out, there is a big lock around the entire `wrapper()` code, so I guess no need to worry about multi-threading for now?
https://github.com/stanfordnlp/dspy/blob/6a316ff95e1fb8900c760b8f9ae4cb5056e11a70/dspy/primitives/assertions.py#L189

theta-lin:
This fixes the issue that originally, `backtrack_handler()` clears `dsp.settings.trace`, which is also used by optimizers such as `BootstrapFewShot`.

Also some ruff linting of `tests/predict/test_retry.py`.

Fixes #1356

### DIFF
--- a/dspy/primitives/assertions.py
+++ b/dspy/primitives/assertions.py
@@ -187,6 +187,8 @@ def backtrack_handler(func, bypass_suggest=True, max_backtracks=2):
     def wrapper(*args, **kwargs):
         error_msg, result = None, None
         with dspy.settings.lock:
+            prev_trace = dspy.settings.trace.copy()
+
             dspy.settings.backtrack_to = None
             dspy.settings.suggest_failures = 0
             dspy.settings.assert_failures = 0
@@ -277,6 +279,7 @@ def backtrack_handler(func, bypass_suggest=True, max_backtracks=2):
                                 "UNREACHABLE: No trace available, this should not happen. Is this run time?",
                             )
 
+            dsp.settings.trace = prev_trace + dsp.settings.trace
             return result
 
     return wrapper
--- a/tests/predict/test_retry.py
+++ b/tests/predict/test_retry.py
@@ -2,6 +2,7 @@ import functools
 import dspy
 from dspy.utils import DummyLM
 from dspy.primitives.assertions import assert_transform_module, backtrack_handler
+from dspy.teleprompt import BootstrapFewShot
 import pydantic
 
 
@@ -25,10 +26,7 @@ def test_retry_simple():
 
     print(lm.get_convo(-1))
     assert lm.get_convo(-1).endswith(
-        "Question: What color is the sky?\n\n"
-        "Previous Answer: red\n\n"
-        "Instructions: Try harder\n\n"
-        "Answer: blue"
+        "Question: What color is the sky?\n\n" "Previous Answer: red\n\n" "Instructions: Try harder\n\n" "Answer: blue"
     )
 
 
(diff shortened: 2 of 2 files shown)


okhat:
thank you @theta-lin !  cc @Shangyint is this OK? thoughts?

theta-lin:
Oops, I think I accidentally used a higher version of pytest locally so that I didn't fail the test. I found that just
```
pytest tests/teleprompt/test_mipro_optimizer.py
```
works fine, but
```
pytest tests/predict/test_retry.py tests/teleprompt/test_mipro_optimizer.py
```
fails. Something funky must be going on, possibly related to global states. I would investigate further.

okhat:
Quoting shangyin after internal discussions that we'll "soon update the docs to warn people not to use assertions like this or undefined behavior may happen" but we can't merge this unfortunately @theta-lin because it's a "rare corner case that would require either restriction on how people can use assert_transform (only at the top level), or do significant change for handling traces for assertions".

PR #1368, opened 2024-08-07
fix(dspy): chromadb_rm embedding function inconsistency

### LINKED ISSUE #1366: InvalidDimensionException: Embedding dimension X does not match collection dimensionality Y
# Description
While following the tutorial [`[01] RAG: Retrieval-Augmented Generation`](https://dspy-docs.vercel.app/docs/tutorials/rag), adapted to my system (that is to say a LM hosted and provided by a local Ollama server and a RM locally hosted via Chroma), I faced an issue of consistency between my embeddings with the following error:
```bash
InvalidDimensionException: Embedding dimension 768 does not match collection dimensionality 384
```

# Package version
python: 3.10.12
dspy-ai: 2.4.13
chromadb: 0.4.24

# Exploration
After some exploration, it appears that even when precising the `embedding_function` as a parameter, it is not used during `chromadb.api.client.Client.get_or_create_collection` call, so the `dspy.retrieve.chroma_db.ChromadbRM.ef` attribute defined during the instanciation of the `dspy.Retrieve` object is not the same as `chromadb.api.models.Collection._embedding_function` defined during the initialization of the collection.

So when you create a `dspy.retrieve.chroma_db.ChromadbRM` object with an `embedding_function` attribute and then you populate it with `dspy.retrieve.chroma_db.ChromadbRM._chromadb_collection.add`, you might get a `chromadb.InvalidDimensionException` (depending on your model compared to `chromadb.utils.embedding_functions.DefaultEmbeddingFunction`, a chroma custom embedding function based on [all-MiniLM-L6-v2 exported as an ONNX model](https://github.com/chroma-core/chroma/blob/6d77a86a2c5110ded924c7ad818eb58faf1fa8c5/chromadb/utils/embedding_functions/onnx_mini_lm_l6_v2.py#L29C1-L34C29).

# Code to reproduce
```python
import dspy
from dspy.retrieve.chromadb_rm import ChromadbRM
from chromadb.utils import embedding_functions

embedding_function = embedding_functions.SentenceTransformerEmbeddingFunction(model_name="all-mpnet-base-v2")

rm = ChromadbRM(
    collection_name="embedding-function-issue",
    persist_directory="./.chroma",
    embedding_function=embedding_function
)

assert rm.ef == rm._chromadb_collection._embedding_function
```

# Expected behaviour
That only the indicated `embedding_function` is used in the pipeline.

# Proposed solution
A PR will be proposed soon to tackle this issue and check the behaviour with associated tests.

Gwenn-LR:
# Description
Fix #1366 

# Content
Use the `embedding_function` during the `chromadb.Collection` initialization and define some tests to check this behaviour.

### DIFF
--- a/dspy/retrieve/chromadb_rm.py
+++ b/dspy/retrieve/chromadb_rm.py
@@ -16,20 +16,6 @@ try:
 except Exception:
     ERRORS = (openai.RateLimitError, openai.APIError)
 
-try:
-    import chromadb
-    import chromadb.utils.embedding_functions as ef
-    from chromadb.api.types import (
-        Embeddable,
-        EmbeddingFunction,
-    )
-    from chromadb.config import Settings
-    from chromadb.utils import embedding_functions
-except ImportError:
-    raise ImportError(
-        "The chromadb library is required to use ChromadbRM. Install it with `pip install dspy-ai[chromadb]`",
-    )
-
 
 class ChromadbRM(dspy.Retrieve):
     """
@@ -87,6 +73,19 @@ class ChromadbRM(dspy.Retrieve):
         client: Optional[chromadb.Client] = None,
         k: int = 7,
     ):
+        try:
+            import chromadb
+            import chromadb.utils.embedding_functions as ef
+            from chromadb.api.types import (
+                Embeddable,
+                EmbeddingFunction,
+            )
+            from chromadb.config import Settings
+            from chromadb.utils import embedding_functions
+        except ImportError:
+            raise ImportError(
+                "The chromadb library is required to use ChromadbRM. Install it with `pip install dspy-ai[chromadb]`",
+            )
         self.ef = embedding_function
         self._init_chromadb(collection_name, persist_directory, client=client)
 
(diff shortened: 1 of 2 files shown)


Gwenn-LR:
I've check the failing check and it seems it is only an import error. It comes from an issue in the test environment where the `chromadb` package is not installed, it might be better to ask the CI pipeline to install package marked as `extras`:

https://github.com/stanfordnlp/dspy/blob/af5186cf07ab0b95d5a12690d5f7f90f202bc86e/.github/workflows/run_tests.yml#L58C1-L59C45

You could switch the `run` line for:
```yml
run: poetry install --no-interaction --all-extras
```

You could also add `llama-index` to poetry extras since the associated tests seems to be ignored currently.

If you wanna, I can do another PR but it won't be today.

Have a nice day!

arnavsinghvi11:
Hi @Gwenn-LR , thanks for the PR. 

i added a couple changes to ensure chromadb is only installed when one is using ChromadbRM. However, the test still fails as `test_chromadb_retriever` relies on some chromadb imports. 

As you pointed out, it would be great to follow the [`test_llama_index_rm` setup to skip the tests](https://github.com/stanfordnlp/dspy/blob/main/tests/retrieve/test_llama_index_rm.py) if the corresponding library is not imported. (also would be fine to add some poetry-extras logic but not super necessary imo).

Issue #1367, opened 2024-08-07
Add support for default_headers in OpenAI client


narengogi:
**Description**

OpenAI Client accepts default_headers parameter, this is used to add headers for logging and tracing and authentication through proxies.
Since the wrapper has to OpenAI SDK compliant, DsPy has to support this parameter

**Exploration**

There's two fixes,
1. Create a named argument to the init method for the OpenAI wrapper
2. Initialize the open client with all the arg

**Proposed solution**

Will raise a PR

PR #1369, opened 2024-08-07
Support Default Headers in OpenAI Client

narengogi:
See https://github.com/stanfordnlp/dspy/issues/1367 for information about the issue

**Tests Done**
- [ X ] Initialized OpenAI LM Client and used it (both chat and completions)
-  [ X ] ran pytest

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -56,11 +56,13 @@ class GPT3(LM):
             model_type: Literal["chat", "text"] = None,
             system_prompt: Optional[str] = None,
             http_client: Optional[httpx.Client] = None,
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Thanks @narengogi !

arnavsinghvi11:
for reference, leaving `default_headers` as None works for openai versions < 1.0 so I just [updated the logic ](https://github.com/stanfordnlp/dspy/commit/d3518d7ee9b717e816893d6dc1a875d687f88780)to set it to an empty dictionary.

narengogi:
Thanks @arnavsinghvi11 for adding backsupport for older versions, I assumed that `None` should work looking at the current OpenAI SDK

PR #1372, opened 2024-08-07
Fix assertion bug with new_signature

Shangyint:
Co-authored-by: chiragshah285
@chiragshah285

### DIFF
--- a/dspy/primitives/assertions.py
+++ b/dspy/primitives/assertions.py
@@ -240,7 +240,7 @@ def backtrack_handler(func, bypass_suggest=True, max_backtracks=2):
                                 for i in range(len(dsp.settings.trace) - 1, -1, -1):
                                     trace_element = dsp.settings.trace[i]
                                     mod = trace_element[0]
-                                    if mod.signature == error_target_module:
+                                    if mod == error_target_module:
                                         error_state = e.state[i]
                                         dspy.settings.backtrack_to = mod
                                         break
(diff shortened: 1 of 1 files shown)


Shangyint:
@arnavsinghvi11 Would there be an issue if we change `target_module` to be the actual module in Assertions[854a259](https://github.com/stanfordnlp/dspy/pull/1372/commits/854a259dd23c1c985054f8aa612e915fa2461ab9)? It seems `assert_transform` will re-assign the wrapped `Retry` module to all modules defined in `__init__` thus we shouldn't have the problem?

arnavsinghvi11:
Thanks @chiragshah285 @Shangyint. This change makes sense, but will require some refactoring to existing assertion notebooks examples as the assertion statements currently rely on the `target_module` mapped to corresponding module signature. I believe this logic was kept originally as with 2 assertion statements referring to the same Module type (e.g. ChainOfThought), Retry was not identifying the correct module to backtrack on and hence relied on the signature identifier. 

will require some further testing to verify if the proposed logic correctly maintains this. Feel free to check out an example notebook using assertions ([LongFormQA](https://github.com/stanfordnlp/dspy/blob/main/examples/longformqa/longformqa_assertions.ipynb) - particularly where it uses `dspy.Suggest` on `target_module=GenerateCitedParagraph` while the module includes 2 chain of thought layers, `dspy.ChainOfThought(GenerateSearchQuery)` and `dspy.ChainOfThought(GenerateCitedParagraph)`

okhat:
What's the urgency and status here?

Issue #1376, opened 2024-08-08
optimization: add dedent for __doc__


grant-d:
A small cost optimization
I can PR, but just checking first that if it's a good/bad idea or some reason not to do it.

When `dspy` generates the `__doc__` prompts, I noticed that it likely wastes an amount of (precious) tokens due to leading whitespace.

```python
class PredictSignature(BaseSignature):
    """
    This
    is
    a
    ...
    ...
    ...
    long
    instruction
    """
```

Suggested optimization:

```python
        # signatures/signature.py

        from textwrap import dedent

        ...

        # The more likely case is that the user has just not given us a type.
        # In that case, we should default to the input/output format.
        if cls.__doc__ is None:
            cls.__doc__ = _default_instructions(cls)

        # New code <--------------------------------------------------------
        if cls.__doc__ is not None:
            cls.__doc__ = dedent(cls.__doc__)
        # <-----------------------------------------------------------------

        # Ensure all fields are declared with InputField or OutputField
        cls._validate_fields()
```

Result:
```python
print(llm.inspect_history(n=3))

# before:
"""
This
    is
    a
    ...
    ...
    ...
    long
    instruction
"""

# after:
"""
This
is
a
...
...
...
long
instruction
"""
```

okhat:
Yeah I think that's a good idea, probably also .strip()?

This is technically a "breaking" change in terms of behavior, so it needs to make it to DSPy 2.5 not 2.4.*

PR #1378, opened 2024-08-09
fix(dspy): trim whitespace from doc strings (for 2.5)

grant-d:
Trim whitespace from doc strings
cc @okhat 
See https://github.com/stanfordnlp/dspy/issues/1376

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -1,6 +1,7 @@
 import ast
 import inspect
 import re
+from textwrap import dedent
 import types
 import typing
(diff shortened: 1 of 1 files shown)


grant-d:
@okhat I was not sure if there was a different merge target for `2.5`, lmk and I will change it

okhat:
Thanks @grant-d , I'll make sure this is in 2.5

okhat:
Alright a related change is now making it in the next 1-2 days, thanks a lot @grant-d !!

PR #1383, opened 2024-08-10
feature(dspy): Add helper methods to `Module` and `Predict` classes

dilarasoylu:
This PR introduces several helper methods for the `Module` and `Predict` classes in order to allow for convenient ways to:
1. Check if two modules (e.g. a student and a teacher program) are "structurally_equivalent", which is a precondition for many teleprompters to work properly,
2. Manage and ensure consistency of the LM(s) used by a program, either through `dspy.settings.lm` or predictor LMs.
Shared below are some example usages. These methods are meant to be used by the other, internal DSPy classes.

**Example use of the structural equivalency methods**
```
    # Ensure that the student program and the teacher program have the same program structure
    student._assert_structural_equivalency(teacher)

    # Ensure that the predictors of the programs point to differenT objects
    assert not student._assert_no_shared_predictor(teacher)
```

**Example use of the Module level LM methods**
```
    # Ensure that the LM consistency is satisfied, push the global LM to the predictors if dspy.settings.lm is set
    program._assert_lm_consistency()
    if dspy.settings.lm:
        program._set_all_predictor_lms(dspy.settings.lm)
```

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -1,11 +1,12 @@
+from __future__ import annotations  # Required for type hints of the class itself, see Predict._is_structurally_equivalent
 import random
+from typing import Optional
 
 from pydantic import BaseModel
 
 import dsp
 from dspy.predict.parameter import Parameter
 from dspy.primitives.program import Module
-
 from dspy.primitives.prediction import Prediction
 from dspy.signatures.signature import ensure_signature, signature_to_template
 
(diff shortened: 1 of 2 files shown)


dspy/predict/predict.py
```diff
@@ -17,6 +17,29 @@ def __init__(self, signature, **config):
         self.config = config
         self.reset()
 
+    def _is_structurally_equivalent(self, other) -> bool:
```
arnavsinghvi11, line 20: return type only has bool, while some conditions return err_msg as well

dspy/predict/predict.py
```diff
@@ -17,6 +17,29 @@ def __init__(self, signature, **config):
...
+        
+        Args:
+            other: The predictor to compare with.
+        
+        Returns:
+            bool: `True` if the predictors are structurally equivalent, `False` otherwise.
+        """
+        # Check if the two predictors are instances of the Predictor class
+        if not isinstance(other, self.__class__):
+            return False
```
arnavsinghvi11, line 31: same comment as above

dspy/predict/predict.py
```diff
@@ -17,6 +17,29 @@ def __init__(self, signature, **config):
...
+
+        # Check if the predictors have matching signatures
+        # TODO: Can we ensure that all the signatures implement the equals method?
```
arnavsinghvi11, line 34: good call here! I believe this is valid for all Signatures inheriting from [SignatureMeta](https://github.com/stanfordnlp/dspy/blob/d3518d7ee9b717e816893d6dc1a875d687f88780/dspy/signatures/signature.py#L209C1-L209C53) (both longform signatures and short ones (input -> output)), which has a [equals method defined](https://github.com/stanfordnlp/dspy/blob/d3518d7ee9b717e816893d6dc1a875d687f88780/dspy/signatures/signature.py#L162)

this likely means L36:L41 can be simplified

dspy/primitives/program.py
```diff
@@ -25,6 +30,99 @@ def __init__(self):
...
+
+    def _has_shared_predictor(self, other: Module) -> bool:
+        """ Check if two modules share a predictor that points to the same Python object.
+        
+        Args:
+            other: The module to compare with.
+        
+        Returns:
+            bool: `True` if the modules share a predictor, `False` otherwise.
+        """
```
arnavsinghvi11, line 65: would be good to add a similar check as L46-L48

arnavsinghvi11 (approved):

dspy/primitives/program.py
dilarasoylu: Ah, this is a great catch, I'll modify the check to ensure there is no predictor overlap irrespective of the order of the predictor...

dspy/predict/predict.py
dilarasoylu: Updated the method to be an assertion method instead.

dspy/predict/predict.py
dilarasoylu: [updated]

dspy/predict/predict.py
dilarasoylu: Updated

dspy/primitives/program.py
dilarasoylu: Updated

Issue #1386, opened 2024-08-12
Error in trying to save compiled dspy program using optimized_program.save(YOUR_SAVE_PATH)


vikrantrathore:
Error in optimized program save. Using the given example of CompiledBaleen from https://github.com/stanfordnlp/dspy/blob/d3518d7ee9b717e816893d6dc1a875d687f88780/docs/docs/tutorials/simplified-baleen.md?plain=1#L186

Trying to save it with following code:

```
def validate_context_and_answer_and_hops(example, pred, trace=None):
    if not dspy.evaluate.answer_exact_match(example, pred): return False
    if not dspy.evaluate.answer_passage_match(example, pred): return False

    hops = [example.question] + [outputs.query for *_, outputs in trace if 'query' in outputs]

    if max([len(h) for h in hops]) > 100: return False
    if any(dspy.evaluate.answer_exact_match_str(hops[idx], hops[:idx], frac=0.8) for idx in range(2, len(hops))): return False

    return True

from dspy.teleprompt import BootstrapFewShot

teleprompter = BootstrapFewShot(metric=validate_context_and_answer_and_hops)
compiled_baleen = teleprompter.compile(SimplifiedBaleen(), teacher=SimplifiedBaleen(passages_per_hop=2), trainset=trainset)
compiled_baleen.save(path="multihop_qa.json")
```

This results in an error in dump_state in https://github.com/stanfordnlp/dspy/blob/d3518d7ee9b717e816893d6dc1a875d687f88780/dspy/primitives/module.py#L117

Following is the error message:

> Bootstrapped 1 full traces after 4 examples in round 0.
> [('generate_query[0]', Predict(StringSignature(context, question -> rationale, query
>     instructions='Write a simple search query that will help answer a complex question.'
>     context = Field(annotation=str required=True json_schema_extra={'desc': 'may contain relevant facts', '__dspy_field_type': 'input', 'prefix': 'Context:'})
>     question = Field(annotation=str required=True json_schema_extra={'__dspy_field_type': 'input', 'prefix': 'Question:', 'desc': '${question}'})
>     rationale = Field(annotation=str required=True json_schema_extra={'prefix': "Reasoning: Let's think step by step in order to", 'desc': '${produce the query}. We ...', '__dspy_field_type': 'output'})
>     query = Field(annotation=str required=True json_schema_extra={'__dspy_field_type': 'output', 'prefix': 'Query:', 'desc': '${query}'})
[...]

PR #1388, opened 2024-08-12
Debugging Issue #1386

CShorten:
Adds save_field_meta default to `Retrieve`, debugging issue #1386.

### DIFF
--- a/docs/docs/building-blocks/6-optimizers.md
+++ b/docs/docs/building-blocks/6-optimizers.md
@@ -122,7 +122,7 @@ optimized_program.save(YOUR_SAVE_PATH)
 
 The resulting file is in plain-text JSON format. It contains all the parameters and steps in the source program. You can always read it and see what the optimizer generated.
 
-You can add `save_field_meta` to additionally save the list of fields with the keys, `name`, `field_type`, `description`, and `prefix` with: `optimized_program.save(YOUR_SAVE_PATH, save_field_meta=True).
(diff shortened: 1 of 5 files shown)


isaacbmiller:
There is also `RetrieveThenRerank` in retrieve.py

isaacbmiller:
Would it be possible to add a test that runs this code and maybe checks that the field info that you want is properly saved?
```python
import dspy

class DumpStateRepro(dspy.Module):
    def __init__(self):
        self.retrieve = dspy.Retrieve(k=1)
        self.predictor = dspy.Predict("question -> answer")

answer_generator = DumpStateRepro()

answer_generator.save("DumpStateRepro", save_field_meta=True)
answer_generator.save("DumpStateRepro", save_field_meta=False)
```

CShorten:
Ok, I think this will fix it, and I like the `save_verbose` naming going forward for consistency / future work on how much optimized program state should be serialized. @isaacbmiller

isaacbmiller:
I do not have time to look at it today. Just merging is fine then

CShorten:
Ok cool, maybe we just add this to `/tests/modules/test_module.py` ?

```python
import dspy

def test_save_method():
    class DumpStateRepro(dspy.Module):
        def __init__(self):
            self.retrieve = dspy.Retrieve(k=1)
            self.predictor = dspy.Predict("question -> answer")

    answer_generator = DumpStateRepro()
    
    answer_generator.save("DumpStateRepro", save_field_meta=True)
    answer_generator.save("DumpStateRepro", save_field_meta=False)
```

PR #1398, opened 2024-08-15
Fix (dspy) Remove extra argument passed to Cohere client

jalammar:
Simple fix to an issue where an exception is thrown when using the Cohere LLM (for prediction, with experimental=True). It was due to an extra parameter (`stop`) being passed to the API. This change checks for it and removes it.

### DIFF
--- a/dsp/modules/cohere.py
+++ b/dsp/modules/cohere.py
@@ -87,8 +87,8 @@ class Cohere(LM):
         if "n" in kwargs.keys():
             kwargs.pop("n")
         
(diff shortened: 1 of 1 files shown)


jalammar:
Regarding the failing "Lint, Test, and Build / Check Ruff Fix (pull_request)" check, running the command it requests only applies changes to pre-existing jupyter notebooks in the repo, not in the file this PR edits.

arnavsinghvi11:
Thanks @jalammar ! will check ruff post-merge

PR #1406, opened 2024-08-18
fix lm reference

haalcala:
fix lm reference when using custom llm


okhat:
Thanks @haalcala ! I think the original here is intentional for legacy mode, which will be deprecated within 1-2 weeks.

When `self.lm` is None (which it usually is), `old_generate` relies on dspy.settings.lm.

Issue #1349, opened 2024-08-02
Pydantic ValidationError in functional._prepare_signature() when using Optionals in Signature 


371258:
https://github.com/stanfordnlp/dspy/blob/6a316ff95e1fb8900c760b8f9ae4cb5056e11a70/dspy/functional/functional.py#L241C1-L245C23

I am facing an issue when using Optionals in my signature. It seems that functional._prepare_signature()  does not recognize a subclass of pydantic.BaseModel once it is wrapped in an Optional. I added a minimal example to explain and reproduce the issue: 

https://github.com/371258/dspy_optional_issue/blob/main/issue.ipynb

Is there something wrong with my approach to declare the signature? 

dspy-ai:  2.4.13
dspy: 0.1.5
pydantic: 2.8.2

PR #1413, opened 2024-08-21
fix(dspy): Allow `Union` with pydantic objects in `TypedPredictor` 

LLukas22:
See  #1349

This uses pydantics [`TypeAdapter`](https://docs.pydantic.dev/latest/api/type_adapter/) to wrap objects into pydantic serializable instances.

This also eliminates the need to prepend single values with a value field in the JSON schema,, e.g. a field of type `list[int]` will now be expressed as `[1,2,3]` instead of `{"value": [1,2,3]}`. 

I also added tests for the most common field types to ensure that they are deserialized properly.


arnavsinghvi11:
Thanks @LLukas22 !

PR #1414, opened 2024-08-21
feature(dspy): refined missing hf error

AlejandroBaron:
The error doesn't specify that you need torch (as `pip install transformers` is not the same as `pip install transformers[torch]`. This makes things a bit clearer


arnavsinghvi11:
Thanks @AlejandroBaron ! added the import message to make it even more clear :)

PR #1421, opened 2024-08-22
feat(dspy): Add input for AWS credentials for AWSProvider, Bedrock and Sagemaker class

joanacmesquitaf:
This pull request introduces enhancements to the **AWSProvider**, **Bedrock**, and **Sagemaker** classes by **adding the capability to input AWS credentials directly**. This update provides greater flexibility beyond using the a profile within the AWS credentials file.

**Motivation:**
The primary goal of this update is to **increase flexibility in managing AWS credentials**.

### DIFF
--- a/dsp/modules/aws_providers.py
+++ b/dsp/modules/aws_providers.py
@@ -1,17 +1,22 @@
 """AWS providers for LMs."""
 
+import importlib.util
 from abc import ABC, abstractmethod
 from typing import Any, Optional
 
 import backoff
+
 from dsp.utils.settings import settings
 
-try:
-    import boto3
-    from botocore.exceptions import ClientError
-    ERRORS = (ClientError,)
+# Check if 'boto3' is available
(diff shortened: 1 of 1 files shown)


dsp/modules/aws_providers.py
```diff
@@ -1,17 +1,20 @@
...
 from abc import ABC, abstractmethod
 from typing import Any, Optional
 
 import backoff
+
 from dsp.utils.settings import settings
 
-try:
-    import boto3
+boto3_spec = importlib.util.find_spec("boto3")
```
arnavsinghvi11, line 11: I think the previous try-except logic suffices? fine to merge after reverting back this logic

joanacmesquitaf: Hello @arnavsinghvi11. I reversed the logic as you suggested but it was necessary to remove the boto3 import here because ruff was failing. We are already importing boto3 in the AWSProvider class constructor.

PR #1432, opened 2024-08-27
Fix AttributeError in MilvusRM: Add missing attribute 'k'

arya-krishna-23:
Problem: The MilvusRM object in our codebase is throwing an AttributeError due to a missing attribute k. This error occurs when attempting to access or use k, which has not been defined within the MilvusRM class.
Solution:

Replaced the existing attribute top_k with k.
error : AttributeError: 'MilvusRM' object has no attribute 'k'


arnavsinghvi11:
Thanks @arya-krishna-23 !

PR #1443, opened 2024-08-30
Rename teleprompt to optimizers for consistency (and reduce confusion)

chenmoneygithub:
Renaming teleprompt to optimizers to be consistent with the documentation. Keeping `teleprompt` directory as a router for backward compatibility.


chenmoneygithub:
@omkar-sh I am closing this PR for now to reduce noise, would like to discuss further about naming and consistency!

PR #1446, opened 2024-09-02
feature(dspy): adding spBLEU and cosine similarity as new metrics

fsndzomga:
Added support for spBLEU (SentencePiece BLEU) and cosine similarity as new evaluation metrics in the dspy module to enhance text similarity and performance analysis.

### DIFF
--- a/dspy/evaluate/metrics.py
+++ b/dspy/evaluate/metrics.py
@@ -2,10 +2,31 @@
 
 import dsp
 
+try:
+    import sacrebleu
+except ImportError:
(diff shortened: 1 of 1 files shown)


CyrusNuevoDia:
This seems cool! Can you move the imports into the functions directly? import torch adds a lot to startup time.

fsndzomga:
Yes, sure. That is a good idea indeed.

fsndzomga:
@CyrusNuevoDia done ! and I added some tests too

PR #1450, opened 2024-09-03
feat: adds baidu ernie modules

juniusxu:


okhat:
Thanks a lot! I think this will not be necessary after DSPy 2.5? Please lmk!

https://github.com/stanfordnlp/dspy/blob/main/examples/migration.ipynb

PR #1451, opened 2024-09-03
refactor(dspy): updating snowflake LM and RM implementations

sfc-gh-alherrera:
Refactoring the Snowflake integration:

- Authentication: now takes authenticated session object instead of user credentials
- LM Models: support for additional models and fine-tuned models by removing strict error handling
- RM Internals: updated RM implementation to use Cortex Search hybrid retrieval service instead of basic similarity functions


arnavsinghvi11:
Thanks @sfc-gh-alherrera !

Issue #1295, opened 2024-07-19
Miprov2 doesn't work with TypedPredictors


eelang-sc:
https://github.com/stanfordnlp/dspy/blob/d8b8909773fc31e72cec093db2f26109590e524e/dspy/propose/grounded_proposer.py#L198

As the TODO line suggests it doesn't work with typed predictors.

a quick fix for me was to change
```
pattern = r"^(.*dspy\.(ChainOfThought|Predict).*)$"  # TODO: make it so that this extends out to any dspy Module
```
to
```
pattern = r"^(.*dspy\.(ChainOfThought|Predict|TypedPredictor).*)$"  # TODO: make it so that this extends out to any dspy Module
```

okhat:
good to get this on the radar of @XenonMolecule

haznai:
Proposed a trivial fix for this suggested change in #1453 


Thanks for your time and effort, people :)

XenonMolecule:
Thanks @haznai this looks great!  I think this should work for ReAct and ProgramOfThought too

PR #1453, opened 2024-09-04
Improve MIPROv2 program aware regex

haznai:
Now supports typed variants of `Predict` and `ChainOfThought` as well.

First mentioned in #1295


XenonMolecule:
Thanks @haznai This looks great!  I think this should work for ReAct and ProgramOfThought too, perhaps we can just patch those in now as well!

okhat:
Will this be enough for TypedPredictors to work? cc @XenonMolecule 

We should test it

okhat:
cc @XenonMolecule

okhat:
Closing in favor of a bunch of newer updates to MIPRO. Thanks a lot for spotting this @haznai !

Issue #788, opened 2024-04-08
Enhance error handling


kylerush:
Absolutely love the work on DSPY. Thank you to everyone for your effort.

I wanted to note that I think the error handing can be improved. It's not very clear most of the time what the error is. For example, if you forget the following:

```
turbo = dspy.OpenAI(model="gpt-3.5-turbo-0125", api_key=os.environ["OPENAI_API_KEY"])
dspy.settings.configure(lm=turbo)
```

You get this error:
```
Failed to run or to evaluate example [EXAMPLE] with <function validate_answer at 0xffff83f0c5e0> due to 'NoneType' object has no attribute 'kwargs'.
```

This led me down many hours of a wild goose chase until I realized I forgot to configure the LLM.

I'd be happy to look into how this specific Exception is handled and maybe suggest a different approach, but just wanted to note this here and get some reactions.

Thanks again!

arnavsinghvi11:
Thanks @kylerush ! Currently, there is a check for if the LM is configured [predict](https://github.com/stanfordnlp/dspy/blob/696f2d2e1f96f173abb36302c71a725d564dfadb/dsp/primitives/predict.py#L59) which ensures the validation before any model generations. 

But it's actually not getting hit since this error comes when you are using the `dspy.BootstrapFewShot` optimizer and the LM is being set for the [teacher](https://github.com/stanfordnlp/dspy/blob/696f2d2e1f96f173abb36302c71a725d564dfadb/dspy/teleprompt/bootstrap.py#L148). When tries to [copy the kwargs](https://github.com/stanfordnlp/dspy/blob/696f2d2e1f96f173abb36302c71a725d564dfadb/dspy/teleprompt/bootstrap.py#L148), it [triggers the error](https://github.com/stanfordnlp/dspy/blob/696f2d2e1f96f173abb36302c71a725d564dfadb/dsp/modules/lm.py#L107) you got `'NoneType' object has no attribute 'kwargs'`

I believe this can be simply resolved with the same check in predict, but we can add to the lm.copy method to make it a universal check, ensuring that an LM is loaded when a) copying the LM object or b) creating generations with it (already handled).

```
if not dsp.settings.lm:
            raise AssertionError("No LM is loaded.")
```

lmk if you had anything else in mind for a broader Exception handling before we push out a PR for this. thanks for the wild goose chase!!

kylerush:
[...]

PR #1458, opened 2024-09-06
Improve error logging with a stacktrace

severi:
Related to issue https://github.com/stanfordnlp/dspy/issues/788

This prints out the example inputs and a stacktrace for the error that occurred. I understand that this change would add more excessive error messages that you have tried to avoid

> The current behavior was designed to limit excessive outputted error messages, partly handled through [how many errors can occur before completely halting execution](https://github.com/stanfordnlp/dspy/blob/5c5e8071916688f462b3b779e0c6adca42f660b8/dspy/evaluate/evaluate.py#L140)

But at least to me this has been the number 1 issue I have been experiencing when testing out dspy for the past week. This change works like a charm in my use case.

If the printed error message seems too excessive with large datasets, then another option would be to add the stacktrace logging behind an env variable that can be set if needed (though my personal preference would be to have the stacktrace visible as the default option and having the possibility to limit the error traces with a similar approach)

```
 if os.environ.get('DEBUG') == 'true':
     dspy.logger.error(f"Error for example in dev set: \t\t {e}\n\twith inputs:\n\t\t{example.inputs()}\n\nStack trace:\n\t{traceback.format_exc()}")
```

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -4,6 +4,7 @@ import sys
 import threading
 import types
 from typing import Any
+import traceback
 
 import pandas as pd
 import tqdm
@@ -53,6 +54,7 @@ class Evaluate:
         max_errors=5,
         return_all_scores=False,
         return_outputs=False,
+        provide_traceback=False,
         **_kwargs,
     ):
         self.devset = devset
@@ -66,7 +68,7 @@ class Evaluate:
         self.cancel_jobs = threading.Event()
         self.return_all_scores = return_all_scores
         self.return_outputs = return_outputs
-
+        self.provide_traceback = provide_traceback
         if "display" in _kwargs:
             dspy.logger.warning(
                 "DeprecationWarning: 'display' has been deprecated. To see all information for debugging,"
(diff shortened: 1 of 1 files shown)


krypticmouse:
Hi @severi 

Thanks for the contribution! Is it possible to make this configurable? 

I think we can keep a variable to define how verbose the traceback should be, something like:
```
class Evaluate:
    def __init__(
        self,
        *,
        devset,
        metric=None,
        num_threads=1,
        display_progress=False,
        display_table=False,
        max_errors=5,
        return_all_scores=False,
        return_outputs=False,
        provide_traceback = False
        **_kwargs,
    ):
        self.devset = devset
        ...
        self.return_outputs = return_outputs
        self.provide_traceback = provide_traceback
        
    def evaluate(...):
        ....
        if self.provide_traceback:
            dspy.logger.error(f"Error for example in dev set: \t\t {e}\n\twith inputs:\n\t\t{example.inputs()}\n\nStack trace:\n\t{traceback.format_exc()}")
        else:
            dspy.logger.error(f"Error for example in dev set: \t\t {e}")
```

krypticmouse:
Also can you resolve the conflicts? I think is just import merging from 2 branches that's all :slightly_smiling_face:

krypticmouse (changes requested):
Some changes needed but mostly looks good!

severi:
Hey! Implemented the requested changes + fixed the merge conflict

severi:
We could also modify the original error message to something like

```
dspy.logger.error(f"Error for example in dev set: \t\t {e}. Set `provide_traceback=True` to see the stack trace.")
```

to make it more visible such an option exists

krypticmouse:
> We could also modify the original error message to something like
> 
> ```
> dspy.logger.error(f"Error for example in dev set: \t\t {e}. Set `provide_traceback=True` to see the stack trace.")
> ```
> 
> to make it more visible such an option exists

I like this better, can we add this?

severi:
> > We could also modify the original error message to something like
> > ```
> > dspy.logger.error(f"Error for example in dev set: \t\t {e}. Set `provide_traceback=True` to see the stack trace.")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to make it more visible such an option exists
> 
> I like this better, can we add this?

done

okhat:
Very cool! Thank you @severi , @krypticmouse !

krypticmouse:
LGTM :ship:

PR #1484, opened 2024-09-12
Fix AzureAISearchRM Error when doing vector/hybrid search

jeffreyrubi:
In hybrid/vector search case, 
when calling client.search, get_embedding() is returning List[VectorQueries]: 
https://github.com/stanfordnlp/dspy/blob/main/dspy/retrieve/azureaisearch_rm.py#L380

It can directly passed to the client.search.

Currently, an extra '[]' is added when passing the vector_query to the client.search:
https://github.com/stanfordnlp/dspy/blob/main/dspy/retrieve/azureaisearch_rm.py#L258

Results in serialization error from AISearch library.

The fix is to remove that extra []. Test is done with positive result after fix.


arnavsinghvi11:
Thanks @jeffreyrubi !

PR #1485, opened 2024-09-12
Add DSPy+finetuning blog post to use cases

rawsh:
Add https://raw.sh/posts/chess_puzzles


okhat:
wasn't it there before?

rawsh:
hmm don't think so. there's a medium post on playing chess but they don't do any optimization.

PR #1490, opened 2024-09-13
fix(evaluate): fix evaluate dict merge for pydantic

MSNP1381:
it had error when your module's output was a pydantic object rather that a dict

### DIFF
--- a/dsp/modules/gpt3.py
+++ b/dsp/modules/gpt3.py
@@ -1,34 +1,43 @@
 import functools
+import importlib.util
 import json
 import logging
 from typing import Any, Literal, Optional, cast
+
 import backoff
 import httpx
-try:
+
+from dsp.modules.cache_utils import CacheMemory, NotebookCacheMemory, cache_turn_on
+from dsp.modules.lm import LM
+from dsp.utils.settings import settings
+
+
+def is_module_installed(module_name):
+    return importlib.util.find_spec(module_name) is not None
+
+
+if is_module_installed("langfuse") and is_module_installed("langfuse.openai"):
     """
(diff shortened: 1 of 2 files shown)


dsp/modules/gpt3.py
```diff
@@ -1,34 +1,44 @@
...
 import backoff
 import httpx
-try:
+
+from dsp.modules.cache_utils import CacheMemory, NotebookCacheMemory, cache_turn_on
+from dsp.modules.lm import LM
+from dsp.utils.settings import settings
+
+
+def is_module_installed(module_name):
```
arnavsinghvi11, line 16: just curious why this is needed. does the try/except block not take care of this check if langfuse is not supported?

dsp/modules/gpt3.py
```diff
@@ -111,6 +122,11 @@ def __init__(
     def _openai_client(self):
         return openai
 
+    # TODO: add test suite to me.
```
arnavsinghvi11, line 125: remove extra code

arnavsinghvi11 (changes requested):
Thanks @MSNP1381 ! left some minor comments to avoid extraneous changes to`gpt3.py` but the changes to `evaluate` LGTM

MSNP1381:
> Thanks @MSNP1381 ! left some minor comments to avoid extraneous changes to`gpt3.py` but the changes to `evaluate` LGTM

i just added the code comments but the merge function is unstable be cause there is varying type of output for modules. My opinion is either we just use pydantic Object or any inherited objects or user have to follow some basic interfaces (as simple as dict interface))

MSNP1381:
@okhat can i talk to you through the email?

PR #1502, opened 2024-09-17
Typo/nit fixes and add tests for BoostrapWithRandomSearch

chenmoneygithub:
Fix a few odds in the BootstrapFewShotWithRandomSearch optimizer:
- Zombie code should be deleted. 
- f-string for print/logging.
- variable names like `program2` are confusing.

Also added basic unit test case for BootstrapFewShotWithRandomSearch.

### DIFF
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -125,15 +125,12 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
                 score = 0 if program._assert_failures > 0 else score
             ######################################################
 
-            print("Score:", score, "for set:", [len(predictor.demos) for predictor in program.predictors()])
-
             if len(scores) == 0 or score > max(scores):
                 print("New best score:", score, "for seed", seed)
                 best_program = program
 
             scores.append(score)
-            print(f"Scores so far: {scores}.")
-            print(f"Best score: {max(scores)}")
+            print(f"Best score so far: {max(scores)}")
 
             score_data.append((score, subscores, seed, program))
 
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -137,16 +137,6 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
 
             score_data.append((score, subscores, seed, program))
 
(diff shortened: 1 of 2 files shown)


dspy/teleprompt/random_search.py
```diff
@@ -48,16 +48,10 @@ def __init__(
         self.max_num_samples = max_bootstrapped_demos
         self.max_errors = max_errors
         self.num_candidate_sets = num_candidate_programs
-        # self.max_num_traces = 1 + int(max_bootstrapped_demos / 2.0 * self.num_candidate_sets)
```
arnavsinghvi11, line 51: comment back in

dspy/teleprompt/random_search.py
```diff
@@ -120,39 +114,28 @@ def compile(self, student, *, teacher=None, trainset, valset=None, restrict=None
...
+            print("Score:", score, "for set:", [len(predictor.demos) for predictor in program.predictors()])
 
             if len(scores) == 0 or score > max(scores):
-                print("New best sscore:", score, "for seed", seed)
-                best_program = program2
+                print("New best score:", score, "for seed", seed)
+                best_program = program
 
             scores.append(score)
-            print(f"Scores so far: {scores}")
```
arnavsinghvi11, line 141: comment back in

arnavsinghvi11 (changes requested):
Hi @chenmoneygithub , let's leave the commented code / variable names as-is for now as we make decisions on how to integrate this going forward. 

The test for RandomSearch is a bit trivial - I would recommend following how the test is setup for `BootstrapFewShot` [here](https://github.com/stanfordnlp/dspy/blob/main/tests/teleprompt/test_bootstrap.py) and adding the validations of multiple optimization program candidates in this test. Let me know if that makes sense

chenmoneygithub:
@arnavsinghvi11 For a public repo with big user base, there shouldn't be any zombie code. Three reasons: 
- Most likely commented code won't be revived in its lifetime.
- Commented code becomes invalid after a short while because it's not executed.
- Only a few people have the context, for other people it's not meaningful.

if something is worth further discussion opening an issue for tracking is a better approach.

PR #1519, opened 2024-09-20
Fix dspy Module's deepcopy

chenmoneygithub:
The current solution has issues if `dspy.BaseModule` contains non-copyable instances. The proposed approach is:

1. Get all attributes, if it is a `BaseModule`, try deep copying, if failed then do recursive call.
2. If the attr is not a `BaseModule`, try deep copying, if failed try shallow copying, if still failed just copying over the reference. 
3. It won't hit infinite recursion because once a module has no attributes that is a `BaseModule`, the recursion stops.

Added unit tests to test it.

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -102,15 +102,50 @@ class BaseModule:
         return [param for _, param in self.named_parameters()]
 
     def deepcopy(self):
-        return copy.deepcopy(self)
+        """Deep copy the module.
+
+        This is a tweak to the default python deepcopy that only deep copies `self.parameters()`, and for other
+        attributes, we just do the shallow copy.
+        """
+        try:
+            # If the instance itself is copyable, we can just deep copy it.
+            # Otherwise we will have to create a new instance and copy over the attributes one by one.
+            return copy.deepcopy(self)
+        except Exception:
+            pass
+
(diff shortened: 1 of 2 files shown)


dspy/primitives/module.py
```diff
@@ -102,15 +102,50 @@ def parameters(self):
...
+        except Exception:
+            pass
+
+        # Create an empty instance.
+        new_instance = self.__class__.__new__(self.__class__)
+        # Set attribuetes of the copied instance.
+        for attr, value in self.__dict__.items():
+            if isinstance(value, BaseModule):
+                try:
+                    # Try deep copying the module to reduce recursion depth.
```
okhat, line 123: Is this needed?

chenmoneygithub: good catch, I should remove this part.

okhat:
This ... is probably OK? Do we want to issue a warning on the first deepcopy that faces trouble and tell the user about it? I worry about silent bugs. People do expect deep copy to give them a real deep copy, but maybe we can explain that it only guarantees a deep copy of DSPy-related attributes.

chenmoneygithub:
@okhat  Yea I think that would be fine, but I will keep monitoring if this is causing trouble for users.

PR #1538, opened 2024-09-24
litellm.py : disable telemetry

bluewin4:
Uses the "LITELLM_LOCAL_MODEL_COST_MAP" variable to turn off an API call when liteLLM is imported. 
Set 'litellm.telemetry = False' as well. 

https://docs.litellm.ai/docs/completion/token_usage#9-register_model
https://github.com/BerriAI/litellm/blob/main/litellm/__init__.py

### DIFF
--- a/dsp/modules/__init__.py
+++ b/dsp/modules/__init__.py
@@ -30,3 +30,4 @@ from .snowflake import *
 from .tensorrt_llm import TensorRTModel
 from .watsonx import *
 from .you import You
+from .litellm import *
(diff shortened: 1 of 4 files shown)


okhat:
Thanks a lot @bluewin4 , could we fix this:

> Ruff fix failed, failing the workflow.
> Please run 'ruff check . --fix-only' locally and push the changes.

okhat:
Hmm hey can you re-create the PR into main, not into the litellm branch which is super outdated? I tried to change it here but it's creating further problems. Also do run ruff if you get a chance. Thanks a lot @bluewin4 !

PR #1548, opened 2024-09-26
Fixes #1534: 'Name' object has no attribute 'ops' error

### LINKED ISSUE #1534: Error: 'Name' object has no attribute 'ops'
When using ProgramOfThought and code is generated with named conditionals like this:

```python
Previous Code:
from datetime import datetime

data = {
    "ExpenseReport": "002",
    "ExpenseAmount": "USD30.0",
    "AuthorizationLevel": "94",
    "ExpenseDate": "2024-01-17",
    "SubmittedDate": "2024-02-14",
    "WorkLocation": "US",
    "ExpenseCategory": "hotel",
    "ReceiptAttached": "yes"
}

# Convert amounts to float for comparison
expense_amount = float(data["ExpenseAmount"].replace("USD", ""))
authorization_level = int(data["AuthorizationLevel"])

# Convert dates to datetime objects
expense_date = datetime.strptime(data["ExpenseDate"], "%Y-%m-%d")
submitted_date = datetime.strptime(data["SubmittedDate"], "%Y-%m-%d")
days_difference = (submitted_date - expense_date).days

# Check each policy
policy_1_violated = (expense_amount > 100.00 and authorization_level < 100) or (expense_amount > 25000.00)
policy_2_violated = ((submitted_date.weekday() >= 5) or (days_difference > 30 and data["WorkLocation"] == "US") or (days_difference > 60 and data["WorkLocation"] == "Europe"))
policy_3_violated = (data["ExpenseCategory"] == "hotel" and data["ReceiptAttached"].lower() != "yes")

# Determine status_reason
if policy_1_violated:
    status_reason = "Policy 1 Violated"
elif policy_2_violated:
    status_reason = "Policy 2 Violated"
elif policy_3_violated:
    status_reason = "Policy 3 Violated"
else:
    status_reason = "Approved"

status_reason
```
The following error message is generated.
`Error: 'Name' object has no attribute 'ops'`

This can be fixed by adding the 'elif isinstance(condition, ast.Name):' statement in `python_interpreter.py` as shown below.

```python
    def _execute_condition(self, condition):
        if isinstance(condition, ast.BoolOp):
            if isinstance(condition.op, ast.And):
                results = [self._execute_ast(value) for value in condition.values]
                return all(results)
            elif isinstance(condition.op, ast.Or):
                results = [self._execute_ast(value) for value in condition.values]
                return any(results)
            else: #TODO - add any other BoolOps missing
                raise InterpreterError(f"Boolean operator {condition.op} is not supported")
        elif isinstance(condition, ast.Name):
            return self._execute_name(name=condition)
```

arthurgreef:
This pull request fixes issue #1534 that causes an exception in the Python interpreter when a named object is used in a conditional expression.

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -357,6 +357,8 @@ class PythonInterpreter:
                 return any(results)
             else: #TODO - add any other BoolOps missing
                 raise InterpreterError(f"Boolean operator {condition.op} is not supported")
+        elif isinstance(condition, ast.Name):
+            return self._execute_name(name=condition)
         elif isinstance(condition, ast.Compare):
             if len(condition.ops) > 1:
                 raise InterpreterError("Cannot evaluate conditions with multiple operators")
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -188,6 +188,8 @@ class PythonInterpreter:
             return self._execute_call(expression)
         elif isinstance(expression, ast.Compare):
             return self._execute_condition(expression)
+        elif isinstance(expression, ast.BoolOp):
+            return self._execute_condition(expression)
         elif isinstance(expression, ast.Constant):
             # Constant -> just return the value
             return expression.value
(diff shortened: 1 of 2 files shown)


arnavsinghvi11:
hey @arthurgreef , does #1550 fix this? Just pushed out a PR to cover this more comprehensively.

PR #1553, opened 2024-09-27
support TEI

xiayouran:
support TEI(Text Embeddings Inference),     For detailed information, visit: [https://github.com/huggingface/text-embeddings-inference](https://github.com/huggingface/text-embeddings-inference).


okhat:
Thanks @xiayouran ! I'm adding Arnav who has been thinking about vectorizers. I suspect this part can be handled by LiteLLM for us?

xiayouran:
Sorry, does that mean we should use `litellm.embedding` in place of `requests.post`? @okhat

PR #1555, opened 2024-09-27
fix typo in lm.py

IgorKhramtsov:
Looks like there is a typo in the name of env variable, in the documentation it is referenced as `DSP_CACHEDIR` [faqs.md](https://github.com/stanfordnlp/dspy/blob/b3ecf879cc18d6c15a2b8654d217cc34a66489fe/docs/docs/faqs.md?plain=1#L90)

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -13,7 +13,7 @@ try:
         litellm.telemetry = False
 
     from litellm.caching import Cache
-    disk_cache_dir = os.environ.get('DSPY_CACHEDIR') or os.path.join(Path.home(), '.dspy_cache')
+    disk_cache_dir = os.environ.get('DSP_CACHEDIR') or os.path.join(Path.home(), '.dspy_cache')
     litellm.cache = Cache(disk_cache_dir=disk_cache_dir, type="disk")
 
 except ImportError:


okhat:
Hey @IgorKhramtsov, awesome to see a PR again from you. I guess we're migrating, so keeping it as DSP would be strange hmm. Maybe we should accept both for now in the code?

IgorKhramtsov:
@okhat ah, that makes sense then. I brought up this inconsistency because we use dspy in AWS lambda and after the recent update it started crashing. Since the "breaking change" for lambda deploys is already introduced, what do you think about renaming all `DSP_CACHEDIR` to `DSPY_CACHEDIR` and `DSP_NOTEBOOK_CACHEDIR` to `DSPY_NOTEBOOK_CACHEDIR` and update the `faqs.md` to reference new names? If that sounds good to you I can prepare a PR.

okhat:
@IgorKhramtsov Thanks a lot. I thought more about this. I think we should keep the code unchanged but definitely update the docs. Specifically, DSP_CACHEDIR should still be responsible for old clients (including dspy.OpenAI, dspy.ColBERTv2, etc.) and DSPY_CACHEDIR should be responsible for the new dspy.LM client.

In the AWS lambda deployment, perhaps you should disable both DSP_* and DSPY_*.

Let me know if this makes sense. I would definitely really appreciate your help with updating this :D

PR #1568, opened 2024-10-01
Add `num_retries` to signature_opt_typed

tkellogg:
This adds a `num_retries` option to
`signature_opt_typed.optimize_signature()`. This is passed through to both `TypedChainOfThought` and `TypedPredictor`.

**Why?** Because I was stuck. I kept getting this error message:

> Value error, Never propose a signature already in the list above.: proposed_signature (error type: value_error)

I can get around this error by raising the model's temperature as well as setting this to `num_retries=10`.

Note: I see comments about not supporting temperature because it breaks parsing. While that's true, having temp be zero also leads to getting completely stuck because there's no simple solution for common errors like this. With high enough temperature and enough retries, I should be able to unstick myself from any situation.

### DIFF
--- a/dspy/teleprompt/signature_opt_typed.py
+++ b/dspy/teleprompt/signature_opt_typed.py
@@ -150,7 +150,7 @@ def optimize_signature(
     prompt_model=None,
     initial_prompts=2,
     verbose=False,
-    num_retries: int = 3,
+    max_retries: int = 3,
 ) -> dspy.Program:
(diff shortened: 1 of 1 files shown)


okhat:
Thank you @tkellogg ! This looks like a very useful addition to me but I'll defer to @thomasahle if he has a chance to check, since he built all the typed pieces. Will merge in 2 days otherwise.

PR #1575, opened 2024-10-02
added deprecation warning

arnavsinghvi11:


okhat:
Thanks @arnavsinghvi11 ! I think the notice should be directed at someone who is looking to build their own custom client. It should tell them that they probably don't need to if it's either supported in LiteLLM or if it has an OpenAI-compatible API endpoint. It should tell them to read the migration document. It should tell them that sub-classing `dspy.LM` (NOT `dsp.LM` as before) is also a good last resort.

PR #1577, opened 2024-10-03
Update intro.ipynb to use 2.5 syntax

rsyi:
dspy.OpenAI -> dspy.LM

### DIFF
--- a/intro.ipynb
+++ b/intro.ipynb
@@ -87,17 +87,17 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


arnavsinghvi11:
Hi @rsyi , 

Currently, the intro.ipynb notebook has cached runs with the pre-dspy.2.5 LM configuration for `gpt-3.5-turbo`, and modifying that will break the notebook's caches. We may have plans to update the cache in future versions to support the v2.5 dspy.LM configuration (but will defer to @okhat on this), but I believe we can close this PR for now.

PR #1585, opened 2024-10-05
Feat/port tests to dspy lm client

mikeedjones:
This PR attempts to port all tests currently using DummyLM to a new DummyLM which inherits from `dspy.LM` - hence migrating the test suite to use the new `ChatAdapter`.

It also replicates all current tests which use the renamed `DspDummyLM(dsp.LM)` into a folder `tests/DSP_LM`, to be deleted when 2.6 is released and `dsp.LM` is deprecated.

The tests for the to-be-deprectaed `MIPRO` have not been migrated to use `dspy.LM`.

I also included a fix for predictors which return `Literal` types (or any type the origin of which does not have a `__name__` attribute) with a small tweak to `dspy/adapters/chat_adapter.py:get_annotation_name`. Without this change `tests/functional/test_functional.py:test_literal.*` fail. Can also skip those tests and create another PR.

This PR also adds an auto-used fixture which resets the `dspy.settings` to default after each test without which some tests were interdependent.


mikeedjones:
Discovered a cute gotcha - if you configure the LM (`dspy.settings.configure(lm=lm)`) after initializing your modules you get a different prompt to if you configure your LM after initializing your modules.

```python
    pot = ProgramOfThought(BasicQA)
```

```python
```

dspy/adapters/chat_adapter.py
```diff
@@ -79,82 +90,90 @@ def format_fields(fields):
...
-        if hasattr(annotation, '__name__'):
+        if hasattr(annotation, "__name__"):
             return annotation.__name__
         else:
             return str(annotation)
     else:
-        args_str = ', '.join(get_annotation_name(arg) for arg in args)
-        return f"{origin.__name__}[{args_str}]"
+        args_str = ", ".join(get_annotation_name(arg) for arg in args)
+        return f"{get_annotation_name(origin)}[{args_str}]"
```
mikeedjones, line 147: Change to account for the origin of `Literal` not having a `__name__` attribute.

okhat:
Thanks so much, @mikeedjones !!! This is amazing. Great catch on the issue.

dsp/utils/settings.py
```diff
@@ -25,26 +46,7 @@ def __new__(cls):
...
-                release=0,
-                bypass_assert=False,
-                bypass_suggest=False,
-                assert_failures=0,
-                suggest_failures=0,
-                langchain_history=[],
-                experimental=False,
-                backoff_time = 10
-            )
+            config = DEFAULT_CONFIG
```
okhat, line 49: I like the refactor but there's a subtle risk here. This is now a reference to a global variable, and I fear somehow that we should make a deep copy here instead. (Arguably it's a singleton so who knows, maybe it's OK now, but I'd like to be sure it's a unique object)

okhat:
@mikeedjones This is super awesome. I'm tempted to merge as-is but I left a comment to be resolved up about DEFAULT_CONFIG.

More importantly, though, I see that the current tests are checking the output string, e.g. `[[ ## name ## ]]`, which may well change frequently in 2.5 until we have 2.6 which is when we won't adjust the default adapters anymore.

More generally, basically we should test that the adapter's `parse` of the `response` is unchanged, not that the string is unchanged, if you know what I mean.

dsp/utils/settings.py
mikeedjones: ahh good point. Changed that thanks!

mikeedjones:
Yeah, I agree - I just tried to replicate old tests as closely as I could - maybe worth writing a dummy adapter as well as a dummy LM along with a test suite for `ChatAdapter`? Also have this issue where the lines passed to `DummyLM` would have to change as the ChatAdapter changes.

tests/dsp_LM/examples/test_baleen.py
```diff
@@ -0,0 +1,122 @@
...
+    dataset = HotPotQA(train_seed=1, train_size=20, eval_seed=2023, dev_size=50, test_size=0)
+    # Tell DSPy that the 'question' field is the input. Any other fields are labels and/or metadata.
+    trainset = [x.with_inputs("question") for x in dataset.train]
+    devset = [x.with_inputs("question") for x in dataset.dev]
+    return trainset, devset
+
+
+# @pytest.mark.slow_test
+# TODO: Find a way to make this test run without openai
+def _test_baleen():
```
okhat, line 59: Wait what's this file? Was it there before? Not sure we want this.

tests/dsp_LM/retrieve/integration_test_pgvectorrm.py
```diff
@@ -0,0 +1,94 @@
+"""Instructions:
```
okhat, line 1: I think the whole `retrieve` folder at `tests/dsp_LM/retrieve/` probably doesn't need to be under `dsp_LM`?

okhat:
OK just reiterating this looks great to me (except I'm not sure I understand the implications of `test_baleen.py`) but we can't the parts with hard-coded prompt strings and output strings can't be merged directly as they would break on any changes to ChatAdapter. And I expect changes to happen in ChatAdapter.

We could introduce a dummy adapter but at that point what are we really testing? In any case, if making a dummy adapter makes sense to you and makes it easy to get this update overall merged, it sounds good to me in the short-to-medium term.

tests/dsp_LM/examples/test_baleen.py
mikeedjones: From this PR: https://github.com/stanfordnlp/dspy/pull/451/files#diff-a1699b14a8933ad094957aac992107cfac4a4a54a7bd20fb0efca6d0cb751fb1

tests/dsp_LM/retrieve/integration_test_pgvectorrm.py
mikeedjones: `tests/dsp_LM/retrieve/test_llama_index_rm.py` uses `DspyDummyLM` but the tests are skipped in the CI.

mikeedjones:
Maybe we remove the history comparisons and pass a list of `field_name:value` dicts into `DummyLM` which then has some output formatter? When a dev makes changes to ChatAdapter they have to update DummyLM's output formatter, doesn't seem like too much of a lift. Might make a good smoke test that their changes to `ChatAdapter` are working as intended?

okhat:
> Maybe we remove the history comparisons
Sure. Later we can figure out how to test this.

> pass a list of field_name:value dicts into DummyLM which then has some output formatter
Hmm, ideally that formatter is grabbed directly from the adapter I guess? Basically the test in essence will just check that parse(format_output_values(values)) == values, right?

I don't want us to maintain two different copies of the same thing for the tests' sake.

Btw I'm totally happy to have us merge this PR without this bit about formatting altogether, then we can discuss that part as a second PR. Your call. It's fine this way too.

mikeedjones:
I guess the formatter for `DummyLM` would be the inverse of `ChatAdapter.parse` so maybe not the easiest to pull out directly. Probably possible using the example formatter.

I think merge and i'll make a PR removing the history comparison and adding input now :)

mikeedjones:
https://github.com/stanfordnlp/dspy/pull/1595 <- can close this in favour?

PR #1590, opened 2024-10-07
updated assertions error msg to specify target_module

arnavsinghvi11:


docs/api/assertions.md
```diff
@@ -101,12 +101,14 @@ We can declare these validation checks through `dspy.Suggest` statements (as we
 dspy.Suggest(
     len(query) <= 100,
     "Query should be short and less than 100 characters",
+    target_module=self.generate_query
```
okhat, line 104: Wait I don't think we want the examples to be verbose like this. It's nice that they were implicit some of the time.

dspy/primitives/assertions.py
```diff
@@ -248,7 +248,7 @@ def wrapper(*args, **kwargs):
                                 dspy.settings.backtrack_to = dsp.settings.trace[-1][0]
 
                             if dspy.settings.backtrack_to is None:
-                                dspy.logger.error("Specified module not found in trace")
+                                dspy.logger.error("Module not found in trace. If passing a DSPy Signature, please specify the intended module for the assertion (e.g., use `dspy.ChainOfThought(GenerateAnswer)` instead of `GenerateAnswer`).")
```
okhat, line 251: The error is good except the example, I think you wanna say use `self.my_module`. With the current error, people may pass a new object of `dspy.ChainOfThought` literally.

PR #1594, opened 2024-10-07
Refactor finetuning implementation to be 2.5 compatible

isaacbmiller:
cc @dilarasoylu 

relevant anysacle notebook is https://github.com/anyscale/templates/pull/381


isaacbmiller:
cc @SolitaryThinker

examples/finetuning/finetune_test.ipynb
```diff
@@ -0,0 +1,163 @@
+{
```
dilarasoylu, line 1: Can remove this file

examples/finetuning/finetuning_demo.ipynb
```diff
@@ -0,0 +1,1175 @@
+{
```
dilarasoylu, line 1: This file can also be removed -- it needs to be updated

dspy/clients/anyscale.py
```diff
@@ -0,0 +1,367 @@
...
+        raise NotImplementedError("Method `status` is not implemented.")
+
+
+def finetune_anyscale(
+        job: FinetuneJobAnyScale,
+        model: str,
+        method: TrainingMethod,
+        train_data: List[Dict[str, Any]],
+        train_kwargs: Optional[Dict[str, Any]]=None,
+    ) -> List[str]:
```
dilarasoylu, line 87: Should the return hint be str?

dspy/clients/lm.py
```diff
@@ -60,9 +87,84 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
+        """Start model fine-tuning, if supported."""
+        # Fine-tuning is experimental and requires the experimental flag
+        from dspy import settings as settings
+        err = "Fine-tuning is an experimental feature."
+        err += " Set `dspy.settings.experimental` to `True` to use it."
+        assert settings.experimental, err
+
+        # Initialize the finetune job
+        FinetuneJobClass = get_provider_finetune_job_class(provider=provider)
+        finetune_job = FinetuneJobClass(
```
dilarasoylu, line 141: Should we also pass `train_method` here? nit. Let's pick between `method` vs. `train_method` and change all the functions to use the same name

isaacbmiller: Lets use train method

dspy/adapters/chat_adapter.py
```diff
@@ -67,6 +67,23 @@ def parse(self, signature, completion, _parse_values=True):
 
         return fields
 
+    def format_completion(self, signature, outputs):
```
okhat, line 70: What's the goal here? It creates a lot of logic duplication, i.e. editing the format now requires editing one more place.

dspy/clients/anyscale.py
```diff
@@ -0,0 +1,366 @@
...
+        # Call the super's cancel method after the custom cancellation logic, 
+        # so that the future can be cancelled
+        # super().cancel()
+
+    def status(self):
+        """Get the status of the finetune job."""
+        raise NotImplementedError("Method `status` is not implemented.")
+
+
+def finetune_anyscale(
```
SolitaryThinker, line 81: Nice, this is what I would have done as well!

SolitaryThinker (approved):
LGTM!

dspy/adapters/chat_adapter.py
isaacbmiller: Used to generate prompt/completion pair given a trace.

Is there a better way to do it?

okhat: Yes, discussed offline

dspy/predict/predict.py
```diff
@@ -16,6 +18,35 @@ def warn_once(msg: str):
...
 
 
+#-------------------------------------------------------------------------------
+#    Helper functions
+#-------------------------------------------------------------------------------
+
+def assert_structural_equivalency_for_predictors(
+        predictor1: object,
+        predictor2: object,
+    ) -> Optional[AssertionError]:
```
okhat, line 28: Bad output type? This function returns None.

dspy/primitives/program.py
```diff
@@ -25,6 +211,26 @@ def __init__(self):
     def __call__(self, *args, **kwargs):
         return self.forward(*args, **kwargs)
 
+    def set_lm(self, lm: LM) -> Optional[AssertionError]:
```
okhat, line 214: bad output type again, and also the import of annotations seems unnecessary

dspy/primitives/program.py
```diff
@@ -1,9 +1,195 @@
...
+
+_INFO_LM_PRED = """
+    LM for {pname} is an instance of ’{lname}’"""
+
+
+#-------------------------------------------------------------------------------
+#    Helper functions
+#-------------------------------------------------------------------------------
+
+def assert_structural_equivalency_for_programs(
```
okhat, line 54: These functions are cluttering this file. Do we even need them? If yes, do they need to be in this file? They seem like nice utils to have in a different file.

The code "smell" here that gives away that they don't need to be here is that the author decided to have the divisioning below, which we don't really like to have in DSPy:

```python
-------------------------------------------------------------------------------
#    Helper functions
#-------------------------------------------------------------------------------
```

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,377 @@
...
+
+    # If the global LM is being used, set it to the LMs of the copied teacher
+    # program predictors to to avoid handling the same edge cases later
+    if dspy.settings.lm:
+        teacher._set_all_predictor_lms(dspy.settings.lm)
+
+    return teacher
+
+
+# TODO: fix docstring
```
okhat, line 81: Having no docstring is better than having a bad docstring. The library is still evolving, I'd vote for not writing long docstrings in general until super stable.

okhat: This is especially true for internal methods like this, which are not consumed by our users.

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,377 @@
...
+        metric: Optional[Callable[
+            [Example, Prediction, Optional[List]], Union[bool, int, float]
+        ]] = None,
+        num_threads = 1,
+        sampling_round: int = 0,
+        sampling_temperature: Optional[float] = 0.9,
+        sampling_temperature_delta: float = 0.001,
+        max_errors: int = 0
+    ) -> Union[List[Dict], AssertionError]:
+    """ Bootstrap data for the given sampling round.
```
okhat, line 266: I'm still super surprised about the amount of docstrings here. Who is the intended audience here? If this needs to be in the docs, users won't find it here. But I think this is for internal consumption?

Let's write self-documenting code. The code should use functions and variables and structure whose content tells us what's happening. Use comments only to document unobvious decisions. Don't write a docstring that's longer than the implementation.

dspy/teleprompt/random_search.py
```diff
@@ -85,6 +85,7 @@ def compile(self, student, *, teacher=None, trainset, valset=None, restrict=None
                     max_labeled_demos=self.max_labeled_demos,
                     teacher_settings=self.teacher_settings,
                     max_rounds=self.max_rounds,
+                    max_errors=self.max_errors,
```
okhat, line 88: This is awesome, thank you!

dspy/clients/self_hosted.py
```diff
@@ -0,0 +1,34 @@
+from typing import Dict, Any
+
+from dspy.clients.openai import is_openai_model
+
+
+#-------------------------------------------------------------------------------
```
okhat, line 6: This file should be a class, that inherits from some ABC that defines the right interface. Right now, we're essentially expressing a class via 3 disconnected functions.

dspy/clients/lm.py
```diff
@@ -60,9 +87,76 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
+            **self.kwargs,
+            **kwargs,
+        }
+        return self.__class__(**kwargs)
+    
     def inspect_history(self, n: int = 1):
         _inspect_history(self, n)
 
+    def launch(self):
+        """Send a request to the provider to launch the model, if needed."""
```
okhat, line 106: Bad design here. We should be getting a Provider object from the user, for example, and just calling `provider.launch()` or something like that. If statements per provider will not scale.

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,377 @@
...
+        pred_ind_to_name = {
+            ind: name for ind, (name, _) in enumerate(program.named_predictors())
+        }
+
+    # Build the prompt-completion data
+
+    adapter = dspy.settings.adapter or dspy.ChatAdapter()
+    data = []
+
+    # TODO: Make sure that this works for multi-stage pipelines
```
isaacbmiller, line 155: can't forget this

dspy/clients/lm.py
```diff
@@ -60,9 +87,76 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
+        ) -> FinetuneJob:
+        """Start model fine-tuning, if supported."""
+        # Fine-tuning is experimental and requires the experimental flag
+        from dspy import settings as settings
+        err = "Fine-tuning is an experimental feature."
+        err += " Set `dspy.settings.experimental` to `True` to use it."
+        assert settings.experimental, err
+
+        # Initialize the finetune job
+        FinetuneJobClass = get_provider_finetune_job_class(provider=provider)
```
okhat, line 140: again I'd like to avoid `get_provider_finetune_job_class` these functions are pretty brittle

dspy/clients/lm.py
```diff
@@ -151,3 +246,139 @@ def _inspect_history(lm, n: int = 1):
...
+        _PROVIDER_OPENAI: FinetuneJobOpenAI,
+    }
+
+    # Get the FinetuneJob class for the provider
+    _CLS = _PROVIDER_TO_FINETUNE_JOB_CLASS[provider]
+
+    return _CLS
+
+
+def get_provider_finetune_function(provider: str) -> callable:
```
okhat, line 305: there's a ton of boilerplate here that definitely shouldn't need to exist IMO

dspy/clients/lm.py
```diff
@@ -151,3 +246,139 @@ def _inspect_history(lm, n: int = 1):
...
+        job=job,
+        model=model,
+        provider=provider,
+        train_data=train_data,
+        train_kwargs=train_kwargs,
+        train_method=train_method,
+    )
+
+
+def finetune(
```
okhat, line 359: more boilerplate that shouldn't need to exist IMO

dspy/predict/predict.py
dilarasoylu: Returns an `AssertionError` if the assert statement fails -- type hint is accounting for that. Return type is optional `AssertionError`, because it returns `None` if the assert statements succeed.

isaacbmiller: It doesn't return it throws. Im in favor of leaving the type off this one

dspy/primitives/program.py
isaacbmiller: I think primitives/utils.py and primitives/constants.py can be good homes for these or just all in primitives/utils.py

dspy/clients/self_hosted.py
chenmoneygithub: I would rather just delete this file until we have real code logic

dspy/predict/predict.py
chenmoneygithub: Yes this type hint is a bit confusing, we can omit the type hints here.

dspy/predict/predict.py
```diff
@@ -16,6 +18,35 @@ def warn_once(msg: str):
...
+    """
+    # Assert that the objects are predictors
+    assert isinstance(predictor1, Predict)
+    assert isinstance(predictor2, Predict)
+
+    # Assert that the predictors have equivalent signatures
+    assert predictor1.signature.equals(predictor2.signature)
+
+
+#-------------------------------------------------------------------------------
```
chenmoneygithub, line 46: we can delete this comment as the code self explains. 

btw it's generally discouraged in industry to use `#-----` for comment: https://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments because the format can be hard to read on different editor, like emacs

dspy/primitives/program.py
chenmoneygithub: `primitives/utils.py` is the right location.

The type hints is wrong here as well.

chenmoneygithub (commented):
Nice work! 

Agreed with @okhat, the PR can use some refactoring to have the finetuning code in a standalone mode.

dspy/clients/lm.py
```diff
@@ -60,9 +87,76 @@ def __call__(self, prompt=None, messages=None, **kwargs):
         
         return outputs
 
+    def copy(self, **kwargs):
```
chenmoneygithub, line 90: copy with updated args is strange, essentially here we are just calling the constructor. I would rather provide a function `get_config` to get all the args, and pass to constructor.

isaacbmiller: I agree but lm.copy is an established pattern. This is actually a duplicate function from the copy function already in lm. Removing it

okhat: This is actually a pretty established pattern iirc, both inside DSPy and in other libraries. It's a non-inplace way to modify an object (getting a copy with overwrites).

dspy/clients/finetune.py
```diff
@@ -0,0 +1,133 @@
...
+    not_started = "not_started"
+    pending = "pending"
+    running = "running"
+    succeeded = "succeeded"
+    failed = "failed"
+    cancelled = "cancelled"
+
+
+"""Dictionary mapping training methods to the data keys they require."""
+TRAINING_METHOD_TO_DATA_KEYS = {
```
chenmoneygithub, line 49: I am unsure how generic this can apply to different platforms. We can keep it with this PR, but may be changed in the future

dspy/clients/lm.py
```diff
@@ -59,10 +81,65 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
         
         return outputs
-
+    
     def inspect_history(self, n: int = 1):
         _inspect_history(self, n)
 
+    def launch(self):
+        """Send a request to the provider to launch the model, if needed."""
+        if is_self_hosted_model(self.model):
```
chenmoneygithub, line 90: let's delete the `self_hosted_model_launch` part, we won't have it ready any time soon.

dspy/clients/lm.py
```diff
@@ -59,10 +81,65 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
+            self_hosted_model_launch(self.model, self.launch_kwargs)
+        elif is_anyscale_model(self.model):
+            anyscale_model_launch(self.model, self.launch_kwargs)
+        msg = f"`launch()` is called for the auto-launched model {self.model}"
+        msg += " -- no action is taken!"
+        logger.info(msg)
+
+    def kill(self):
+        """Send a request to the provider to kill the model, if needed."""
+        if is_self_hosted_model(self.model):
```
chenmoneygithub, line 100: same here

dspy/clients/lm.py
```diff
@@ -151,3 +229,114 @@ def _inspect_history(lm, n: int = 1):
...
+    model: str,
+    train_data: List[Dict[str, Any]],
+    train_kwargs: Optional[Dict[str, Any]]=None,
+    provider: str = "openai",
+    train_method: TrainingMethod = TrainingMethod.SFT,
+) -> Union[str, ValueError]:
+    return finetune(
+        job=job,
+        model=model,
+        provider=provider,
```
chenmoneygithub, line 310: minor issue - let's keep the arg order same as definition.

dspy/clients/lm.py
```diff
@@ -151,3 +229,114 @@ def _inspect_history(lm, n: int = 1):
...
+        job.set_result(lm)
+    except Exception as err:
+        logger.error(err)
+        job.set_result(err)
+
+
+# TODO: Perhaps we shouldn't directly cache the data
+# TODO: Add DiskCache, ignore job
+
+def cached_finetune(
```
chenmoneygithub, line 299: curious - what does this `cached_finetune ` do? from the code it's just a tunnel to `finetune` method

dspy/clients/openai.py
```diff
@@ -0,0 +1,403 @@
...
+    TrainingStatus,
+    validate_finetune_data,
+    save_data,
+)
+
+# Provider name
+PROVIDER_OPENAI = "openai"
+
+# List of model IDs
+_MODEL_IDS = [
```
chenmoneygithub, line 21: This is hard to maintain, as openai releases pretty quickly. We can do something like:
```
client = OpenAI()
models = client.models.list()
```
when we need to get the model list.

dspy/clients/openai.py
```diff
@@ -0,0 +1,403 @@
...
+        model: str,
+        method: TrainingMethod,
+        train_data: List[Dict[str, Any]],
+        train_kwargs: Optional[Dict[str, Any]]=None,
+    ) -> str:
+    train_kwargs = train_kwargs or {}
+    train_method = TrainingMethod.SFT  # Note: This could be an argument; ignoring method
+
+    # Validate train data and method
+    logger.info("[Finetune] Validating the formatting of the data")
```
chenmoneygithub, line 133: I am curious about a thing - does openai finetuning itself have logging? I haven't tried that.

dspy/clients/anyscale.py
```diff
@@ -0,0 +1,347 @@
...
+def verify_dataset(dataset: List[dict[str, Any]]) -> bool:
+    """Verify the training arguments before starting training."""
+    dataset_validation = openai_data_validation(dataset)
+
+    if dataset_validation:
+        logger.error(f"Dataset validation failed: {dataset_validation}")
+        return False
+
+    # TODO: Is this still useful?
+    convo_lens = check_message_lengths(dataset)
```
chenmoneygithub, line 161: this variable is unused, I would just delete that along with the `check_message_lengths` method, and rely on OpenAI to report the usage.

dspy/predict/predict.py
```diff
@@ -1,21 +1,20 @@
...
 import random
-from functools import lru_cache
 
 from pydantic import BaseModel
 
 import dsp
 from dspy.predict.parameter import Parameter
-from dspy.primitives.prediction import Prediction
 from dspy.primitives.program import Module
+from dspy.primitives.prediction import Prediction
```
chenmoneygithub, line 10: Is this order change expected? the old one seems to sort correctly

dspy/primitives/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .example import *
 from .prediction import *
 from .program import *
-from .python_interpreter import *
+from .python_interpreter import *
```
chenmoneygithub, line 5: undo this change

dspy/primitives/program.py
```diff
@@ -1,18 +1,13 @@
+
```
chenmoneygithub, line 1: nit: delete the empty line

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
+from typing import Any, Callable, Dict, List, Optional, Union
+
+import dspy
+from dspy.utils.logging import logger
+from dspy.evaluate.evaluate import Evaluate
+from dspy.primitives.example import Example
+from dspy.primitives.program import Program
+from dspy.primitives.prediction import Prediction
+
+_INFO_DEFAULT_TEACHER = """No teacher provided. Using a copy of the student \
```
chenmoneygithub, line 10: These messages are only used once, we can inline them

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+    for pred_ind, (pred, inputs, outputs) in enumerate(trace):
+        # Get the demos from the predictor if exclude_demos is False
+        demos = [] if exclude_demos else pred.demos
+        messages = adapter.format(pred.signature, demos, inputs)
+        messages.append(adapter.format_turn(signature=pred.signature, values=outputs, role="assistant", incomplete=False))
+        data.append(messages)
+
+    return data
+
+def dummy_metric(example, pred, trace=None, frac=1.0):
```
chenmoneygithub, line 119: let's delete this, and just check if metric function is None.

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+    data = []
+    for example in dataset:
+        data_dict = process_example(example, 0, program, metric)
+        if data_dict is not None:
+            data.append(data_dict)
+    
+    return data
+
+
+def process_example(example: Any, example_ind: int, program: Callable, metric: Optional[Callable] = None) -> Dict[str, Any]:
```
chenmoneygithub, line 152: I think this is called `bootstrap_one_example()` (or similar) in the `BootstrapFewshot`optimizer, maybe keep the consistency?

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+    # print("Processing example:", example_ind)
```
chenmoneygithub, line 153: nit: delete this

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+    with dspy.context(trace=[]):
+        # print("Running program...", example_ind)
```
chenmoneygithub, line 155: same here

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+        try:
+            prediction = program(**example.inputs())
+        except Exception as e:
+            print(f"Error processing example {example_ind}: {e}")
+            return None
+        # print("Getting trace...", example_ind)
```
chenmoneygithub, line 161: hmm

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+        trace = dspy.settings.trace
+        # print("Getting score...", example_ind)
+        score = metric(example, prediction, trace) if metric else None
+
+    data_dict = {
```
chenmoneygithub, line 166: Do we want to only bootstrap data that is over a threshold? Right now we assume the bootstrapping succeeds

dspy/teleprompt/finetune_teleprompter.py
```diff
@@ -0,0 +1,237 @@
...
+    }
+    if metric:
+        data_dict['score'] = score
+    
+    return data_dict
+
+# TODO: If we can ensure to pass the "round" information every time a call is
+# issued to an LM, we can make repetitive un-cached calls to the same LM without
+# modifying it's temperature. This function can be removed then.
+def bootstrap_data_for_round(
```
chenmoneygithub, line 180: Do we need this? Looks like this is just changing the temperature and call `bootstrap_data`, we should be able to write this code in the optimizer implementation.

chenmoneygithub (changes requested):
Thanks for the update! Most comments are on nits.

dspy/adapters/chat_adapter.py
```diff
@@ -67,6 +67,9 @@ def parse(self, signature, completion, _parse_values=True):
 
         return fields
 
+    def format_turn(self, signature, values, role, incomplete=False):
```
okhat, line 70: This function is weirdly recursive / it's wrong right?

okhat:
Besides the things discussed on Slack, this is good to merge except for `lm.py` which I think @dilarasoylu is handling.

okhat:
Incredible work on this @dilarasoylu & @isaacbmiller !!

dspy/clients/anyscale.py
```diff
@@ -0,0 +1,347 @@
...
+        # Call the super's cancel method after the custom cancellation logic, 
+        # so that the future can be cancelled
+        # super().cancel()
+
+    def status(self):
+        """Get the status of the finetune job."""
+        raise NotImplementedError("Method `status` is not implemented.")
+
+
+def finetune_anyscale(
```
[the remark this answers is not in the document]
dilarasoylu: Thank you @chenmoneygithub! This is under the works, but will be part of the next PR.

dspy/clients/finetune.py
dilarasoylu: This is true -- our `TRAINING_METHOD` is actually a misnomer for `DATA_METHOD`. We will have the correct abstraction in the next PR...

dspy/clients/finetune.py
```diff
@@ -0,0 +1,133 @@
...
+        return dict(
+            model=self.model,
+            train_data=self.train_data,
+            train_kwargs=self.train_kwargs,
+            provider=self.provider,
+            train_method=self.train_method,
+        )
+
+    def __str__(self):
+        return f"FinetuningJob({self.to_dict()})"
```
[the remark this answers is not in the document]
dilarasoylu: Argh, a version of this didn't include `train_data` in the past, so it was just printing high level information about the Job. Removing it now.

dspy/clients/lm.py
dilarasoylu: (In next PR)

dspy/clients/lm.py
dilarasoylu: (Deleted, thank you!)

dspy/clients/lm.py
dilarasoylu: (Also deleted)

dspy/clients/lm.py
dilarasoylu: Next PR again

dspy/clients/lm.py
dilarasoylu: Next PR resolves this

dspy/clients/lm.py
dilarasoylu: Oh, yes, this is meant to be the cached version of a function that actually does the work. Similar to the relationship between the following two functions: 

```
@functools.lru_cache(maxsize=None)
def cached_litellm_completion(request):
    return litellm_completion(request, cache={"no-cache": False, "no-store": False})


def litellm_completion(request, cache={"no-cache": True, "no-store": True}):
    kwargs = ujson.loads(request)
    return litellm.completion(cache=cache, **kwargs)
```

dspy/clients/lm.py
dilarasoylu: Changed

dspy/clients/lm.py
dilarasoylu: Next PR again

dspy/predict/predict.py
dilarasoylu: Thank you for the catch, reverted

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: Done!

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: Removed -- originally meant for public use

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: Reverting

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: Changed

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: As in based on a score? We leave that to the caller

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: This was used elsewhere -- isn't needed in this PR. Removing.

dspy/teleprompt/finetune_teleprompter.py
dilarasoylu: Was originally meant for user facing, removed

dspy/clients/lm.py
```diff
@@ -1,29 +1,51 @@
...
+        ):
+        # Remember to update LM.copy() if you modify the constructor!
         self.model = model
         self.model_type = model_type
         self.cache = cache
+        self.launch_kwargs = launch_kwargs or {}
         self.kwargs = dict(temperature=temperature, max_tokens=max_tokens, **kwargs)
         self.history = []
 
+        # TODO: This is error prone!
```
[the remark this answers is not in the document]
dilarasoylu: Moving to internal discussion -- we can perform an exact change here. Otherwise this will trigger in any model string that include `o1-` as a substring.

dspy/clients/anyscale.py
dilarasoylu: Deleted, @isaacbmiller feel free to revert if this was needed

dspy/clients/anyscale.py
```diff
@@ -0,0 +1,347 @@
+from typing import Any, Dict, List, Optional
```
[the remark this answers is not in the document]
dilarasoylu: Also curious about this, could be better to move all these modules elsewhere

dspy/adapters/chat_adapter.py
```diff
@@ -67,6 +67,9 @@ def parse(self, signature, completion, _parse_values=True):
...
+        return format_turn(signature, values, role, incomplete)
```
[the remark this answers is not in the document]
dilarasoylu: Yes, we use it in bootstrap data functions. Also think this part can be improved -- will move to internal discussion.

dspy/primitives/program.py
dilarasoylu: (Looks like we moved them to the teleprompter file, resolving)

dspy/clients/self_hosted.py
dilarasoylu: (Looks like deleted)

PR #1604, opened 2024-10-08
Remove legacy code from client/lm.py and reformat

chenmoneygithub:
As litellm becomes an official dependency of DSPy, we no longer need the hacky import handling code. 

Also reformat the code for better readability.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -1,51 +1,46 @@
-import os
-import uuid 
-import ujson
 import functools
-from pathlib import Path
+import os
+import uuid
 from datetime import datetime
+from pathlib import Path
 
-try:
-    import warnings
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", category=UserWarning)
-        if "LITELLM_LOCAL_MODEL_COST_MAP" not in os.environ:
-             os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"
-        import litellm  
-        litellm.telemetry = False
+import litellm
+import ujson
+from litellm.caching import Cache
 
-    from litellm.caching import Cache
-    disk_cache_dir = os.environ.get('DSPY_CACHEDIR') or os.path.join(Path.home(), '.dspy_cache')
-    litellm.cache = Cache(disk_cache_dir=disk_cache_dir, type="disk")
+disk_cache_dir = os.environ.get("DSPY_CACHEDIR") or os.path.join(Path.home(), ".dspy_cache")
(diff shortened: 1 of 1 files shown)


dspy/clients/lm.py
```diff
@@ -1,51 +1,46 @@
...
-    class LitellmPlaceholder:
-        def __getattr__(self, _): raise ImportError("The LiteLLM package is not installed. Run `pip install litellm`.")
+if "LITELLM_LOCAL_MODEL_COST_MAP" not in os.environ:
+    os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"
 
-    litellm = LitellmPlaceholder()
 
 class LM:
-    def __init__(self, model, model_type='chat', temperature=0.0, max_tokens=1000, cache=True, **kwargs):
+    def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, cache=True):
```
okhat, line 20: Thanks Chen! Why remove kwargs? The kwargs here are important. They're at the level of the LM object. The kwargs in`__call__` are at the level of the individual call. Both are needed.

chenmoneygithub: ohh the `kwargs` is not really used in `__init__` method, and the only place it is used is inside `__call__` method, where we join it with `kwargs` of `__call__` method. This is not ideal because we are not drawing an explicit boundary between these two `kwargs`, so users don't know which one they should use. We might want to explicitly list out model args like `temperature` and `max_tokens` for clarity.

okhat: Thank you @chenmoneygithub ! It's important that users can set kwargs in `__init__`, which become defaults for `__call__`. The `kwargs` passed to `__call__` overwrite any defaults from `__init__`. It was well-defined and commonly used. We should add it back IMO.

chenmoneygithub: sure thing! Let me add it back.

dspy/clients/lm.py
```diff
@@ -1,51 +1,47 @@
...
 class LM:
-    def __init__(self, model, model_type='chat', temperature=0.0, max_tokens=1000, cache=True, **kwargs):
+    def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, cache=True, **kwargs):
         self.model = model
         self.model_type = model_type
         self.cache = cache
-        self.kwargs = dict(temperature=temperature, max_tokens=max_tokens, **kwargs)
+        self.temperature = temperature
+        self.max_tokens = max_tokens
+        self.kwargs = kwargs
```
okhat, line 26: This used to be:

```python
```

okhat: I think that was cleaner?

PR #1613, opened 2024-10-11
feat(dspy): extend incomplete generations

mikeedjones:
If the output parsed by `ChatAdapter.parse` does not include all the fields required by the signature, the lm will be called again up to `max_extensions` times, with the thus-far completed output added to the input fields for the next generation.

Logic is wrapped up as a decorator to allow the logic to be reused by future `Adaper`s.

Added tests for extended generation.

`max_extensions` defaults to 1 to maintain current behaviour.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -114,7 +114,6 @@ class Predict(Module, Parameter):
             *_, last_key = self.extended_signature.fields.keys()
             self.extended_signature = self.extended_signature.with_updated_fields(last_key, prefix=prefix)
 
-
     def __call__(self, **kwargs):
         return self.forward(**kwargs)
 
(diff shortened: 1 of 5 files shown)


dspy/predict/predict.py
okhat, line 1: Just to double check: this file is not being changed materially right? Just formatting?

dspy/predict/predict.py
```diff
@@ -247,14 +254,21 @@ def new_generate(lm, signature, example, max_depth=6, **kwargs):
...
-    
+    adapter = dspy.settings.adapter or dspy.ChatAdapter()
+    return adapter(
+        lm,
+        lm_kwargs=lm_kwargs,
+        signature=signature,
+        demos=demos,
+        inputs=inputs,
+        _parse_values=_parse_values,
+        max_extensions=dspy.settings.max_extensions or 0,
```
mikeedjones, line 266: @okhat this is the only real change in this file - everything else is linting/formatting from rust

PR #1614, opened 2024-10-11
Minor fixes on MiproV2 optimizer

chenmoneygithub:
This is just a minor fix while I am trying to understand deeper about how MiproV2 works.  It's not touching the functionality, but fix the lint and dead code. Change summary:

- Convert long string to multiple lines to comply to our 120 max width requirement. In the future we should enable lint check for it.
- Remove dead code.
- Function calls should set arguments in the order as definition (at best effort). 
- Some general lint fixes.


chenmoneygithub:
@klopsahlong This PR doesn't touch the functionality, just fixes the readability. Please let me know if you have questions, thanks!

PR #1617, opened 2024-10-12
Add a tutorial for summarization task

karthikscale3:
This PR adds a tutorial for the summarization task that is built by composing a Metric program and a Summarization program.


okhat:
Looks great. Is it missing imports?

You can say `dspy.Evaluate` for example btw.

karthikscale3:
> Looks great. Is it missing imports?
> 
> You can say `dspy.Evaluate` for example btw.

Good catch! fixed it

okhat:
Thanks a lot @karthikscale3 !!! Merged.

PR #1621, opened 2024-10-14
Fix: 'LM' object has no attribute 'copy' 

xiayouran:
Bug info:
```python
  File "/data/liyanpeng/pywork/dspy/dspy/teleprompt/bootstrap.py", line 84, in compile
    self._bootstrap()
  File "/data/liyanpeng/pywork/dspy/dspy/teleprompt/bootstrap.py", line 154, in _bootstrap
    success = self._bootstrap_one_example(example, round_idx)
  File "/data/liyanpeng/pywork/dspy/dspy/teleprompt/bootstrap.py", line 210, in _bootstrap_one_example
    raise e
  File "/data/liyanpeng/pywork/dspy/dspy/teleprompt/bootstrap.py", line 182, in _bootstrap_one_example
    lm = lm.copy(temperature=0.7 + 0.001 * round_idx) if round_idx > 0 else lm
AttributeError: 'LM' object has no attribute 'copy'
```
This problem occurs when parameter `max_rounds` is greater than 1, so I tried to fix it by adding a `copy` function to `LM`.


dspy/clients/lm.py
```diff
@@ -63,6 +63,11 @@ def __call__(self, prompt=None, messages=None, **kwargs):
     def inspect_history(self, n: int = 1):
         _inspect_history(self, n)
 
+    def copy(self, **kwargs):
```
okhat, line 66: Hey @xiayouran ! Thanks a lot. I think this function is not correct right now.

We should probably get all the fields automatically from Python with `.__dict__` or something similar. But anyway, right now we're missing `model_type, temperature, max_tokens, cache` in the initializer, and we're likely going to miss other parameters if they're added, so this should be automated a bit.

xiayouran: Yes, @okhat I may have lost some parameters by doing so, but I updated this part of the code to use the `__dict__` attribute you suggested, so I think there should be no more missing parameters.
```suggestion
    def copy(self, **kwargs):
        kwargs = {**self.__dict__, **kwargs}
        return self.__class__(**kwargs)
```

PR #1622, opened 2024-10-14
MIPRO Bootstrap + Evaluation Parallelism

XenonMolecule:
This PR includes two main updates:

1. Enables nested parallelism in Evaluate.  Before the way that signal handling for user terminations was incompatible with nested threading.  If a user program used multi-threading (for instance calling BootstrapFewshotWithRandomSearch as a meta-optimization within a program they were evaluating or optimizing) then Evaluate would throw an exception.  This nested multithreading now works.
2. MIPRO bootstraps in parallel.  Theoretically this change can be applied to BootstrapFewshotWithRandomSearch as well.  Now all runs of the program to bootstrap the fewshot demonstrations occur at the same time.  This is only enabled by setting `bootstrap_parallel=True` in the MIPROv2 class (which is False by default).  The main TODO is to fix printing on this, because currently all the evaluation bars print simultaneously and make for a bit of an ugly printout.

I would appreciate a more thorough review of this PR since it touches functionality outside of my area of expertise (by modifying Evaluate), and because I am a relative novice when it comes to multithreading in python.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -1,10 +1,11 @@
 import contextlib
-import signal
 import sys
 import threading
 import traceback
 import types
 from typing import Any
+from joblib import Parallel, delayed
+import contextvars
 
 import pandas as pd
 import tqdm
@@ -94,45 +95,61 @@ class Evaluate:
         reordered_devset = []
         job_cancelled = "cancelled"
 
-        # context manger to handle sigint
-        @contextlib.contextmanager
-        def interrupt_handler_manager():
-            """Sets the cancel_jobs event when a SIGINT is received."""
-            default_handler = signal.getsignal(signal.SIGINT)
-
-            def interrupt_handler(sig, frame):
-                self.cancel_jobs.set()
-                dspy.logger.warning("Received SIGINT. Cancelling evaluation.")
-                default_handler(sig, frame)
-
-            signal.signal(signal.SIGINT, interrupt_handler)
-            yield
-            # reset to the default handler
-            signal.signal(signal.SIGINT, default_handler)
-
         def cancellable_wrapped_program(idx, arg):
             # If the cancel_jobs event is set, return the cancelled_job literal
             if self.cancel_jobs.is_set():
                 return None, None, job_cancelled, None
             return wrapped_program(idx, arg)
 
-        with ThreadPoolExecutor(max_workers=num_threads) as executor, interrupt_handler_manager():
-            futures = {executor.submit(cancellable_wrapped_program, idx, arg) for idx, arg in devset}
-            pbar = tqdm.tqdm(total=len(devset), dynamic_ncols=True, disable=not display_progress)
+        try:
+            with ThreadPoolExecutor(max_workers=num_threads) as executor:
+                futures = {executor.submit(cancellable_wrapped_program, idx, arg) for idx, arg in devset}
+                pbar = tqdm.tqdm(total=len(devset), dynamic_ncols=True, disable=not display_progress, file=sys.stdout)
 
-            for future in as_completed(futures):
-                example_idx, example, prediction, score = future.result()
+                for future in as_completed(futures):
+                    if self.cancel_jobs.is_set():
+                        break  # Exit the loop if jobs are cancelled
+                    try:
+                        example_idx, example, prediction, score = future.result()
+                    except KeyboardInterrupt:
+                        self.cancel_jobs.set()
+                        dspy.logger.warning("Received KeyboardInterrupt. Cancelling evaluation.")
+                        # Optionally cancel remaining futures
+                        for f in futures:
+                            f.cancel()
+                        break
+                    except Exception as e:
+                        # Handle exceptions in worker threads
+                        with self.error_lock:
+                            self.error_count += 1
+                            current_error_count = self.error_count
+                        if current_error_count >= self.max_errors:
+                            raise e
+                        if self.provide_traceback:
+                            dspy.logger.error(
+                                f"Error for example in dev set: \t\t {e}\n\twith inputs:\n\t\t{arg}\n\nStack trace:\n\t{traceback.format_exc()}"
+                            )
+                        else:
+                            dspy.logger.error(
+                                f"Error for example in dev set: \t\t {e}. Set `provide_traceback=True` to see the stack trace."
+                            )
+                        continue  # Skip this future and continue
 
-                # use the cancelled_job literal to check if the job was cancelled - use "is" not "=="
-                # in case the prediction is "cancelled" for some reason.
-                if prediction is job_cancelled:
-                    continue
+                    if prediction is job_cancelled:
+                        continue
 
-                reordered_devset.append((example_idx, example, prediction, score))
-                ncorrect += score
-                ntotal += 1
-                self._update_progress(pbar, ncorrect, ntotal)
-            pbar.close()
+                    reordered_devset.append((example_idx, example, prediction, score))
+                    ncorrect += score
+                    ntotal += 1
+                    self._update_progress(pbar, ncorrect, ntotal)
+                pbar.close()
+        except KeyboardInterrupt:
+            self.cancel_jobs.set()
+            dspy.logger.warning("Received KeyboardInterrupt. Cancelling evaluation.")
+            # Optionally cancel remaining futures
+            for f in futures:
+                f.cancel()
+            raise
 
         if self.cancel_jobs.is_set():
             dspy.logger.warning("Evaluation was cancelled. The results may be incomplete.")
@@ -164,21 +181,22 @@ class Evaluate:
         return_outputs = return_outputs if return_outputs is not None else self.return_outputs
         results = []
 
+        # Initialize context variable for the stack
+        stack_var = contextvars.ContextVar('stack', default=dspy.settings.main_stack)
+
         def wrapped_program(example_idx, example):
-            # NOTE: TODO: Won't work if threads create threads!
-            thread_stacks = dspy.settings.stack_by_thread
-            creating_new_thread = threading.get_ident() not in thread_stacks
-            if creating_new_thread:
-                thread_stacks[threading.get_ident()] = list(dspy.settings.main_stack)
+            # Use the context variable to manage the stack
+            stack = stack_var.get()
+            # No need to check for thread identity or manually manage stacks
 
             try:
-                prediction = program(**example.inputs())
-                score = metric(
-                    example,
-                    prediction,
-                )  # FIXME: TODO: What's the right order? Maybe force name-based kwargs!
+                # Set the context variable for the current thread
+                token = stack_var.set(list(stack))  # Copy the current stack
 
-                # increment assert and suggest failures to program's attributes
+                prediction = program(**example.inputs())
+                score = metric(example, prediction)
+
+                # Increment assert and suggest failures to program's attributes
                 if hasattr(program, "_assert_failures"):
                     program._assert_failures += dspy.settings.get("assert_failures")
                 if hasattr(program, "_suggest_failures"):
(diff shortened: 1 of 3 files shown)


dspy/evaluate/evaluate.py
```diff
@@ -94,45 +93,61 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
+                    if self.cancel_jobs.is_set():
+                        break  # Exit the loop if jobs are cancelled
+                    try:
+                        example_idx, example, prediction, score = future.result()
+                    except KeyboardInterrupt:
+                        self.cancel_jobs.set()
+                        dspy.logger.warning("Received KeyboardInterrupt. Cancelling evaluation.")
+                        # Optionally cancel remaining futures
+                        for f in futures:
+                            f.cancel()
```
mikeedjones, line 117: Not sure how this is optional?

mikeedjones:
Which exception was thrown with a nested threadpool? An MRE/tests to avoid regressions would be great :)

dspy/evaluate/evaluate.py
```diff
@@ -94,45 +93,61 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
-                self._update_progress(pbar, ncorrect, ntotal)
-            pbar.close()
+        try:
+            with ThreadPoolExecutor(max_workers=num_threads) as executor:
+                futures = {executor.submit(cancellable_wrapped_program, idx, arg) for idx, arg in devset}
+                pbar = tqdm.tqdm(total=len(devset), dynamic_ncols=True, disable=not display_progress, file=sys.stdout)
+
+                for future in as_completed(futures):
+                    if self.cancel_jobs.is_set():
+                        break  # Exit the loop if jobs are cancelled
```
mikeedjones, line 109: What happens to the other futures?

dspy/evaluate/evaluate.py
XenonMolecule: Good point I'll modify this comment

dspy/evaluate/evaluate.py
XenonMolecule: My understanding of how the futures work is that when I call f.cancel() below it will prevent all future jobs that haven't started yet from starting.  Then for the remaining jobs that are already in progress we catch them here and terminate as soon as they finish.

Was the implementation with signal better?  This seemed to be what was causing the exception, but maybe necessary to simultaneously terminate all threads?

XenonMolecule:
> Which exception was thrown with a nested threadpool? An MRE/tests to avoid regressions would be great :)

`2024-10-14T18:10:56.119310Z [error    ] Error for example in dev set:            signal only works in main thread of the main interpreter
        with inputs:
                Example({'context': 'a boy is not concentrating on a machine', 'question': 'Can we logically conclude for sure that a boy is not concentrating on a typewriter?'}) (input_keys={'context', 'question'})

Stack trace:
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/evaluate/evaluate.py", line 175, in wrapped_program
    prediction = program(**example.inputs())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/primitives/program.py", line 26, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/testing/tasks/scone.py", line 68, in forward
    new_prog = optimizer.compile(dspy.ChainOfThought("context,question->answer"), trainset=self.trainset, valset=self.testset)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/teleprompt/random_search.py", line 118, in compile
    score, subscores = evaluate(program, return_all_scores=True)
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/evaluate/evaluate.py", line 211, in __call__
    reordered_devset, ncorrect, ntotal = self._execute_multi_thread(
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/evaluate/evaluate.py", line 119, in _execute_multi_thread
    with ThreadPoolExecutor(max_workers=num_threads) as executor, interrupt_handler_manager():
  File "/opt/miniconda3/envs/dspy_maintenance/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/Users/michaelryan/Documents/School/Stanford/misc/dspy_updates/dspy/dspy/evaluate/evaluate.py", line 108, in interrupt_handler_manager
    signal.signal(signal.SIGINT, interrupt_handler)
  File "/opt/miniconda3/envs/dspy_maintenance/lib/python3.12/signal.py", line 58, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: signal only works in main thread of the main interpreter`

dspy/evaluate/evaluate.py
```diff
@@ -94,45 +93,61 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
+                    try:
+                        example_idx, example, prediction, score = future.result()
+                    except KeyboardInterrupt:
+                        self.cancel_jobs.set()
+                        dspy.logger.warning("Received KeyboardInterrupt. Cancelling evaluation.")
+                        # Cancel remaining futures
+                        for f in futures:
+                            f.cancel()
+                        break
```
mikeedjones, line 118: do you need to re-raise the KeyboardInterrupt here? If there's some other logic following on from `Evaluate` does the process get stopped or do you have to fire the interrupt again?

dspy/evaluate/evaluate.py
```diff
@@ -94,45 +93,61 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
+                                f"Error for example in dev set: \t\t {e}\n\twith inputs:\n\t\t{arg}\n\nStack trace:\n\t{traceback.format_exc()}"
+                            )
+                        else:
+                            dspy.logger.error(
+                                f"Error for example in dev set: \t\t {e}. Set `provide_traceback=True` to see the stack trace."
+                            )
+                        continue  # Skip this future and continue
+
+                    if prediction is job_cancelled:
+                        continue
```
mikeedjones, line 137: nice to return early so maybe handle this case first?

dspy/evaluate/evaluate.py
```diff
@@ -94,45 +93,61 @@ def _execute_multi_thread(self, wrapped_program, devset, num_threads, display_pr
...
+                    ntotal += 1
+                    self._update_progress(pbar, ncorrect, ntotal)
+                pbar.close()
+        except KeyboardInterrupt:
+            self.cancel_jobs.set()
+            dspy.logger.warning("Received KeyboardInterrupt. Cancelling evaluation.")
+            # Cancel remaining futures
+            for f in futures:
+                f.cancel()
+            raise
```
mikeedjones, line 150: maybe wrap this up as a local function since the same as above.

dspy/evaluate/evaluate.py
mikeedjones: I think this is better - any event based signaling using `threading.Event` would require the whatever is spawning Threads to also listen for that event so not really viable. Left some other comments/nits :)

okhat:
@XenonMolecule I love this and @mikeedjones thanks for the review :D 

I want to merge this and to use it. I see failing tests. Is that expected? Are there blockers for merge?

dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -400,7 +402,7 @@ def _bootstrap_fewshot_examples(
                 teacher_settings=self.teacher_settings,
                 seed=seed,
                 metric_threshold=self.metric_threshold,
-                rng=self.rng,
+                parallel_bootstrapping=self.parallel_bootstrapping,
```
okhat, line 405: interesting, i gotta check what the semantics are of defaulting to self.attribute.....

okhat: btw why False by default?

dspy/teleprompt/utils.py
```diff
@@ -352,8 +353,44 @@ def create_n_fewshot_demo_sets(
...
+        return seed, predictor_demos
+
+    if parallel_bootstrapping:
+        # Use ThreadPoolExecutor to parallelize the creation of candidate sets
+        num_threads = total_candidate_sets  # As per your request
+
+        # Dictionary to store the results in order
+        seed_to_result = {}
+
+        with ThreadPoolExecutor(max_workers=num_threads) as executor:
```
okhat, line 370: this code is probably wrong for a very very subtle reason that probably has zero effect for @XenonMolecule 's usecase but will have for some other people

Launching threads in DSPy is tricky, because (like dspy.Evaluate does) you need to handle dspy.settings.stack_by_tid or whatever it's called

okhat: We do have a good oppty here to refactor dspy.Evaluate's thread launcher into its own reuseable parallelizer for dspy and then use it here too

dspy/teleprompt/mipro_optimizer_v2.py
XenonMolecule: Partially it's in case it breaks something I didn't test for and partially because the printing is currently ugly, but I can change the logic to support num_threads

dspy/teleprompt/utils.py
XenonMolecule: This probably wasn't sufficient but I used ctrl-f on the repo and used GitHub search and both times the only use of `dspy.settings.stack_by_thread` was in Evaluate.py?  Might've missed it somewhere else though

XenonMolecule:
> @XenonMolecule I love this and @mikeedjones thanks for the review :D
> 
> I want to merge this and to use it. I see failing tests. Is that expected? Are there blockers for merge?

The failing tests only started when I pushed a commit that changed comments, so not sure what happened.  Looking into it.

dspy/teleprompt/mipro_optimizer_v2.py
XenonMolecule: How do we want to handle `num_threads`?  Theoretically BootstrapFewshotWithRandomSearch will call Evaluate which means if num_threads=6 one interpretation is allowing 6*6=36 simultaneous threads, but this is likely not what a user expects.  We could allow two different variables for num_threads (num_threads_eval, num_threads_bootstrap), or we could set each to the sqrt of num_threads (with max bootstrap threads being limited to num_candidates+3)

okhat: Oh, hmm you're mentioning BootstrapFewshotWithRandomSearch? I thought we're parallelizing across BootstrapFewShot. Is that wrong?

dspy/teleprompt/utils.py
okhat: Hey @XenonMolecule , I don't get the response. I'm saying we cannot launch threads with ThreadProolExecutor plainly. It's missing the important steps that dspy.Evaluate does when launching a thread. Not sure if you addressed this comment above (I think no).

XenonMolecule: No I don't think I addressed this comment fully, my point was that I wasn't sure where else `dspy.settings.stack_by_thread` was being used, so I was unsure of the expected behavior for this setting.  It looked like it was only used in Evaluate.py and nowhere else in the repo.  Maybe we can chat offline about how `dspy.settings.stack_by_thread` needs to be handled.

dspy/teleprompt/mipro_optimizer_v2.py
XenonMolecule: Currently I am parallelizing **calls to** BootstrapFewShot rather than BootstrapFewShot itself.  I see what you are saying though and can refactor to that instead.  MIPRO has a util function that basically runs BootstrapFewshotWithRandomSearch just without the final evaluation, so my comment earlier about threads within threads is not a concern if we pick to parallelize EITHER **calls to** BootstrapFewShot or BootstrapFewShot.  Parallelizing BootstrapFewShot has the benefit that it can work for both RandomSearch and MIPRO out of the box.

dspy/teleprompt/utils.py
isaacbmiller: This will very subtly ruin strange things if not handled correctly. Speaking from experience but don't let this one fall to the wayside.

dspy/teleprompt/mipro_optimizer_v2.py
okhat: >Parallelizing BootstrapFewShot has the benefit that it can work for both RandomSearch and MIPRO out of the box.

Yes, that's the better one.

dspy/teleprompt/utils.py
okhat: Yes, let's make sure we handle threading correctly.

dilarasoylu: > We do have a good oppty here to refactor dspy.Evaluate's thread launcher into its own reuseable parallelizer for dspy and then use it here too

This would be nice (not sure if it should be a part of this PR) Moreover, it would be even nicer if all the logic for handling various control signals can be handled by this parallelizer, ensuring that we don't deal with them every time we make use of parallelism in DSPy.

Issue #1615, opened 2024-10-12
How can I check the current version of the dspy package? 


baiyin:
I've tried the following methods without success

```python
import dspy
print(dspy.__version__) 
```

and

```python
import importlib.metadata
importlib.metadata.version('dspy')
```

okhat:
Thank you @baiyin ! @hmoazam is working on adding `dspy.__version__`

okhat:
If this is urgent, you can get this information from pip:

```python
import importlib.metadata

# Get the version of the 'dspy' package
version = importlib.metadata.version('dspy')
print(f'dspy version: {version}') 
```

Output:
dspy version: 2.5.6

PR #1642, opened 2024-10-17
Fix the package version and name

chenmoneygithub:
The version in pyproject.toml and setup.py are wrong, we should use `2.5.13.dev0` for now. Basically we should write it as the next minor release + `.dev0`, which is a standard way (not officially documented though).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,9 +4,9 @@ build-backend = "setuptools.build_meta"
 
 [project]
 #replace_package_name_marker
-name="dspy-ai"
+name="dspy"
 #replace_package_version_marker
-version="2.5.8"
+version="2.5.12"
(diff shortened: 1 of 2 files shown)


okhat:
See #1615. Can we automate this in the release workflow perhaps?

chenmoneygithub:
@okhat We can update this version frequently to match the actual pypi version, just need to standardize it as part of the release process. We can set up automated release pipeline that handles this version as well, but IMO manually updating it is not cumbersome. 

Speaking of the automated release, I am thinking about automatically publishing to pypi when we push a new github tag. Similar to https://github.com/keras-team/keras-hub/blob/master/.github/workflows/publish-to-pypi.yml, does it work for you?

PR #1647, opened 2024-10-18
Feature/human-in-the-loop-teleprompt

burtenshaw:
## 📝 Changes Description

...

## ✅ Contributor Checklist

## ⚠️ Warnings

### DIFF
--- a/dspy/teleprompt/__init__.py
+++ b/dspy/teleprompt/__init__.py
@@ -12,3 +12,4 @@ from .signature_opt_bayesian import BayesianSignatureOptimizer
 from .teleprompt import *
 from .teleprompt_optuna import *
 from .vanilla import *
+from .human import *
--- a/dspy/teleprompt/human.py
+++ b/dspy/teleprompt/human.py
@@ -1,0 +1,201 @@
+import random
+
+from IPython.utils import data
+
+from dspy.evaluate.evaluate import Evaluate
+from dspy.teleprompt.teleprompt import Teleprompter
+
+from .bootstrap import BootstrapFewShot
+from .vanilla import LabeledFewShot
+
+# TODO: Don't forget dealing with the raw demos.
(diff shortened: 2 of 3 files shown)


okhat:
Hey @burtenshaw ! This looks super awesome. Can we help with upgrading the notebook to 2.5?

From glancing at the new code, I can't immediately understand which part involves human input. Is it happening due to the API of `argilla_dataset`?

burtenshaw:
@okhat Thanks for checking this out!

> Can we help with upgrading the notebook to 2.5?
That would be great. 

> Is it happening due to the API of argilla_dataset?
Correct. The notebook isn't clear on this. When I push the records to argilla at the beggining I assign all of the answers from GSM8k to annotator responses using a `mapping`: 

```python
rg_dataset.records.log(
    [dict(example) for example in trainset], mapping={"answer": "answer.response"}
)
```

I did this to skip manual annotation from the notebook and make it a standalone example, but I can see it's easy to miss. How do you think I should improve it?

okhat:
Re: 2.5, it's actually pretty simple --- https://github.com/stanfordnlp/dspy/blob/main/examples/migration.ipynb

We can figure out those details once the bigger Qs are out of the way. Thanks a ton!

burtenshaw:
Ok @okhat. I'll try to give more context.

 In the the `BootstrapFewShotWithHuman` class, the human reviews the few shot examples` answers. This is useful when the data relates to a specific domain.

Within the notebook example, the human isn't really adding anything. Because GSM8k has gold standard labels.

I'm using GSM8k as an example dataset, but in reality the user would start from a dataset that was unlabelled. Probably in a specific domain, that the human reviewer understood.

okhat:
Awesome, so the human reviews the bootstrapped labels coming out of the LM? And then edits them?

burtenshaw:
> Awesome, so the human reviews the bootstrapped labels coming out of the LM? And then edits them?

Yes. That's correct. The human could select samples and answers.

PR #1655, opened 2024-10-20
Merge branch 'main' of https://github.com/stanfordnlp/dspy

drawal1:
## 📝 Changes Description

...

## ✅ Contributor Checklist

## ⚠️ Warnings


okhat:
Amazing. Having some discussions on Discord at https://discord.com/channels/1161519468141355160/1294140517764042794/1297707179054207028

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+
+from pydantic import BaseModel
+from pydantic_core import PydanticUndefined
+from pydantic.fields import FieldInfo
+
+import dspy
+from dspy import InputField, OutputField, Signature
+
+
+class TypedPredictorSignature:
```
okhat, line 12: Starting to leave notes here. FIrst thought: we don't want to change the normal interface of DSPy, e.g. no distinct classes for all inputs and for all outputs.

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        if prefix_instructions:
+            prefix_instructions += "\n\n"
+        instructions = prefix_instructions + "Use only the available information to extract the output fields.\n\n"
+        dspy_fields = {}
+        for field_name, field in pydantic_class_for_dspy_input_fields.model_fields.items():
+            if field.default and 'typing.Annotated' in str(field.default):
+                raise ValueError(f"Field '{field_name}' is annotated incorrectly. See 'Constraints on compound types' in https://docs.pydantic.dev/latest/concepts/fields/")
+
+            is_default_value_specified, is_marked_as_optional, inner_field = cls._process_field(field)
```
okhat, line 35: Interesting to see special support for optional fields hmm, do we need optional fields? Or are Optional[.] types enough?

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+            if is_marked_as_optional:
+                if field.default is None or field.default is PydanticUndefined:
+                    field.default = 'null'
+                field.description = inner_field.description
+                field.examples = inner_field.examples
```
okhat, line 40: interesting to see examples and metadata per field

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+                field.metadata = inner_field.metadata
+                field.json_schema_extra = inner_field.json_schema_extra
+            else:   
+                field.validate_default = False
+
+            if field.default is PydanticUndefined:
+                raise ValueError(
+                    f"Field '{field_name}' has no default value. Required fields must have a default value. "
```
okhat, line 66: ah interesting concept here, requiring a default value for every required output field, why?

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+                    "Change the field to be Optional or specify a default value."
+                )
+            
+            output_field = OutputField(desc=field.description if field.description else "")
+            dspy_fields[field_name] = (field.annotation, output_field)
+
+            instructions += f"When extracting '{field_name}':\n"
+            instructions += f"If it is not mentioned in the input fields, return: '{field.default}'. "
```
okhat, line 74: Interesting choice of words. "When extracting" and "return". Something to keep in mind.

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+
+            examples = field.examples
+            if examples:
+                quoted_examples = [f"'{example}'" for example in examples]
```
okhat, line 78: quotes will fail on complex values?

tests/functional/test_typed_predictor_signature.py
```diff
@@ -0,0 +1,82 @@
...
+    @field_validator("name", mode="wrap")
+    @staticmethod
+    def validate_name(name, handler):
+        try:
+            return handler(name)
+        except ValidationError:
+            return 'INVALID_NAME'
+
+    name: Annotated[str,
+                    Field(default='NOT_FOUND', max_length=15,
```
okhat, line 21: why use Annotated instead of assignment = Field(...)?

dspy/functional/typed_predictor_signature.py
drawal1: Agreed

tests/functional/test_typed_predictor_signature.py
drawal1: From Pydantic docs - "In case you use field constraints with compound types, an error can happen in some cases. To avoid potential issues, you can use Annotated:"

See https://docs.pydantic.dev/latest/concepts/fields/#validation-alias

dspy/functional/typed_predictor_signature.py
```diff
@@ -0,0 +1,130 @@
...
+                instructions += f"Example values are: {', '.join(quoted_examples)} etc. "
+
+            if field.metadata:
+                constraints = [meta for meta in field.metadata if 'Validator' not in str(meta)]
+                if field.json_schema_extra and 'invalid_value' in field.json_schema_extra:
+                    instructions += f"If the extracted value does not conform to: {constraints}, return: '{field.json_schema_extra['invalid_value']}'."
```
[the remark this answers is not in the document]
drawal1: Detecting missing and invalid values is a "must-have" requirement for production-quality apps. Here, I am basically trying to avoid the multiple retries that the framework makes when the extracted value does not match the constraints specified in the Pydantic BaseModel. 

The LM may be incorrectly extracting the parameter or the user has specified an invalid value. In both cases, the system should not hallucinate the closest value that matches the actual input. Rather, it should flag the invalid field and let the application handle the error with a proper error-correction workflow ("Provided value was invalid. It must be... Here are some examples... Did you mean...?")

dspy/functional/typed_predictor_signature.py
drawal1: Only issue I have seen is the LM sometimes returning string values enclosed in quotes. For example, returning "'John Doe'", instead of "John Doe". But this is easy to strip. It correctly extracts numbers without putting them in quotes but I guess the prompt could be enhanced to detect if the example values should be quoted. Any other suggestions on how this could be improved?

Also,  re. complex types - isn't that what I am testing with complex Input and Output BaseModels? So may be the code for pulling apart complex types and building a custom signature is unavoidable. Thoughts?

dspy/functional/typed_predictor_signature.py
drawal1: Seems to work. Any pitfalls here? Are you thinking if the field name was "extracted_name" or field.default was "return"? Not quite sure how to get around that

dspy/functional/typed_predictor_signature.py
drawal1: Users may provide input without passing the value of a required parameter. So how to indicate that the required value was missing in the input, without throwing an exception? And without forcing the LM to hallucinate a required missing value?

Luckily, you can specify that Pydantic default value should not be subject to validation. This means that regardless of the type (str, int, float, ...), you can specify a default value of "NOT_FOUND" and this signature will correctly detect and return it without hallucinating. 

No wrap validators necessary, which in any case don't distinguish between invalid and missing

dspy/functional/typed_predictor_signature.py
drawal1: these are definitely valuable signals. We should leverage Pydantic field metadata instead of duplicating in DSPy signature fields

dspy/functional/typed_predictor_signature.py
drawal1: Pydantic allows multiple ways to specify a field schema, including Optional property. It could be on the field spec inside the annotation or outside. IMO, we should not let users guess what pydantic field metadata we do and don't support. If we can support it, we should

dspy/functional/typed_predictor_signature.py
drawal1: The unique nature of agentic applications is that we cannot assume user input is being validated in the UI for required input. Combine this with LM ability to hallucinate when forced to do so...

dspy/functional/typed_predictor_signature.py
drawal1: Update: I am no longer using examples. It turns out LLM's will incorrectly return an example as the parameter value instead of "NOT_FOUND"

isaacbmiller:
Closing as stale - superseded by adapter based structured outputs

Issue #556, opened 2024-03-05
How do I optimise for F1 score?


umarbutler:
The documentation states that '[f]or simple tasks, [a metric] could be just "accuracy" or "exact match" or "F1 score". This may be the case for simple classification or short-form QA tasks', yet it does not clarify how F1 scores can be used to optimize programs. Is that possible? And if so, how?

The `metric` function that is passed to `teleprompter.compile` seems to take `gold` and `pred` as inputs, which are single classifications, so I am unable to see how you could calculate an F1 score based off of that. It would be really helpful to have that ability since not all tasks should be optimised based on accuracy.

umarbutler:
It would be great if we could define a function that indicates whether a prediction is a true positive, false positive, true negative or false negative and then another function could be passed that takes a confusion matrix and calculates a score based on that. I don't just want to optimize F1 scores, I also want to be able to optimize for F-beta scores, MCC, precision, recall and other metrics calculatable based on confusion matricies.

Another way would be to input a `sklearn.metrics` function and a function for normalising values into `True` and `False`. That would also very helpful.

okhat:
https://dspy-docs.vercel.app/docs/building-blocks/metrics

umarbutler:
@okhat This does not answer my question. Please see:
> The documentation states that '[f]or simple tasks, [a metric] could be just "accuracy" or "exact match" or "F1 score". This may be the case for simple classification or short-form QA tasks', yet it does not clarify how F1 scores can be used to optimize programs. Is that possible? And if so, how?
> The metric function that is passed to teleprompter.compile seems to take gold and pred as inputs, which are single classifications, so I am unable to see how you could calculate an F1 score based off of that. It would be really helpful to have that ability since not all tasks should be optimised based on accuracy.

KevinGregory:
https://discord.com/channels/1161519468141355160/1161519642985111593/1234543339404267620

chrico-bu-uab:
@umarbutler I too had a need for access to the full confusion matrix as well as scoring by a better performance metric (semantic f1 is effectively just confidence-weighted accuracy), so I made these two classes in the code below.

https://github.com/chrico-bu-uab/dspy_confusion/blob/main/dspy/evaluate/confusion.py

fyears:
Hi I am also wondering how to compute the f1 score and optimize the prompt based on the f1 score, since I am doing a mult-label classification task. 

[...]

PR #1660, opened 2024-10-21
Confusion matrix-based metrics

chrico-bu-uab:
Since the existing frameworks don't allow for confusion matrix-based evaluators or optimization, I created additional evaluator and optimization classes (`Confusion` and `MCCBootstrapFewShotWithRandomSearch`, respectively).

These additions serve as a solution to this issue: [https://github.com/stanfordnlp/dspy/issues/556](url)

### DIFF
--- a/dspy/evaluate/confusion.py
+++ b/dspy/evaluate/confusion.py
@@ -45,10 +45,11 @@ class Confusion:
         return match.group(0) if match else None
 
     def construct_matrix(self, preds, devset):
-        # use answers from devset to get weights
-        weight = {k: 1 / v for k, v in Counter([self.extract(arg) for _, arg in devset]).items()}
-
         labels = self.labels
+
+        # use answers from devset to get weights
+        weight = {k: 1 / v for k, v in Counter([self.extract(arg) for _, arg in devset]).items()} \
+            if self.use_class_weight else {k: 1 for k in labels}
 
         # Initialize the confusion matrix
         confusion_matrix = np.zeros((len(labels), len(labels)), dtype=np.float64)
--- a/dspy/evaluate/confusion.py
+++ b/dspy/evaluate/confusion.py
@@ -1,10 +1,10 @@
+import random
 import re
-from collections import Counter
+
 import numpy as np
+
 from dspy import LabeledFewShot, BootstrapFewShot
 from dspy.evaluate.evaluate import *
-import random
-
 from dspy.teleprompt.teleprompt import Teleprompter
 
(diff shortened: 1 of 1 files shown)


stevegbrooks:
Very much support this PR! How do we get attention from a maintainer?

chrico-bu-uab:
Have we tried `import maintainer_attention from dspy`?

okhat:
(looks like importing maintainer_attention works)

chrico-bu-uab:
I haven't tried making each example instance carry a set of examples yet. I still want separate predictions for each example, though, and to potentially parallelize their evaluations. It's just the metrics that need to be calculated all at once.

I'm not sure I understand the question about optimization. I mimicked the Evaluator and BootstrapFewShotWithRandomSearch classes pretty exactly (I think), so however those two interact with optimization should have transferred over to my classes.

I just thought confusion matrices are required often enough to warrant a dedicated framework, or even be the default metric. I rarely use straight accuracy for any ML. But I may be old-school.

I appreciate the feedback and understand if having the user handle this instead makes more sense for the community.

stevegbrooks:
Hi @okhat - also thanks for that explanation.

Would you be able to post some example code to show how to do this: "Just make each dspy.Example carry a set of examples and handle that in your metric and your program. Then the normal dspy.Evaluate works just fine, and so do all optimizers."?

piemasbi:
Thanks for all the nice posts, and thanks to @chrico-bu-uab for the code with the new classes. 

If I understand correctly, the current optimizers use the _mean_ of the metric value (which is computed on a per-example basis, let's call it the "sample metric") over the train/validation set for the optimization. What would be nice to have would be to have the optimizers computing other summary statistics on the sample metrics (for example, sensitivity, precision, f1-score, etc).

I see the issue with this, since it is not always meaningful to have summary metrics other than the mean for many sample metrics (think about the semantic f1 score, for example). Still, it would make sense to support optimizing for other summary metrics, especially when datasets are imbalanced.

Hope this makes sense, otherwise please correct me, and helps to pinpoint the issue.

stevegbrooks:
> Hi @okhat - also thanks for that explanation.
> 
> Would you be able to post some example code to show how to do this: "Just make each dspy.Example carry a set of examples and handle that in your metric and your program. Then the normal dspy.Evaluate works just fine, and so do all optimizers."?

Hi @okhat  just pinging you on this. Is this something you have offhand? Because I dont see any obvious way to implement it. 

Thanks!

PR #1664, opened 2024-10-21
Bump version

chenmoneygithub:
Bump version to be consistent with pypi release.


okhat:
Are you extremely fast @chenmoneygithub or is this automatic?

PR #1668, opened 2024-10-21
Update 6-optimizers.md

amanshresthaa:
Page no longer exists.
https://dspy-docs.vercel.app/deep-dive/optimizers/


okhat:
Thanks @amanshresthaa ! I agree we need to fix this link but I don't think this is the best edit?

krypticmouse:
I think the best way to handle this is to add redirects for root page to their corresponding 1st sub page or add `index.md` file for each folder maybe.

PR #1682, opened 2024-10-23
Fix error in Evaluate with display_table=True with outputs that cannot be converted to dict

dbczumar:
If I write a DSPy module that returns an object that is not coercable to `dict`, I cannot currently call `Evaluate()` on the module with `display_table=True` because `display_table()` assumes that all outputs are dictionaries. This PR fixes the issue and adds a test.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -227,7 +227,11 @@ class Evaluate:
             results = [(example, prediction, score) for _, example, prediction, score in predicted_devset]
 
         data = [
-            dict(example) | {"prediction": prediction, "correct": score}
+            (
+                merge_dicts(example, prediction) | {"correct": score}
+                if isinstance(prediction, dict)
+                else dict(example) | {"prediction": prediction, "correct": score}
+            )
             for _, example, prediction, score in predicted_devset
         ]
 
@@ -274,6 +278,23 @@ class Evaluate:
             return round(100 * ncorrect / ntotal, 2), results
 
         return round(100 * ncorrect / ntotal, 2)
+
+
+def merge_dicts(d1, d2) -> dict:
+    merged = {}
+    for k, v in d1.items():
+        if k in d2:
+            merged[f"example_{k}"] = v
+        else:
+            merged[k] = v
+
+    for k, v in d2.items():
+        if k in d1:
+            merged[f"pred_{k}"] = v
+        else:
+            merged[k] = v
+
+    return merged
 
 
 def truncate_cell(content) -> str:
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -227,7 +227,8 @@ class Evaluate:
             results = [(example, prediction, score) for _, example, prediction, score in predicted_devset]
 
         data = [
-            merge_dicts(example, prediction) | {"correct": score} for _, example, prediction, score in predicted_devset
+            dict(example) | {"prediction": prediction, "correct": score}
+            for _, example, prediction, score in predicted_devset
         ]
 
         result_df = pd.DataFrame(data)
@@ -273,23 +274,6 @@ class Evaluate:
             return round(100 * ncorrect / ntotal, 2), results
 
         return round(100 * ncorrect / ntotal, 2)
-
-
-def merge_dicts(d1, d2) -> dict:
-    merged = {}
-    for k, v in d1.items():
-        if k in d2:
-            merged[f"example_{k}"] = v
-        else:
-            merged[k] = v
-
-    for k, v in d2.items():
-        if k in d1:
-            merged[f"pred_{k}"] = v
-        else:
-            merged[k] = v
-
-    return merged
 
 
 def truncate_cell(content) -> str:
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -226,10 +226,17 @@ class Evaluate:
         if return_outputs:  # Handle the return_outputs logic
             results = [(example, prediction, score) for _, example, prediction, score in predicted_devset]
 
+        def prediction_is_dictlike(prediction):
+            try:
+                dict(prediction)
+                return True
+            except Exception:
+                return False
+
         data = [
             (
                 merge_dicts(example, prediction) | {"correct": score}
-                if isinstance(prediction, dict)
+                if prediction_is_dictlike(prediction)
                 else dict(example) | {"prediction": prediction, "correct": score}
             )
             for _, example, prediction, score in predicted_devset
--- a/tests/evaluate/test_evaluate.py
+++ b/tests/evaluate/test_evaluate.py
@@ -124,6 +124,7 @@ def test_evaluate_call_bad():
 @pytest.mark.parametrize(
     "program_with_example",
     [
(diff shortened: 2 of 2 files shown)


tests/evaluate/test_evaluate.py
```diff
@@ -120,14 +121,38 @@ def test_evaluate_call_bad():
...
+    "program_with_example",
+    [
+        (Predict("question -> answer"), new_example("What is 1+1?", "2")),
+        (
+            # Create a program that extracts entities from text and returns them as a list,
+            # rather than returning a Predictor() wrapper. This is done intentionally to test
+            # the case where the program does not output a dictionary-like object because
+            # Evaluate() has failed for this case in the past
+            lambda text: TypedPredictor("text: str -> entities: List[str]")(text=text).entities,
+            dspy.Example(text="United States", entities=["United States"]).with_inputs("text"),
```
dbczumar, line 134: @okhat This appears to be a valid setup for DSPy evaluation - i.e. all other parts of evaluation work with such a `lambda` except for displaying table output. Let me know if there's a strong reason to disallow this and require all DSPy modules / programs to return `Prediction()` objects

tests/evaluate/test_evaluate.py
```diff
@@ -120,14 +121,38 @@ def test_evaluate_call_bad():
...
+@pytest.mark.parametrize(
+    "program_with_example",
+    [
+        (Predict("question -> answer"), new_example("What is 1+1?", "2")),
+        (
+            # Create a program that extracts entities from text and returns them as a list,
+            # rather than returning a Predictor() wrapper. This is done intentionally to test
+            # the case where the program does not output a dictionary-like object because
+            # Evaluate() has failed for this case in the past
+            lambda text: TypedPredictor("text: str -> entities: List[str]")(text=text).entities,
```
dbczumar, line 133: This case fails on `main`:

```
...
test_evaluate.py:163:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../dspy/evaluate/evaluate.py:229: in __call__
    data = [
../../dspy/evaluate/evaluate.py:230: in <listcomp>
    merge_dicts(example, prediction) | {"correct": score} for _, example, prediction, score in predicted_devset
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

d1 = Example({'text': 'United States', 'entities': ['United States']}) (input_keys={'text'}), d2 = ['United States']

    def merge_dicts(d1, d2) -> dict:
        merged = {}
        for k, v in d1.items():
            if k in d2:
                merged[f"example_{k}"] = v
            else:
                merged[k] = v

>       for k, v in d2.items():
E       AttributeError: 'list' object has no attribute 'items'

../../dspy/evaluate/evaluate.py:286: AttributeError
================================================================================= warnings summary =================================================================================
../../../miniconda3/envs/default/lib/python3.10/site-packages/pydantic/_internal/_config.py:291
  /Users/corey.zumar/miniconda3/envs/default/lib/python3.10/site-packages/pydantic/_internal/_config.py:291: PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.9/migration/
    warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)

../../../miniconda3/envs/default/lib/python3.10/site-packages/wandb/env.py:16
  /Users/corey.zumar/miniconda3/envs/default/lib/python3.10/site-packages/wandb/env.py:16: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
    from distutils.util import strtobool

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================= short test summary info ==============================================================================
FAILED test_evaluate.py::test_evaluate_display_table[True-True-program_with_example1] - AttributeError: 'list' object has no attribute 'items'
FAILED test_evaluate.py::test_evaluate_display_table[True-False-program_with_example1] - AttributeError: 'list' object has no attribute 'items'
FAILED test_evaluate.py::test_evaluate_display_table[True-1-program_with_example1] - AttributeError: 'list' object has no attribute 'items'
FAILED test_evaluate.py::test_evaluate_display_table[False-True-program_with_example1] - AttributeError: 'list' object has no attribute 'items'
FAILED test_evaluate.py::test_evaluate_display_table[False-False-program_with_example1] - AttributeError: 'list' object has no attribute 'items'
FAILED test_evaluate.py::test_evaluate_display_table[False-1-program_with_example1] - AttributeError: 'list' object has no attribute 'items'

```

dspy/evaluate/evaluate.py
```diff
@@ -227,7 +227,8 @@ def wrapped_program(example_idx, example):
             results = [(example, prediction, score) for _, example, prediction, score in predicted_devset]
 
         data = [
-            merge_dicts(example, prediction) | {"correct": score} for _, example, prediction, score in predicted_devset
+            dict(example) | {"prediction": prediction, "correct": score}
```
okhat, line 230: Thanks so much @dbczumar ! Everything looks great to me, besides the following questions.

Is the moving of `prediction` under a key called "prediction" intentional? What if the prediction has many fields? How will that get displayed? I imagine it breaks the niceness of columns, unless the table displayer is smart and shows some kind of sub-columns.

I do like getting rid of `merge_dicts` which is not needed in 3.9+.

dbczumar: @okhat Got it - makes perfect sense :). For dict-like outputs, I've restored the preexsting behavior. For non-dict-like outputs, I think including the full value in a `predictions` column makes sense.

dspy/evaluate/evaluate.py
```diff
@@ -226,8 +226,20 @@ def wrapped_program(example_idx, example):
...
+            try:
+                dict(prediction)
+                return True
+            except Exception:
+                return False
+
         data = [
-            merge_dicts(example, prediction) | {"correct": score} for _, example, prediction, score in predicted_devset
+            (
+                merge_dicts(example, prediction) | {"correct": score}
```
okhat, line 238: I vaguely presuppose that merge_dicts is there for legacy reasons, and that `example | prediction` may have the same effect, is that false? (it may be false)

okhat: oh, i see, merge_dicts does some kind of disambiguation instead of overwriting

dbczumar: Yeah, exactly

PR #1689, opened 2024-10-25
Add callback mechanism

B-Step62:
**Change Overview**:
Adding a robust callback mechanism to provide users an ability to inject custom hooks before/after executing DSPy components. The main use case right now is [MLflow Tracing](https://mlflow.org/docs/latest/llms/tracing/index.html), but this opens up more use cases. The callback design has been discussed offline.  

**Tests**:
- Unit test: `pytest tests/`
- Manual test with prototype MLflow tracing ([PR](https://github.com/mlflow/mlflow/pull/13543))


okhat:
Looks great to me! Thanks a ton @B-Step62 !

PR #1690, opened 2024-10-25
Add support for native parallel execution in DSPy

krypticmouse:
This pull request introduces significant refactoring and new functionality to the `dspy` package. The most important changes include the introduction of a `ParallelExecutor` class, updates to the `Evaluate` class to use this new executor, and the addition of a `batch` method in the `Module` class.

### Parallel Execution Enhancements:

* **`dspy/utils/parallelizer.py`**: Introduced the `ParallelExecutor` class to handle multi-threaded execution with error handling and progress tracking.
* **`dspy/evaluate/evaluate.py`**: Refactored the `Evaluate` class to use the `ParallelExecutor` for both single-threaded and multi-threaded execution, simplifying the code and removing redundant threading and error handling logic.

### New Functionality:

* **`dspy/primitives/program.py`**: Added a `batch` method to the `Module` class to process multiple examples in parallel using the `ParallelExecutor`. This method supports error handling and optional traceback information.


dspy/predict/parallel.py
```diff
@@ -0,0 +1,83 @@
...
+            num_threads=num_threads,
+            display_progress=True,
+            max_errors=self.max_errors,
+            provide_traceback=self.provide_traceback,
+        )
+
+        def process_pair(pair):
+            module, example = pair
+            thread_stacks = dspy.settings.stack_by_thread
+            creating_new_thread = threading.get_ident() not in thread_stacks
```
okhat, line 44: Should this code be here? I'd think this stuff about thread_stacks should be in parallelizer.py. Thoughts?

krypticmouse: Yea I've been juggling on this rn, it's not duplicated though I've removed the stuff from evaluate and added it here.

dspy/primitives/program.py
```diff
@@ -91,6 +93,43 @@ def activate_assertions(self, handler=backtrack_handler, **handler_args):
 
     #     return new_copy
 
+    def batch(
+        self,
+        examples: List[dspy.Example],
+        batch_size: int = 32,
+        max_errors: int = 10,
+        return_failed_examples: bool = False,
```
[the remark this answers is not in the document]
krypticmouse: Yea I think that makes sense!!

dspy/primitives/program.py
```diff
@@ -94,6 +95,43 @@ def activate_assertions(self, handler=backtrack_handler, **handler_args):
...
+
+        :param examples: List of dspy.Example instances to process.
+        :param batch_size: Number of threads to use for parallel processing.
+        :param max_errors: Maximum number of errors allowed before stopping execution.
+        :param return_failed_examples: Whether to return failed examples and exceptions.
+        :param provide_traceback: Whether to include traceback information in error logs.
+        :return: List of results, and optionally failed examples and exceptions.
+        """
+        # Create a list of execution pairs (self, example)
+        exec_pairs = [(self.forward, example.inputs()) for example in examples]
```
okhat, line 117: should this be just self instead of self.forward, thoughts?

krypticmouse: Yep can be self as well! Both work, `.forward` states function more explicitly so used that for readability but can switch to self.

okhat:
One important note: I think we should deepcopy the stack when we're forking a thread! Not just copy, but deep copy, because the lists in the values are mutable.

dspy/primitives/program.py
okhat: i prefer self

okhat:
Also important: we need to handle creating threads inside threads. This means every thread needs to "inherit" / fork / deepcopy the stack from its parent, not from the main stack.

ArslanS1997:
Hi do I use dspy.Parallel?

Issue #1540, opened 2024-09-25
How to get LangFuse trace_id


raghavbali:
Hey dspy team,
I am trying to use langfuse tracer to trace my usage and most importantly capture cost. The place where I am struggling at the moment is getting hold of ``trace_id`` for each request that is generated with LLMs (eg: openai). I am able to get to the whole list of traces using ``TracerObject.langfuse.fetch_traces()`` and then using the ``data`` attribute to get to ``total_cost`` but this requires a whole lot of processing to filter the actual traces I am looking for. 

Is there a cleaner way I can make use of ``fetch_trace_id()`` method without creating a custom tracer? From what I understand the default implementation is not passing any id and expecting LangFuse to generate random trace ids.

PS: _Kudos on the Amazing work in getting this integration in place_

okhat:
Thanks @raghavbali ! Can you tag the langfuse team? You can find who committed the relevant parts.

isaacbmiller:
@karthikscale3

karthikscale3:
I know the names are confusing, but I maintain Langtrace and not Langfuse 😄. But in case, you want to monitor cost, run experiments, visualize eval scores or capture checkpoints, Langtrace has native support for DSPy. You can find more information [here](https://docs.langtrace.ai/supported-integrations/llm-frameworks/dspy#dspy)

isaacbmiller:
Oops sorry about that @karthikscale3 !

isaacbmiller:
@xucailiang

xucailiang:
> @xucailiang

May I know how you use langfuse in dspy? Here are some simple usage instructions. I hope this helps you. [langfuse.md](https://github.com/stanfordnlp/dspy/blob/main/docs/api/language_model_clients/langfuse.md)

raghavbali:
Thanks for the quick response everyone.
@xucailiang : I am following the instructions on the linked README and using the following snippet to get to the most recent traces
```python
# capturing cost and latency
# start and end times are simple time-stamps captured before and after the dspy predict
total_time = end_time-start_time
tracers = langfuse.langfuse.fetch_traces()
total_cost = sum([t.total_cost for t in tracers.data if t.timestamp>=start_time and t.timestamp<end_time])
```

Not a very clean way and hence I wanted to get hold of the trace_id without going through the additional step of creating a customer tracer

xucailiang:
[...]

PR #1186, opened 2024-06-21
feature(dspy): dspy-integration-with-langfuse

xucailiang:
As mentioned in issue [#1179](https://github.com/stanfordnlp/dspy/issues/1179) , I wrote some code to integrate langfuse.
 when configured correctly, It will support the display of data from OpenAI, AzureOpenAI and Ollama on the Langfuse dashboard

PR #1714, opened 2024-10-29
Feature langfuse tracker update

xucailiang:
### Changes Overview
1. Modified the Langfuse integration in gpt3.py to use decorators for better code organization and readability.https://github.com/stanfordnlp/dspy/pull/1186#discussion_r1816503177

2. Enhanced langfuse_tracker.py with additional metric collection methods for improved statistics tracking.(https://github.com/stanfordnlp/dspy/issues/1540?notification_referrer_id=NT_kwDOBHJY27QxMjU3MDM3NzAzODo3NDYwMjcxNQ)

3. If you're using LiteLLM, you can enable Langfuse's excellent metric collection features by simply adding these configurations:[langfuse docs](https://langfuse.com/docs/integrations/litellm/tracing)
```
litellm.success_callback = ["langfuse"]
litellm.failure_callback = ["langfuse"]
```

### Key Features
- Streamlined integration process
- Enhanced metric collection capabilities
- Ready-to-use statistics tracking functions

### Usage
For LiteLLM users, simply add the callback configurations to enable Langfuse integration and utilize the pre-built metric collection functions from langfuse_tracker.py.

### Related Files
gpt3.py
langfuse_tracker.py
/tests/trackers


xucailiang:
@arnavsinghvi11 @okhat 

Made some updates to the Langfuse integration - would love your thoughts on this!

Cheers

dsp/trackers/langfuse_tracker.py
```diff
@@ -1,18 +1,26 @@
...
 import logging
 import os
-from langfuse.client import Langfuse, StatefulTraceClient, StatefulSpanClient, StateType
 from dsp.trackers.base import BaseTracker
 
+try:
+    from langfuse.client import Langfuse
+    from langfuse.decorators import observe
+except ImportError:
+    def observe():
```
marcklingen, line 12: Great idea to switch to the decorator to track any LLM used via dspy instead of relying on the langfuse openai intgeration. I'd suggest to capture token counts and model parameters as these otherwise get lost when making this change: https://langfuse.com/docs/sdk/python/decorators#log-any-llm-call

xucailiang: Hit the mark!
If an exception occurs here, it means that the langfuse package is not currently installed. So is it unnecessary for me to obtain the indicator data here? And users can also obtain indicator data through the `history` parameter without installing any tracking tools. 

So I am a little unsure whether to collect indicators here.

Is there anything I haven't considered? Please point it out to me, thank you!

marcklingen: Sorry for the confusion, I added the feedback to this line as I recognized that you switched to the `observe` decorator instead of the openai integration when reading this. I was not hinting at that any change at this line is necessary but more in general that you should log out additional llm call information as shown in the docs to maintain these metrics/

xucailiang:
> Thanks @xucailiang ! This seems incompatible with 2.5
> 
> https://github.com/stanfordnlp/dspy/blob/main/examples/migration.ipynb

Sorry, can I get more detailed information? I noticed that dspy2.5 integrates liteLLM, which makes me excited. Then the way to integrate langfuse is even simpler. This PR will also provide some more convenient functions in langfuse_tracker to count llm indicator data.

okhat:
Basically you're editing gpt3.py but that file will be deleted next week and is not very actively used by anyone, since it's deprecated.

xucailiang:
> Basically you're editing gpt3.py but that file will be deleted next week and is not very actively used by anyone, since it's deprecated.

Yes, I changed the import method of langfuse in `gpt3.py`. It is a good idea to delete the old modules next week, but considering that some old projects may not be able to migrate directly to the new version, I made this change, but I can also cancel this part of the change if necessary. For the new LM, the changes in the langfuse tracker are also meaningful, and some methods will be provided here to facilitate users to obtain LLM parameters.
I don't know if I missed anything

PR #1731, opened 2024-10-31
Add tenacity as a dependency to support LM retries with LiteLLM

dbczumar:
Add tenacity as a dependency to support LiteLLM retries

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,6 +137,7 @@ magicattr = "^0.1.6"
 litellm = "1.49.1"
 diskcache = "^5.6.0"
 json-repair = "^0.30.0"
+tenacity = ">=8.2.3"
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^6.2.5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,8 @@ dependencies = [
     "magicattr~=0.1.6",
     "litellm",
     "diskcache",
-    "json-repair"
+    "json-repair",
+    "tenacity>=8.2.3",
 ]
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -43,7 +43,8 @@ dependencies = [
     "magicattr~=0.1.6",
     "litellm",
     "diskcache",
-    "json-repair"
+    "json-repair",
+    "tenacity>=8.2.3",
```
dbczumar, line 47: LiteLLM uses `tenacity==8.2.3` for testing: https://github.com/BerriAI/litellm/blob/b73039b28337c04ac87eea8e0be11186f142f38f/requirements.txt#L46.

I've confirmed that 8.2.3 *and* 9.0.0 appear to work just fine via the following script using an Azure OpenAI + Databricks endpoint with an aggressive rate limit of 1QPM:

```
import litellm
import uuid


litellm.set_verbose=True
for i in range(100):
    print(litellm.completion(
        model="databricks/coreyazureopenai",
        messages=[{"role": "user", "content": "You are a helpful assistant." + uuid.uuid4().hex}],
        retry_strategy="exponential_backoff_retry",
        num_retries=10,
    ))
```

Accordingly, `>=8.2.3` appears appropriate

requirements.txt
```diff
@@ -1,17 +1,18 @@
...
 litellm<=1.49.1
+magicattr~=0.1.6
 openai
 optuna
 pandas
 pydantic~=2.0
 regex
 requests
 structlog
+tenacity>=8.2.3
```
dbczumar, line 16: This is the only material change. Everything else is just linter formatting

PR #1729, opened 2024-10-30
[WIP] Implement async 

isaacbmiller:
Beginning of an async forward call implementation.

Max num threads for my mac is 2666, and at 2666+ threads (max) async does start to outperform threading

PR #1734, opened 2024-11-01
(Feature) Async Support

CyrusOfEden:
🥁 the long awaited async implementation for DSPy 🔥 

All dspy.predict.* programs can now be used async!

```python
dspy.settings.configure(async_mode=True)
```

All programs are now awaitable:

```python
predict = dspy.Predict("question -> answer")
result = await predict(question="...")
assert result.answer == "..."
```

Custom programs must implement the `aforward` method, which has the same method signature as `forward`. Developers must take care not to block the event loop with synchronous function calls.

```python
class MyProgram(dspy.Module):
  async def aforward(...):
    ...
```

### Notable Changes

0. Added a @with_async_callbacks impl based off of @dbczumar's @with_callbacks impl
1. Retrievers now have a default async implementation that delegates to a threadpool
2. New LRU cache for completions, with a default maxlen of 10M. Can be overwritten with: `dspy.settings.configure(request_cache=LMRequestLRUCache(maxsize: int = ...)`. This prevents unbounded memory growth.
3. GLOBAL_HISTORY updated from a list to a deque with a maxlen of 10M, to prevent unbounded memory growth.

### Minor Changes

1. Request caching previously serialized, deserialized, then serialize the request JSON. I've removed the unnecessary deserialization step. 

### Notes

It looks like the ReAct tests are broken in main, and they're broken here too.

### Acknowledgements

Kudos to @isaacbmiller for his prototype in #1729, this wouldn't have been possible without him!


okhat:
This is AWESOME, thanks a lot @CyrusOfEden !!

Are you using a linter that's more opinionated than the default linting behavior in the library? I think a large fraction of the 24 edited files are just edited for style and that makes review / history a bit more complex than it needs to be for this PR.

Relatedly, there are some outdated ReAct tests that were commented out, we can keep them commented out until they're updated.

examples/async/dspy_async_impl.py
```diff
@@ -0,0 +1,62 @@
...
+    thread_main(num_items=num_items)
+    dspy.settings.configure(async_mode=True)
+    # print("Sleeping for 60 seconds")
+    # time.sleep(60)
+    asyncio.run(async_main(num_items))
+
+# edge cases I can think of:
+# - multiple modules in the chain
+# - async gets stuck/fails
+# - what happens to settings inside each async module?
```
okhat, line 62: These are great questions.

Threading in DSPy needs care, because of the singleton dspy.settings. We handle that carefully when launching threads, e.g. see the PR #1690 for how this worked before and how it will work after the PR is merged.

How is async thinking about this?

okhat:
tl;dr

1. I'd love a version of this PR that focuses on content updates. If you prefer, we could merge a (less aggressive) style reformat PR first for the relevant files.
2. For us to merge, we need to consider the implications on complex programs that involve, e.g. `with dspy.context` calls. Alternatively, I'm fine disallowing any .context and .configure calls in async programs *as long as* it throws an error, so people don't forget.

CyrusOfEden:
> I think a large fraction of the 24 edited files are just edited for style

Only files that have had changes have had the formatting applied, so there's no superfluous formatting included in the PR. 

My editor should've picked up the formatting rules, I can check in the afternoon today.

examples/async/dspy_async_impl.py
CyrusOfEden: These were Isaac's 😅

dspy/clients/lm.py
```diff
@@ -163,54 +225,135 @@ def copy(self, **kwargs):
         return new_instance
 
 
-@functools.lru_cache(maxsize=None)
-def cached_litellm_completion(request, num_retries: int):
-    return litellm_completion(
-        request,
+class LMRequestLRUCache(OrderedDict):
```
mikeedjones, line 228: is there an off-the-shelf async lru cache which can be used? https://github.com/aio-libs/async-lru for example

dspy/predict/predict.py
```diff
@@ -210,6 +192,49 @@ def forward(self, **kwargs):
...
+        if new_signature is not None:
+            signature = new_signature
+
+        if not all(k in kwargs for k in signature.input_fields):
+            present = [k for k in signature.input_fields if k in kwargs]
+            missing = [k for k in signature.input_fields if k not in kwargs]
+            print(f"WARNING: Not all input fields were provided to module. Present: {present}. Missing: {missing}.")
+
+        assert isinstance(lm, dspy.LM), "Async mode is only supported for dspy.LM"
+        completions = await get_adapter()(lm, config, signature, demos, kwargs, _parse_values=self._parse_values)
```
mikeedjones, line 228: deduplicate the wrapper code?

dspy/predict/react.py
```diff
@@ -94,6 +116,35 @@ def format(trajectory_: dict[str, Any], last_iteration: bool):
...
+    async def aforward(self, **input_args):
+        trajectory = {}
+
+        def format(trajectory_: dict[str, Any], last_iteration: bool):
+            adapter = dspy.settings.adapter or dspy.ChatAdapter()
+            signature_ = dspy.Signature(f"{', '.join(trajectory_.keys())} -> x")
+            return adapter.format_fields(signature_, trajectory_, role="user")
+
+        for idx in range(self.max_iters):
+            pred = await self.react(
```
mikeedjones, line 132: deduplicate the wrapper code?

examples/async/dspy_async_impl.py
```diff
@@ -0,0 +1,62 @@
...
+    print(f"Total time (threaded): {end_time - start_time} seconds")
+
+
+if __name__ == "__main__":
+    num_items = 500
+    dspy.settings.configure(async_mode=False)
+    thread_main(num_items=num_items)
+    dspy.settings.configure(async_mode=True)
+    # print("Sleeping for 60 seconds")
+    # time.sleep(60)
```
mikeedjones, line 56: rm comments?

examples/async/dspy_async_impl.py
mikeedjones: Is the async execution of mutiple modules not just up to the implementation of the parent module?

mikeedjones (commented):
We've been using this branch in our fastAPI app and working really well so far - amazing work!

Our only concern rn is future drift between the sync and async implementations since the pre/post processing code is duplicated - have called out where I think it could be packaged up as common functions.

ArslanS1997:
Hi how can I use this, is this merged?

dspy/clients/lm.py
CyrusOfEden: I thought of that first but because Omar wanted the async and sync to share the same cache I implemented it this way

CyrusOfEden:
Closing this in favour of #1806

swiftdevil:
#1806 uses multi-threading to achieve an async-like behavior. Does this change use true async (ie, run in a single thread)? If so, then this PR would be preferable in terms of adherence to async principles.

PR #1735, opened 2024-11-01
Add `dspy.Embedding`

chenmoneygithub:
Very simple `dspy.Embedding` supports:

- Hosted embedding model, we use litellm to send the request, and only take the embedding fields from the response.
- Custom embedding model, users can pass a custom callable to `dspy.Embedding`, and the output is just the custom callable's output.

Added unit test for both scenarios. 

Confirmed that the cache works:

```
(dspy) *[dspy-embedding][~/Documents/mlflow_team/dspy]$ python3
Python 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 10:07:17) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import dspy;import time;embedder = dspy.Embedding("openai/text-embedding-ada-002")
>>> start=time.time();geez = embedder(["oh my god", "holy geese and elephant", "ohhhh hoover dam"]);print("TOTAL TIME: ", time.time()-start)
TOTAL TIME:  0.963151216506958
>>> start=time.time();geez = embedder(["oh my god", "holy geese and elephant", "ohhhh hoover dam"]);print("TOTAL TIME: ", time.time()-start)
TOTAL TIME:  0.009987115859985352
>>> start=time.time();geez = embedder(["oh my god", "holy geese and elephant", "ohhhh hoover dam"]);print("TOTAL TIME: ", time.time()-start)
TOTAL TIME:  0.010275125503540039
```

PR #1739, opened 2024-11-01
WIP - dspy.RM/retrieve refactor

arnavsinghvi11:
refactoring for cleaner retriever interface in DSPy


dspy/retrieve/embedder.py
```diff
@@ -0,0 +1,16 @@
+from typing import Callable, List, Optional 
+
+
+class Embedder:
```
arnavsinghvi11, line 4: #1735 will go here once merged @chenmoneygithub

dspy/clients/embedding.py
```diff
@@ -41,21 +42,27 @@ class Embedding:
...
+        embedder = dspy.Embedding(embedding_function=my_embedder)
         embeddings = embedder(["hello", "world"])
 
         assert embeddings.shape == (2, 10)
         ```
     """
 
-    def __init__(self, model):
-        self.model = model
+    def __init__(self, embedding_model: Union[str, Callable[[List[str]], List[List[float]]]] = 'text-embedding-ada-002', embedding_function: Optional[Callable[[List[str]], List[List[float]]]] = None):
```
chenmoneygithub, line 57: Why do we need this change? It's a common pattern to have model be string for hosted model, and callable for custom model, as most model instances are callable themselves. 

Btw, if we add type annotation, we shouldn't assume the arg type of the callable, which could be easily violated by users' custom function.

dspy/retrieve/retrieve.py
```diff
@@ -16,13 +18,20 @@ def single_query_passage(passages):
         passages_dict["passages"] = passages_dict.pop("long_text")
     return Prediction(**passages_dict)
 
+@lru_cache(maxsize=None)
+def warn_once(msg: str):
+    logging.warning(msg)
```
chenmoneygithub, line 23: We have standardized the logging pattern to be: 

```
import logging

logger.info(...)
```

For example: https://github.com/stanfordnlp/dspy/blob/943ee02d4cf9a630c89136ff6dc6b9f6ebf805dd/dspy/utils/callback.py#L12

dspy/retrieve/retrieve.py
```diff
@@ -16,13 +18,20 @@ def single_query_passage(passages):
...
 
 class Retrieve(Parameter):
     name = "Search"
     input_variable = "query"
     desc = "takes a search query and returns one or more potentially relevant passages from a corpus"
 
     def __init__(self, k=3, callbacks=None):
+        warn_once(
+            "Existing retriever integrations under dspy/retrieve inheriting `dspy.Retrieve` are deprecated and will be removed in DSPy 2.6+. \n"
```
chenmoneygithub, line 32: We can be specific here: will be removed in DSPy 2.7 release. It's usually nice to give people a clear timeline.

dspy/retriever/__init__.py
```diff
@@ -0,0 +1 @@
+from .retriever import Retriever
```
chenmoneygithub, line 1: Nit: add an empty line at EOF

dspy/retriever/colbertv2_retriever.py
```diff
@@ -0,0 +1,76 @@
+from typing import Any, Union, Optional
```
chenmoneygithub, line 1: we should be able to delete the suffix `_retriever` from the file name.

dspy/retriever/colbertv2_retriever.py
```diff
@@ -0,0 +1,76 @@
...
+import dspy
```
chenmoneygithub, line 2: Usually it's preferred to do direct imports `from dspy.retriever.retriever import Retriever` than `import dspy`, which is more robust against circular dependencies.

dspy/retriever/colbertv2_retriever.py
```diff
@@ -0,0 +1,76 @@
...
+from dsp.utils import dotdict
+import requests
+import functools
+
+class ColBERTv2(dspy.Retriever):
+    """
+    ColBERTv2 Retriever for retrieval of top-k most relevant text passages for given query.
+
+    Args:
+        post_requests (bool): Determines if POST requests should be used
```
chenmoneygithub, line 12: The arglist is outdated.

dspy/retriever/colbertv2_retriever.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Example:
+        from dspy.retriever.colbertv2_retriever import ColBERTv2
+        results = ColBERTv2(url='http://20.102.90.50:2017/wiki17_abstracts')(query, k=5).passages
+        print(results)
+    """
+    def __init__(self, url: str = "http://0.0.0.0", port: Optional[Union[str, int]] = None, post_requests: bool = False):
+        super().__init__(embedder=None)
+        self.post_requests = post_requests
+        self.url = f"{url}:{port}" if port else url
```
chenmoneygithub, line 27: Is this port required? If so let's raise an ValueError for clear action.

dspy/retriever/colbertv2_retriever.py
```diff
@@ -0,0 +1,76 @@
...
+@CacheMemory.cache
+def colbertv2_post_request_v2(url: str, query: str, k: int):
+    headers = {"Content-Type": "application/json; charset=utf-8"}
+    payload = {"query": query, "k": k}
+    res = requests.post(url, json=payload, headers=headers, timeout=10)
+
+    return res.json()["topk"][:k]
+
+
+@functools.cache
```
chenmoneygithub, line 70: why do we need this `_wrapped` function? seems it's not adding extra logic except from some cache, call we move the annotations to `colbertv2_post_request_v2`?

dspy/clients/embedding.py
```diff
@@ -1,8 +1,8 @@
 import litellm
 import numpy as np
+from typing import Callable, List, Union, Optional
 
-
-class Embedding:
+class Embedder:
```
chenmoneygithub, line 5: I chose name `Embedding` because that seems to be the name for most packages:

- Langchain uses `XXXEmbeddings`
- LlamaIndex uses `XXXEmbedding`
- OpenAI python SDK uses `client.embeddings.create`

I didn't do a thorough search, Is there some package that is using the name `Embedder`?

dspy/retriever/colbertv2_retriever.py
arnavsinghvi11: just an optional i believe. our examples typically use `'http://20.102.90.50:2017/wiki17_abstracts'` where the port can be specified in the URL

dspy/retriever/colbertv2_retriever.py
arnavsinghvi11: yea i think we can simplify it but will defer to @okhat as this was existing behavior from colbertv2.py

PR #1745, opened 2024-11-03
added the log size increase and update the documentation for the signatures

sushmanthreddy:
1. Updating the documentation for th signatures 
2. Increased the logo size and moved the communities and faq right side for better design


docs/mkdocs.yml
```diff
@@ -81,12 +81,27 @@ nav:
...
+  - '' : '' 
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
+  - '' : ''
```
krypticmouse, line 93: Is this intentional? What is this doing?

docs/mkdocs.yml
```diff
@@ -1,4 +1,4 @@
-site_name: DSPy
+site_name: DSPy Documentation
```
krypticmouse, line 1: @okhat might be able to judge this better but I think we can keep this as DSPy? It's a mix of docs and landing page and everything in between imo

docs/docs/stylesheets/extra.css
```diff
@@ -0,0 +1,33 @@
+/* Custom styles for logo */
+.md-logo {
+    width: auto !important;
```
krypticmouse, line 3: Avoid !important whenever you can and make sure this file doesn't break existing style.

krypticmouse (commented):
LGTM but left some comments!

docs/docs/stylesheets/extra.css
sushmanthreddy: yeah doesn't break any existing style actually it just makes logo bigger
Screenshot 2024-11-06 at 5 19 54 PM

sushmanthreddy: @okhat has told me to do, this way actually for better design

docs/mkdocs.yml
sushmanthreddy: yeah @okhat  told me to do that way to create space  between tutorial and community ,FAQ

PR #1749, opened 2024-11-04
added TypedChainOfThoughtWithHint

saifulhaq-hyprbots:
Code for TypedChainOfThoughtWithHint
@okhat


okhat:
Thank you @saifulhaq-hyprbots ! TypedPredictors are deprecated and are not needed any more. You can pass types to normal predictors! They work out of the box.

PR #1753, opened 2024-11-05
LMs: retry with exponential backoff for a limited set of error codes

dbczumar:
LMs: retry with exponential backoff for a limited set of error codes


dspy/clients/lm.py
```diff
@@ -32,7 +33,7 @@ def __init__(
         cache: bool = True,
         launch_kwargs: Optional[Dict[str, Any]] = None,
         callbacks: Optional[List[BaseCallback]] = None,
-        num_retries: int = 3,
```
dbczumar, line 35: 3 retries is insufficient to overcome rate limiting in production conditions

dspy/clients/lm.py
```diff
@@ -32,7 +33,7 @@ def __init__(
...
+        num_retries: int = 8,
```
dbczumar, line 36: Empirically, 8 retries translates to ~ 1 minute of wall clock time, which should be sufficient to overcome rate limiting in most cases

dspy/clients/lm.py
```diff
@@ -174,13 +175,55 @@ def cached_litellm_completion(request, num_retries: int):
...
+        model_list=[
+            {
+                "model_name": model,
+                "litellm_params": {
+                    "model": model,
+                },
+            }
+        ],
+        retry_policy=retry_policy,
+    )
```
dbczumar, line 224: LiteLLM Routers appear to be the only mechanism allowing exponential backoff *and* configurable retry codes. Docs: https://docs.litellm.ai/docs/routing

PR #1760, opened 2024-11-05
Fixes for OpenAI / Azure OpenAI compatibility with LiteLLM router

dbczumar:
Fixes for OpenAI / Azure OpenAI compatibility with LiteLLM router

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -296,10 +296,10 @@ def cached_litellm_text_completion(request, num_retries: int):
 
 def litellm_text_completion(request, num_retries: int, cache={"no-cache": True, "no-store": True}):
     kwargs = ujson.loads(request)
-    model = kwargs.pop("model").split("/", 1)
-    text_completion_model_name = f"text-completion-openai/{model}"
-
-    api_config = _extract_provider_api_config(model=kwargs["model"], **kwargs)
+    model = kwargs.pop("model")
+    api_config = _extract_provider_api_config(model=model, llm_kwargs=kwargs)
+    model_name = model.split("/", 1)[-1]
+    text_completion_model_name = f"text-completion-openai/{model_name}"
 
     # Build the prompt from the messages.
     prompt = "\n\n".join([x["content"] for x in kwargs.pop("messages")] + ["BEGIN RESPONSE:"])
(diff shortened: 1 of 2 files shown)


dspy/clients/lm.py
```diff
@@ -207,19 +258,29 @@ def _get_litellm_router(model: str, num_retries: int) -> Router:
...
+        "model": model,
+    }
+    if api_config.api_key is not None:
+        litellm_params["api_key"] = api_config.api_key
+    if api_config.api_base is not None:
+        litellm_params["api_base"] = api_config.api_base
+    if api_config.api_version is not None:
+        litellm_params["api_version"] = api_config.api_version
+    if api_config.azure_ad_token is not None:
+        litellm_params["azure_ad_token"] = api_config.azure_ad_token
```
dbczumar, line 275: @okhat For OpenAI and Azure OpenAI, API keys must be defined in the Router's `model_list` directly, due to initialization logic in https://github.com/BerriAI/litellm/blob/0fe8cde7c78d6b975b936f7802a4124b58bd253a/litellm/router.py#L3999-L4001

dspy/clients/lm.py
```diff
@@ -235,26 +296,18 @@ def cached_litellm_text_completion(request, num_retries: int):
...
-
-    # Extract the provider and model from the model string.
-    # TODO: Not all the models are in the format of "provider/model"
-    model = kwargs.pop("model").split("/", 1)
-    provider, model = model[0] if len(model) > 1 else "openai", model[-1]
-    text_completion_model_name = f"text-completion-openai/{model}"
-
-    # Use the API key and base from the kwargs, or from the environment.
-    api_key = kwargs.pop("api_key", None) or os.getenv(f"{provider}_API_KEY")
-    api_base = kwargs.pop("api_base", None) or os.getenv(f"{provider}_API_BASE")
```
dbczumar, line 247: These are now part of the router definition, so they don't need to be passed to the `text_completion()` API

dspy/clients/lm.py
```diff
@@ -164,6 +165,55 @@ def copy(self, **kwargs):
...
+    azure_ad_token: Optional[str]
+
+
+def _extract_provider_api_config(model: str, llm_kwargs: Dict[str, Any]) -> _ProviderAPIConfig:
+    """
+    Extract the API configurations from the specified LLM keyword arguments (`llm_kwargs`) for the
+    provider corresponding to the given model.
+
+    Note: The API configurations are removed from the specified `llm_kwargs`, if present, mutating
+    the input dictionary.
```
dbczumar, line 188: This avoids duplicating code in `litellm_text_completion`, which pops the api key and base from the user-specified kwargs

PR #1763, opened 2024-11-06
Add Support for List[Image]

ryanh-ai:
When providing `List[dspy.Image]` for input field, the `ChatAdapter` was formatting the list of images as a string and overflowing the input context for LM's from multiple providers.

Changes were done very quickly to unblock active work, but this is tested locally across Bedrock and OpenAI providers.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -151,6 +151,11 @@ def _format_field_value(field_info: FieldInfo, value: Any, assume_text=True) -> 
     Returns:
       The formatted value of the field, represented as a string.
     """
+    # Check if it's a List[Image] type
+    origin = get_origin(field_info.annotation)
+    args = get_args(field_info.annotation)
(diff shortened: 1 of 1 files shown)


ryanh-ai:
@isaacbmiller FYI, per Discord thread on this issue

okhat:
Thanks a ton @ryanh-ai ! I guess in general do we need more complex collections of images? Will people also want to do dict[str, list[Image]] and all kinds of crazy things?

If we only wanna support list[Image], then this PR is good.

isaacbmiller:
I think that people will at a minimum want to do things like dict[str, Image] which I am not sure even works currently.

I think this PR shouldn't be merged and I can work on a scalable solution for adapters for all different types including Image

ryanh-ai:
Agree above on the dictionary! 

I also found an edge case when length of image list is 1.  I will update the patch when I adjust that, but not invest too much more in this until you determine best path forward.  I'm happy to test as you do!

Issue #1619, opened 2024-10-13
LiteLLM error message


fullstackwebdev:
using the litellm openai adapter to override base url and simply connecting creates a large dump of LiteLLM error messages, rendering the terminal useless on large programs

1. setup an open ai compatible endpoint
2. set your API_BASE point
3. use the 'openai/MODELNAME' syntax

simple app

```python
PORT = os.getenv('PORT', 6002)
API_BASE = f"http://localhost:{PORT}/v1/"
MODEL_NAME = "openai/unsloth/gemma-2-27b-it-bnb-4bit"
logging.getLogger("dspy").setLevel(logging.ERROR) # doesn't work
logging.getLogger("dspy").propagate = False # doesn't work 

lm = dspy.LM(MODEL_NAME, api_key="..", api_base=API_BASE)
dspy.configure(lm=lm)

lm("What is 2+2?", temperature=0.9)
```

output

```
Provider List: https://docs.litellm.ai/docs/providers

['2 + 2 = 4']

Provider List: https://docs.litellm.ai/docs/providers


Provider List: https://docs.litellm.ai/docs/providers


Provider List: https://docs.litellm.ai/docs/providers

```

![image](https://github.com/user-attachments/assets/b8236631-603b-427f-a505-07a318664ff9)

arnavsinghvi11:
Hi @fullstackwebdev , does the discussion from this [LiteLLM issue](https://github.com/BerriAI/litellm/issues/5598#issuecomment-2339245304) help? (`litellm.suppress_debug_info = True`)

PR #1768, opened 2024-11-06
add flag to suppress litellm logs in dspy.LM

arnavsinghvi11:
default behavior to suppress litellm logs to avoid logging for every dspy.LM call.

related to #1619 and https://github.com/BerriAI/litellm/issues/5598#issuecomment-2339245304

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -6,7 +6,7 @@ import uuid
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from typing import Any, Dict, List, Literal, Optional
-
+import dspy
 import litellm
 import ujson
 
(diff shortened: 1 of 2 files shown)


dspy/clients/lm.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         launch_kwargs: Optional[Dict[str, Any]] = None,
         callbacks: Optional[List[BaseCallback]] = None,
         num_retries: int = 3,
+        suppress_debug_info: bool = True,
```
okhat, line 36: I'd remove this since it doesn't actually control things, i.e. users can't do dspy.configure() and get suppress to turn off

okhat: after dealing with this one way or another feel free to merge :D

PR #1772, opened 2024-11-07
Fix ruff error

isaacbmiller:
cc @okhat


okhat:
it's a very ruff error

PR #1777, opened 2024-11-08
Change the way of enable/disable litellm logging

chenmoneygithub:
We introduced litellm logging suppression in #1768, but we were configuring `litellm` inside the `__init__` call. So if users do:

```
lm = dspy.LM("...")
dspy.settings.configure(suppress_debug_info=False)
lm("some query")
```

Then the logging won't be turned on. In this PR I am introducing a pair of util functions: `dspy.clients.enable_litellm_logging()` and `dspy.client.disable_litellm_logging()` to turn on/off the litellm logging dynamically.

### DIFF
--- a/dsp/utils/settings.py
+++ b/dsp/utils/settings.py
@@ -1,6 +1,6 @@
 import threading
-from copy import deepcopy
 from contextlib import contextmanager
+from copy import deepcopy
 
 from dsp.utils.utils import dotdict
 
@@ -24,7 +24,6 @@ DEFAULT_CONFIG = dotdict(
     experimental=False,
     backoff_time=10,
     callbacks=[],
-    suppress_debug_info=True,
 )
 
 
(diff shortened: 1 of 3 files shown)


okhat:
Thank you, @chenmoneygithub! I like this! But should it just be dspy.configure(litellm_logging=True/False)?

If you think the answer is no, feel free to merge this!

chenmoneygithub:
@okhat Thanks! I took a second look at `dspy.settings.configure()`, and seems right now it's just doing updating the state dict instead of running actual configuration code:
```
    def configure(self, inherit_config: bool = True, **kwargs):
        """Set configuration settings.

        Args:
            inherit_config (bool, optional): Set configurations for the given, and use existing configurations for the rest. Defaults to True.
        """
        if inherit_config:
            config = {**self.config, **kwargs}
        else:
            config = {**kwargs}

        self.__append(config)
```

So for now I feel it's better not change this pattern. But in the future (after we deprecate the `dsp` directory) we may want to have a centralized namespace like `dspy.settings` that is a collection of dynamic configs, e.g., `dspy.settings.disable_litellm_logging()` and `dspy.settings.enable_verbose_logging()`.

PR #1806, opened 2024-11-15
(Feature) dspy.asyncify

CyrusOfEden:
🥁 the long awaited async implementation for DSPy (Take 2) 🔥

Since doing #1734 we realized a full async rewrite has too much complexity, particularly vis a vis optimization, so we decided to scope it down to inference-time async only.

**TL;DR:** Use `dspy.asyncify` to make your program async!

```python
predict = dspy.asyncify(dspy.Predict("question -> answer"))
result = await predict(question="...")
assert result.answer == "..."
```

This will run the dspy program in a separate thread and await its result. By default, the limit of spawned threads is 8. Think of this like a worker pool (like ThreadPoolExecutor). If you have 8 in-flight programs and call it once more, the 9th call will wait until one of the 8 returns.

You can configure this using the new `async_max_workers` setting, e.g. to set it to 16:

```python
dspy.settings.configure(async_max_workers=16)
```

This will be shared by all async calls.

**Recommended practice is to set async_max_workers once at the entry point of your program.** If you configure it multiple times, then the last value will be used as the global async max workers value for all async calls.


CyrusOfEden:
Weird, those tests pass locally for me

requirements.txt
```diff
@@ -15,3 +15,5 @@ requests
 tenacity>=8.2.3
 tqdm
 ujson
+anyio==4.6.2.post1
```
chenmoneygithub, line 18: we can remove the version requirement, our code should work with any latest version

dspy/utils/asyncify.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if cls._limiter is None or cls._limiter.total_tokens != dspy.settings.async_capacity:
+            cls._limiter = CapacityLimiter(dspy.settings.async_capacity)
+
+        return cls._limiter
+
+
+def asyncify(
+    program: Callable[T_ParamSpec, T_Retval],
+) -> Callable[T_ParamSpec, Awaitable[T_Retval]]:
```
chenmoneygithub, line 32: The type annotation part is not really readable. I am never a fan of Python type annotation, so I would rather just delete the type annotation for simplicity. But will let @okhat make the decision.

dspy/utils/asyncify.py
```diff
@@ -0,0 +1,33 @@
+import sys
+from typing import Awaitable, Optional, TypeVar, Callable
+
+from anyio import CapacityLimiter
+from asyncer import asyncify as __asyncify
```
chenmoneygithub, line 5: It's generally discouraged to use double leading underscore: https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

Let's just call it `_asyncify` or `import asyncer`, which is pretty fast too.

dspy/utils/asyncify.py
```diff
@@ -0,0 +1,33 @@
...
+        if cls._limiter is None or cls._limiter.total_tokens != dspy.settings.async_capacity:
+            cls._limiter = CapacityLimiter(dspy.settings.async_capacity)
+
+        return cls._limiter
+
+
+def asyncify(
+    program: Callable[T_ParamSpec, T_Retval],
+) -> Callable[T_ParamSpec, Awaitable[T_Retval]]:
+    return __asyncify(program, abandon_on_cancel=True, limiter=AsyncLimiter.get())
```
chenmoneygithub, line 33: what happens if we have multiple calls, each with a different value of `limiter`?

tests/predict/test_program_of_thought.py
```diff
@@ -1,4 +1,4 @@
-import textwrap
+import pytest
```
chenmoneygithub, line 1: hmm is this intended?

tests/utils/test_asyncify.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def set_async_capacity(capacity: int):
+        dspy.settings.configure(async_capacity=capacity)
+
+    async def assert_capacity(capacity: int, wait: float = 0.01, buffer: float = 0.1):
+        set_async_capacity(capacity)
+        t1 = time()
+        await run_n_tasks(capacity, wait)
+        t2 = time()
+        assert t2 - t1 < wait + buffer
```
chenmoneygithub, line 41: could this be flaky?

dspy/utils/asyncify.py
CyrusOfEden: In what cases would that happen?

The `async_max_workers` flag should be set once at program startup

tests/utils/test_asyncify.py
CyrusOfEden: Yeah a little bit, thoughts on how to make it better?

dspy/utils/asyncify.py
CyrusOfEden: Done!

tests/predict/test_program_of_thought.py
CyrusOfEden: Nope

dspy/utils/asyncify.py
chenmoneygithub: it's not very likely, but if the user do weird things like modifying `dspy.settings.configure(async_max_workers=200)`  between async calls, what would happen? As long as it's not causing dangerous behavior we should be good to go

dspy/utils/asyncify.py
CyrusOfEden: Removed!

requirements.txt
CyrusOfEden: Done!

okhat (approved):

mikeedjones:
Is a full async rewrite still planned at some point? I think there was still value in the `aforward` method of Predict when writing custom modules compared to being limited to N concurrent threads.

CyrusOfEden:
@mikeedjones will defer to @okhat on that one, more than happy to write it, I (as an engineer) think there's value in it, but form an ML researcher's perspective the value is less clear (to me).

ArslanS1997:
Hi how can I get it work with a custom module?

CyrusOfEden:
@ArslanS1997

```python
class YourCustomModule(dspy.Module):
  ...

module = YourCustomModule(...)
async_module = dspy.asyncify(module)
```

Jdogtherock:
how does it handle possible rate limit errors?

CyrusOfEden:
@Jdogtherock the same way the rest of DSPy does(n't)?

CyrusNuevoDia:
Heya @Stijn-B!

For the time being I'd recommend increasing async_max_workers to 200 and your server memory accordingly.

swiftdevil:
> For the time being I'd recommend increasing async_max_workers to 200 and your server memory accordingly.
> For the time being

Is the goal of this change to eventually use true async within this module, as defined in #1734? I'm building a program that makes _tons_ of calls to inference endpoints. Spamming thousands of waiting threads is not a viable solution for a production system. I'm not sure why you'd want to do this for a research system, for that matter... async is way, _way_ faster (and scalable) than multi-threading in this use case.

CyrusNuevoDia:
@swiftdevil yes

mirandaconrado:
@CyrusNuevoDia what would be the right way to start a conversation on native async vs using threads?

kalanyuz:
@CyrusNuevoDia would this allow ReAct tool calling to support Anthropic's [MCP servers](https://modelcontextprotocol.io/quickstart/client)? Tool calls would require await to the tool functions. Thinking this might be a good way for DSPy to gain access to hundreds of well-built tools instantly.

PR #1804, opened 2024-11-15
Add deployment guide

chenmoneygithub:
Two options: 1) FastAPI 2) Mlflow.

PR #1807, opened 2024-11-16
Add deployment guide

chenmoneygithub:
Reopen #1804 for clean commit history.


docs/docs/tutorials/deployment/deploy_dspy_model.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+In the above code, we call `dspy.asyncify(dspy.ChainOfThought("question -> answer"))` to convert the dspy program to run
+in async mode, which is required by high-throughput FastAPI deployments. This under the hood runs the dspy program in a
+separate thread and await its result. By default, the limit of spawned threads is 8. Think of this like a worker pool
+(like ThreadPoolExecutor). If you have 8 in-flight programs and call it once more, the 9th call will wait until one of the
+8 returns.
+
+You can configure the async capacity using the new `async_max_workers` setting, e.g. to set it to 16, which will be shared
+by all async calls:
```
mikeedjones, line 76: I think this is [what fastAPI does under the hood](https://fastapi.tiangolo.com/async/#very-technical-details) when you define a sync path operation function. I'm not sure what dspy is adding here by having its own round of "stick this in a threadpool and await"? Maybe something I'm missing sorry! 

Since asyncer and fastapi are both from tiangolo it might even be the case he's using asyncer under the hood to run the path operation function :D

PR #1821, opened 2024-11-18
.load_demos(**kwargs) for dynamically loading demos

CyrusOfEden:
In preparation for `BootstrapKNNFewShotWithRandomSearch`, the change is:

```python
demos = kwargs.pop("demos", self.demos) # old
demos = kwargs.pop("demos", None) or self.load_demos(**kwargs) # new
```

Which should let enable us to do dynamic demo loading at runtime (i.e. test-time compute)

### DIFF
--- a/dspy/predict/retry.py
+++ b/dspy/predict/retry.py
@@ -53,7 +53,6 @@ class Retry(Predict):
     def __call__(self, **kwargs):
         copy.deepcopy(kwargs)
         kwargs["_trace"] = False
-        kwargs.setdefault("demos", self.demos if self.demos is not None else [])
 
         # perform backtracking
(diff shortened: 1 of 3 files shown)


dspy/predict/retry.py
```diff
@@ -53,7 +53,6 @@ def forward(self, *, past_outputs, **kwargs):
     def __call__(self, **kwargs):
         copy.deepcopy(kwargs)
         kwargs["_trace"] = False
-        kwargs.setdefault("demos", self.demos if self.demos is not None else [])
```
CyrusOfEden, line 56: This line actually doesn't affect the downstream code, at least I don't think so, since the downstream code does the same thing. Also, tests pass, so this is good.

Issue #1624, opened 2024-10-14
ImportError: cannot import name 'Cache' from 'litellm.caching' (unknown location)


HusainMkd:
I can't seem to import DSpy anymore (on Databricks)... I've tried downgrading the DSpy version, restarting the cluster, installing litellm - nothing seems to be working - Is there a breaking change somewhere?
Below is my error message

```python
ImportError: cannot import name 'Cache' from 'litellm.caching' (unknown location)
File <command-86610133237801>, line 5
      3 AZ_GPT4o_URI = os.getenv('AZ_GPT4o_URI')
      4 # from litellm.caching import Cache
----> 5 import dspy
      6 import pandas as pd
      7 import mlflow
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-33e20db8-6c63-45a6-a796-81c039e2d548/lib/python3.11/site-packages/dspy/__init__.py:8
      6 from .retrieve import *
      7 from .signatures import *
----> 8 from .utils.logging import logger, set_log_output
     10 # Functional must be imported after primitives, predict and signatures
     11 from .functional import *  # isort: skip
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-33e20db8-6c63-45a6-a796-81c039e2d548/lib/python3.11/site-packages/dspy/utils/__init__.py:1
----> 1 from .dummies import *
      2 from .logging import *
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-33e20db8-6c63-45a6-a796-81c039e2d548/lib/python3.11/site-packages/dspy/utils/dummies.py:11
      9 from dsp.utils.utils import dotdict
     10 from dspy.adapters.chat_adapter import field_header_pattern, format_fields
---> 11 from dspy.clients.lm import LM
     14 class DSPDummyLM(DSPLM):
     15     """Dummy language model for unit testing purposes subclassing DSP LM class."""
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-33e20db8-6c63-45a6-a796-81c039e2d548/lib/python3.11/site-packages/dspy/clients/__init__.py:1
----> 1 from .lm import LM
File /local_disk0/.ephemeral_nfs/envs/pythonEnv-33e20db8-6c63-45a6-a796-81c039e2d548/lib/python3.11/site-packages/dspy/clients/lm.py:9
      7 import litellm
      8 import ujson
----> 9 from litellm.caching import Cache
[...]

PR #1833, opened 2024-11-20
Make litellm version more flexible

rogeriochaves:
hey there!

I wanted to install dspy in a project where I also use litellm more explicitly, but unfortunately I couldn't update litellm version independently, I'm then adding the caret^ to flexibilize the minor version of it

now, that may be a risk as indeed I've seen litellm having quite few breaking changes in the past, @krypticmouse you were the one to change that from ^ to a explicit version before, do you remember what was the reason for that? https://github.com/stanfordnlp/dspy/commit/f5a781ca1e443a445cf8cbab8e6f0c8580375fae

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -8523,4 +8523,4 @@ weaviate = ["weaviate-client"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9,<3.13"
-content-hash = "0675fa1dd1f44d63bafaffc6c2dfb384f2d016cfa20db379c19515ca182e867c"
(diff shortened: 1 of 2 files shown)


krypticmouse:
Hi @rogeriochaves 

Thank You so much for the contribution!! Yes the issue was with importing the caching modeling which kept breaking the test in CI. See: #1624

rogeriochaves:
okay thanks! What do you think then moving forward, should we unpin as that issue was fixed?

krypticmouse:
I think we can and iirc @chenmoneygithub checked the latest version and it was running fine. 

cc: @okhat

PR #1848, opened 2024-11-23
Removing duplicate sentence in signature docs

b-d055:
Noticed a duplicate sentence in the docs for signatures.


okhat:
Thanks @b-d055 !

PR #1856, opened 2024-11-25
Allow react.Tool to wrap methods

tkellogg:
The big reason for this is to pass parameters out-of-band, e.g. a user_id to ensure the LLM doesn't get the wrong data.

The unit test includes a usage, you can't use it as a decorator this way, but it works.

The alternative, of course, is to have a very long function and have all the tools be nested functions. It works, but can lead to some very long functions. I prefer long classes over long functions.


tkellogg:
Alright, it's ready for merge

okhat:
Amazing, thank you @tkellogg !

PR #1862, opened 2024-11-26
[BUG] Avoid serializing Callable kwargs during Completion

### LINKED ISSUE #1759: Using Azure token provider in dspy.LM
A good practice when authenticating with Azure OpenAI is using token providers.
Following then example of:
https://docs.litellm.ai/docs/providers/azure#azure-ad-token-refresh---defaultazurecredential

The reason for token provider vs. getting token is that the token expires with some time limit.

We get a type error that the token provider is not JSON serializable. 
image

This was also experienced by @arthurgreef in https://github.com/stanfordnlp/dspy/issues/1561#issue-2554542697.
Although "fixed" by getting the token instead of providing the callable that handles refreshing of tokens.

I assume this is a byproduct of caching, but the behaviour is still present with cache=False parameter.

bahtman:
Fixes #1759

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -84,7 +84,8 @@ class LM(BaseLM):
         cache = kwargs.pop("cache", self.cache)
         messages = messages or [{"role": "user", "content": prompt}]
         kwargs = {**self.kwargs, **kwargs}
-
+        callable_kwargs = {k: v for k, v in kwargs.items() if isinstance(v, Callable)}
+        kwargs = {k: v for k, v in kwargs.items() if not isinstance(v, Callable)}
         # Make the request and handle LRU & disk caching.
         if self.model_type == "chat":
             completion = cached_litellm_completion if cache else litellm_completion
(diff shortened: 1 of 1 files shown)


okhat:
Wow super cool, thanks @bahtman ! I'll review this more carefully but I think we can merge soon

okhat:
Hmm, will the lru_cache work for callables? Do we need to add "ignore" parameter for the lru_cache to ignore callable_kwargs?

bahtman:
I was able to run the LM with a callable token provider. I’m unfamiliar with how the lru cache works, so unsure whether it actually caches the parameter.

This is a workaround so we don’t serialize it. But since we don’t serialize in the new Embed, it might be a relic? If that is the case we should just revert and remove the ujson part.

bahtman:
I did some digging. lru cache works as long as parameters are hashable (has __hash__() method)
https://docs.python.org/3/library/functools.html#functools.lru_cache

This is the case for callables, but not mutable containers like lists and dicts. This is probably the reason for the serialization.

PR #1874, opened 2024-11-29
Streaming

CyrusNuevoDia:
`dspy.streamify` can be used to convert the dspy program to a streaming mode. This is useful when you want to stream
the intermediate outputs (i.e. O1-style reasoning) to the client before the final prediction is ready. This uses
asyncify under the hood and inherits the execution semantics.

The deltas of *every module in the program* are streamed directly with no processing and then once the final prediction is ready it is yielded.

### Here's how it works for deployment

```python
from fastapi.responses import StreamingResponse

streaming_dspy_program = dspy.streamify(dspy.ChainOfThought("question -> answer"))

@app.post("/predict/stream")
async def stream(question: Question):
    async def generate():
        async for value in streaming_dspy_program(question=question.text):
            if isinstance(value, dspy.Prediction):
                data = {"prediction": value.labels().toDict()}
            elif isinstance(value, litellm.ModelResponse):
                data = {"chunk": value.json()}
            yield f"data: {ujson.dumps(data)}\n\n"
        yield "data: [DONE]\n\n"

    return StreamingResponse(generate(), media_type="text/event-stream")

# Since you're often going to want to stream the result of a DSPy program as server-sent events,
# we've included a helper function for that, which is equivalent to the code above.

from dspy.utils.streaming import streaming_response

@app.post("/predict/stream")
async def stream(question: Question):
    stream = streaming_dspy_program(question=question.text)
    return StreamingResponse(streaming_response(stream), media_type="text/event-stream")
```

### Changes

- New in-memory LMRequestLRUCache with a default max size of 10_000_000.

### Notes

- No intermediate details are streamed with a cache hit on the in-memory LRU cache because we have the final result instantly. Streaming should work with in-memory cache turned off, which enables the LiteLLM cache


okhat:
This looks AMAZING, thanks @CyrusNuevoDia ! I'm just wrapping my head around the caching improvements here (which I quite like so far) and then will merge

CyrusNuevoDia:
@okhat 

Before: Infinite LRU caches (unbounded memory growth in prod with cache=True)
After: Bounded LRU caches

Before: Had to serialize/deserialize JSON to cache properly 
After: Just need to serialize (to compute the hash key)

Anything else I can help clarify?

bahtman:
Hi @CyrusNuevoDia. Sick feature, have been waiting for this long :)
Do you know whether this fixes the issue with serializing callable kwargs?
If it does not, you might be able to implement the fix built in #1862

dspy/utils/caching.py
```diff
@@ -12,3 +18,58 @@ def create_subdir_in_cachedir(subdir: str) -> str:
...
+    def pop(self, request: dict, default=None):
+        key = self.cache_key(request)
+        return super().pop(key, default)
+
+    @staticmethod
+    def cache_key(request: Union[dict, str]) -> str:
+        params = request
+        if isinstance(request, dict):
+            params = {k: v for k, v in request.items() if not callable(v)}
+        return sha256(ujson.dumps(params, sort_keys=True).encode()).hexdigest()
```
dbczumar, line 75: @CyrusNuevoDia @bahtman Can we handle pydantic models here as well, by converting them to JSON objects? This will be necessary for supporting structured outputs - https://github.com/stanfordnlp/dspy/pull/1881/files#diff-ee21f597435b8e775fd162c84aaa2c7a544120407baa26c12b3ff26d0d31a7dbR43-R44

bahtman: Would the pydantic model be a part of the dictionary request object or the entirety of it? Do we want to cache it or just avoid it like the callable parameters? (Above implementation ignores pydantic models because of the callable check)

dbczumar: Great question! I think we should cache it, I.e. it should be included in `params`. It will be part of the dictionary request object, e.g. an entry like `params["response_format"] = <Pydantic object>`

bahtman: Can we assume its always "response_format" and we just check for that and convert it to dict?

dbczumar: I think it'd be a bit more future proof (and also totally sufficient) to check for `isinstance(v, pydantic.BaseModel)`

docs/docs/tutorials/deployment/index.md
```diff
@@ -54,14 +56,45 @@ async def predict(question: Question):
...
+
+# Since you're often going to want to stream the result of a DSPy program as server-sent events,
+# we've included a helper function for that, which is equivalent to the code above.
+
+from dspy.utils.streaming import streaming_response
+
+@app.post("/predict/stream")
+async def stream(question: Question):
+    stream = streaming_dspy_program(question=question.text)
+    return StreamingResponse(streaming_response(stream), media_type="text/event-stream")
```
dbczumar, line 84: Awesome example! Not sure if you have bandwidth, but it would be great to introduce test coverage like this as well by:

1. Extending https://github.com/stanfordnlp/dspy/blob/main/tests/test_utils/server/litellm_server.py to introduce a streaming endpoint

2. Writing a test case that uses `litellm_test_server` + streaming, similar to https://github.com/stanfordnlp/dspy/blob/97ca14cdae61b14bb1c2d520ba559e710d2cb7f6/tests/clients/test_lm.py#L9

dspy/utils/caching.py
```diff
@@ -12,3 +18,58 @@ def create_subdir_in_cachedir(subdir: str) -> str:
     subdir = os.path.abspath(subdir)
     os.makedirs(subdir, exist_ok=True)
     return subdir
+
+
+class LRUCache(OrderedDict):
```
dbczumar, line 23: @bahtman Can we extend some of the testing in https://github.com/stanfordnlp/dspy/blob/main/tests/caching/test_caching.py to verify that in-memory caching works properly?

dspy/utils/caching.py
bahtman: To avoid another passthrough I would prefer a staticmethod `convert_pydantic` that check for `isinstance(v, pydantic.BaseModel)` and if true returns `v.schema()` else `v`. Then we can just call that while we do the dict comprehension.

bahtman: `params = {k: self.convert_pydantic(v) for k, v in request.items() if not callable(v)}`

bahtman: Ahh. This might give us issues, since v is callable when a pydantic. Either we build an or statement or we just do 2 passthroughs (Later is probably more clean at this point)

dspy/utils/caching.py
CyrusNuevoDia: @dbczumar I had this realization today — any reason why we're not just using joblib? I think it was because it was in the dsp folder (and we're trying to move away from that), but joblib seemed like it handled a lot of things well out of the box.

okhat: @CyrusNuevoDia What do you think was handled by joblib? Joblib handled everything strictly worse than a basic DiskCache. In 2.4 and earlier, we combined joblib with manual lru_cache.

CyrusNuevoDia: @okhat oh thanks, I wasn't privy to that context. Makes sense!

I liked how it automatically wrote/read from disk — maybe this LRUCache + DiskCache is the way to go then?

docs/docs/tutorials/deployment/index.md
CyrusNuevoDia: @dbczumar would love your help on this, I'm stuck on the `miprov2-knn` branch for benchmarking atm

dbczumar: Done :)

dspy/utils/streaming.py
```diff
@@ -0,0 +1,87 @@
...
+            data = {"prediction": {k: v for k, v in value.items(include_dspy=False)}}
+            yield f"data: {ujson.dumps(data)}\n\n"
+        elif isinstance(value, litellm.ModelResponse):
+            data = {"chunk": value.json()}
+            yield f"data: {ujson.dumps(data)}\n\n"
+        elif isinstance(value, str) and value.startswith("data:"):
+            # The chunk value is an OpenAI-compatible streaming chunk value,
+            # e.g. "data: {"finish_reason": "stop", "index": 0, "is_finished": True, ...}",
+            # so yield it directly
+            yield value
```
dbczumar, line 84: This is necessary in order to write a DSPy program that consumes a streaming response from an LLM provider (e.g OpenAI) and produces a response containing its predictions. In this case, the LLM provider chunks are streamed through the program output as JSON objects.

dspy/utils/streaming.py
```diff
@@ -0,0 +1,87 @@
...
+    used by a service as an API response to a streaming request.
+
+    Args:
+        streamer: An async generator that yields values from a DSPy program output stream.
+    Returns:
+        An async generator that yields OpenAI-compatible streaming response chunks.
+    """
+    async for value in streamer:
+        if isinstance(value, Prediction):
+            data = {"prediction": {k: v for k, v in value.items(include_dspy=False)}}
```
dbczumar, line 75: The previous logic that relied on `labels()` failed with:

```
self = Prediction(
    output_text='Hello!'
), key = '_input_keys'

    def __getattr__(self, key):
        if key.startswith("__") and key.endswith("__"):
            raise AttributeError
        if key in self._store:
            return self._store[key]
>       raise AttributeError(f"'{type(self).__name__}' object has no attribute '{key}'")
E       AttributeError: 'Prediction' object has no attribute '_input_keys'
```

Prediction is an instance of Example that lacks `_input_keys`. Fortunately, because we know that Prediction is an output wrapper, it's safe to assume that there are no inputs & call `items()` directly.

bahtman:
@dbczumar 
Do you see my comments?

dspy/utils/caching.py
dbczumar: One issue I hadn't accounted for here is thread safety. This cache is likely to be called in parallel during optimization or evaluation. The implementation needs to be thread safe. I suspect relying on a threadsafe cache (e.g. continuing to use functools) is probably best. @CyrusNuevoDia thoughts on removing the caching portions from this PR and getting the streaming functionality merged independently?

rohitgarud:
Hi @dbczumar, Will it be possible to apply assertions/suggestions on the partially streamed output so that we can stop streaming if the assertion fails and reduce the token usage by trying again with prompt added due to assertion failure? This might be particularly useful if we are applying multiple assertions to assess the output according to multiple dimensions

CyrusNuevoDia:
@rohitgarud great idea! out of scope for this current version but would be awesome

CyrusNuevoDia:
@dbczumar merged in your request cache logic, getting this error on tests — any idea how to fix it?

```
ImportError while importing test module '/Users/knrz/Git/stanfordnlp/dspy/tests/caching/test_caching.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/caching/test_caching.py:10: in <module>
    from tests.test_utils.server import litellm_test_server, read_litellm_test_server_request_logs
E   ModuleNotFoundError: No module named 'tests.test_utils'
____________________________________________________________________________________ ERROR collecting tests/clients/test_lm.py ____________________________________________________________________________________
ImportError while importing test module '/Users/knrz/Git/stanfordnlp/dspy/tests/clients/test_lm.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/clients/test_lm.py:8: in <module>
    from tests.test_utils.server import litellm_test_server
E   ModuleNotFoundError: No module named 'tests.test_utils'
_______________________________________________________________________________________ ERROR collecting tests/reliability ________________________________________________________________________________________
../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
.venv/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:184: in exec_module
    exec(co, module.__dict__)
tests/reliability/conftest.py:6: in <module>
    from tests.conftest import clear_settings
E   ImportError: cannot import name 'clear_settings' from 'tests.conftest' (/Users/knrz/Git/stanfordnlp/dspy/.venv/lib/python3.9/site-packages/tests/conftest.py)
```

dspy/utils/caching.py
CyrusNuevoDia: Fixed by merging in your cache

dbczumar:
> @dbczumar merged in your request cache logic, getting this error on tests — any idea how to fix it?
> 
> ```
> ImportError while importing test module '/Users/knrz/Git/stanfordnlp/dspy/tests/caching/test_caching.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> ../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> tests/caching/test_caching.py:10: in <module>
>     from tests.test_utils.server import litellm_test_server, read_litellm_test_server_request_logs
> E   ModuleNotFoundError: No module named 'tests.test_utils'
> ____________________________________________________________________________________ ERROR collecting tests/clients/test_lm.py ____________________________________________________________________________________
> ImportError while importing test module '/Users/knrz/Git/stanfordnlp/dspy/tests/clients/test_lm.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> ../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> tests/clients/test_lm.py:8: in <module>
>     from tests.test_utils.server import litellm_test_server
> E   ModuleNotFoundError: No module named 'tests.test_utils'
> _______________________________________________________________________________________ ERROR collecting tests/reliability ________________________________________________________________________________________
> ../../../.rye/py/cpython@3.9.18/lib/python3.9/importlib/__init__.py:127: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> <frozen importlib._bootstrap>:1030: in _gcd_import
>     ???
> <frozen importlib._bootstrap>:1007: in _find_and_load
>     ???
> <frozen importlib._bootstrap>:986: in _find_and_load_unlocked
>     ???
> <frozen importlib._bootstrap>:680: in _load_unlocked
>     ???
> .venv/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:184: in exec_module
>     exec(co, module.__dict__)
> tests/reliability/conftest.py:6: in <module>
>     from tests.conftest import clear_settings
> E   ImportError: cannot import name 'clear_settings' from 'tests.conftest' (/Users/knrz/Git/stanfordnlp/dspy/.venv/lib/python3.9/site-packages/tests/conftest.py)
> ```

Thanks @CyrusNuevoDia ! I'll push some updates to remove caching changes from this PR. My reasoning is that streaming and caching aren't directly related (I added some test coverage to verify that caching works properly with streaming, though). We can make adjustments to caching in future PRs.

dspy/dsp/utils/settings.py
```diff
@@ -17,6 +18,7 @@
     backoff_time=10,
     callbacks=[],
     async_max_workers=8,
+    send_stream=None,
```
dbczumar, line 21: This is the only meaningful change in the file - everything else is a linter adjustment

MohammedAlhajji:
Thanks @CyrusNuevoDia This looks great. Appreciate it.

PR #1877, opened 2024-12-01
Replace localhost links in the doc

B-Step62:
Replace broken links that host to localhost (`127.0.0.1`) to the correct ones.


okhat:
Thank you @B-Step62 ! The link updated here is outdated. Fixed on main.

PR #1879, opened 2024-12-01
Add on_tool_start/end callbacks

B-Step62:
Update the `BaseCallback` class to handle `Tool` for tool calling.


okhat:
Thanks so much @B-Step62 !!

PR #1885, opened 2024-12-04
Allow programmatically setting litellm cache

jasper-xian:


okhat:
Thank you Jasper. We're changing a lot of things around the cache, so I will not merge this until it's clearer where things go.

Please do feel free to use this in your own experiments in the time being.

PR #1889, opened 2024-12-05
Improve DSPy saving

chenmoneygithub:
We are reworking DSPy saving in order to make it more robust:

- Introduce whole model saving, users can specify `state_only=False` to save the model architecture along with the state. This is done via cloudpickle.
- Improve the state-only saving to support json and pickle, which resolves the problem that some demos are not json-serializable.
- Add unit test on both saving format. 

From the user perspective, it means they have 3 journeys on saving a DSPy model.

Method 1: Save everything.

```
# Saving.
predict = dspy.Predict("question->answer")
optimizer = dspy.BootstrapFewShot(max_bootstrapped_demos=4, max_labeled_demos=4, max_rounds=5, metric=some_metric)
compiled_predict = optimizer.compile(predict, trainset=trainset)
compiled_predict.save("/tmp/my_model", save_program=True)   # `my_model` is a directory.

# Loading.
loaded_predict = dspy.load("/tmp/my_model")
```

Method 2 (our old approach): Save state with Json

```
# Saving.
predict = dspy.Predict("question->answer")
optimizer = dspy.BootstrapFewShot(max_bootstrapped_demos=4, max_labeled_demos=4, max_rounds=5, metric=some_metric)
compiled_predict = optimizer.compile(predict, trainset=trainset)
compiled_predict.save("/tmp/my_model/model.json", save_program=False)

# Loading.
loaded_predict = dspy.Predict("question->answer")
loaded_predict.load("/tmp/my_model/model.json")
```

Method 3: Save state with cloudpickle

```
# Saving.
predict = dspy.Predict("question->answer")
optimizer = dspy.BootstrapFewShot(max_bootstrapped_demos=4, max_labeled_demos=4, max_rounds=5, metric=some_metric)
compiled_predict = optimizer.compile(predict, trainset=trainset)
compiled_predict.save("/tmp/my_model/model.pkl", save_program=False)

# Loading.
loaded_predict = dspy.Predict("question->answer")
loaded_predict.load("/tmp/my_model/model.pkl")
```


dspy/predict/predict.py
```diff
@@ -136,17 +137,22 @@ def _load_state_legacy(self, state):
...
         """Load a saved state from a file.
-        
+
         Args:
             path (str): Path to the saved state file
-            return_self (bool): If True, returns self to allow method chaining. Default is False for backwards compatibility.
-        
+            use_legacy_loading (bool): Whether to use the legacy loading method. Only use it when you are loading a
+                saved state from a version of DSPy prior to v2.5.3.
+            use_json (bool): Whether to load the state from a json file. If False, the state is loaded from a pickle
```
okhat, line 147: Do we need a flag for `use_json`? Can we infer this from the file path?

We should probably get rid of `return_self` and just always assume it's true?

dspy/primitives/module.py
```diff
@@ -160,13 +167,105 @@ def load_state(self, state, use_legacy_loading=False):
...
+        if state_only:
+            state = self.dump_state(save_field_meta)
+            state["metadata"] = metadata
+            if use_json:
+                if path.suffix != ".json":
+                    raise ValueError(
+                        f"`path` must be a json file when `state_only=True` and `use_json=True`, but received: {path}"
+                    )
+                with open(path, "w") as f:
+                    f.write(ujson.dumps(state, indent=2))
```
okhat, line 203: Is there an unhandled `else` case here?

dspy/primitives/module.py
```diff
@@ -160,13 +167,105 @@ def load_state(self, state, use_legacy_loading=False):
...
                 param.load_state(state[name])
 
-    def save(self, path, save_field_meta=False):
-        with open(path, "w") as f:
-            f.write(ujson.dumps(self.dump_state(save_field_meta), indent=2))
-
-    def load(self, path, use_legacy_loading=False):
-        with open(path) as f:
-            self.load_state(ujson.loads(f.read()), use_legacy_loading=use_legacy_loading)
+    def save(self, path, save_field_meta=False, state_only=True, metadata=None, use_json=True):
```
okhat, line 170: Just thinking out loud: I don't like that we already have so many flags, since this is our opportunity to simplify save/load a lot, rather than to make it more complicated from a DX standpoint.

More actionable: I don't like that state_only is a "double negative". Turning it _off_ does _more_ work. It should probably be reversed, e.g. `save_program=False` by default.

Separately, do we need `use_json`? It seems to me that we use JSON _iff_ `save_program == False`.

chenmoneygithub: that makes sense! We allow users to pickle the state when `save_program == False` and `use_json = False`, which handles the problem of serializing data like `datetime.date`. 

What about this:

- I will remove the `metadata` arg, which is not useful now.
- I will remove the `use_json` arg, and use the file suffix (.json or .pkl) to determine the format for saving state.
- Replace `state_only` by `save_program`.

dspy/predict/predict.py
chenmoneygithub: good call, we don't need that. I am changing to code to read the suffix to determine the saving/loading format.

dspy/primitives/module.py
chenmoneygithub: refactored the code, now it should be more clear

dspy/predict/predict.py
```diff
@@ -136,17 +138,20 @@ def _load_state_legacy(self, state):
 
         return self
 
-    def load(self, path, return_self=False):
+    def load(self, path, use_legacy_loading=False, return_self=False):
```
okhat, line 141: Does this method even need to exist?

chenmoneygithub: We can delete it since we have made our new plan for backward compatibility.

PR #1892, opened 2024-12-05
fix the 404 issues of "cheatsheet" urls

keyuchen21:
the original url shows "404 - Not found"
update with the new one


keyuchen21:
the original url "https://dspy.ai/docs/cheatsheet" shows
image

I fixed it with the new urls

PR #1896, opened 2024-12-06
Integrate cachetools for in-memory LM caching, including unhashable types & pydantic

### LINKED ISSUE #1759: Using Azure token provider in dspy.LM

dbczumar:
Integrate cachetools for in-memory LM caching, including unhashable types & pydantic

Fixes #1759

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -4,10 +4,13 @@ import os
 import threading
 import uuid
 from datetime import datetime
+from hashlib import sha256
 from typing import Any, Dict, List, Literal, Optional
 
 import litellm
+import pydantic
 import ujson
+from cachetools import LRUCache, cached
 
 from dspy.adapters.base import Adapter
 from dspy.clients.openai import OpenAIProvider
@@ -92,7 +95,7 @@ class LM(BaseLM):
             completion = cached_litellm_text_completion if cache else litellm_text_completion
 
         response = completion(
-            request=ujson.dumps(dict(model=self.model, messages=messages, **kwargs)),
+            request=dict(model=self.model, messages=messages, **kwargs),
             num_retries=self.num_retries,
         )
         outputs = [c.message.content if hasattr(c, "message") else c["text"] for c in response["choices"]]
@@ -153,7 +156,11 @@ class LM(BaseLM):
         thread = threading.Thread(target=thread_function_wrapper)
         model_to_finetune = self.finetuning_model or self.model
         job = self.provider.TrainingJob(
-            thread=thread, model=model_to_finetune, train_data=train_data, train_kwargs=train_kwargs, data_format=data_format
+            thread=thread,
+            model=model_to_finetune,
+            train_data=train_data,
+            train_kwargs=train_kwargs,
+            data_format=data_format,
         )
         thread.start()
 
@@ -212,8 +219,48 @@ class LM(BaseLM):
         return new_instance
 
 
-@functools.lru_cache(maxsize=None)
-def cached_litellm_completion(request, num_retries: int):
+def request_cache(maxsize: Optional[int] = None):
+    """
+    A threadsafe decorator to create an in-memory LRU cache for LM inference functions that accept
+    a dictionary-like LM request. An in-memory cache for LM calls is critical for ensuring
+    good performance when optimizing and evaluating DSPy LMs (disk caching alone is too slow).
+
+    Args:
+        maxsize: The maximum size of the cache.
+
+    Returns:
+        A decorator that wraps the target function with caching.
+    """
+    # NB: cachetools doesn't support maxsize=None; it recommends using float("inf") instead
+    cache = LRUCache(maxsize=maxsize or float("inf"))
+    cache_lock = threading.Lock()
+
+    def cache_key(request: Dict[str, Any]) -> str:
+        # Transform Pydantic models into JSON-convertible format and exclude unhashable objects
+        params = {
+            k: (v.dict() if isinstance(v, pydantic.BaseModel) else v) for k, v in request.items() if not callable(v)
+        }
+        return sha256(ujson.dumps(params, sort_keys=True).encode()).hexdigest()
+
+    def decorator(func):
+        @cached(
+            cache=cache,
+            key=lambda request, *args, **kwargs: cache_key(request),
+            # Use a lock to ensure thread safety for the cache when DSPy LMs are queried
+            # concurrently, e.g. during optimization and evaluation
+            lock=cache_lock,
+        )
+        @functools.wraps(func)
+        def wrapper(request: dict, *args, **kwargs):
+            return func(request, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+@request_cache(maxsize=None)
+def cached_litellm_completion(request: Dict[str, Any], num_retries: int):
     return litellm_completion(
         request,
         cache={"no-cache": False, "no-store": False},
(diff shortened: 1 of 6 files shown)


dspy/clients/lm.py
```diff
@@ -212,47 +219,82 @@ def copy(self, **kwargs):
...
+        return sha256(ujson.dumps(params, sort_keys=True).encode()).hexdigest()
+
+    def decorator(func):
+        @cached(
+            # NB: cachetools doesn't support maxsize=None; it recommends using float("inf") instead
+            cache=LRUCache(maxsize=maxsize or float("inf")),
+            key=lambda request, *args, **kwargs: cache_key(request),
+            # Use a lock to ensure thread safety for the cache when DSPy LMs are queried
+            # concurrently, e.g. during optimization and evaluation
+            lock=threading.Lock(),
```
dbczumar, line 249: cachetools provides thread safety natively

dspy/clients/lm.py
```diff
@@ -212,47 +219,82 @@ def copy(self, **kwargs):
...
+        params = {
+            k: (v.dict() if isinstance(v, pydantic.BaseModel) else v) for k, v in request.items() if not callable(v)
+        }
+        return sha256(ujson.dumps(params, sort_keys=True).encode()).hexdigest()
+
+    def decorator(func):
+        @cached(
+            # NB: cachetools doesn't support maxsize=None; it recommends using float("inf") instead
+            cache=LRUCache(maxsize=maxsize or float("inf")),
+            key=lambda request, *args, **kwargs: cache_key(request),
```
dbczumar, line 246: This is the key advantage of cachetools. Unlike `lru_cache`, it allows us to define a cache key using a subset of function arguments, rather than forcing all arguments to be hashed / JSON-encoded and then decoded afterwards. Encoding / decoding is infeasible for callables.

dspy/clients/lm.py
```diff
@@ -212,47 +219,82 @@ def copy(self, **kwargs):
...
+def cached_litellm_completion(request: Dict[str, Any], num_retries: int):
     return litellm_completion(
         request,
         cache={"no-cache": False, "no-store": False},
         num_retries=num_retries,
     )
 
 
-def litellm_completion(request, num_retries: int, cache={"no-cache": True, "no-store": True}):
-    kwargs = ujson.loads(request)
```
dbczumar, line 225: We no longer have to serialize / deserialize `request` within the `litellm_completion` and `litellm_text_completion` calls

tests/caching/test_caching.py
```diff
@@ -88,3 +89,51 @@ def test_lm_calls_are_cached_across_interpreter_sessions(litellm_test_server, te
...
+    api_base, server_log_file_path = litellm_test_server
+
+    lm_with_unhashable_callable = dspy.LM(
+        model="openai/dspy-test-model",
+        api_base=api_base,
+        api_key="fakekey",
+        # Define a callable kwarg for the LM to use during inference
+        azure_ad_token_provider=lambda *args, **kwargs: None,
+    )
+    lm_with_unhashable_callable("Query")
```
dbczumar, line 124: Fails on `main` with:

```
        )
E       TypeError: <function test_lm_calls_support_unhashable_types.<locals>.<lambda> at 0x31204d5a0> is not JSON serializable
```

tests/caching/test_caching.py
```diff
@@ -88,3 +89,51 @@ def test_lm_calls_are_cached_across_interpreter_sessions(litellm_test_server, te
...
+    class ResponseFormat(pydantic.BaseModel):
+        response: str
+
+    lm = dspy.LM(
+        model="openai/dspy-test-model",
+        api_base=api_base,
+        api_key="fakekey",
+        response_format=ResponseFormat,
+    )
+    lm("Query")
```
dbczumar, line 139: Fails on `main` with:

```
TypeError: <class 'tests.caching.test_caching.test_lm_calls_support_pydantic_models.<locals>.ResponseFormat'> is not JSON serializable
```

dspy/clients/lm.py
```diff
@@ -212,47 +219,82 @@ def copy(self, **kwargs):
         return new_instance
 
 
-@functools.lru_cache(maxsize=None)
-def cached_litellm_completion(request, num_retries: int):
+def request_cache(maxsize: Optional[int] = None):
```
dbczumar, line 222: @okhat @bahtman @CyrusNuevoDia Thoughts on this approach? See inline comments discussing advantages below

tests/clients/test_lm.py
```diff
@@ -1,24 +1,77 @@
...
         model="azure/dspy-test-model",
         api_base=api_base,
         api_key="fakekey",
+        model_type="chat",
     )
-    azure_openai_lm("azure openai query")
+    assert azure_openai_lm("azure openai query") == expected_response
+
+
+def test_text_lms_can_be_queried(litellm_test_server):
```
dbczumar, line 31: Since we're making changes to `litellm_text_completion` as well, we should have some coverage for LM queries with `model_type="text"`

tests/clients/test_lm.py
```diff
@@ -1,24 +1,77 @@
...
+    api_base, server_log_file_path = litellm_test_server
+
+    lm_with_unhashable_callable = dspy.LM(
+        model="openai/dspy-test-model",
+        api_base=api_base,
+        api_key="fakekey",
+        # Define a callable kwarg for the LM to use during inference
+        azure_ad_token_provider=lambda *args, **kwargs: None,
+    )
+    lm_with_unhashable_callable("Query")
```
dbczumar, line 62: Fails on `main` with:

```
E       TypeError: <function test_lm_calls_support_unhashable_types.<locals>.<lambda> at 0x31204d5a0> is not JSON serializable
```

tests/clients/test_lm.py
```diff
@@ -1,24 +1,77 @@
...
+    class ResponseFormat(pydantic.BaseModel):
+        response: str
+
+    lm = dspy.LM(
+        model="openai/dspy-test-model",
+        api_base=api_base,
+        api_key="fakekey",
+        response_format=ResponseFormat,
+    )
+    lm("Query")
```
dbczumar, line 77: Fails on `main` with:

```
TypeError: <class 'tests.caching.test_caching.test_lm_calls_support_pydantic_models.<locals>.ResponseFormat'> is not JSON serializable
```

dspy/clients/lm.py
CyrusNuevoDia: Looks cool! 

Could set default `maxsize = float("inf")` here

CyrusNuevoDia:
Looks awesome! Is there a way to have a global cache that we can dump/load?

dbczumar:
> Looks awesome! Is there a way to have a global cache that we can dump/load?

Totally! We can add that if / when we need it by leveraging cachetools `LRUCache.items()` method

CyrusNuevoDia:
Awesome, lgtm! Appreciate you 🙏

dspy/clients/lm.py
CyrusNuevoDia: Can you do a global `dspy.settings.request_cache default = LRUCache(maxsize=10_000_000)` and then have this function pull from that?

CyrusNuevoDia (changes requested):
One quick fix and lgtm

dspy/clients/lm.py
dbczumar: With this naming, it could be confused with the disk cache though, right? It seems like we'd want some unified way to refer to both caches, or more distinctive naming. Thoughts?

dspy/clients/lm.py
okhat: This is not a blocker for merge, but I'm slightly uneasy about locking. Maybe it's required for thread safety, but since it's happening for every single LM call it's a bit worrisome.

dbczumar: Thanks Omar! Functools uses a Python lock as well (Rlock). Ill follow up with a small PR to use Rlock instead of Lock.

Issue #1899, opened 2024-12-07
Support more ast types in Signature typed parsing


thomasahle:
On the website, you [see this example](https://dspy.ai/#__tabbed_3_3): 
```python
import dspy
dspy.configure(lm=dspy.LM('gpt-4o-mini-2024-07-18'))

# Define the DSPy module for classification. It will use the hint at training time, if available.
signature = dspy.Signature("text -> label").with_updated_fields('label', type_=Literal[tuple(CLASSES)])
classify = dspy.ChainOfThoughtWithHint(signature)

# Optimize via BootstrapFinetune.
optimizer = dspy.BootstrapFinetune(metric=(lambda x, y, trace=None: x.label == y.label), num_threads=24)
optimized = optimizer.compile(classify, trainset=trainset)

optimized_classifier(text="What does a pending cash withdrawal mean?")
```

This would have been a lot cleaner as
```
signature = dspy.Signature(f"text -> label:Literal{CLASSES}")
```

We already has support for most of this.
If you see https://github.com/stanfordnlp/dspy/blob/a8d11075daf7b39b890bf47d8d9d4aa8b1cc9b14/dspy/signatures/signature.py#L402 you just have to add a case for `ast.Constant`, like this:

```python
    if isinstance(node, ast.Constant):
        return node.value
```

Then you get
```
In[1]: dspy.Signature(f"text -> label:Literal[1,2]")
Out[1]:
StringSignature(text -> label
    instructions='Given the fields `text`, produce the fields `label`.'
    text = Field(annotation=str required=True json_schema_extra={'__dspy_field_type': 'input', 'prefix': 'Text:', 'desc': '${text}'})
    label = Field(annotation=Literal[1, 2] required=True json_schema_extra={'__dspy_field_type': 'output', 'prefix': 'Label:', 'desc': '${label}'})
)
```
and
```
In[1]: CLASSES = ["good", "bad"]
In[2]: dspy.Signature(f"text -> label:Literal{CLASSES}")
Out[2]:
StringSignature(text -> label
    instructions='Given the fields `text`, produce the fields `label`.'
    text = Field(annotation=str required=True json_schema_extra={'__dspy_field_type': 'input', 'prefix': 'Text:', 'desc': '${text}'})
    label = Field(annotation=Literal['good', 'bad'] required=True json_schema_extra={'__dspy_field_type': 'output', 'prefix': 'Label:', 'desc': '${label}'})
)
```

okhat:
[...]

PR #1900, opened 2024-12-07
Added support for more types

thomasahle:
Adds support for all python `ast` types in string signatures, like `Signature("text -> label:Literal['a', 'b']")`.
Fixes https://github.com/stanfordnlp/dspy/issues/1899


thomasahle:
Ah, I forgot it needs to work in Python 3.9.
NoneType was only added in 3.10:
https://docs.python.org/3/library/types.html#types.NoneType

PR #1905, opened 2024-12-08
Caching: Use reentrant locks, don't discard callables (or any other unhashable object in the future) from the cache key

dbczumar:
Caching: Use reentrant locks, don't discard callables (or any other unhashable object in the future) from the cache key

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -299,7 +299,6 @@ def cached_litellm_completion(request: Dict[str, Any], num_retries: int):
 
 
 def litellm_completion(request: Dict[str, Any], num_retries: int, cache={"no-cache": True, "no-store": True}):
-    print("RUNNING NOT CACHED", request)
     return litellm.completion(
         num_retries=num_retries,
         cache=cache,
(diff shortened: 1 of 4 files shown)


dspy/clients/lm.py
```diff
@@ -233,9 +233,30 @@ def request_cache(maxsize: Optional[int] = None):
...
+        """
+
+        def transform_value(value):
+            if isinstance(value, type) and issubclass(value, pydantic.BaseModel):
+                return value.schema()
+            elif isinstance(value, pydantic.BaseModel):
+                return value.dict()
+            else:
+                try:
+                    return hash(value)
```
dbczumar, line 255: Most functions appear to be hashable. This is tested in `test_lm.py`

dspy/clients/lm.py
```diff
@@ -233,23 +233,56 @@ def request_cache(maxsize: Optional[int] = None):
...
         @functools.wraps(func)
         def wrapper(request: dict, *args, **kwargs):
-            return func(request, *args, **kwargs)
+            try:
+                key = cache_key(request)
+                return func_cached(key, request, *args, **kwargs)
+            except Exception:
+                # If the cache key cannot be computed (e.g. because it contains an unhahsable
+                # value), fall back to the uncached version of the function
+                return func(request, *args, **kwargs)
```
dbczumar, line 285: @okhat Let me know if you think we should print a warning here, since falling back to the disk cache is slower.

PR #1912, opened 2024-12-09
allow non ascii characters in compiled prompts

### LINKED ISSUE #1911: Compiled models are saved using ascii characters only
The compiled prompts are converted into unicode special characters when saved.

The simple fix is to pass the optional ensure_ascii parameter to json.dumps

RawEnchilada:
Adds an optional ensure_ascii parameter to saving a module, which is consistent which json.dumps declarations.

closes #1911

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -160,9 +160,9 @@ class BaseModule:
                 # `use_legacy_loading` is only applicable for BaseModule instances.
                 param.load_state(state[name])
 
-    def save(self, path, save_field_meta=False):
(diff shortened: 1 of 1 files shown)


chenmoneygithub (changes requested):
Thanks for the PR! 

Can we use a try-except block instead of introducing a new arg? IIUC the loading should always work, only saving could cause issue on ascii-only platform. In general we are cautious at introducing new args to public APIs, the longer the arglist, the more confusing it becomes.

okhat:
Just so I understand. Do you want to allow ensure_ascii=False? Should we just make it False by default?

PR #1918, opened 2024-12-10
Caching fix - only try/cache key creation

dbczumar:
Caching fix - only try/cache key creation

### DIFF
--- a/tests/caching/test_caching.py
+++ b/tests/caching/test_caching.py
@@ -111,6 +111,7 @@ def test_lm_calls_are_cached_in_memory_when_expected(litellm_test_server, tempor
     assert len(request_logs) == 2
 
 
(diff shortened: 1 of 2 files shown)


tests/caching/test_caching.py
```diff
@@ -150,3 +150,16 @@ def test_lm_calls_with_callables_are_cached_as_expected():
         lm_without_callable("Query")
 
         assert mock_completion.call_count == 2
+
+
+def test_lms_called_expected_number_of_times_for_cache_key_generation_failures():
```
dbczumar, line 155: Fails on main:

```
FAILED test_caching.py::test_lms_called_expected_number_of_times_for_cache_key_generation_failures - AssertionError: assert 2 == 1
```

PR #1922, opened 2024-12-10
Feature/caching

hmoazam:
One single caching interface which has two levels of cache - in memory lru cache and fanout (on disk)

### DIFF
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -15,10 +15,12 @@ from dspy.clients import * # isort: skip
 from dspy.adapters import * # isort: skip
 from dspy.utils.logging_utils import configure_dspy_loggers, disable_logging, enable_logging
 from dspy.utils.asyncify import asyncify
+from dspy.utils.cache import Cache
 
 settings = dsp.settings
 
 configure_dspy_loggers(__name__)
+
 
 # LM = dsp.LM
 
@@ -64,6 +66,13 @@ You = dsp.You
 configure = settings.configure
 context = settings.context
 
+CACHE_DIR = os.environ.get("DSPY_CACHEDIR") or os.path.join(Path.home(), ".dspy_cache")
+CACHE_LIMIT = int(os.environ.get("DSPY_CACHE_LIMIT", 1e10))  # 10 GB default
+MEM_CACHE_LIMIT = float(os.environ.get("DSPY_CACHE_LIMIT", float("inf"))) # unlimited by default
+
+# Initialize the cache
+dspy_cache = Cache(directory=CACHE_DIR, disk_size_limit=CACHE_LIMIT, mem_size_limit=MEM_CACHE_LIMIT)
+settings.configure(cache=dspy_cache)
 
 import dspy.teleprompt
 
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -19,7 +19,7 @@ from dspy.utils.callback import BaseCallback, with_callbacks
 
 from .base_lm import BaseLM
 
-from .cache import dspy_cache_decorator
+from dspy.utils.cache import dspy_cache_decorator
 
 logger = logging.getLogger(__name__)
 
@@ -219,13 +219,16 @@ class LM(BaseLM):
 
         return new_instance
 
-
-@dspy_cache_decorator()
 def cached_litellm_completion(request: Dict[str, Any], num_retries: int):
-    return litellm_completion(
-        request,
-        num_retries=num_retries,
-    )
+    from dspy import settings
+    @dspy_cache_decorator(settings.cache)
+    def cached_litellm_completion_inner(request: Dict[str, Any], num_retries: int):
+        return litellm_completion(
+            request,
+            cache={"no-cache": False, "no-store": False},
+            num_retries=num_retries,
+        )
+    return cached_litellm_completion_inner(request, num_retries)
 
 
 def litellm_completion(request: Dict[str, Any], num_retries: int, cache={"no-cache": True, "no-store": True}):
@@ -235,13 +238,15 @@ def litellm_completion(request: Dict[str, Any], num_retries: int, cache={"no-cac
         **request,
     )
 
-
-@dspy_cache_decorator()
 def cached_litellm_text_completion(request: Dict[str, Any], num_retries: int):
-    return litellm_text_completion(
-        request,
-        num_retries=num_retries,
-    )
+    @dspy_cache_decorator()
+    def cached_litellm_text_completion_inner(request: Dict[str, Any], num_retries: int):
+        return litellm_text_completion(
+            request,
+            cache={"no-cache": False, "no-store": False},
+            num_retries=num_retries,
+        )
+    return cached_litellm_text_completion_inner(request, num_retries)
 
 
 def litellm_text_completion(request: Dict[str, Any], num_retries: int):
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -1,4 +1,3 @@
-import functools
 import logging
 import os
 import threading
@@ -19,6 +18,8 @@ from dspy.clients.utils_finetune import DataFormat, infer_data_format, validate_
 from dspy.utils.callback import BaseCallback, with_callbacks
 
 from .base_lm import BaseLM
+
+from .cache import dspy_cache_decorator
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


dspy/__init__.py
```diff
@@ -64,6 +66,13 @@
 configure = settings.configure
 context = settings.context
 
+CACHE_DIR = os.environ.get("DSPY_CACHEDIR") or os.path.join(Path.home(), ".dspy_cache")
```
okhat, line 69: Note to self: .dspy_cache is already used for litellm, this PR should be swapped (it uses .dspy_litellm_cache for the existing cache but that cache doesn't exist at that new path)

dspy/clients/lm.py
```diff
@@ -218,84 +219,16 @@ def copy(self, **kwargs):
...
-
-@request_cache(maxsize=None)
 def cached_litellm_completion(request: Dict[str, Any], num_retries: int):
-    return litellm_completion(
-        request,
-        cache={"no-cache": False, "no-store": False},
-        num_retries=num_retries,
-    )
+    from dspy import settings
+    @dspy_cache_decorator(settings.cache)
```
okhat, line 224: hmm this is a strange pattern; is the inner function redefined every time? does this mean the decorator... works? I guess it does, but yeah it's a bit overkill probably?

dspy/clients/lm.py
```diff
@@ -305,17 +238,18 @@ def litellm_completion(request: Dict[str, Any], num_retries: int, cache={"no-cac
...
-    return litellm_text_completion(
-        request,
-        num_retries=num_retries,
-        cache={"no-cache": False, "no-store": False},
-    )
+    @dspy_cache_decorator()
+    def cached_litellm_text_completion_inner(request: Dict[str, Any], num_retries: int):
+        return litellm_text_completion(
+            request,
+            cache={"no-cache": False, "no-store": False},
```
okhat, line 246: This passes cache to the function below. But for some reason we removed the cache kwarg below? Why?

dspy/utils/cache.py
```diff
@@ -0,0 +1,123 @@
...
+        params = {k: transform_value(v) for k, v in request.items()}
+        return sha256(ujson.dumps(params, sort_keys=True).encode()).hexdigest()
+    
+    def get(self, request: Dict[str, Any]) -> Any:
+        try:
+            # try/except in case can't compute key
+            key = self.cache_key(request)
+        except Exception:
+            return None
+        with self.lock:
```
okhat, line 69: Hmm do we need to lock just to read?

okhat: if we can remove this lock, we may wanna do something like `if key := self.memory_cache.get()` to avoid double access

dspy/clients/lm.py
hmoazam: The default (both kwargs true) should be added back to `litellm_text_completion` function (I forgot to fix this)

dspy/utils/cache.py
hmoazam: I would think reads are thread safe but the [docs](https://cachetools.readthedocs.io/en/latest/index.html#:~:text=Note%20Please%20be%20aware%20that%20all%20these%20classes%20are%20not%20thread%2Dsafe.%20Access%20to%20a%20shared%20cache%20from%20multiple%20threads%20must%20be%20properly%20synchronized%2C%20e.g.%20by%20using%20one%20of%20the%20memoizing%20decorators%20with%20a%20suitable%20lock%20object.) aren't clear.

dspy/clients/lm.py
hmoazam: It does work but definitely a weird pattern. Not 100% sure if it has any negative side effects.

okhat:
Recording some thoughts here.

1. We need to make sure that caching behavior makes sense if the function being cached raises an Exception. 
2. We need to make sure that the caching behavior makes sense for LiteLLM. Specifically, LiteLLM's default caching is a bit nuanced: it does not cache api_* keys (which is important) and it does not cace metadata (like cost), as far as I can tell.

dspy/clients/lm.py
```diff
@@ -89,15 +86,18 @@ def __call__(self, prompt=None, messages=None, **kwargs):
...
-            num_retries=self.num_retries,
-        )
+        if cache:
+            @cache_decorator(keep=litellm.Cache()._get_relevant_args_to_use_for_cache_key())
+            def cached_completion(**kwargs):
+                return completion(**kwargs, cache={"no-cache": False, "no-store": False})
+            
+            wrapped_completion = cached_completion
+
+        response = wrapped_completion(model=self.model, messages=messages, num_retries=self.num_retries, **kwargs)
```
dbczumar, line 99: @hmoazam This might be the culprit explaining why caching isn't working properly in our tests. We seem to be constructing a new LiteLLM cache & completion function with `@cache_decorator` every time an LM is called

dspy/utils/cache.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, directory, disk_size_limit, mem_size_limit):
+        """
+        Args:
+            directory: The directory where the disk cache is stored.
+            disk_size_limit: The maximum size of the disk cache (in bytes).
+            mem_size_limit: The maximum size of the in-memory cache (in number of items).
+        """
+
+        self.memory_cache = LRUCache(maxsize=mem_size_limit)
+        self.fanout_cache = FanoutCache(shards=16, timeout=2, directory=directory, size_limit=disk_size_limit)
```
dbczumar, line 44: @hmoazam As I understand it, part of the motivation for this PR is to enable users to dump the current state of the cache for a particular DSPy session.

Is the `FanoutCache` introduced for that purpose? If so, since `FanoutCache` shares the same `directory` as the `litellm.cache`, what extra utility is it providing?

dbczumar: Separately, I think we should define the behavior for dumping / loading a cache. 

For example, consider the case where I've already run "Program 1" as a Python script / notebook. Now, I'm running "Program 2" in a separate notebook / Python script invocation. I'm using the default DSPy cache directory for both program executions.

If I dump the cache after I run Program 2, do I also want to include cache contents from Program 1? Or, do I only want to include cache contents from Program 1, since that's my "current session"? Should users have a choice?

Once we align on the desired semantics, we can figure out how to make adjustments to the implementation.

cc @okhat

okhat: @dbczumar For disk caching, we will migrate from LiteLLM's caching to FanoutCache. The former has proved limiting when users need many read/write processes and threads in parallel.

For a short period of time, we might keep the LiteLLM cache _as well_. This means that highly active users that upgrade versions regularly will have an implicit seamless migration of their caches, because LiteLLM's requests will remain cached and will implicitly transfer to the FanoutCache. (This behavior is not required, and in principle there are other ways to more intentionally allow migration, but explicit migration is almost never worthwhile for caches.)

Cache dumping is fairly simple. It saves the current in-memory Python session, unless the user triggers a reset for the in-memory cache. (The current in-memory session draws on saved caches on disk, as usual.)

dbczumar: @okhat @hmoazam Apologies for the delay.

I took a look at this further and found that diskcache does not natively handle objects that can't be pickled (like lambdas), but LiteLLM does. There's a nontrivial amount of logic that goes into this. It may be easier to add support for FanoutCache to LiteLLM, rather than implement our own diskcache wrapper. Are there other reasons that we need to move away from LiteLLM?

Separately, it's not clear to me that dumping the in-memory Python session is what we want. If I understand correctly, the user's objective is to run a workload and store all of the cached values for that workload on disk. That way, another user can set `DSPY_CACHEDIR` to that disk location and automatically use the cached values. Dumping memory isn't quite the same thing because users might have memory limits set on their cache. Further, dumping memory to a separate location than the disk cache (as is done in this PR) complicates the experience of loading the cache (I can't just set `DSPY_CACHEDIR` to the desired location - I have to worry about loading this pickled memory state as well). To solve both of these problems, I'd propose an API like the following:

```
dspy.settings.cache.start_recording(location="/location/of/cache/that/I/want/to/share")
# Run workload
dspy.settings.cache.stop_recording() # Optional
```

When recording is enabled, results will be dual written to the main disk cache location (e.g. the one specified by `DSPY_CACHEDIR`) *and* the user-specified disk cache location. This way, recipients of the cache who want to reuse its values don't have to reason about loading stuff from disk vs memory - the snapshotted (recorded) cache looks just like any other disk cache (because it is :D).

hmoazam: Thanks @dbczumar! I've been looking into the failing tests and was still figuring out the part with pickling. I'll let Omar comment on any other reasons for moving away from litellm's cache, but the main motivations were for us to have more options e.g. use fanoutcache, and offer additional methods like saving and loading. 

We originally envisioned the user journey for saving/loading as the following: 
1. User executes a program, assuming caching is turned on, calls are cached both in memory and on disk. 
2. User decides they want to port some subset of the cache, so they run a new session of the notebook, executing only the calls they wish to export, which end up in the in-memory cache, and can then be dumped out with the save method. This means they do not have to reset the disk cache or create another one. 
3. User calls the load method with the path to the saved cache, which loads it into memory 

I do think that we should keep the 2 levels of caching, and I do see we have a gap with point 3, in that we should also populate the disk cache with those values explicitly. Your suggestion does help us achieve a similar experience for the users, but my main concern is whether we would run into issues simply because of the pickling limitations. I haven't gone through all the litellm code yet, but with your proposed approach we still face that limitation.

dbczumar: @hmoazam We can also add Fanout to LiteLLM directly. My motivation here is that LiteLLM handles pickling on top of diskcache directly. The problem with building our own wrapper over diskcache is that diskcache does not handle pickling directly - i.e. we won't be able to save many valid LiteLLM request objects (e.g. requests with lambdas) to *disk* unless we reuse LiteLLM's pickle compatibility logic (which is private / internal) or roll our own. This is a problem even *without* memory pickling (two separate issues).

Regarding save / load, I think this is what the memory / dumping logic in your PR (and my proposed `start_recording()` API) is attempting to handle. We can still wrap LiteLLM to achieve this; I don't think save / load is sufficient motivation for rewriting the wrapper around diskcache to achieve this (why not wrap litellm.caching.Cache, which saves us all the pickle + disk headaches?)

> User decides they want to port some subset of the cache, so they run a new session of the notebook, executing only the calls they wish to export, which end up in the in-memory cache, and can then be dumped out with the save method. This means they do not have to reset the disk cache or create another one.

- By definition *porting* means creating another cache (let me know if I'm missing something

- From a solution standpoint, "dump memory" is an *indirect* way of achieving portability. It relies on the assumption that all of the data in the new session is cached in memory and that, therefore, dumping memory will create the desired "port" (copy) of the cache. However, this may not be the case if the user has set memory limits on the cache (in which case some of the data from the session is only on disk and never gets ported), and it also relies on the user clearing memory before running the session to ensure a "clean port" (which can be inefficient if most of the data is already in memory, and only a small subset needs to be excluded from the port). "start_recording()" is more explicit, doesn't require clearing a memory cache, and doesn't risk data loss from the cache port due to configured memory limits.

@okhat

PR #1923, opened 2024-12-11
Update search.py

MaxwellSalmon:
Two edits:

1. Removed `not` from line in `retrieveEnsemblewithMetadata`, as the logic was faulty: using a reranker if it has *not* been set.
2. Retrieving with metadata was not possible - `retrievewithMetadata` was not called and `**kwargs` was missing.

Additionally, using `WeaviateRM` I believe that metatada will never be returned either way, due to the line `parsed_results = [result.properties[self._weaviate_collection_text_key] for result in results.objects]`, as it only returns the collection key text. An issue should be created if my understanding is true.

### DIFF
--- a/dsp/primitives/search.py
+++ b/dsp/primitives/search.py
@@ -122,7 +122,7 @@ def retrieveEnsemblewithMetadata(
     queries = [q for q in queries if q]
 
     if len(queries) == 1:
-        return retrieve(queries[0], k)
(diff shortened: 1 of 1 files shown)


okhat:
Thanks @MaxwellSalmon ! This whole file was removed for 2.6, was just some outdated stuff :D

PR #1943, opened 2024-12-16
Simplify built-in modules (remove new_* & extended_signature). Prepare for assertions v2.

okhat:


asparagus:
Where can I read about assertions v2?

marfago:
I have just updated from v2.5 to 2.6 and my code does not work anymore becuase `dspy.Assert` and `dspy.Suggest` do not exist anymore. Can you please share some documentation to port them from v2.5 to v2.6?

Nasreddine:
> @okhat I have just updated from v2.5 to 2.6 and my code does not work anymore becuase `dspy.Assert` and `dspy.Suggest` do not exist anymore. Can you please share some documentation to port them from v2.5 to v2.6?

I think you should revert to v2.5, as I did, since assertions in v2 are not yet complete.

I wonder why assertions were removed while v2 is still not fully implemented.

marfago:
@Nasreddine Thank you for confirming. I have already reverted to the previous version. I agree with you, this is a regression of 2.6.0 with respect to 2.5.x (the entire assertion logic code has been commented out and the feature removed).

I would also add that the documentation is quite messy right now between 2.4/2.5/2.6  and it is not really clear what is the recommended way to solve problems.

PR #1955, opened 2024-12-18
Improve literal format adherence in ChatAdapter, JSONAdapter

dbczumar:
Improve literal format adherence in ChatAdapter

**Testing**

Confirmed that existing reliability tests pass with GPT-4o and Llama3.1 70B Instruct


dspy/adapters/utils.py
```diff
@@ -103,6 +103,26 @@ def find_enum_member(enum, identifier):
...
+    args = get_args(annotation)
+    if origin is None:
+        if hasattr(annotation, "__name__"):
+            return annotation.__name__
+        else:
+            return str(annotation)
+
+    if origin is Literal:
+        args_str = ", ".join(
+            _quoted_string_for_literal_type_annotation(a) if isinstance(a, str) else get_annotation_name(a)
```
dbczumar, line 117: @okhat This is slightly trickier than using `repr()`, since `repr()` inserts undesired backslashes that are likely to confuse an LM. I think the approach proposed in the PR should hopefully reduce LM confusion with strings containing quotes

PR #1959, opened 2024-12-19
Introduce `dspy.Refine`, `dspy.BestOfN` and fix thread safety for `dspy.ChainOfThoughtWithHint`

chenmoneygithub:
`dspy.Refine` is a new API we are adding to support feedback-based retry. More details coming soon.


younes-io:
I'm really interested in this... I mean in terms of specs!

dspy/refine/metrics.py
```diff
@@ -0,0 +1,35 @@
...
+class BoolMetric(Metric):
+    def __init__(self, name, description, fn):
+        super().__init__(name, description, fn)
+        self.type_description = "This is a bool metric, true if the metric looks good, false otherwise."
+
+    def __repr__(self):
+        return f"{super().__repr__()}\nMetric type:{self.type_description}"
+
+
+class FloatMetric(Metric):
```
zbambergerNLP, line 29: Perhaps worth including lower and upper bounds for numeric scores? These can be specified concretely in pydantic fields (https://docs.pydantic.dev/latest/concepts/fields/#numeric-constraints)

dspy/refine/refine.py
```diff
@@ -0,0 +1,128 @@
...
+
+from dspy.predict.chain_of_thought import ChainOfThought
+from dspy.primitives.program import Module
+from dspy.refine.feedback import GenerateFeedback
+from dspy.refine.utils import get_traces
+from dspy.signatures.field import InputField
+
+
+class Refine(Module):
+    def __init__(self, module, metrics, metric_thresholds=None, max_iter=3):
```
zbambergerNLP, line 13: The type for metric thresholds is very vague. Can you add type annotations/a description to articulate this better? 
Also, wouldn't metadata/parameters related to the metric be better defined when initializing the metric itself (i.e., define per metric what is the minimum allowed performance)?

dspy/refine/metrics.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, name, description, fn):
+        super().__init__(name, description, fn)
+        self.type_description = "This is a float metric, the higher the value the better."
```
zbambergerNLP, line 32: What if we're dealing with a metric like perplexity or a bias metric, where smaller values are preferred? I imagine that it'd be best to add support for this via some addition boolean initialization parameter for `larger_is_better`.

dspy/refine/metrics.py
```diff
@@ -0,0 +1,35 @@
+import inspect
+
+
+class Metric:
```
zbambergerNLP, line 4: Perhaps worth adding an integer/categorical Metric sub-type? This would accommodate the trend of using concrete rubrics for score assignment (i.e., each discrete category/score has a string description for when an input should be classified as that category/score).

dspy/refine/feedback.py
```diff
@@ -0,0 +1,27 @@
...
+    module_inputs: dict = InputField(desc="The inputs of the DSPy module")
+    module_outputs: dict = InputField(desc="The outputs of the DSPy module")
+    source_code: str = InputField(desc="The source code of the DSPy module")
+    feedback: dict[str, list[str]] = OutputField(
+        desc="Feedback for the DSPy module in general, along with feedback for each submodule in the DSPy model, only "
+        "provide feedback for attributes in `__init__` method that is a built-in class of dspy. The key should be the "
+        "attribute name, e.g., `self.cot` or `self.predict`. If the attribute is a list, write the key as "
+        "`self.cots[0]`, `self.predicts[1]` and so on. The feedback should be "
+        "a list of strings, corresponding to each score function in `metrics`."
+    )
```
zbambergerNLP, line 27: This signature is very complex, and presumably accepts a very long input. Can small models (e.g., GPT-4o-mini) also use this reliably to generate feedback? Or is this meant more for larger/reasoning oriented models?

zbambergerNLP (commented):
Really cool PR!

A project I am working on used to leverage the dspy.Suggest/Assert functionality (before the recent deprecation), and this looks like an amazing substitute. 

I added some comments/feedback/questions, which I hope will help in building this. Let me know if I can be of service :D

dspy/refine/metrics.py
chenmoneygithub: you can take the opposite number, similar to how loss function works in pytorch.

dspy/refine/refine.py
chenmoneygithub: threshold is not an attribute of metric, it only makes sense at Refine/Assert time.

chenmoneygithub:
@zbambergerNLP  Thanks for your comments! 

Usually we don't recommend community dropping comments on PR, a much better way to contribute is trying this API out in your workflow, and let us know what's good/bad (ideally with a colab link for reproducing). We will have this API merged and released in 1-2 weeks, stay tuned!

dspy/refine/feedback.py
chenmoneygithub: I would say try it out! This doesn't really go crazy. Multi-hop RAG provides way longer prompt in fact.

okhat:
Thanks a lot @zbambergerNLP !

@chenmoneygithub , @zbambergerNLP is not just a community member, he's been building some cool things in the broader team :D

dspy/refine/refine.py
```diff
@@ -0,0 +1,128 @@
...
+                predict.original_forward = predict.forward
+
+            partial_forward = partial(
+                forward_with_feedback, dspy_refine_feedback=feedback, dspy_refine_last_trace=last_trace
+            )
+            # Patch the `forward` method to the `forward_with_feedback` methd with partial values of feedback and
+            # last_trace.
+            predict.forward = MethodType(partial_forward, predict)
+
+    def _undo_patch_predict_call_with_feedback(self, named_predicts):
```
okhat, line 73: ah this is not thread safe

PR #1964, opened 2024-12-19
Enforcing 3 Demos in MIPROv2 Meta Prompt

klopsahlong:
### Previous Behavior:
In some cases, only 1 demo was included in the meta-prompt. This occurred when few-shot examples were generated with a size of 1 (since the size is randomly chosen between 1: max_bootstrap_fewshot), and when max_bootstrap_fewshot is set to 1.

### New behavior:
We wanted for the # of few-shot examples in the prompt to always equal 3 to avoid potential overfitting of the generated instruction to a single example.

### The fix: 
We will essentially did the following: 
prompt_demos = (my_demos + all_other_demos)[:min(3, len(my_demos + all_other_demos))]

Which will give us 3 demos for all prompts, unless there are less than 3 demos total across all bootstrapped few-shot examples. Note that we keep the demos empty for a single instruction generation, just in case demos end up being counter productive in the meta prompt.

### A few other minor additions: 
- Renamed some variables for legibility
- Added back in the generated module description to the meta-prompt, which summarizes what the module we're generating an instruction for is responsible for

### DIFF
--- a/dspy/propose/grounded_proposer.py
+++ b/dspy/propose/grounded_proposer.py
@@ -5,7 +5,7 @@ from dspy.propose.dataset_summary_generator import create_dataset_summary
 from dspy.propose.utils import create_example_string, create_predictor_level_history_string, strip_prefix, get_dspy_source_code
 from dspy.teleprompt.utils import get_signature, get_prompt_model
 
-from dspy.propose.propose_base import Proposer
+from .propose_base import Proposer
 
 # Hardcoded variables (TODO: update)
 MAX_INSTRUCT_IN_HISTORY = 5  # 10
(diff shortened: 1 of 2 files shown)


dspy/propose/grounded_proposer.py
```diff
@@ -385,15 +401,19 @@ def propose_instruction_for_predictor(
         # Generate a new instruction for our predictor, using the temperature specified for this round
         original_temp = self.prompt_model.kwargs["temperature"]
 
+        epsilon = self.rng.uniform(0.01, 0.05)
+        modified_temp = T + epsilon
+
         with dspy.settings.context(lm=self.prompt_model):
-            self.prompt_model.kwargs["temperature"] = T
+            self.prompt_model.kwargs["temperature"] = modified_temp
```
okhat, line 408: this is not how we'd wanna do this btw, this affects the LM object in-place

normally we'd do a .copy(temperature=...)

PR #1988, opened 2024-12-25
Add async support to `Module` class

### LINKED ISSUE #1975: Async support in dspy.ReAct
`dspy.asyncify` is a good start but it doesn't feel idiomatic. Would be cool if we had an AsyncModule where the `forward` method is async or alternatively add an `aforward` method to `dspy.Module`. The reason this would be useful is

1. The user will be able to control parallelization inside their `Module`'s forward and
2. This will enable async tool calls with `dspy.ReAct` for example
3. Better integration with async frameworks such as FastAPI etc..

At the moment, using `dspy.asyncify` on a Module doesn't allow easy async programming inside the forward method because you can't use the `async` keyword, you have to resort to interacting with the active asyncio loop etc..

younes-io:
This PR adds native async support to DSPy's `Module` class, making it easier to work with async operations. Instead of using `asyncify`, you can now write `async` modules naturally:

- Added `aforward` method for async implementations
- Detection of async arguments
- Works with both sync and async code
- Fully tested with different async patterns

Please, let me know your feedback on this initial async implementation..

Fixes #1975

### DIFF
--- a/dspy/primitives/program.py
+++ b/dspy/primitives/program.py
@@ -1,8 +1,13 @@
 import magicattr
+import inspect
+import asyncio
+from typing import Any, Union, Awaitable, TypeVar, Optional, List, Callable
 
 from dspy.predict.parallel import Parallel
 from dspy.primitives.module import BaseModule
 from dspy.utils.callback import with_callbacks
+
+T = TypeVar('T')
 
 
 class ProgramMeta(type):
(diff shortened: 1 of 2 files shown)


okhat:
Thanks a lot @younes-io ! Tagging @CyrusNuevoDia , the async expert here :D

dspy/primitives/program.py
```diff
@@ -18,9 +23,117 @@ def __init__(self, callbacks=None):
...
+                        async_args.append(await arg)
+                    else:
+                        async_args.append(arg)
+                
+                async_kwargs = {}
+                for k, v in kwargs.items():
+                    if is_async_arg(v):
+                        async_kwargs[k] = await v
+                    else:
+                        async_kwargs[k] = v
```
CyrusNuevoDia, line 70: Can you make this use gather so they solve concurrently?

CyrusNuevoDia:
Thanks @younes-io this is great!

2 questions --

1. Can this directly use an Async LM instead of a sync LM?
2. Can you test this with the sync optimizers?

CyrusNuevoDia (changes requested):

younes-io:
> Thanks @younes-io this is great!
> 
> 2 questions --
> 
> 1. Can this directly use an Async LM instead of a sync LM?
> 2. Can you test this with the sync optimizers?

Thank you for looking into this.. 

The short answer would be:

1 - No, I need an Async LM interface for this
2 - Yes, it should work but I need to write tests for this

This commit is just the first step.. I'm trying to do this step by step and write tests to avoid regressions because it could get tricky

PR #1993, opened 2024-12-27
Added missing project dependency `pillow`.

CodeReclaimers:


okhat:
Thanks @CodeReclaimers ! It's an optional dependency only required when using Images

PR #2003, opened 2025-01-01
Add a description to ReAct module's input fields

itay1542:
added a short description in the ReAct output fields because some models (including big ones like the latest claude-3-5-sonnet) sometimes output the ReAct keys without the "next" string, for example:
```
[[ ## next_thought ## ]]

[[ ## tool_name ## ]] <-- missing "next"

[[ ## next_tool_args ## ]]
```
This should fix a lot of instances of these errors:
`ValueError: Expected dict_keys(['next_thought', 'next_tool_name', 'next_tool_args']) but got dict_keys([])`


okhat:
Hey @itay1542 ! This is SUPER helpful, thank you.

I didn't know that claude 3.5 sonnet struggled with this. That's a very useful report.

I think we must fix that, but I don't think the fundamental fix is by changing the ReAct descriptions. I think that has to happen at the Adapter layer instead. A good adapter should ensure this kind of thing passes for all popular models out of the box, and should fix the issue on retry if it fails the first time.

I will keep this PR open as a "draft" so that we can prioritize adding this unit test (sonnet ReAct) to our adapters in 2.6

PR #2004, opened 2025-01-01
AsyncLM

itay1542:
Hey @okhat, thanks for this repository its really super.

One step towards proper async functionality is having an `AsyncLM`, similar to how OpenAI has their Client and AsyncClient.
I know this PR has some code duplication issues, however I think its better to have a separate async LM rather than handling both sync and async in the `LM`'s `__call__`.

This PR complements #1988 and lays a foundation for async modules to use async lm client


dspy/clients/async_lm.py
```diff
@@ -0,0 +1,138 @@
...
+import litellm
+from anyio.streams.memory import MemoryObjectSendStream
+from litellm.types.router import RetryPolicy
+
+import dspy
+from dspy.clients.lm import LM, request_cache
+from dspy.utils import with_callbacks
+
+
+class AsyncLM(LM):
```
Goldziher, line 13: Perhaps add ABC and decorate?

dspy/clients/async_lm.py
```diff
@@ -0,0 +1,138 @@
...
+    @with_callbacks
+    def __call__(self, prompt=None, messages=None, **kwargs) -> Awaitable:
```
Goldziher, line 15: This should be async to allow for the caller to use this in an async fashion, no?

```python
async def __call__(...)
```

PR #2009, opened 2025-01-03
Add shuffle to dataset split in `_set_and_validate_datasets`in MIPROv2

Haknt:
### Summary
Added `random.shuffle` to `_set_and_validate_datasets` for better randomness in train/validation split when no validation set is provided.

### Notes
- Improves representativeness of validation set.
- Open to feedback on whether this responsibility fits within the method's scope.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -242,7 +242,7 @@ class MIPROv2(Teleprompter):
                 )
             valset_size = min(1000, max(1, int(len(trainset) * 0.80)))
             cutoff = len(trainset) - valset_size
-            random.shuffle(trainset)
+            self.rng.shuffle(trainset)
             valset = trainset[cutoff:]
             trainset = trainset[:cutoff]
         else:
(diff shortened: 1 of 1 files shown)


okhat:
Thanks @Haknt ! I have a few different thoughts. One is: shuffling should ideally use an RNG object to avoid messing with the global seed, but maybe this is not already respected in the current MIPRO implementation? (can't recall how it was done)

Another one is that ideally the user pre-shuffles their data. It's a little easier to reason about the current behavior as the first 20% vs last 80% but I do see the value of shuffling "just in case" too, so a bit conflicted.

Haknt:
Thanks for the feedback! (1) You’re absolutely right, using `random.shuffle` directly does affect the global RNG state, and I overlooked this.

I’ve updated the code to use the existing `self.rng` for shuffling, ensuring it doesn’t interfere with the global state and aligns with the seed parameter for reproducibility. Now, `self.rng.shuffle()` is used instead of `random.shuffle()`.

(2) I'm thinking out loud, In the ideal scenario, the user should provide the validation set explicitly. If they don’t, it’s likely they are either unaware or prefer not to handle it themselves. In both cases, shuffling would be beneficial to ensure a representative split.

Alternatively, we could make the validation set mandatory and throw an error if it’s not provided. This would ensure users are deliberate about their validation strategy.

Let me know your thoughts!

PR #2021, opened 2025-01-07
React function default args

stantonius:
Adds the ability to provide default args for function calling (TOPIC 2). Has worked well for my tests thus far.

Additionally, I added `private_defaults` that I found useful as a concept to avoid including the function args in the trajectory (in case they included sensitive info like credentials)

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -41,6 +41,9 @@ jobs:
         python-version: ["3.9"]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


okhat:
This is super awesome! @stantonius Thanks a lot! Is this meant to be a draft for now?

stantonius:
Sorry I left as a draft meaning to come right back to it (wasn't sure about the error handling of the `private_defaults`). 

I can take a look at `inspect` as an alternative option

PR #2025, opened 2025-01-08
Update multi_chain_comparison.py

zbambergerNLP:
Encountered a type error locally since my signature was returning a float (score), which doesn't have the `strip()` method. Once we cast this answer to a string, things work smoothly. 

NOTE: It seems like the answers provided within the completions only contain the final output field, but this does not generalize to cases where the signature has multiple output fields.


okhat:
Thanks @zbambergerNLP ! I'm happy to merge this but I realize that the implementation of this module is itself super outdated and was reasonable for an old generation of models but nowadays I'd probably do it differently.

I will merge nonetheless to unblock you but will push a new implementation in the coming weeks.

PR #2031, opened 2025-01-09
Simplify Parallel Executor

CyrusNuevoDia:


tests/evaluate/test_evaluate.py
```diff
@@ -68,44 +68,6 @@ def test_multithread_evaluate_call():
...
-
-    with pytest.raises(KeyboardInterrupt):
-        ev = Evaluate(
-            devset=devset,
-            metric=answer_exact_match,
-            display_progress=False,
-            num_threads=2,
-        )
-        score = ev(program)
-        assert score == 100.0
```
CyrusNuevoDia, line 106: This effectively got moved to `tests/utils/test_parallelizer.py`

okhat:
Thanks a lot @CyrusNuevoDia ! Is this ready to merge

CyrusNuevoDia:
It works... can you think of any tricky testing scenarios?

dspy/utils/parallelizer.py
```diff
@@ -28,180 +29,102 @@ def __init__(
...
                 if self.provide_traceback:
-                    logger.error(
-                        f"Error processing item {item}: {e}\nStack trace:\n{traceback.format_exc()}"
-                    )
+                    logger.error(f"Error processing item {item}: {e}\nStack trace:\n{traceback.format_exc()}")
                 else:
                     logger.error(
                         f"Error processing item {item}: {e}. Set `provide_traceback=True` to see the stack trace."
                     )
+                with self._lock:
```
okhat, line 56: I would probably move this above the logging

dspy/utils/parallelizer.py
```diff
@@ -28,180 +29,102 @@ def __init__(
...
-        self.cancel_jobs = threading.Event()
+        self._lock = threading.Lock()
 
     def execute(self, function, data):
         wrapped_function = self._wrap_function(function)
-        if self.num_threads == 1:
-            return self._execute_isolated_single_thread(wrapped_function, data)
-        else:
-            return self._execute_multi_thread(wrapped_function, data)
+        exec_type = "multi" if self.num_threads != 1 else "single"
```
okhat, line 36: would could `self._execute_single_thread if self.num_threads == 1 else self._execute_multi_thread`?

dspy/utils/parallelizer.py
```diff
@@ -28,180 +29,102 @@ def __init__(
...
+        executor = getattr(self, f"_execute_{exec_type}_thread")
+        return executor(wrapped_function, data)
 
     def _wrap_function(self, function):
+        from dspy.dsp.utils.settings import thread_local_overrides
+
         # Wrap the function with error handling
         def wrapped(item):
-            if self.cancel_jobs.is_set():
```
okhat, line 44: I'm not entirely sure, but would this save us some delays in practice?

dspy/utils/parallelizer.py
```diff
@@ -28,180 +29,102 @@ def __init__(
...
-                signal.signal(signal.SIGINT, interrupt_handler)
-                try:
-                    yield
-                finally:
-                    signal.signal(signal.SIGINT, default_handler)
-            else:
-                # If not in the main thread, skip setting signal handlers
-                yield
-
-        def cancellable_function(parent_overrides, index_item):
```
okhat, line 147: If I recall correctly, this was really important. It seems to have gotten lost in the (otherwise extremely neat) refactor.

When launching multiple threads, we want each thread to inherit the *parent* thread's local overrides.

dspy/utils/parallelizer.py
```diff
@@ -28,180 +29,102 @@ def __init__(
...
-
-            for future in as_completed(futures):
-                index, result = future.result()
-
-                if result is job_cancelled:
-                    continue
+    def _execute_multi_thread(self, function, data):
+        pbar = self._create_pbar(data)
+        total_score = 0
+        total_processed = 0
```
okhat, line 106: Somewhere here, we'd need to have something like:

```python
parent_overrides = thread_local_overrides.overrides.copy()
```

and then we should pass `parent_overrides` in `data`, so that `wrapped(item)` can handle using the *parent* thread's overrides, not the new child's overrides.

okhat:
Two isses:

- Like in the comments above, we need to carefully handle parent overrides.
- Will also need to update requirements.txt until Hanna moves us to purely pyproject.toml

PR #2199, opened 2025-01-20
Weaviate multitenancy support

Timtech4u:
The changes introduce multitenancy support to the WeaviateRM class. A new optional parameter tenant_id is added to the constructor, allowing the user to specify a tenant ID. This tenant ID can also be provided as a keyword argument during query execution. When a tenant ID is specified, all retrieval requests are scoped to the given tenant ID, ensuring tenant-specific data access within a multitenant Weaviate instance.

### DIFF
--- a/docs/docs/deep-dive/retrieval_models_clients/WeaviateRM.md
+++ b/docs/docs/deep-dive/retrieval_models_clients/WeaviateRM.md
@@ -20,6 +20,8 @@ WeaviateRM(
 ```
 
 ## Using Multitenancy
+Multi-tenancy allows a collection to efficiently serve isolated groups of data. Each "tenant" in a multi-tenant collection can only access its own data, while sharing the same data structure and settings.
+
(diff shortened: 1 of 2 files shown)


okhat:
Thanks @Timtech4u ! Requesting review from @CShorten  :D

CShorten:
Thanks so much for adding this @Timtech4u and the tag @okhat! 

Approved!

Made a few small changes:

Caught one missing `.with_tenant` and added two documentation comments explaining the `tenant_id` constructor argument and a general blurb about what Multi-Tenancy is.

=== Broader Comments ===

Going forward I don't think this is the best pattern for Multi-Tenancy and DSPy Modules in Production -- probably better to pass `tenant_id` in the forward pass versus as internal state to the Module. This looks good for now, but will keep an eye on this as it evolves. 

cc @hsm207

CShorten (approved):

PR #6755, opened 2025-01-27
Fix poetry lock

chenmoneygithub:
resolve #6644

PR #6877, opened 2025-01-28
Update litellm version

onel:


onel:
Already covered by #6755

PR #7261, opened 2025-01-29
Infer optimizer

arnavsinghvi11:
BootstrapFewShot defines the behavior behind the "how" in few-shot example selection by differentiating examples and selecting those that meet the metric criteria as "passing examples" in the optimized prompt. 

BootstrapInfer takes a next step in exploring the "why" by inducing learned rules from the selected few-shot examples and adding supplementary guidance within the optimized prompt. The optimizer explores various sets of complementary rules + fewshot example combinations, and outputs a final optimal prompt with the original DSPy program instruction, selected few-shot examples and additional natural language rules to guide the task.


dspy/teleprompt/bootstrap_infer.py
```diff
@@ -0,0 +1,117 @@
+import copy
+import random
+import numpy as np
+from dspy.teleprompt import BootstrapFewShot
+from dspy.evaluate.evaluate import Evaluate
+import dspy
+    
+class BootstrapFewShotInfer(BootstrapFewShot):
```
chenmoneygithub, line 8: Can we add docstring to explain what this optimizer do? like what's the key improvement from BTFS, and when this optimizer should be used.

dspy/teleprompt/bootstrap_infer.py
```diff
@@ -0,0 +1,117 @@
...
+            num_threads=self.num_threads,
+            max_errors=self.max_errors,
+            display_table=False,
+            display_progress=True,
+            return_all_scores=True
+        )
+        score, _ = evaluate(program, metric=self.metric)
+        return score
+
+class RulesInductionProgram(dspy.Module):
```
chenmoneygithub, line 96: maybe just call this `RuleExtractor`?

chenmoneygithub (changes requested):
@arnavsinghvi11 Interesting one! A few high-level comments:

- could you install ruff, and enable vscode/cursor auto formatting with ruff? There are many lint issues, but most of them can be automatically resolved.
- can we add some explanation on this optimizer with code examples? With a quick scan, I am wondering if this optimizer is only suitable for use cases where there are strong patterns.

PR #7610, opened 2025-01-31
Preventing overwriting of the input var

swiftdevil:
Fixes a bug where the exception handler was using the already-formatted input, which was causing it to choke on the retry attempt. This PR is for the async branch, another will be opened for master.


CyrusNuevoDia:
@swiftdevil huh, interesting, how were you getting this error? 

Seems innocuous enough though, let's find a nice name for that var like sig_inputs instead

PR #7621, opened 2025-01-31
Fix react bug in 2.6.0 that cause tools without type hint to fail

kalanyuz:
## Problem
When using functions without type hints as tools in ReAct, the code would throw an error when trying to access `arg_types[k]` since `arg_types` would be empty. This made it impossible to use simple functions without type annotations as tools, even for basic operations.

## Solution
Modified the tool argument parsing logic to:
1. First check if the tool has `arg_types` attribute
2. Then check if the specific argument key exists in `arg_types`
3. Only attempt type validation for arguments that have explicit type hints
4. Pass through arguments without type hints as-is

Added a test case `test_tool_calling_without_typehint` that demonstrates:
- A simple function `foo(a, b)` without type hints can be used as a tool
- Arguments are correctly passed through without type validation
- The tool executes successfully and returns expected results
- Previous tests, including the ones that uses pydantic args, does not break

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -10,7 +10,13 @@ from dspy.utils.callback import with_callbacks
 
 
 class Tool:
-    def __init__(self, func: Callable, name: str = None, desc: str = None, args: dict[str, Any] = None):
+    def __init__(
+        self,
+        func: Callable,
+        name: str = None,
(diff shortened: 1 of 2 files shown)


okhat:
Hey, great catch. Not sure when this regression happened. Will review soon.

PR #7649, opened 2025-02-03
feat(lm): add support for o3-mini and openai reasoning models

kalanyuz:
# Use max_completion_tokens for OpenAI reasoning models (o1/o3)

## Problem
The OpenAI o3 model family deprecated `max_tokens` in favor of `max_completion_tokens`. Additionally, both o1 and o3 models (including their mini variants) are part of the same "reasoning model" family that require specific configuration (temperature=1.0, tokens >= 5000). This change is does this behind the scenes so that it wouldn't introduce any backward compatibility issues.

## Changes
1. Updated model family detection to use regex pattern matching for all variations:
   - Base models: `o1`, `o3`
   - Mini variants: `o1-mini`, `o3-mini`
   - Any other variants starting with o1/o3
   
2. Unified parameter handling for reasoning models:
   - Use `max_completion_tokens` instead of `max_tokens` for all reasoning models
   - Simplified error message to refer to "reasoning models" as a family

3. Added comprehensive tests:
   - Verify parameter requirements are enforced
   - Check correct parameter naming (`max_completion_tokens` vs `max_tokens`)
   - Test various model name patterns

## Testing
The changes are verified by new test cases in `test_clients/test_lm.py`:
- `test_reasoning_model_requirements`: Verifies temperature and token requirements
- `test_reasoning_model_token_parameter`: Tests parameter naming across different model variants

## Example Usage
```python
# Old way (would fail for o3 models)
lm = dspy.LM("openai/o3", max_tokens=5000, temperature=1.0)

# New way (works for all reasoning models)
lm = dspy.LM("openai/o3-mini", max_tokens=5000, temperature=1.0)
# Internally uses max_completion_tokens for all o1/o3 family models

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -1,6 +1,7 @@
 import functools
 import logging
 import os
+import re
 import threading
 import uuid
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


kalanyuz:
@okhat Please have a look at this one

PR #7674, opened 2025-02-03
Make each retrieved doc contain full information

chenmoneygithub:
Previously every field of the DatabricksRM output is a list, so users need to pack them together. To ease the workflow, we are packing them inside DatabricksRM.


chenmoneygithub:
Not going to merge it since it breaks backward compatibility and general contract.

PR #7684, opened 2025-02-03
fix: resolve litellm dependencies

hung-phan:
This PR is to update litellm to make it works with uvicorn 0.29.

cc @CyrusNuevoDia since you did sth for the pyproject.yml previously

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ httpx
 jinja2
 joblib~=1.3
 json-repair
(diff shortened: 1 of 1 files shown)


CyrusNuevoDia:
hey @hung-phan can you also please update pyproject.yaml? This way poetry tests can work

hung-phan:
@CyrusNuevoDia can you elaborate a bit on what update you want? I see we already updated litellm there.

PR #7768, opened 2025-02-06
Change drop down title

B-Step62:
Change tracing drop down title from "optional" to "recommended"


chenmoneygithub (approved):
Thanks Yuki!

PR #7771, opened 2025-02-06
update: update litellm

hung-phan:
This PR is to update litellm dev dependencies

cc @CyrusNuevoDia


hung-phan:
updated @CyrusNuevoDia

PR #7794, opened 2025-02-09
feature/support-lm-reasoning-field

xucailiang:
‘ reasoning_field’ is a custom field, so it can be adapted to any LLM with a reasoning process, or it can be completely ignored if needed.

The test script is as follows:

```
import dspy

lm = dspy.LM('deepseek/deepseek-reasoner', api_key='sk-xxx', reasoning_field='reasoning_content')
dspy.configure(lm=lm)

if __name__ == '__main__':
    react = dspy.ChainOfThought("question -> answer")
    pred = react(question="Who is Apple's CEO?")
    
    lm.inspect_history()
    print("answer", pred.answer)
```

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -70,8 +70,7 @@ def _inspect_history(history, n: int = 1):
 
         if reasoning_content_list:
             print(_red("Reasoning From LLM:"))
-            for c in reasoning_content_list:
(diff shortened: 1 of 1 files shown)


isaacbmiller:
thanks for the contribution - i haven't seen this field before.

which provider(s) are including this instead of the <think> tags

PR #7818, opened 2025-02-18
Remove pandas from required deps list

chenmoneygithub:
Installing dspy==2.6.3 gives users a crazily long list of deps, while most of them are light, `pandas` is a big one. This PR removes pandas from the necessary list, which should be all right because there are only two usages of pandas:

1. In datasets, where it is just for type hints. 
2. In `dspy.Evaluate`, for the purpose of displaying the eval table. This is actually useful, but we can prompt users to install pandas if they want to use `dspy.Evaluate`. The downside is people get prompted once, but considering many users (especially users productionize with DSPy) don't need `dspy.Evaluate`, this may be worth it.

### DIFF
--- a/dspy/datasets/dataloader.py
+++ b/dspy/datasets/dataloader.py
@@ -1,8 +1,10 @@
 import random
 from collections.abc import Mapping
-from typing import List, Tuple, Union
+from typing import TYPE_CHECKING, List, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


chenmoneygithub:
We will hold this PR until DSPy 3.0, right now we don't want to break people's tutorial/script because of removing deps.

Issue #7835, opened 2025-02-21
RuntimeError: uvloop does not support Windows at the moment


chakravarthik27:
### What happened?

When installing dspy on Windows using `pip install dspy`, a runtime error occurs, preventing the package from being installed.

```shell
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [15 lines of output]
      Traceback (most recent call last):
        File "C:\Users\USER\AppData\Local\Programs\Python\Python39\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 389, in <module>
          main()
        File "C:\Users\USER\AppData\Local\Programs\Python\Python39\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
        File "C:\Users\USER\AppData\Local\Programs\Python\Python39\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 143, in get_requires_for_build_wheel
          return hook(config_settings)
        File "C:\Users\USER\AppData\Local\Temp\pip-build-env-3efo0lc5\overlay\Lib\site-packages\setuptools\build_meta.py", line 334, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
        File "C:\Users\USER\AppData\Local\Temp\pip-build-env-3efo0lc5\overlay\Lib\site-packages\setuptools\build_meta.py", line 304, in _get_build_requires
          self.run_setup()
        File "C:\Users\USER\AppData\Local\Temp\pip-build-env-3efo0lc5\overlay\Lib\site-packages\setuptools\build_meta.py", line 320, in run_setup
          exec(code, locals())
        File "<string>", line 8, in <module>
      RuntimeError: uvloop does not support Windows at the moment
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

### Steps to reproduce

- open the cmd prompt
- execute the `pip install dspy`

### DSPy version

latest

PR #7838, opened 2025-02-21
 Fix RuntimeError: uvloop Incompatibility on Windows

chakravarthik27:
#7835 uvloop Incompatibility on Windows due to litellm[proxy]

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -5,18 +5,17 @@ import re
 import textwrap
 from collections.abc import Mapping
 from itertools import chain
-
 from typing import Any, Dict, Literal, NamedTuple
 
 import pydantic
 from pydantic.fields import FieldInfo
 
 from dspy.adapters.base import Adapter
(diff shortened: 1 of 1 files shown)


okhat:
Hey @chakravarthik27 ! Thanks so much for this.

Two thoughts to start. First, the tests are failing. Could you check? Second, can you revert the style changes to the first two files. (The change to dspy/.internal_dspyai/setup.py will break some scripts!)

poetry.lock
```diff
@@ -2,102 +2,107 @@
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.4.4"
+version = "2.4.6"
```
chenmoneygithub, line 5: This poetry.lock has many unrelated changes, which is likely caused by the mismatch between your dev environment and the required version in the original poetry.lock, could you run `poetry install` before `poetry lock`? thank you!

chakravarthik27:
Hi @okhat , I checked in my local. It was fine.

PR #7841, opened 2025-02-22
docs: Add support for class methods in API reference

AkeemMcLennon:
## Context
This pull request adds support for class methods in the api documentation by replacing calls to `inspect.isfunction()` with the more inclusive `inspect.isroutine()`. In particular, the `Image`, `Prediction` and `Signature` primitives have several relevant functions that were not being captured by the previous method. 

It also excludes functions defined by external modules (e.g. Pydantic), but it seems there is some other mechanism that also filters them out somewhere, since they do not appear in the existing documentation despite being present in the generated markdown. 

## Before
image

## After

image

### DIFF
--- a/docs/docs/api/primitives/Image.md
+++ b/docs/docs/api/primitives/Image.md
@@ -4,14 +4,11 @@
     handler: python
     options:
         members:
-            - copy
-            - dict
-            - json
-            - model_copy
-            - model_dump
-            - model_dump_json
-            - model_post_init
+            - from_PIL
+            - from_file
+            - from_url
             - serialize_model
+            - validate_input
         show_source: true
         show_undocumented_members: true
         show_root_heading: true
--- a/docs/scripts/generate_api_docs.py
+++ b/docs/scripts/generate_api_docs.py
@@ -76,12 +76,10 @@ API_MAPPING = {
     ],
 }
 
-def is_documented_method(obj):
-    is_routine = inspect.isroutine(obj)
-    if not is_routine:
-        return False
+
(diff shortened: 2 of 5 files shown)


docs/docs/api/primitives/Image.md
```diff
@@ -4,14 +4,11 @@
     handler: python
     options:
         members:
-            - copy
-            - dict
```
chenmoneygithub, line 8: is the removal expected?

docs/scripts/generate_api_docs.py
```diff
@@ -76,6 +76,19 @@
     ],
 }
 
+def is_documented_method(obj):
+    is_routine = inspect.isroutine(obj)
+    if not is_routine:
+        return False
+    name = obj.__name__
+    # Exclude methods defined in external libraries
+    module = getattr(obj, "__module__", "")
```
chenmoneygithub, line 85: I don't think this is required, during traversal we only look at things defined inside dspy/

docs/scripts/generate_api_docs.py
```diff
@@ -76,6 +76,19 @@
     ],
 }
 
+def is_documented_method(obj):
```
chenmoneygithub, line 79: `is_documented_method` sounds like we are checking if the method has a docstring, probably go with `should_document_method`?

chenmoneygithub (changes requested):
Thanks for the PR, this is pretty good!

chenmoneygithub (approved):

Issue #7853, opened 2025-02-26
[Bug] runner.js is not bundled into the tarball on PyPI as of 2.6.6


tevino:
### What happened?

The "agents" example from the official website failed due to a failure of `evaluate_math`.

```python3
def evaluate_math(expression: str):
    return dspy.PythonInterpreter({}).execute(expression)
```

```
Failed to execute: No output from Deno subprocess. Stderr: error: Module not found "/dspy/.venv/lib/python3.12/site-packages/dspy/primitives/runner.js".
```

### Observation

Related implementation from dspy:

```python3
class PythonInterpreter:
    r"""
    PythonInterpreter that runs code in a sandboxed environment using Deno and Pyodide.

    Prerequisites:
    - Deno (https://docs.deno.com/runtime/getting_started/installation/).

    Example Usage:
    ```python
    code_string = "print('Hello'); 1 + 2"
    interp = PythonInterpreter()
    output = interp(code_string)
    print(output)  # If final statement is non-None, prints the numeric result, else prints captured output
    interp.shutdown()
    ```
    """

    def __init__(
        self,
        deno_command: Optional[List[str]] = None
    ) -> None:
        if isinstance(deno_command, dict):
            deno_command = None  # no-op, just a guard in case someone passes a dict
        self.deno_command = deno_command or [
            "deno", "run", "--allow-read", self._get_runner_path()
        ]
        self.deno_process = None

    def _get_runner_path(self) -> str:
        current_dir = os.path.dirname(os.path.abspath(__file__))
        return os.path.join(current_dir, "runner.js")  # <--- looking for runner.js
```

This file should be bundled into the release of dspy, but it's not: `/dspy/primitives/runner.js`.

As of 2.6.6 [the release from PyPI](https://pypi.org/project/dspy/2.6.6/#dspy-2.6.6.tar.gz) is still broken, there's no `primitives/runner.js` in the tarball.

SHA256 `8fb08bb8009c51d3f972856c19620cab2cc5e2b5a4c379005441bd81267ce905`

The same issue was reported and fixed here: https://github.com/stanfordnlp/dspy/issues/7490


### Steps to reproduce

```python3
import dspy

dspy.PythonInterpreter({}).execute("1+1")
```

**Output**:
```
Traceback (most recent call last):
  File "/dspy/main.py", line 3, in <module>
    dspy.PythonInterpreter({}).execute("1+1")
[...]

Issue #7490, opened 2025-01-30
[Bug] runner.js not bundled in fresh install on latest RC tags


seanbyrne88:
### What happened?

Working off a fresh install on the latest RC, runner.js is not bundled and errors out when running Program of Thought. 

### Steps to reproduce

Fresh install and run any program of thought application. 

### DSPy version

2.6.0rc8

okhat:
Thanks @seanbyrne88 ! Tagging @arnavsinghvi11

arnavsinghvi11:
Hi @seanbyrne88 , thanks for sharing!

Could you share the stack trace you're seeing and the minimal example using PoT you're running? (also, was this working for you before on any past versions?)
I just tried a fresh install with dspy 2.6.0rc8 on a new env and it worked from my end, but I'm wondering if maybe deno is not configured correctly for you so happy to investigate further.

seanbyrne88:
Thanks will put something together tonight. 

I pip installed from the release sha and it looked like PoT expected runner to be in the primitives directory. Confirmed that deno works and I dropped runner.js in there manually and it did work after that. Will put together an example though.

Installed using pip and pyenv.

seanbyrne88:
Used example from [here](https://dspy.ai/deep-dive/modules/program-of-thought/) on 2.6.0

```
#Example Usage
#Define a simple signature for basic question answering
generate_answer_signature = dspy.Signature("question -> answer")
generate_answer_signature.attach(question=("Question:", ""))

# Pass signature to ProgramOfThought Module
pot = dspy.ProgramOfThought(generate_answer_signature)


question = 'Sarah has 5 apples. She buys 7 more apples from the store. How many apples does Sarah have now?'
result = pot(question=question)

print(f"Question: {question}")
print(f"Final Predicted Answer (after ProgramOfThought process): {result.answer}")
```

When I execute this I get:
```
Error in code execution
```

When I step through using a debugger I see in `program_of_thought.py`
The error variable on looping has the value:
`No output from Deno subprocess. Stderr: Module not found: file:///envs/dspy-repro/lib/python3.12/site-packages/dspy/primitives/runner.js`

seanbyrne88:
When I used poetry like in the CI process it adds runner.js. 

Are you able to repro with a pip install from pypi now that the latest release is there?

okhat:
Thanks @seanbyrne88 ! We're looking into this

arnavsinghvi11:
[...]

PR #7855, opened 2025-02-26
bundle files with dspy installation

arnavsinghvi11:
fix for #7853  and #7490 

ensures all github files are in the dspy installation package so `runner.js` can be recognized. 

Tagging @chenmoneygithub for review on the proper pypi package installation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,9 @@ dependencies = [
 [tool.setuptools.packages.find]
 include = ["dspy*"]
 
+[tool.setuptools]
+include-package-data = true
+
 [project.optional-dependencies]
 anthropic = ["anthropic>=0.18.0,<1.0.0"]
 chromadb = ["chromadb~=0.4.14"]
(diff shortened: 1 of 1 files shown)


chenmoneygithub (changes requested):
Thanks for the PR! We should probably go with the following style to give us control over extra files to include instead of including everything:

```
[tool.setuptools.packages.find]
include = ["dspy*"]

[tool.setuptools]
include-package-data = true
```

chenmoneygithub (approved):

chenmoneygithub:
I pushed a commit to explicitly list out the non-python files, testing result:

```
(always-clean) [package_bundling][~/Documents/mlflow_team/dspy/dist]$ ls dspy-2.6.6/dspy/primitives
__init__.py           module.py             python_interpreter.py
assertions.py         prediction.py         runner.js
example.py            program.py            tool.py
```

I can see the `runner.js` in the extracted directory.

PR #7861, opened 2025-02-27
Fix save predict model with LM

### LINKED ISSUE #7852: [Bug] save() is broken when predict has lm instance
### What happened?

save() is broken when predict has lm assigned. See code below for reproducing.

To fix it, we can serialize the model name, then at loading time, reconstruct the lm instance by calling `dspy.LM(saved_model_name)`

### Steps to reproduce

```
>>> import dspy
>>> predict = dspy.Predict("q->a")
>>> predict.lm = dspy.LM("openai/gpt-4o-mini")
>>> predict.save("test_lm_saving.json", save_program=False)
```

Raises error:

```
>>> predict.save("test_lm_saving.json", save_program=False)
  File "/Users/chen.qian/Documents/mlflow_team/dspy/dspy/primitives/module.py", line 218, in save
    f.write(ujson.dumps(state, indent=2))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: <dspy.clients.lm.LM object at 0x10563c2c0> is not JSON serializable

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/chen.qian/Documents/mlflow_team/dspy/dspy/primitives/module.py", line 220, in save
    raise RuntimeError(
RuntimeError: Failed to save state to test_lm_saving.json with error: <dspy.clients.lm.LM object at 0x10563c2c0> is not JSON serializable. Your DSPy program may contain non json-serializable objects, please consider saving the state in .pkl by using `path` ending with `.pkl`, or saving the whole program by setting `save_program=True`.
```

### DSPy version

2.6.6

TomeHirata:
This PR enables the serialization of dspy.LM to fix the issue of model saving.
Resolve #7852

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -246,6 +246,10 @@ class LM(BaseLM):
                 new_instance.kwargs[key] = value
 
         return new_instance
+    
+    def dump_state(self):
+        # TODO: callbacks cannot be saved. We should consider how to save callbacks.
+        return { k: v for k, v in self.__dict__.items() if k not in ["provider", "callbacks", "kwargs"] }
 
 
 def request_cache(maxsize: Optional[int] = None):
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -248,8 +248,8 @@ class LM(BaseLM):
         return new_instance
     
     def dump_state(self):
-        # TODO: callbacks cannot be saved. We should consider how to save callbacks.
-        return { k: v for k, v in self.__dict__.items() if k not in ["provider", "callbacks", "kwargs"] }
+        state_keys = ["model", "model_type", "cache", "cache_in_memory", "num_retries", "finetuning_model", "launch_kwargs", "train_kwargs"]
+        return { key: getattr(self, key) for key in state_keys } | self.kwargs
 
 
 def request_cache(maxsize: Optional[int] = None):
--- a/tests/predict/test_predict.py
+++ b/tests/predict/test_predict.py
@@ -34,11 +34,18 @@ def test_reset_method():
 
 def test_lm_after_dump_and_load_state():
     predict_instance = Predict("input -> output")
-    predict_instance.lm = "lm_state"
+    lm = dspy.LM(
+        model="openai/gpt-4o-mini", 
+        model_type="chat",
+        temperature=1,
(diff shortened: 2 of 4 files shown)


dspy/clients/lm.py
```diff
@@ -246,6 +246,10 @@ def copy(self, **kwargs):
                 new_instance.kwargs[key] = value
 
         return new_instance
+    
+    def dump_state(self):
+        # TODO: callbacks cannot be saved. We should consider how to save callbacks.
```
chenmoneygithub, line 251: good catch! actually callbacks doesn't belong to the state, it's part of runtime config, we don't need to save it with this `dump_state`

dspy/clients/lm.py
```diff
@@ -246,6 +246,10 @@ def copy(self, **kwargs):
...
+        return { k: v for k, v in self.__dict__.items() if k not in ["provider", "callbacks", "kwargs"] }
```
chenmoneygithub, line 252: let's explicitly list out the attributes to save for clarity.

tests/predict/test_predict.py
```diff
@@ -34,11 +34,18 @@ def test_reset_method():
...
+        temperature=1,
+        max_tokens=100,
+        num_retries=10,
+    )
+    predict_instance.lm = lm
     dumped_state = predict_instance.dump_state()
     new_instance = Predict("input -> output")
     new_instance.load_state(dumped_state)
-    assert new_instance.lm == "lm_state"
+    assert new_instance.lm.dump_state() == lm.dump_state()
```
chenmoneygithub, line 48: minor comment - here we are assuming `dump_state()` is working correctly, let's add a few `assert` check to verify the dumped state contains the desired fields with the correct value.

chenmoneygithub (commented):
Great work! 

Left some minor comments, we can merge after addressing those.

tests/predict/test_predict.py
TomeHirata: Good call. Added the assertion here and also added a unit test for the LM side.

TomeHirata:
@chenmoneygithub Thank you for your review. Can you take another look?

chenmoneygithub (approved):
LGTM!

Issue #7867, opened 2025-02-27
[Bug] ModuleNotFoundError: No module named 'dspy.predict'


bnsh:
### What happened?

When you `import dspy` with `dspy-ai==2.6.7` it just fails immediately with `ModuleNotFoundError: No module named 'dspy.predict'`

### Steps to reproduce

Here's my [gist](https://gist.github.com/bnsh/36890f62b5e47f69c2131f91132d0c16) [https://gist.github.com/bnsh/36890f62b5e47f69c2131f91132d0c16](https://gist.github.com/bnsh/36890f62b5e47f69c2131f91132d0c16)... You run it like so:

`make build-2.6.7 # Will build (and run test.py with dspy-ai==2.6.7)` and
`make build-2.6.6 # Will build (and run test.py with dspy-ai==2.6.6)`

What you get with 2.6.7 is:

```
Step 6/6 : RUN python3 /tmp/test.py
 ---> Running in 686554c33a98
Traceback (most recent call last):
  File "/tmp/test.py", line 4, in <module>
    import dspy
  File "/usr/local/lib/python3.12/site-packages/dspy/__init__.py", line 1, in <module>
    from dspy.predict import *
ModuleNotFoundError: No module named 'dspy.predict'
The command '/bin/sh -c python3 /tmp/test.py' returned a non-zero code: 1
make: *** [Makefile:4: build-2.6.7] Error 1
```

Notice the error isn't from the test.py code, it's from the dspy __init__ code.

Whereas with 2.6.6:

```
Step 6/6 : RUN python3 /tmp/test.py
 ---> Running in cac353cc3506
2.6.5
```

### DSPy version

2.6.7

TomeHirata:
Thank you for the report. We yanked 2.6.7 and will fix the package configuration. For the time being, please use 2.6.6.

PR #7869, opened 2025-02-27
Fix pyproject.toml to include package data

### LINKED ISSUE #7853: [Bug] runner.js is not bundled into the tarball on PyPI as of 2.6.6

TomeHirata:
This PR fixes #7853 and #7490 without causing #7867.
Confirmed the behavior by running `python -m build` on a clean clone of the DSPy repo.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "wheel"]
+requires = ["setuptools==62.3.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -51,10 +51,12 @@ dependencies = [
 ]
 
 [tool.setuptools.packages.find]
-include = ["dspy*"]
+where = ["."]
+include = ["dspy", "dspy.*"]
+exclude = ["tests", "tests.*"]
 
-[tool.setuptools]
-include-package-data = true
+[tool.setuptools.package-data]
+dspy = ["primitives/*.js"]
 
 [project.optional-dependencies]
 anthropic = ["anthropic>=0.18.0,<1.0.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools==62.3.0", "wheel"]
+requires = ["setuptools>=40.8.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]


pyproject.toml
```diff
@@ -51,10 +51,12 @@ dependencies = [
 ]
 
 [tool.setuptools.packages.find]
-include = ["dspy*"]
+where = ["."]
+include = ["dspy", "dspy.*"]
+exclude = ["tests", "tests.*"]
```
chenmoneygithub, line 56: curious - since we do explicit `include`, do we still need this `exclude` part? 

Non-blocking, feel free to merge!

chenmoneygithub (approved):
thanks for the PR! LGTM

pyproject.toml
TomeHirata: That's not necessary (I confirmed "tests" are not included when removing this line), but mostly for the safeguard. We do the same configuration in mlflow: https://github.com/mlflow/mlflow/blob/master/pyproject.release.toml#L139

PR #7872, opened 2025-02-27
feat: Added support for non-image mimes in image_url content

glesperance:
Added support for non-image mimes in image_url content.

As offputting as this sounds this is how some models -- eg Gemini[1] & Anthropic[2] -- work for giving non-image files to the LM.

[1] https://docs.litellm.ai/docs/providers/gemini#usage---pdf--videos--etc-files
[2] https://docs.litellm.ai/docs/providers/anthropic#usage---pdf


okhat:
@isaacbmiller thoughts on merging?

PR #7876, opened 2025-02-28
Add testing before publishing to pypi

chenmoneygithub:
Test the `import dspy` in the process of publishing to pypitest to verify the wheel building is successful.

In the future we may want to execute some unit/e2e testing before publishing to pypi.

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -56,7 +56,7 @@ jobs:
           source test_env/bin/activate
           # Install the locally built wheel and testing dependencies
           pip install dist/*.whl pytest
-          python -m "import dspy; print(dspy.__version__)"
+          python -c "import dspy; print(dspy.__version__)"
           deactivate
       # Publish to test-PyPI
       - name: Publish distribution 📦 to test-PyPI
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_release.yml
```diff
@@ -41,11 +41,24 @@ jobs:
...
         run: python3 -m build
+      # Test the locally built wheel
+      - name: Create test environment
+        run: python -m venv test_env
+      - name: Test package installation and functionality
+        run: |
+          source test_env/bin/activate
+          # Install the locally built wheel and testing dependencies
+          pip install dist/*.whl pytest
+          python -c "import dspy; print(dspy.__version__)"
```
TomeHirata, line 59: Did this line actually fail when we removed `pip install dist/*.whl pytest` above? I wonder if this still works without the installation due to the flat structure of this repo.

chenmoneygithub: Great question!

The answer is it will fail. The reason is activating virtual env changes the sys path, which doesn't include the working dir . any more. Without virtual env, or using the conda env includes the working dir in sys path, so even if we don't do installation the import still works.

TomeHirata: Thank you for the answer! Can we add comments for the context of the venv step above?

TomeHirata (approved):
LGTM

PR #7878, opened 2025-02-28
Refactor ProgramOfThought and add unit tests

TomeHirata:
This PR 
- Refactored the logic of ProgramOfThought not to run program when the code is invalid
- Add unit tests for the failure cases of ProgramOfThought and PythonInterpreter
- Add guidance for deno installation
- Fixed error handling in PythonInterpreter


tests/predict/test_program_of_thought.py
```diff
@@ -21,7 +24,7 @@ def test_pot_code_generation():
     assert res.answer == "2"
 
 
-def test_pot_code_generation_with_error():
+def test_pot_code_generation_with_one_error():
```
chenmoneygithub, line 27: Since we are here, can we add code to skip the `test_pot_code_generation` when `deno` is not installed? It's quite distractive that people clone DSPy => run pytest test/ => find errors without clearly telling them to install missing packges.

Something like the following should work:

```

def is_deno_available():
    return shutil.which("deno") is not None

@pytest.mark.skipif(not is_deno_available(), reason="Deno is not installed or not in PATH")
def test_pot_code_generation():

**Review by chenmoneygithub — COMMENTED** [review]
Nice work! Left some minor comments

File: dspy/predict/program_of_thought.py [hunk 2]
```diff
@@ -6,11 +7,22 @@
 from dspy.primitives.program import Module
 from dspy.primitives.python_interpreter import PythonInterpreter
 
+_logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
TomeHirata: Agreed, let me fix here

dspy/predict/program_of_thought.py
```diff
@@ -148,35 +165,40 @@ def parse_code(self, code_data):
...
             field_name: kwargs[field_name] for field_name in self.input_fields
         }
         code_data = self.code_generate(**input_kwargs)
-        parsed_code, error = self.parse_code(code_data)
-        # FIXME: Don't try to execute the code if it didn't parse
-        code, output, error = self.execute_code(parsed_code)
-        hop = 0
-        while hop < self.max_iters and error:
-            print("Error in code execution")
+        code = output = error = None
```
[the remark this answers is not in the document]
TomeHirata: Got it, this should be ok since we init vars with None. But let me define them separately.

tests/predict/test_program_of_thought.py
TomeHirata: Agree that's better. Actually the main ProgramOfThought flow also have the problem. I at least improved the experience by including Deno error message in the RuntimeException so that people can see the error message. But it'd be better if we could make it more explicit.

TomeHirata: Added the skip marker and more instructions.

docs/docs/api/models/LM.md
```diff
@@ -6,6 +6,7 @@
         members:
             - __call__
             - copy
+            - dump_state
```
TomeHirata, line 9: note: this is just a left over from a previous PR

dspy/predict/program_of_thought.py
```diff
@@ -148,35 +175,42 @@ def parse_code(self, code_data):
...
+                output, error = self._execute_code(code)
             hop += 1
-            if hop == self.max_iters:
-                print("Max hops reached. Error persists.")
-                return None
         input_kwargs.update({"final_generated_code": code, "code_output": output})
         answer_gen_result = self.generate_answer(**input_kwargs)
-        return answer_gen_result
\ No newline at end of file
+        self.interpreter.shutdown()
```
TomeHirata, line 215: Added shutdown logic for the deno subprocess in order not to have many remaining subprocesses.

tests/primitives/test_python_interpreter.py
```diff
@@ -1,4 +1,10 @@
-from dspy.primitives.python_interpreter import PythonInterpreter
+import shutil
+import pytest
+from dspy.primitives.python_interpreter import PythonInterpreter, InterpreterError
+
+# This test suite requires deno to be installed. Please install deno following https://docs.deno.com/runtime/getting_started/installation/
+if shutil.which("deno") is None:
+    pytestmark = pytest.mark.skip(reason="Deno is not installed or not in PATH")
```
chenmoneygithub, line 7: Let's use this instead of the magic global var `pytestmark` for clarity:

```
if shutil.which("deno") is None:
    pytest.skip("Deno is not installed or not in PATH", allow_module_level=True)
```

chenmoneygithub (approved):
Solid work! Approved with a minor comment

PR #7892, opened 2025-03-03
Add ChainOfDraft module

SerjSmor:
https://arxiv.org/pdf/2502.18600

### DIFF
--- a/dspy/predict/__init__.py
+++ b/dspy/predict/__init__.py
@@ -1,6 +1,7 @@
 from dspy.predict.aggregation import majority
 from dspy.predict.best_of_n import BestOfN
 from dspy.predict.chain_of_thought import ChainOfThought
+from dspy.predict.chain_of_draft import ChainOfDraft
 from dspy.predict.chain_of_thought_with_hint import ChainOfThoughtWithHint
 from dspy.predict.knn import KNN
 from dspy.predict.multi_chain_comparison import MultiChainComparison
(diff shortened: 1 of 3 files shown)


nullhack:
Hi, @chenmoneygithub, I've read the paper and I was excited when I saw this PR so fast after the release of the paper. 

I've done some tests and the idea of keeping a shorter reasoning really helps.

That said, I agree that it is quite similar to ChainOfThought. May I suggest something else instead?

What if we could change the 'reasoning' definition of the ChainOfThought on the fly during instantiation using keywords?

e.g.

```python

prefix = "draft: Think step by step, but only keep a minimum draft for each thinking step, with 5 words at most"

CoT = dspy.ChainOfThought("question -> answer", reasoning_desc="draft", reasoning_prefix=prefix)
```

This would change the standard CoT beehavior and accommodate the results of the paper without introducing unnecessary abstractions

chenmoneygithub:
@nullhack Yea I can see changing prefix being useful, you can actually achieve the same goal by writing a `rationale_type`, see the code [here](https://github.com/stanfordnlp/dspy/blob/583ba3dc151d1e4135a04462f4057be5f84f4799/dspy/predict/chain_of_thought.py#L14C9-L14C86): 

```
rationale_type = rationale_type or dspy.OutputField(prefix=prefix, desc=desc)
```

The variable name `rationale_type` is a bit confusing and vague though, we may want to improve it.

nullhack:
Yes, the rationale_type will work. But I agree that it is confusing the way it is

SerjSmor:
Thank you for your comments @chenmoneygithub @nullhack @arnavsinghvi11, after using the library for some time, I'm learning the internal bit by bit, so your comments are helpful.

Would it be worthwhile to implement what @nullhack suggested? adding both `reasoning_desc` and `reasoning_prefix` to the method parameters? It seems like an elegant solution.

PR #7894, opened 2025-03-03
Fix: avoid PythonInterpreter(deno) leak

innomentats:
PythonInterpreter(deno) allocated by `ProgramOfThought.execute_code()` never gets closed, leading to memory leak. Simply shut it down.


chenmoneygithub:
@innomentats Thanks for the PR, nice catch! Coincidentally seems it's covered in another PR: #7878

PR #7896, opened 2025-03-04
Enhance type hints of adapters

TomeHirata:
This PR updates type hints as well as some fixes

- Add type hints to public methods of Adapter
- Fix inversed dependency of BaseAdapter; BaseAdapter should not know anything about JsonAdapter
- Removed unused utils

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -1,10 +1,11 @@
 from abc import ABC, abstractmethod
-from typing import Type, Union, Any, Optional
+from typing import Type, Any, Optional, TYPE_CHECKING
 
 from dspy.adapters.types import History
 from dspy.utils.callback import BaseCallback, with_callbacks
-from dspy.clients.lm import LM
 from dspy.signatures.signature import Signature
+if TYPE_CHECKING:
+    from dspy.clients.lm import LM
 
 class Adapter(ABC):
     def __init__(self, callbacks: Optional[list[BaseCallback]] = None):
@@ -17,7 +18,7 @@ class Adapter(ABC):
         cls.format = with_callbacks(cls.format)
         cls.parse = with_callbacks(cls.parse)
 
-    def __call__(self, lm: LM, lm_kwargs: dict[str, Any], signature: Type[Signature], demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
+    def __call__(self, lm: "LM", lm_kwargs: dict[str, Any], signature: Type[Signature], demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
         inputs_ = self.format(signature, demos, inputs)
         inputs_ = dict(prompt=inputs_) if isinstance(inputs_, str) else dict(messages=inputs_)
 
@@ -47,15 +48,15 @@ class Adapter(ABC):
 
 
     @abstractmethod
-    def format(self, signature: Union[str, Type[Signature]], demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
+    def format(self, signature: Type[Signature], demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
         raise NotImplementedError
 
     @abstractmethod
-    def parse(self, signature: Union[str, Type[Signature]], completion: str) -> dict[str, Any]:
+    def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
         raise NotImplementedError
 
     @abstractmethod
-    def format_finetune_data(self, signature: Union[str, Type[Signature]], demos: list[dict[str, Any]], inputs: dict[str, Any], outputs: dict[str, Any]) -> dict[str, list[Any]]:
+    def format_finetune_data(self, signature: Type[Signature], demos: list[dict[str, Any]], inputs: dict[str, Any], outputs: dict[str, Any]) -> dict[str, list[Any]]:
         raise NotImplementedError
     
     @abstractmethod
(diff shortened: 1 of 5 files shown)


dspy/adapters/base.py
```diff
@@ -17,60 +17,56 @@ def __init_subclass__(cls, **kwargs) -> None:
...
+    def format(self, signature: Union[str, Type[Signature]], demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
         raise NotImplementedError
 
     @abstractmethod
-    def parse(self, signature, completion):
+    def parse(self, signature: Union[str, Type[Signature]], completion: str) -> dict[str, Any]:
         raise NotImplementedError
 
-    def format_finetune_data(self, signature, demos, inputs, outputs):
+    @abstractmethod
```
chenmoneygithub, line 58: let's remove the `@abstractmethod` annotation for this method, we don't enforce all child classes to implement this method

dspy/adapters/base.py
```diff
@@ -17,60 +18,56 @@ def __init_subclass__(cls, **kwargs) -> None:
...
+    def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
         raise NotImplementedError
 
-    def format_finetune_data(self, signature, demos, inputs, outputs):
+    @abstractmethod
+    def format_finetune_data(self, signature: Type[Signature], demos: list[dict[str, Any]], inputs: dict[str, Any], outputs: dict[str, Any]) -> dict[str, list[Any]]:
+        raise NotImplementedError
+    
+    @abstractmethod
+    def format_fields(self, signature: Type[Signature], values: dict[str, Any], role: str) -> str:
```
chenmoneygithub, line 63: we can remove this method, chances are that users don't need this abstraction, and just write everything in `format`.

dspy/adapters/base.py
```diff
@@ -17,60 +18,56 @@ def __init_subclass__(cls, **kwargs) -> None:
...
         raise NotImplementedError
 
-    def format_turn(self, signature, values, role, incomplete=False, is_conversation_history=False):
-        pass
+    @abstractmethod
```
chenmoneygithub, line 66: ditto: remove this annotation

dspy/adapters/chat_adapter.py
```diff
@@ -31,7 +34,23 @@ class FieldInfoWithName(NamedTuple):
 
 
 class ChatAdapter(Adapter):
-    def format(self, signature: Signature, demos: list[dict[str, Any]], inputs: dict[str, Any]) -> list[dict[str, Any]]:
+    def __init__(self, callbacks: Optional[list[BaseCallback]] = None):
+        super().__init__(callbacks)
+        self.json_adapter = None
```
chenmoneygithub, line 39: Is this `None` assignment useful? can we create a JsonAdapter() upon failure detection?

dspy/adapters/chat_adapter.py
```diff
@@ -115,7 +134,7 @@ def format_fields(self, signature, values, role):
         }
         return format_fields(fields_with_values)
 
-    def format_turn(self, signature, values, role, incomplete=False, is_conversation_history=False):
+    def format_turn(self, signature: Type[Signature], values: dict[str, Any], role: str, incomplete=False, is_conversation_history=False) -> dict[str, Any]:
```
chenmoneygithub, line 137: nit: missing type hint for `incomplete` and `is_conversation_history `

chenmoneygithub (changes requested):
Thanks for the PR! Nice catch on bad reference to JSONAdapter from the BaseAdapter! 

A quick note - we don't enforce or forbid type hints in DSPy, which is arguably more useful than causing trouble.

dspy/adapters/chat_adapter.py
TomeHirata: This is avoiding fields of ChatAdapter from being changed during the method call. I think `if self.json_adapter is not None` is better than `if hasattr(self, "json_adapter")`

dspy/adapters/base.py
TomeHirata: format_finetune_data is used in BootstrapFinetune. Is it ok for other adapters not to implement this method?

dspy/adapters/base.py
TomeHirata: We cannot remove this method since this is used by ReAct

chenmoneygithub: gotcha, it's actually a bad practice to have a predefined module reference a method that has no guarantee to exist.  Let's keep `format_fields ` here but remove the `@abstractmethod` annotation.

dspy/adapters/base.py
chenmoneygithub: yes that's okay, many users won't use `BootstrapFinetune ` in their workflow. We allow users to pick what's useful for them when customizing their workflows, irrelevant parts can break.

TomeHirata:
Copy from our internal discussion for type hints:
I agree that we don't need type hints for all class/methods in general! I thought it'd be helpful for adapters since 1) it uses inheritance so having consistent signatures between parent and child classes is important 2) it partially used type hints."

dspy/adapters/base.py
TomeHirata: +1, we should not assume the existence of format_fields if it might not exist

dspy/adapters/chat_adapter.py
chenmoneygithub: ohh we don't need to have `self.json_adapter`, can we create a new instance every time?

TomeHirata: We could, but I think it's better to create the object only once to avoid unnecessary overhead.

chenmoneygithub (approved):
Feel free to merge after https://github.com/stanfordnlp/dspy/pull/7896#discussion_r1978503018 is addressed, nice fix!

PR #7898, opened 2025-03-04
Fix package name var in github actions

TomeHirata:


chenmoneygithub (approved):

.github/workflows/build_and_release.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       - name: Update package name in pyproject.toml
         run: sed -i '/#replace_package_name_marker/{n;s/name *= *"[^"]*"/name="dspy"/;}' pyproject.toml
       - name: Update package name in metadata.py
```
harupy, line 92: What is this step for?

TomeHirata: DSPy has two packages with the same content: "dspy" and "dspy-ai". This workflow replaces the name of the package dynamically.

PR #7900, opened 2025-03-04
Fix progress bar and tutorial of BootstrapFewShot

TomeHirata:
The current BootstrapFewShot progress bar has an issue that when there are many training examples compared with necessary demonstrations, the progress bar does not proceed.
This PR updated the progression bar implementation to show two progress bars: one is for the number of bootstrapped examples and another is for the number of training samples processed.
Also fixed the tutorial for BootstrapFewShot.

image

### DIFF
--- a/docs/docs/deep-dive/optimizers/bootstrap-fewshot.md
+++ b/docs/docs/deep-dive/optimizers/bootstrap-fewshot.md
@@ -49,12 +49,16 @@ class CoT(dspy.Module):
         return self.prog(question=question)
 ```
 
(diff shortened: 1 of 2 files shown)


TomeHirata:
I also had the same concern that Omar raised there. To solve both issues, it would be good if we could have two progress bars; one is for the training set and another is for bootstrapping.

okhat:
Thank you @TomeHirata ! I think this file is due for a near-complete refactor because we want to parallelize it intelligently and perhaps some other changes.

TomeHirata:
@okhat I see, will we change the way to display progress in the upcoming refactoring? If so we can do the improvement in the refactor! But if we continue to display progress in a similar way and just let the optimizer run bootstrapping in parallel, this improvement can still help I guess.

PR #7904, opened 2025-03-04
Run test before publishing to pypi

chenmoneygithub:
We've been messing up releases between 2.6.6 and 2.6.9, which could've been prevented with a decent test before publishing to pypi. This PR adds the following check before publishing to pypi:

- Check the metadata is valid after the release.
- All tests in tests/predict should pass, which has a decent coverage over primitives, predict, adapters and LM.

See screenshot below for an example run (testpypi only) from my fork:
image


TomeHirata (commented):
Greet validation to prevent the issues from happening again!!

PR #7912, opened 2025-03-05
enh: bump `lancedb` to 0.19.0 and unblock Python 3.13

stevapple:
This PR resolved the incompatibility of `lancedb` (originally fixing to `numpy` v1) and `fastembed` (newer versions requiring `numpy` v2) and thus unblocked Python 3.13 support.

We have long been using Python 3.13 with DSPy (before the 2.6.6 release which no longer allows that) so it should be already supported. `lancedb` is an optional dependency and will not affect users of DSPy only.


okhat:
Thanks so much @stevapple ! cc @chenmoneygithub  We should probably get rid of a lot of the "optional" dependencies that aren't doing anything in practice...

PR #7917, opened 2025-03-06
Don't suppress exception in eval_candidate_program

TomeHirata:
Currently, eval_candidate_program method suppresses any exception from evaluation and returns 0. The behavior is confusing since users cannot notice their evaluation metric is not correct and see the suboptimal result without realizing the optimizer actually fails to optimize a given metric. This PR stops suppressing exceptions and adds unit tests to this util.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -467,7 +467,7 @@ class MIPROv2(Teleprompter):
         adjusted_num_trials = (num_trials + num_trials // minibatch_full_eval_steps + 1) if minibatch else num_trials
         logger.info(f"== Trial {1} / {adjusted_num_trials} - Full Evaluation of Default Program ==")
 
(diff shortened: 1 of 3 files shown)


okhat:
Thanks @TomeHirata ! Hmm let's think about this. We do want to make sure MIPRO doesn't fail even if one of the generated programs is bad at the task. When you're optimizing, it's often the case that some of the attempts at producing optimized programs are bad. You don't want to crash the whole process, because the point of optimization is to find candidates that aren't bad...

TomeHirata:
@okhat Thank you for clarifying the context. When one of the generated program is bad at solving the problem, does it raise an exception? I thought it would just get a bad score and the exception here is caused by a bad metric specification. 

If receiving an exception here is an acceptable case, we need to fix the error handling a bit since it still failed [here](https://github.com/stanfordnlp/dspy/blob/main/dspy/teleprompt/mipro_optimizer_v2.py#L470) when I tried. The util always returns float while it should consider arguments.

PR #7918, opened 2025-03-06
Update tool.py to include **kwargs

ButterflyAtHeart:
Before the tool call of a function containing **kwargs had to be done like this: {"kwargs": {"foo":"bar"} now {"foo":bar"} is directly supported.

### DIFF
--- a/dspy/primitives/tool.py
+++ b/dspy/primitives/tool.py
@@ -42,7 +42,8 @@ class Tool:
         self.args = args
         self.arg_types = arg_types
         self.arg_desc = arg_desc
-
+        self.has_var_kwargs = False
+        
         self._parse_function(func, arg_desc)
 
     def _resolve_pydantic_schema(self, model: type[BaseModel]) -> dict:
@@ -86,6 +87,8 @@ class Tool:
 
         # Use inspect.signature to get all arg names
         sig = inspect.signature(annotations_func)
+        # set has_var_kwargs to True if the function has a **kwargs parameter
+        self.has_var_kwargs = any([sig.parameters[k].kind == 4 for k in sig.parameters.keys()])
         # Get available type hints
         available_hints = get_type_hints(annotations_func)
         # Build a dictionary of arg name -> type (defaulting to Any when missing)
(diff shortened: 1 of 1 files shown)


dspy/primitives/tool.py
```diff
@@ -116,7 +119,10 @@ def _parse_function(self, func: Callable, arg_desc: dict[str, str] = None):
     def __call__(self, **kwargs):
         for k, v in kwargs.items():
             if k not in self.args:
-                raise ValueError(f"Arg {k} is not in the tool's args.")
+                if self.has_var_kwargs:
+                    continue
```
okhat, line 123: Thanks so much @ButterflyAtHeart ! Is `continue` the right thing to do here?

I think the right thing to do might be more like "find the name of the kwargs argument" and then "wrap the non-matching arguments into a dict with that kwargs name".

Maybe we need to test this with a real function that takes `**kwargs` as inputs and describes those kwargs in the docstring.

ButterflyAtHeart: The issue with checking for a kwargs argument is that needs to happen before the loop, because `kwargs["kwargs"]` doesn't necessarily exist so changing the dictionary `kwargs` while iterating raises an error. This would add quite a lot of code. I think continue is a more elegant solution as it just says we do not need to do any further tests for this keyword argument.  Maybe adding a comment there before `continue` would be good. So if you agree we should stick with continue I would add this comment. 

Here is the test I have used to test this:
```
import dspy

def func(x:int, y=1, **kwargs:dict):
    """This is a test function.
    Args:
        x (int): The first argument.
        y (int): The second argument.
    """
    return "Completed."

f = dspy.Tool(func)
f(x=0, y = 0, z=2)
```
If you think there need to be any other cases included into the test. Please let me now.

TomeHirata:
Hi, @ButterflyAtHeart. What is the use case for defining a function with `kwargs` for tools calling agents? To my understanding, tool parameters need to be defined explicitly and passed to LLMs ([OAI spec](https://platform.openai.com/docs/guides/function-calling?api-mode=chat#defining-functions)). I'm not aware of the valid usecase of defining the tool function with arbitrary keyword arguments.

dspy/primitives/tool.py
TomeHirata: @ButterflyAtHeart Let's make sure the insufficient arguments are detected:

```python
import dspy

def func(x:int, **kwargs:dict):
    """This is a test function.
    Args:
        x (int): The first argument.
    """
    return x

f = dspy.Tool(func)
f(z=2)
```

TomeHirata:
Incorporate this change in https://github.com/stanfordnlp/dspy/pull/8102

PR #7922, opened 2025-03-07
dspy.Retry module and a conversation feedback adapter

gilad12-coder:
This PR introduces a new Retry module and ConversationalFeedbackAdapter to enhance DSPy's validation capabilities:

**1. Added Conversational Feedback Approach:**
- Implemented `ConversationalFeedbackAdapter` that extends ChatAdapter to inject feedback history 
- Created conversation-based refinement workflow with clear assistant-user message pairs
- Added system message enhancement for error analysis and improvement guidance
- Structured feedback in a natural conversation format that models can better utilize for refinement

**2. Implemented Retry Module with Validator Support:**
- Created a flexible `Retry` module that wraps any DSPy module (Predict or ChainOfThought)
- Added validator-based feedback collection system:
  - Support for multiple validator functions with consistent interface
  - Error message aggregation and formatting for more effective feedback
  - Best-prediction tracking to ensure graceful fallback to partial solutions
- Implemented conversation-simulated refinement loop to improve outputs through iterative feedback
- Added detailed metadata tracking for monitoring and debugging refinement progress

**3. Enhanced Configurability and Debugging:**
- Added support for verbose mode with detailed logging of validation results
- Implemented configurable maximum attempts with sensible defaults
- Added fail-on-invalid option for strict validation requirements
- Created comprehensive metadata capture of refinement history
- Implemented error handling with instructive feedback for common issues like missing fields

**4. Comprehensive Documentation and Type Safety:**
- Added detailed class and method docstrings explaining usage patterns
- Included practical example code demonstrating validator definition and module usage
- Implemented proper type hints throughout both modules
- Created clear, structured code comments explaining complex logic sections
- Added input validation and error checking for robust behavior

PR #7926, opened 2025-03-09
Adds Improved Error Handling, Documentation, and Tests For The Refine Module

cezarc1:
Of note: Adds configurable `fail_count` parameter to `Refine` to control how many errors are tolerated before throwing. By default we will throw once if all of the underlying module's calls fails.

Adds in-line documentation to both `Refine` and `BestOfN` and in the cheatsheet.


okhat:
This is amazing, thanks so much @cezarc1 ! Timing is interesting because we're considering how to act on this and #7922 together. We will merge both PRs, but the question is what's the final design that emerges from that.

cezarc1:
@okhat . I see! I missed #7922. As I think it was mentioned in #7922 I still see the benefits of`Refine` and `BestofN` still existing as my understanding is that the retry mechanism in that PR requires a new adapter, while `Refine` and `BestOfN` do not require a new adapter in order to work.

PR #7932, opened 2025-03-10
Update lm_local.py

zbambergerNLP:
Without ignoring the "hugggingface/" prefix during launch, the model is not properly recognized within the huggingface hub. This change fixes that error.


okhat:
Thank you so much @zbambergerNLP ! Merging this and tagging @Ziems to see how this interacts with his changes

PR #7941, opened 2025-03-12
Fix lint problems of workflows so that we can remove .prettierignore in the future

chenmoneygithub:
We can remove .prettierignore to keep our top level clean, but we need to fix the lint issue of github workflows first. This PR only fixes the lint, and the actual removal will happen in the next PR.


okhat:
@hmoazam thoughts?

hmoazam:
So the reason for that was that the release workflow throws an error about not being able to modify the workflows themselves. It turned out this was due to prettify formatting the workflows. So in general yes we can remove this, but it will mean any time we edit any of the workflows, we will have to ensure they follow prettify standards locally otherwise the workflows will break.

okhat:
thank you Hanna,,, ok @chenmoneygithub thoughts?

chenmoneygithub:
> release workflow throws an error about not being able to modify the workflows themselves

@hmoazam  Hmm I am not following this, I don't see the release workflow modifies the yml file, was it in an older version?

chenmoneygithub:
@hmoazam I am reverting the removal of `.prettierignore` in this PR, when you get back, we can check again why it caused error in earlier release.

PR #7944, opened 2025-03-12
Remove diskcache from dependencies

TomeHirata:
diskcache is not used in the main logic.

```
$ git grep diskcache
-> no result
```


TomeHirata:
litellm uses the library

PR #7964, opened 2025-03-15
Adds Improved Error Handling, Documentation, and Tests For The BestOfN Module 

cezarc1:
Adds configurable `fail_count` parameter to the `BestOfN` module to control how many errors are tolerated before throwing. By default we will now throw once if all of the underlying module's calls fails. This is the same logic as for `Refine` which was added in #7926.

Updates mkdocs documentation for both `BestofN` and `Refine` modules, as well as adding a basic tutorial section. Confirmed that the docs compile by running `$ mkdocs build` and then `$ mkdocs serve`.


okhat:
This is incredible. Thank you @cezarc1 !!

PR #7978, opened 2025-03-18
[WIP] patch training dataset type mismatch error

arnavsinghvi11:
When running finetuning [notebooks](https://dspy.ai/tutorials/classification_finetuning/), SFTTrainer runs into `error: can only concatenate list (not "Tensor") to list` 

This change ensures the `train_dataset` is structured to have list of dictionaries of the tokenized texts to enable training.


dilarasoylu:
Thank you so much @arnavsinghvi11! This might be caused by a `transformers` library version issue. Updated the notebook instructions here #8044.

PR #7989, opened 2025-03-20
Fix example in refine

chenmoneygithub:
The format doesn't work well with mkdocs:

Before fix:
image

After fix:
image

Also it has a logical bug on `return 1.0 if len(pred.answer) == 1 else 0.0`, which should be `return 1.0 if len(pred.answer.split()) == 1 else 0.0` to check for number of tokens instead of characters.


cezarc1:
@chenmoneygithub Thanks for catching that!

PR #7991, opened 2025-03-21
Refactor evaluate and introduce construct_result_table method

TomeHirata:
We want to make `result_df` of the evaluate call available outside of the `Evaluate.__call__` method. In this PR, we refactored the logic flow and introduced `construct_result_table` method that will be patched by tracing tools.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -1,6 +1,6 @@
 import logging
 import types
-from typing import TYPE_CHECKING, Any, Callable, List, Optional
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple
 
 if TYPE_CHECKING:
     import pandas as pd
@@ -85,7 +85,6 @@ class Evaluate:
         self.return_outputs = return_outputs
         self.provide_traceback = provide_traceback
         self.failure_score = failure_score
-        self.last_result_df = None
 
     @with_callbacks
     def __call__(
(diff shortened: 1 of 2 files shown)


dspy/evaluate/evaluate.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         metric: Optional[Callable] = None,
         num_threads: int = 1,
         display_progress: bool = False,
-        display_table: bool = False,
+        display_table: Tuple[bool, int] = False,
```
okhat, line 57: Hmm I don't think this should be a Tuple. Maybe a Union?

dspy/evaluate/evaluate.py
```diff
@@ -108,8 +109,8 @@ def __call__(
                 `self.num_threads`.
             display_progress (bool): Whether to display progress during evaluation. if not provided, use
                 `self.display_progress`.
-            display_table (bool): Whether to display the evaluation results in a table. if not provided, use
-                `self.display_table`.
+            display_table (Tuple[bool, int]): Whether to display the evaluation results in a table. if not provided, use
+                `self.display_table`. If a number is passed, the evaluation results will be truncated to that number.
```
okhat, line 113: Should just be the number (max rows)

okhat:
LGMT, thanks so much @TomeHirata . Left minor comment.

dspy/evaluate/evaluate.py
TomeHirata: If we'll drop the support for bool here, let's show a warning for one cycle and change the behavior?

TomeHirata:
Thanks @okhat. Can you take another look?

Issue #7793, opened 2025-02-09
[Feature] Add support for deepseek 'reasoning_content' parameter


xucailiang:
### Question 1:

When using the deepseek-reasoner model, we obtain a `reasoning_content` parameter that represents the reasoning process of the model. 

However, while working with the ChainOfThought() module, I noticed that the content of reasoning_content does not match the content wrapped in [** reasoning **].

How should we handle this inconsistency? Are there any specific guidelines or fixes to ensure alignment between reasoning_content and the [** reasoning **] section?

### Question 2:
Currently, when using the deepseek-reasoner model, there is no straightforward way to access the model's internal reasoning results (`reasoning_content`). 

This would allow users to clearly view the model's reasoning process, improving transparency and debugging capabilities.

A simple and feasible approach would be to include the `reasoning_content` parameter in the `inspect_history()` function. 

This would allow users to clearly view the model's reasoning process, improving transparency and debugging capabilities.

Thank you for your contributions to this great project. 

If possible, I would like to know your thoughts on this issue. I have already written a small portion of the code, and I am willing to make changes to the rest if needed.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Additional Context

Here is the log of my testing:

```
    react = dspy.ChainOfThought("question -> answer")
    pred = react(question="Who is Apple's CEO?")
```

log:
```
>>>litellm resp: 

[...]

PR #7994, opened 2025-03-21
[WIP] feat: add reasoning flag in `dspy.LM`

lxdlam:
## Summary

- Add a `reasoning_model` parameter in `dspy.LM` to indicate if the LM is a reasoning model.
- Add checks that if the `lm` is set with `reasoning_model == True`, we use plain predictor instead of the CoT prompted predictor.

## Explanation

This PR addresses how _reasoning models_, a new category of models capable of effective self-CoT, are changing the ecosystem. Many best practice guides recommend removing CoT prompts to reduce token usage, and even recommend avoiding them in OpenAI's [documentation](https://platform.openai.com/docs/guides/reasoning-best-practices).

Since the `ChainOfThought` module is a core component of dspy, completely refactoring all clients would cause significant breaking changes. Instead, I've implemented two separate predictors with routing based on the `lm` type.

This PR should be considered primarily as a proposal. Given how rapidly the ecosystem is evolving, self-CoT models may soon become the standard, potentially making the CoT module obsolete. I recognize that the current solution isn't ideal, but it offers a path forward.

### DIFF
--- a/dspy/predict/chain_of_thought.py
+++ b/dspy/predict/chain_of_thought.py
@@ -17,14 +17,14 @@ class ChainOfThought(Module):
         extended_signature = signature.prepend("reasoning", rationale_type, type_=str)
 
         self.plain_predict = dspy.Predict(signature, **config)
-        self.cot_predict = dspy.Predict(extended_signature, **config)
+        self.predict = dspy.Predict(extended_signature, **config)
 
     def forward(self, **kwargs):
         # Keep same logic with `dspy.Predict`
-        lm = kwargs.pop("lm", self.lm) or settings.lm
+        lm = kwargs.pop("lm", getattr(self, "lm", None)) or settings.lm
         assert isinstance(lm, BaseLM), "No LM is loaded."
 
         # Custom models that subclassing `BaseLM` don't have this parameter
         if getattr(lm, "reasoning_model", False):
             return self.plain_predict(**kwargs)
-        return self.cot_predict(**kwargs)
+        return self.predict(**kwargs)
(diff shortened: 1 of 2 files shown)


shiosai:
Apparently LiteLLM has already (some) support to detect the reasoning part so it would be probably better if it could just be used?: https://github.com/stanfordnlp/dspy/issues/7793

lxdlam:
> Apparently LiteLLM has already (some) support to detect the reasoning part so it would be probably better if it could just be used?: #7793

Good point. We can utilize it but the parameter should be retained since there are native client libraries or sometimes we're not on the track of the latest LiteLLM.

shiosai:
> Good point. We can utilize it but the parameter should be retained since there are native client libraries or sometimes we're not on the track of the latest LiteLLM.

True. I guess it might be not possible (in an easy way) for DSPy to know in advance whether a certain model will have reasoning or not. In any case, if the model has a reasoning part it would be good to not throw it away even when using "plain" predictor.  Im curious if there's a general design decision by the DSPy maintainers.

lxdlam:
> > Good point. We can utilize it but the parameter should be retained since there are native client libraries or sometimes we're not on the track of the latest LiteLLM.
> 
> True. I guess it might be not possible (in an easy way) for DSPy to know in advance whether a certain model will have reasoning or not. In any case, if the model has a reasoning part it would be good to not throw it away even when using "plain" predictor. Im curious if there's a general design decision by the DSPy maintainers.

I think it's not an issue only suffered by DSPy since the CoT part for current reasoning models is all different: in different response fields, in different naming, even [obfuscated](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking?q=extended#understanding-thinking-blocks) or [omitted](https://community.openai.com/t/how-to-print-o1-reasoning-token/964057/12).

Given no standard way to process these fields, manually marking the model as a reasoning model seems the most practical solution by now.

PR #7996, opened 2025-03-21
Refactor DSPy adapters to make it more extensible

chenmoneygithub:
We are reworking DSPy adapters for extensibility. For most users this change shouldn't cause backward compatibility issues, but if your workflow explicitly calls child methods of DSPy adapters, you need to make adjustments. 

The goal here is with DSPy 3.0, we want Adapter to be a customizable interface rather than some tribal knowledge. We acknowledge that it will be common for users willing to write their own adapter to adjust to their LLMs and workflows. However, the current adapter doesn't have a decent abstraction, and to write a custom adapter users need to understand the source code, and go through a tedious debugging process without guidelines.

In this PR, we are trying to standardize the dspy Adapters, and open a few hooks for people to override during customization. We are aware that there is no single standard that fits all use cases, but trying to hit a stage where we don't over-simplify or over-engineer the base DSPy Adapter/

In a nutshell, we are making the following breakdown of Adapters:

- Adapter 
  - format(): formats the type-based inputs into LM multiturn messages
    - System messages: The high level description of the task, and LM I/O format.
      - Fields description: `format_field_description()`
      - LM input/output structure description: `format_field_structure()`
      - task description: `format_task_description`
    - Few-shot examples (demo): multiturn few-shot examples
      - user message (inputs of demo): `format_user_message_content()`
      - assistant message (outputs of demo): `format_assistant_message_content()`
    - Conversation history: multiturn conversation history
      - user message (inputs of history message): `format_user_message_content()`
      - assistant message (outputs of history message): `format_assistant_message_content()` 
    - Current input: the actual question/input
      - user message: `format_user_message_content()`
  - parse(): parse the LM response to type-based outputs. No sub-hook for `parse()` because it varies for different adapters.

Note that `format_user_message_content()` and `format_assistant_message_content()` are used in multiple places. Users can override any level of hooks for customization. 

We will publish a guide on how to customize Adapter with concrete use cases after landing this PR.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -57,70 +57,197 @@ class Adapter(ABC):
         demos: list[dict[str, Any]],
         inputs: dict[str, Any],
     ) -> list[dict[str, Any]]:
+        """Format the input messages for the LM call.
+
+        This method converts the DSPy structured input along with few-shot examples and conversation history into
+        multiturn messages as expected by the LM. For custom adapters, this method can be overridden to customize
+        the formatting of the input messages.
+
+        In general we recommend the messages to have the following structure:
+        ```
+        [
+            {"role": "system", "content": system_message},
+            # Begin few-shot examples
+            {"role": "user", "content": few_shot_example_1_input},
+            {"role": "assistant", "content": few_shot_example_1_output},
+            {"role": "user", "content": few_shot_example_2_input},
+            {"role": "assistant", "content": few_shot_example_2_output},
+            ...
+            # End few-shot examples
+            # Begin conversation history
+            {"role": "user", "content": conversation_history_1_input},
+            {"role": "assistant", "content": conversation_history_1_output},
+            {"role": "user", "content": conversation_history_2_input},
+            {"role": "assistant", "content": conversation_history_2_output},
+            ...
+            # End conversation history
+            {"role": "user", "content": current_input},
+        ]
+
+        And system message should contain the field description, field structure, and task description.
+        ```
+
+
+        Args:
+            signature: The DSPy signature for which to format the input messages.
+            demos: A list of few-shot examples.
+            inputs: The input arguments to the DSPy module.
+
+        Returns:
+            A list of multiturn messages as expected by the LM.
+        """
         inputs_copy = dict(inputs)
         messages = []
         system_message = (
             f"{self.format_field_description(signature)}\n"
             f"{self.format_field_structure(signature)}\n"
-            f"{self.format_objective(signature)}"
+            f"{self.format_task_description(signature)}"
         )
         messages.append({"role": "system", "content": system_message})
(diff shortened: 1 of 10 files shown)


dspy/adapters/chat_adapter.py
```diff
@@ -54,40 +40,127 @@ def __call__(
...
+        `ChatAdapter` requires input and output fields to be in their own sections, with section header using markers
+        `[[ ## field_name ## ]]`.
+
+        For example:
+        ```
+        import dspy
+
+        class MySignature(dspy.Signature):
+            text: str = dspy.InputField(description="The text to analyze")
+            context: str = dspy.InputField(description="The context of the text")
```
okhat, line 54: Why this long docstring that will go stale as soon as someone changes the adapter slightly?

dspy/adapters/base.py
```diff
@@ -42,55 +44,271 @@ def __call__(
...
+                by the LM.
+
+        Returns:
+            A string that contains the user message content.
+        """
+        messages = [prefix]
+        for k, v in signature.input_fields.items():
+            value = inputs[k]
+            formatted_field_value = format_field_value(field_info=v, value=value)
+            messages.append(f"[[ ## {k} ## ]]\n{formatted_field_value}")
```
okhat, line 193: This is _very_ specific to chat adapter. Why does it belong in base.py?

dspy/adapters/base.py
```diff
@@ -42,55 +44,271 @@ def __call__(
...
+            f"{self.format_field_description(signature)}\n"
+            f"{self.format_field_structure(signature)}\n"
+            f"{self.format_task_description(signature)}"
+        )
+        messages.append({"role": "system", "content": system_message})
+        messages.extend(self.format_demos(signature, demos))
+        conversation_history = self.format_conversation_history(signature, inputs_copy)
+        if conversation_history:
+            # Conversation history and current input
+            messages.extend(conversation_history)
```
okhat, line 111: How do we handle cases where the *demos* have multi-turn history?

chenmoneygithub: it will be a single turn entity with history as one field, for example:

```
User message:

This is an example of the task, though some input or output fields are not supplied.

[[ ## question ## ]]
What is the capital of France?

[[ ## history ## ]]
[{"question": "What is the capital of Germany?", "answer": "Berlin"}]

Respond with the corresponding output fields, starting with the field `[[ ## reasoning ## ]]`, then `[[ ## answer ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.


Assistant message:

[[ ## reasoning ## ]]
Not supplied for this particular example. 

[[ ## answer ## ]]
Paris

```

dspy/adapters/chat_adapter.py
chenmoneygithub: simplified for easy maintenance!

PR #8001, opened 2025-03-24
Fix a bug in Teleprompter.get_params

TomeHirata:
Fix the bug in `get-params` method and change the test to capture the issue.


TomeHirata:
Since this is a hotfix for an existing broken method, let me quickly merge it.

PR #8003, opened 2025-03-24
Change the output interface of evaluate

TomeHirata:
In this PR, we change the output interface of `Evaluate.__call__`.
Instead of returning either score, (score, outputs), (score, scores, outputs) based on arguments, it will always return dspy.Prediction containing the following fields:

- score: A float percentage score (e.g., 67.30) representing overall performance
- all_scores: a list of float scores for each example in devset
- all_outputs: a list of (example, prediction, score) tuples for each example in devset
- result_table: a pandas DataFrame containing the evaluation results.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -112,7 +112,7 @@ class Evaluate:
             
             - all_scores: a list of float scores for each example in devset
             
-            - return_outputs: a list of (example, prediction, score) tuples for each example in devset
+            - all_outputs: a list of (example, prediction, score) tuples for each example in devset
 
             - result_table: a pandas DataFrame containing the evaluation results.
 
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -56,8 +56,6 @@ class Evaluate:
         display_progress: bool = False,
         display_table: Union[bool, int] = False,
         max_errors: int = 5,
-        return_all_scores: bool = False,
-        return_outputs: bool = False,
         provide_traceback: bool = False,
         failure_score: float = 0.0,
         **kwargs,
@@ -71,8 +69,6 @@ class Evaluate:
             display_table (Union[bool, int]): Whether to display the evaluation results in a table. 
                 If a number is passed, the evaluation results will be truncated to that number before displayed. 
             max_errors (int): The maximum number of errors to allow before stopping evaluation.
-            return_all_scores (bool): Whether to return scores for every data record in `devset`.
-            return_outputs (bool): Whether to return the dspy program's outputs for every data in `devset`.
             provide_traceback (bool): Whether to provide traceback information during evaluation.
             failure_score (float): The default score to use if evaluation fails due to an exception.
         """
@@ -82,8 +78,6 @@ class Evaluate:
         self.display_progress = display_progress
         self.display_table = display_table
         self.max_errors = max_errors
-        self.return_all_scores = return_all_scores
-        self.return_outputs = return_outputs
         self.provide_traceback = provide_traceback
         self.failure_score = failure_score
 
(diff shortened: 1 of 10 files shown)


dspy/evaluate/evaluate.py
```diff
@@ -183,15 +164,13 @@ def process_item(example):
...
-            return round(100 * ncorrect / ntotal, 2), [score for *_, score in results]
-        if return_outputs:
-            return round(100 * ncorrect / ntotal, 2), results
-
-        return round(100 * ncorrect / ntotal, 2)
+        
+        return dspy.Prediction(
+            score=round(100 * ncorrect / ntotal, 2),
+            all_scores=[score for *_, score in results],
+            all_outputs=results,
```
TomeHirata, line 171: maybe we don't need this. We can simply return a prediction for each example.

dspy/primitives/prediction.py
```diff
@@ -90,6 +90,15 @@ def __ge__(self, other):
         elif isinstance(other, Prediction):
             return self.__float__() >= float(other)
         raise TypeError(f"Unsupported type for comparison: {type(other)}")
+    
+    def __eq__(self, other):
+        if isinstance(other, (float, int)):
+            return self.__float__() == other
+        elif isinstance(other, Prediction):
+            return self.__float__() == float(other)
```
chenmoneygithub, line 98: nit: shall we do `float(self) == float(other)` for consistency?

chenmoneygithub (approved):
Solid work! LGTM with one minor comment. 

Let's talk offline for potential breakage, and align on the release schedule.

dspy/primitives/prediction.py
TomeHirata: I guess this should be consistent with how __ge__ or __le__ are implemented?

chenmoneygithub (approved):

Nasreddine:
please merge the PR to be able to get the individual example level evaluation score. This will be useful for mlflow tracing.

dspy/evaluate/evaluate.py
```diff
@@ -41,6 +41,18 @@ def HTML(x: str) -> str:  # noqa: N802
 logger = logging.getLogger(__name__)
 
 
+class EvaluationResult(dspy.Prediction):
```
chenmoneygithub: minor nit: use direct import instead of `dspy.Prediction`

dspy/evaluate/evaluate.py
```diff
@@ -41,6 +41,18 @@ def HTML(x: str) -> str:  # noqa: N802
...
+    """
+    A class that represents the result of an evaluation.
+    It is a subclass of `dspy.Prediction` that contains the following fields
+
+    - score: An float value (e.g., 67.30) representing the overall performance
```
chenmoneygithub, line 50: probably we can override the `__init__()` to include `score` and `results` for clarity?

dspy/evaluate/evaluate.py
```diff
@@ -41,6 +41,18 @@ def HTML(x: str) -> str:  # noqa: N802
...
+    - results: a list of (example, prediction, score) tuples for each example in devset
+    """
+    def __repr__(self):
+        return f"EvaluationResult(score={self.score}, results=`{len(self.results)} results`)"
```
chenmoneygithub: I feel a bit unnatural when reading "results=128 results", maybe we can do something like below?

```
 return f"EvaluationResult(score={self.score}, results=<list of {len(self.results)} items>)"
```

TomeHirata: Yeah, I was wondering what is the least ugly message. Let me use <> style

dspy/primitives/prediction.py
```diff
@@ -98,6 +98,15 @@ def __ge__(self, other):
             return self.__float__() >= float(other)
         raise TypeError(f"Unsupported type for comparison: {type(other)}")
 
+    def __eq__(self, other):
+        if isinstance(other, (float, int)):
+            return self.__float__() == other
+        elif isinstance(other, Prediction):
```
okhat: Hmm this is really dangerous! It's a really bad idea. Why did we add this?

TomeHirata: This was added mainly for two reasons:
- To support the existing logic that compares the equality of evaluation outputs. Users may also do `eval(program_1)==eval(program_2)` in their code
- Completeness for the comparison operator of dspy.Prediction. It is not consistent if `>=` works for `dspy.Prediction` but `==` does not

PR #8011, opened 2025-03-25
Introduce two step adapter

TomeHirata:
This PR introduces a new adapter that calls a reasoning model to produce a text and then let a smaller model parse the text and extract structured data.

## Quick testing

**Math subset**
Two step adapter using o3-mini as the main LM and 4o-mini as an extractor increases the performance on 50 sample subsets of the algebra dataset.

Chat adapter (o3-mini)
43 / 50
Two-step adapter (o3-mini + 4o-mini)
46 / 50

**GSM8K**
In the local testing using o3-mini as the main LM and 4o-mini as an extractor for 20 samples of GSM8K valset, it correctly resolved all issues without optimization.
image


dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    def _format_input(self, signature: Signature, values: dict[str, Any]) -> str:
+        parts = []
+        
+        for name in signature.input_fields.keys():
+            if name in values:
+                parts.append(f"{name}: {values[name]}")
+                
+        return "\n".join(parts)
+    
+    def _format_demo(self, signature: Signature, values: dict[str, Any]) -> list[dict[str, str]]:
```
chenmoneygithub, line 116: I am curious about one thing - the few-shot example's output part (assistant message) could be different from the main LM's output, since we don't require the output format from the main LM. Could this be a problem? One way to verify is running optimizer on a simple CoT, and use this two-step adapter, then compare the main LM's output between the original and optimized CoT. 

My guess is this should be all right, since reasoning model is pretty strong at understanding user intention, so this slight mismatch won't affect the response quality.

chenmoneygithub (commented):
Great work! Left some minor comments.

dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    adapter = dspy.TwoStepAdapter(dspy.LM('openai/gpt-4o-mini'))
+    dspy.configure(lm=lm, adapter=adapter)
+    program = CoT()
+    result = program("What is the capital of France?")
+    print(result)
+    ```
+    """
+    
+    def __init__(self, extraction_model: LM):
+        assert isinstance(extraction_model, LM), "extraction_model must be an instance of LM"
```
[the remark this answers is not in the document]
TomeHirata: Thanks, let's unify this type of code style and enable linter.

dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+        Returns:
+            A list of messages to be passed to the main LM.
+        """
+        messages = []
+
+        # Create a task description for the main LM
+        task_description = self._create_task_description(signature)
+        messages.append({"role": "system", "content": task_description})
+
+        # Format demos in a natural way
```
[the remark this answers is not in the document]
TomeHirata: I guess Isaac intended to emphasize "natural", but agree this comment is not necessary

dspy/adapters/two_step_adapter.py
TomeHirata: Great point, actually the main LM tries to output a text in a semi-structured way like the one below. The good point of this adapter is that it does not require reasoning models to strictly follow a specific format.

```
reasoning: First, we determine Ella's correct answers by subtracting her incorrect answers from the total: 40 − 4 = 36. Next, we compute half of Ella’s score, which is ½ × 36 = 18. Marion's score is stated as 6 more than half of Ella's score; thus, Marion’s score is 18 + 6 = 24. answer: 24
```

dspy/adapters/utils.py
```diff
@@ -237,3 +237,14 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
     else:
         # Neither => enclose in single quotes
         return f"'{s}'"
+    
+def enumerate_fields(fields: dict) -> str:
```
chenmoneygithub, line 241: this method has been renamed to `get_field_description_string` in the same file, can we use it? 

I did the renaming because `enumerate_fields` is a bit vague.

dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+        
+        Args:
+            original_signature: The original signature to extract output fields from
+        
+        Returns:
+            A new Signature type with a text input field and all output fields
+        """
+        # Create new fields dict with 'text' input field and all output fields
+        new_fields = {
+            'text': (str, InputField()),
```
chenmoneygithub, line 145: minor nit: double quotes

dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+from dspy.signatures.field import InputField
+from dspy.signatures.signature import make_signature
+from dspy.clients import LM
+
+class TwoStepAdapter(Adapter):
+    """
+    A two-stage adapter that:
+        1. Uses a simpler, more natural prompt for the main LM
+        2. Uses a smaller LM with chat adapter to extract structured data from the response of main LM
+    This adapter uses a commong __call__ logic defined in base Adapter class.
```
chenmoneygithub, line 16: typo: commong => common

dspy/adapters/two_step_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+    This class is particularly useful when interacting with reasoning models as the main LM since reasoning models
+    are known to struggle with structured outputs.
+
+    Example:
+    ```
+    import dspy
+    lm = dspy.LM(model='openai/o3-mini', max_tokens=10000, temperature = 1.0)
```
chenmoneygithub, line 23: nit: double quotes for consistency

dspy/dsp/utils/settings.py
```diff
@@ -7,6 +7,7 @@
 DEFAULT_CONFIG = dotdict(
     lm=None,
     adapter=None,
+    backup_adapter=None,
```
chenmoneygithub, line 10: is this intentional? I remember this is in another pending PR 🤔

chenmoneygithub (commented):
We should be good to go after addressing the comments! Meanwhile, can we include a comparison score between running optimizer + chatadapter and running optimizer + two-step optimizer in the PR description?

dspy/dsp/utils/settings.py
TomeHirata: This shouldn't be necessary, I guess Isaac added this but didn't use it in the two step adapter implementation.

dspy/adapters/utils.py
TomeHirata: Good catch, let me use that instead!

okhat:
Merging after several updates. My take is that this will not really help o3-mini very much, but it *does* help unreliable reasoning models like R1 and (what I verified is) that `together_ai/deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B` benefits hugely. (On 30 MATH-all tasks, 43% vs 70% accuracy via math-verify.)

PR #8014, opened 2025-03-26
Add usage tracking

chenmoneygithub:
Per community request, we are adding usage tracking for DSPy program. As an overview:

- We track the LM usage in the openai (litellm) format for each program call
- Local cache hit will cause no usage
- If multiple LMs are used in the process, they will be tracked separately. 
- The returned usage is a dictionary from lm name to openai-format usage data. The usage data is the merge result across all LM calls in one single program pass.

To use usage tracking, follow the code example below:

```
import dspy

dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini", cache=False), track_usage=True)

predict = dspy.ChainOfThought("question -> answer")
output = predict(question="What is the capital of France?")
print(output.get_lm_usage())
```

This usage tracking also works for multi-module program, for example:

```
import dspy

dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini", cache=False), track_usage=True)

class MyProgram(dspy.Module):
    def __init__(self):
        self.predict1 = dspy.ChainOfThought("question -> answer")
        self.predict2 = dspy.ChainOfThought("question, answer -> score")

    def __call__(self, question: str) -> str:
        score = self.predict2(question=question, answer=answer)
        return score

program = MyProgram()
output = program(question="What is the capital of France?")
print(output.get_lm_usage())
```

The output is the combined output for the end to end program call. 

This usage tracking also works with multithreading:

```
class MyProgram(dspy.Module):
    def __init__(self, lm):
        self.lm = lm
        self.predict1 = dspy.ChainOfThought("question -> answer")
        self.predict2 = dspy.ChainOfThought("question, answer -> score")

    def __call__(self, question: str) -> str:
        with dspy.settings.context(lm=self.lm):
            score = self.predict2(question=question, answer=answer)
            return score

dspy.settings.configure(track_usage=True)
program1 = MyProgram(lm=dspy.LM("openai/gpt-4o-mini", cache=False))
program2 = MyProgram(lm=dspy.LM("openai/gpt-3.5-turbo", cache=False))

parallelizer = dspy.Parallel()

results = parallelizer(
    [
        (program1, {"question": "What is the meaning of life?"}),
        (program2, {"question": "why did a chicken cross the kitchen?"}),
    ]
)
```
it is thread-safe, each program's output will only see the usage of itself.

### DIFF
--- a/dspy/primitives/program.py
+++ b/dspy/primitives/program.py
@@ -24,7 +24,7 @@ class Module(BaseModule, metaclass=ProgramMeta):
         if settings.track_usage and settings.usage_tracker is None:
             with track_usage() as usage_tracker:
                 output = self.forward(*args, **kwargs)
-                output.lm_usage = usage_tracker.get_total_tokens()
+                output.set_lm_usage(usage_tracker.get_total_tokens())
                 return output
 
         return self.forward(*args, **kwargs)
--- a/dspy/primitives/program.py
+++ b/dspy/primitives/program.py
@@ -24,7 +24,7 @@ class Module(BaseModule, metaclass=ProgramMeta):
         if settings.track_usage and settings.usage_tracker is None:
             with track_usage() as usage_tracker:
                 output = self.forward(*args, **kwargs)
-                output.usage = usage_tracker.get_total_tokens()
+                output.lm_usage = usage_tracker.get_total_tokens()
                 return output
 
         return self.forward(*args, **kwargs)
--- a/dspy/primitives/program.py
+++ b/dspy/primitives/program.py
@@ -1,8 +1,10 @@
 import magicattr
 
+from dspy.dsp.utils.settings import settings
 from dspy.predict.parallel import Parallel
 from dspy.primitives.module import BaseModule
 from dspy.utils.callback import with_callbacks
+from dspy.utils.usage_tracker import track_usage
 
 
 class ProgramMeta(type):
@@ -19,6 +21,12 @@ class Module(BaseModule, metaclass=ProgramMeta):
 
     @with_callbacks
     def __call__(self, *args, **kwargs):
+        if settings.track_usage and settings.usage_tracker is None:
+            with track_usage() as usage_tracker:
+                output = self.forward(*args, **kwargs)
+                output.usage = usage_tracker.get_total_tokens()
+                return output
+
         return self.forward(*args, **kwargs)
 
     def named_predictors(self):
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -87,9 +87,9 @@ class ParallelExecutor:
 
             original = thread_local_overrides.overrides
             thread_local_overrides.overrides = parent_overrides.copy()
-            if original.get("usage_tracker"):
+            if parent_overrides.get("usage_tracker"):
                 # Usage tracker needs to be deep copied across threads so that each thread tracks its own usage
-                thread_local_overrides.overrides["usage_tracker"] = copy.deepcopy(original["usage_tracker"])
+                thread_local_overrides.overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
 
             try:
                 return index, function(item)
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -1,13 +1,13 @@
-import sys
-import tqdm
-import signal
-import logging
-import threading
-import traceback
-import time
 import contextlib
+import logging
+import signal
+import sys
+import threading
+import time
+import traceback
+from concurrent.futures import FIRST_COMPLETED, ThreadPoolExecutor, wait
 
-from concurrent.futures import ThreadPoolExecutor, wait, FIRST_COMPLETED
+import tqdm
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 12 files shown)


dspy/primitives/program.py
```diff
@@ -19,6 +21,12 @@ def __init__(self, callbacks=None):
 
     @with_callbacks
     def __call__(self, *args, **kwargs):
+        if settings.track_usage and settings.usage_tracker is None:
+            with track_usage() as usage_tracker:
+                output = self.forward(*args, **kwargs)
+                output.lm_usage = usage_tracker.get_total_tokens()
```
TomeHirata, line 27: q: what happens if the signature of the program is `xxx -> lm_usage`?

dspy/utils/usage_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Any
+
+from dspy.dsp.utils.settings import settings
+
+
+class UsageTracker:
+    """Tracks LM usage data within a context."""
+
+    def __init__(self):
+        self.usage_data = defaultdict(list)
```
TomeHirata, line 14: nit: if may be helpful if we document what are the key and value of this dict

dspy/utils/usage_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+        for k, v in usage_entry1.items():
+            if k in result:
+                if isinstance(v, dict):
+                    result[k] = self._merge_usage_entries(result[k], v)
+                else:
+                    result[k] = result[k] or 0
+                    result[k] += v if v else 0
+        return result
+
+    def add_usage(self, lm: str, usage_entry: dict):
```
TomeHirata, line 38: nit: shall we use `model` instead of lm? In DSPy codebase, I see lm is mostly used for dspy.LM object

dspy/utils/usage_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+            return dict(usage_entry1)
+
+        result = dict(usage_entry2)
+        for k, v in usage_entry1.items():
+            if k in result:
+                if isinstance(v, dict):
+                    result[k] = self._merge_usage_entries(result[k], v)
+                else:
+                    result[k] = result[k] or 0
+                    result[k] += v if v else 0
```
TomeHirata, line 35: q: do we need to check if the value is either float or int?

dspy/primitives/program.py
chenmoneygithub: aha good point! Although I think no user will do this, we can actually make it more robust by changing property to a method. Let me fix.

dspy/utils/usage_tracker.py
chenmoneygithub: sg!

dspy/utils/usage_tracker.py
chenmoneygithub: got your concern. It's actually meaning the string representation of `dspy.settings.LM`, so I kinda prefer the name `lm` here for consistency.

dspy/utils/usage_tracker.py
chenmoneygithub: good question, it's actually guaranteed to be either dict, numerics or None: https://platform.openai.com/docs/api-reference/chat/object#chat/object-usage

dspy/utils/parallelizer.py
```diff
@@ -88,7 +86,10 @@ def worker(parent_overrides, submission_id, index, item):
             from dspy.dsp.utils.settings import thread_local_overrides
 
             original = thread_local_overrides.overrides
-            thread_local_overrides.overrides = parent_overrides.copy()
+            thread_local_overrides.overrides = copy.deepcopy(parent_overrides)
```
okhat, line 89: For now, only do deepcopy for the usage tracker. But soon, switch to full deepcopy once we change the semantics of `lm.history` or deprecate it.

xaviermehaut:
It seems that using track_usage=True, the get_lm_usage() method returns None for Google models (e.g., gemini-1.5-pro). This is because LiteLLM returns a Usage object for these models.
Example of a return forom liteLLM :

ModelResponse(id='<id>', created=1743326073, model='gemini-1.5-pro', object='chat.completion', system_fingerprint=None, choices=[Choices(finish_reason='stop', index=0, message=Message(content="Rome est la capitale de l'Italie.\n", role='assistant', tool_calls=None, function_call=None, provider_specific_fields=None))], usage=Usage(completion_tokens=10, prompt_tokens=9, total_tokens=19, completion_tokens_details=None, prompt_tokens_details=None), vertex_ai_grounding_metadata=[], vertex_ai_safety_results=[], vertex_ai_citation_metadata=[])

--- Informations d'utilisation via litellm direct (Google) ---
Usage(completion_tokens=10, prompt_tokens=9, total_tokens=19, completion_tokens_details=None, prompt_tokens_details=None)

Process finished with exit code 0

okhat:
Great catch, @xaviermehaut What would you like it to return instead?

xaviermehaut:
I don't know :)
when i try to use the tracking in a simple example using the gemini llm instead og openai  , I've the following :
--- Example 1: Simple Module ---
Question: What is the capital of Italy?
Answer: Rome

LLM Usage Information (Simple Module):
None
------------------------------

--- Example 2: Multi-Module Program ---
Question: What is the capital of France?
Score (program result): 1

LLM Usage Information (Multi-Module Program):
None
------------------------------

which suggests that the set output.set_lm_usage(usage_tracker.get_total_tokens()) in Module is not set correctly because of usage_tracker.get_total_tokens() returned normally by google server...

PR #8033, opened 2025-03-31
Remove inspect_history from COPRO

### LINKED ISSUE #8026: [Feature] I hope to disable the COPRO log when I use prompt_model

When I pass `prompt_model`, the model's logs are always printed.  
I tried setting the logger to `logger.disabled=True`, but it still didn't work.  
Here, it calls `self.prompt_model.inspect_history`, and the logs are directly printed inside this method.

Image
   

### Additional Context

_No response_

TomeHirata:
Remove inspect_history from COPRO optimizer.

Fixes #8026

### DIFF
--- a/docs/docs/faqs.md
+++ b/docs/docs/faqs.md
@@ -116,7 +116,7 @@ If you're dealing with "context too long" errors in DSPy, you're likely using DS
 
 A more general fix is simply increasing the number of `max_tokens` specified to the LM request (e.g. `lm = dspy.OpenAI(model = ..., max_tokens = ...`).
 
-## Set Vorbose
+## Set Verbose Level
 DSPy utilizes the [logging library](https://docs.python.org/3/library/logging.html) to print logs. If you want to debug your DSPy code, set the logging level to DEBUG with the example code below.
 
 ```python
--- a/docs/docs/faqs.md
+++ b/docs/docs/faqs.md
@@ -121,5 +121,12 @@ DSPy utilizes the [logging library](https://docs.python.org/3/library/logging.ht
 
 ```python
 import logging
-logging.basicConfig(level=logging.DEBUG)
+logging.getLogger("dspy").setLevel(logging.DEBUG)
+```
+
+Alternatively, if you want to reduce the amount of logs, set the logging level to WARNING or ERROR.
+
+```python
+import logging
+logging.getLogger("dspy").setLevel(logging.WARNING)
 ```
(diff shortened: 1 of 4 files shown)


okhat:
Thank you @TomeHirata ! Why would we make inspect_history log? It's a function for printing. No one should use it if they don't want to print. Adding logging will clutter the printing with date and time.

dspy/teleprompt/copro_optimizer.py
```diff
@@ -331,7 +329,7 @@ def compile(self, student, *, trainset, eval_kwargs):
                 )
 
             if self.prompt_model:
-                logger.debug(f"{self.prompt_model.inspect_history(n=1)}")
+                self.prompt_model.inspect_history(n=1)
```
chenmoneygithub, line 332: We can probably just remove this logging, I suspect if users will be able to interpret the overwhelming logging...

docs/docs/faqs.md
```diff
@@ -115,3 +115,18 @@ Modules can be frozen by setting their `._compiled` attribute to be True, indica
 If you're dealing with "context too long" errors in DSPy, you're likely using DSPy optimizers to include demonstrations within your prompt, and this is exceeding your current context window. Try reducing these parameters (e.g. `max_bootstrapped_demos` and `max_labeled_demos`). Additionally, you can also reduce the number of retrieved passages/docs/embeddings to ensure your prompt is fitting within your model context length.
 
 A more general fix is simply increasing the number of `max_tokens` specified to the LM request (e.g. `lm = dspy.OpenAI(model = ..., max_tokens = ...`).
+
+## Set Vorbose
```
chenmoneygithub, line 119: typo: verbose

probably we can go with "Set Verbosity Level"

chenmoneygithub (commented):
Thanks for the PR! 

`logger.info` doesn't work well with colorful printing, and this `inspect_history` is mostly used for debugging, so as Omar mentioned `print` is the good fit. For the original issue, I think we can remove the logging completely, and we should ship your `faqs.md` change!

docs/docs/faqs.md
TomeHirata: Good catch, fixed

TomeHirata:
Hi, @okhat @chenmoneygithub. Understood that format change caused by logging is a concern. The colored logging works fine, but I agree that the date and time are a bit clean.
@okhat Do you also agree that we remove this logging from COPRO completely?

PR #8040, opened 2025-04-01
Fix unrecognized api mkdoc options

TomeHirata:
When running `mkdoc serve`, the following warning appears many times. Also, inherited methods don't appear in the DSPy api doc as the `show_inherited_members` option is not recognized. This PR fixes the issue by updating the mkdoc options.

```
DeprecationWarning: Passing extra options directly under options is deprecated. Instead, pass them under options.extra, and
           update your templates. Current extra (unrecognized) options: show_inherited_members, show_undocumented_members?
```

### DIFF
--- a/docs/docs/api/adapters/Adapter.md
+++ b/docs/docs/api/adapters/Adapter.md
@@ -12,11 +12,10 @@
             - format_turn
             - parse
         show_source: true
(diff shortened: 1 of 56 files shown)


chenmoneygithub (approved):
LGTM!

PR #8049, opened 2025-04-07
Fix the type of format_demos

TomeHirata:
Fix the output type hint for Adapter.format_demos


chenmoneygithub (approved):
Thanks for the fix!

PR #8050, opened 2025-04-08
Implement adapter retry for Pydantic Validation Error

### LINKED ISSUE #7693: [Feature] Retries on Pydantic Validation Errors

Maybe I'm wrong but I thought I recalled the old version of DSPY (pre-litellm / pre-adapter era) doing retries for pydantic validation failures, but I'm not seeing any retrying in the codepaths I'm looking at in the stack trace (predict.py, json_adapter.py)
Is there a way built into the new version of DSPY to control retries? Or do I need to impl it into my own modules now?

E.g.:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for SomeSignature
some_list
  List should have at most 3 items after validation, not 4 [type=too_long, input_value=[1,2,3,4], input_type=list]
    For further information visit https://errors.pydantic.dev/2.9/v/too_long
```

I would expect this to feed back into the LLM to let it correct its outputs.

### Additional Context

_No response_

TomeHirata:
This PR introduces RetryAdapter, which 
1. Call the main_adapter.
2. Call the fallback_adapter if the main_adapter fails, c
3. Retry the main_adapter including previous response for `max_retries` times if fallback adapter failed.

Resolve #7693

Alternative design of adapter retry:

1. Implement the decorator method around the adapter call
This design can separate the call function from the retry logic. However, faces difficulty in accessing the previous LLM call output.

2. Implement Retry as a dspy module
This is the previous design before LiteLLM. However, Pydantic validation is a responsibility of adapter abstraction and having the retry logic in this layer should be cohesive.

3. Have the logic in BaseAdapter.call
This approach is not suitable since several different adapters are used in the retry logic instead of retrying one adapter for multiple times.


dspy/adapters/base.py
```diff
@@ -28,25 +32,37 @@ def __call__(
...
+                    values.append(value)
+
+                return values
+            except ValueError:
+                if retry_count >= max_retries:
+                    raise
+                
+                logger.debug("A ValueError occurred while parsing the LM output. Retrying with a new signature.", exc_info=True)
+                if retry_count == 0:
+                    signature = create_signature_for_retry(signature)
```
okhat, line 63: Note to self: How does modifying the signature interact with demonstrations and with optimization?

dspy/dsp/utils/settings.py
```diff
@@ -22,6 +22,7 @@
     disable_history=False,
     track_usage=False,
     usage_tracker=None,
+    adapter_retry_count=0, # number of times to retry the adapter call
```
chenmoneygithub, line 25: For thinking: 

- Shall we just call this `max_retry`? `adapter_retry_count` could be mysterious for DSPy beginners. 
- Shall we default to 1 instead of 0? It would be helpful to collect data on the success ratio of 1 retry, 2 retries when the first call fails. A small model like llama-3b might provide useful data for us.

dspy/adapters/base.py
```diff
@@ -28,25 +32,37 @@ def __call__(
...
+                        value["logprobs"] = output_logprobs
+
+                    values.append(value)
+
+                return values
+            except ValueError:
+                if retry_count >= max_retries:
+                    raise
+                
+                logger.debug("A ValueError occurred while parsing the LM output. Retrying with a new signature.", exc_info=True)
```
chenmoneygithub, line 61: "Retrying with a new signature" is a bit opaque, probably:

```
logger.debug("ValueError while parsing LM output on attempt %d. Retrying", retry_count, exc_info=True)
```

dspy/adapters/base.py
```diff
@@ -28,25 +32,37 @@ def __call__(
...
-            if output_logprobs is not None:
-                value["logprobs"] = output_logprobs
-
-            values.append(value)
-
-        return values
+        retry_count = 0
+        max_retries = max(settings.adapter_retry_count, 0)
+        outputs = None
+        while (True):
```
chenmoneygithub, line 38: nonblocking: I prefer to use a for-loop `for attempt in range(max_retries + 1)` to indicate that this is a loop with cap. Feel free to dismiss if you like the `while` structure better.

dspy/adapters/base.py
```diff
@@ -28,25 +32,37 @@ def __call__(
...
+
+                return values
+            except ValueError:
+                if retry_count >= max_retries:
+                    raise
+                
+                logger.debug("A ValueError occurred while parsing the LM output. Retrying with a new signature.", exc_info=True)
+                if retry_count == 0:
+                    signature = create_signature_for_retry(signature)
+                inputs["previous_response"] = json.dumps(outputs)
```
chenmoneygithub, line 64: should we also include the exception message?

chenmoneygithub (commented):
Nice work! Can we include some demo code + logging/screenshot on retry resolves the pydantic parsing error? Although I am almost sure that this is helpful for small models, I don't have a concrete idea on how robust this retry works.

dspy/dsp/utils/settings.py
TomeHirata: We have num_retries in LM. So I think it's confusing if we can use retry without any prefix. Maybe `max_parse_retries`?

dspy/adapters/base.py
```diff
@@ -8,6 +8,7 @@
 if TYPE_CHECKING:
     from dspy.clients.lm import LM
 
+logger = logging.getLogger(__name__)
```
chenmoneygithub, line 11: seems unused?

dspy/adapters/retry_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+
+logger = logging.getLogger(__name__)
+
+class RetryAdapter(Adapter):
+    """
+    RetryAdapter is an adapter that retries the execution of another adapter for
+    a specified number of times if it fails to parse completion outputs.
+    """
+
+    def __init__(self, main_adapter: Adapter, fallback_adapter: Optional[Adapter] = None, max_retries: int = 3):
```
chenmoneygithub, line 20: Just thinking out loud: is it worth making `fallback_adapter` a list so that we can have a series of retry?

dspy/adapters/retry_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+        lm_kwargs: dict[str, Any],
+        signature: Type[Signature],
+        demos: list[dict[str, Any]],
+        inputs: dict[str, Any],
+    ) -> list[dict[str, Any]]:
+        """
+        Execute main_adapter and fallback_adapter in the following procedure:
+        1. Call the main_adapter.
+        2. If the main_adapter fails, call the fallback_adapter.
+        3. If the fallback_adapter fails, retry the main_adapter including previous response for `max_retries` times.
```
chenmoneygithub, line 45: This step 3 can be a bit counterintuitive for users, as `max_retries` is affecting the main adapter instead of the fallback adapter, but the very first "retry" happens on the fallback adapter. 

Writing down my suggestion for discussion purpose:

- Move the `max_retries` into the individual adapter, and only issue retries on pydantic parsing errors to avoid retrying on runtime errors like malformed request, context limit exceeding or so. 
- In the RetryAdapter,  we make `fallback_adapter` as a list, and in `__call__` method we just run down the list of adapters until hitting no exception.

dspy/adapters/retry_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+            signature,
+            demos,
+            inputs,
+        )
+        outputs.extend(values)
+
+        if len(outputs) == n_completion:
+            return outputs
+        
+        lm_kwargs["n"] = n_completion - len(outputs)
```
chenmoneygithub, line 77: I am wondering if it is possible that for the same request's multiple responses, some responses can be parsed while some cannot. It does sound possible, but I don't have this experience... 

Or should we discard all responses if some of responses fail to be parsed? There is a situation where one response has successful parsing because it is incomplete but parseable, e.g., only have `[## Completion ##]` or so.

dspy/adapters/retry_adapter.py
TomeHirata: Yeah, I was in two minds. Fallback_adapter is more explicit but it would be more flexible if we have a list of adapters.

dspy/adapters/retry_adapter.py
TomeHirata: Discussed internally and decided to keep the successful outputs in the previous steps.

dspy/adapters/retry_adapter.py
TomeHirata: Discussed internally and decided to try the fallback adapter first, then retry the main adapter several times.

dspy/adapters/retry_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    def __init__(self, main_adapter: Adapter, fallback_adapter: Optional[Adapter] = None, max_retries: int = 3):
+        """
+        Initializes the RetryAdapter.
+
+        Args:
+            main_adapter (Adapter): The main adapter to use.
+            fallback_adapter (Optional[Adapter]): The fallback adapter to use if the main adapter fails.
+            max_retries (int): The maximum number of retries. Defaults to 3.
```
chenmoneygithub, line 27: shall we name it to `main_adapter_max_retries` for clarity?

dspy/adapters/utils.py
```diff
@@ -237,3 +240,13 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
     else:
         # Neither => enclose in single quotes
         return f"'{s}'"
+
+def create_signature_for_retry(signature: Type[Signature]):
+    signature = signature.append("previous_response", InputField(
+        prefix="Previous Response",
```
chenmoneygithub, line 246: we can remove the `prefix` field, which is not used in DSPy now

chenmoneygithub (approved):
LGTM! Left two minor comments

thomasnormal:
Btw, there were some more features in [TypedPredictors](https://github.com/stanfordnlp/dspy/blob/943ee02d4cf9a630c89136ff6dc6b9f6ebf805dd/dspy/functional/functional.py#L95) you can consider bringing back:

- [Synthetic examples](https://github.com/stanfordnlp/dspy/blob/943ee02d4cf9a630c89136ff6dc6b9f6ebf805dd/dspy/functional/functional.py#L143): Some times examples are easier to follow than schemas. TypedPredictors would create simple examples of valid outputs and include them in the prompt. It just used another LLM call for this, but there are also projects like [polyfactory](https://pypi.org/project/polyfactory) that can do it programatically.
- [Error reformatting](https://github.com/stanfordnlp/dspy/blob/943ee02d4cf9a630c89136ff6dc6b9f6ebf805dd/dspy/functional/functional.py#L170): Pydantic often gives very hard to read error messages, even for humans. Rewriting those two explain what's actually wrong can help performance a lot. TypedPredictors would take the more general approach of using a different LLM call to explain stack traces and [turn them into advice](https://github.com/stanfordnlp/dspy/blob/943ee02d4cf9a630c89136ff6dc6b9f6ebf805dd/dspy/functional/functional.py#L202).

jukiewiczm:
Hi! This PR seems to have been approved since April. I just wanted to kindly ask - are there any plans on merging this and releasing this feature? That would be a great improvement to the dspy library

immu4989:
Came across this PR from #7693 and noticed the review thread's been quiet
since January — the topic is something I hit regularly in production LLM
pipelines, so I pulled the branch and probed a few behaviors. Existing
test suite passes clean (9/9). Three observations worth surfacing while
the design is being finalized.

**1. Happy-path coverage gap.** The existing tests assert either
"always-fail → raises" or "fallback succeeds immediately." Neither asserts
the core thing `RetryAdapter` is for: main fails N times and then succeeds
on retry N+1. I ran:

```python
state = {"n": 0}
real_parse = main.parse
def flaky_parse(sig, output):
    state["n"] += 1
    if state["n"] <= 2:
        raise ValueError(f"forced failure #{state['n']}")
    return real_parse(sig, output)

with patch.object(main, "parse", side_effect=flaky_parse):
    result = adapter(lm, {"n": 1}, MathSig, [], inputs)

# parse called 3 times → succeeded on attempt 3, result == [{"answer": 42}]
```
Works correctly, but a regression test for this would close the
happy path gap. Happy to send as a follow up commit.

2. On @chenmoneygithub's "fallback_adapter as list" question. I
checked whether RetryAdapter(RetryAdapter(primary, secondary), tertiary)
could serve as an interim multi-fallback pattern. It can't — nesting
fails with NotImplementedError since RetryAdapter is a meta-adapter
and doesn't implement format() / parse() directly:


`NotImplementedError:  (from outer RetryAdapter calling inner.format)`
So if users want a "JSON → XML → Chat" fallback chain (which is a real
production pattern for handling provider quirks), the only path is the
list of fallbacks design you mentioned. That's a +1 from a user
perspective.

3. Side effect on caller's inputs dict. __call__ mutates the
caller's inputs dict in place — after a failing run, the user's dict
has `previous_response` and `error_message` keys appended:


```
inputs = {"question": "6 x 7"}                       # before
# ... adapter raises ...
inputs == {"question": "6 x 7",
           "previous_response": "[[ ## answer ## ]]\n42",
           "error_message": "err"}                    # after
```
(lm_kwargs is also mutated in the general case, though for n=1 the
math comes out unchanged.) Cheap fix: shallow-copy at the top of
__call__. Worth doing — anyone who reuses the inputs dict for logging
or replay will be surprised, and it's a no-op behavior change.

Offering. Happy to send any of:

(a) The regression test for the success-on-retry-N happy path
(b) The inputs.copy() / lm_kwargs.copy() fix at __call__ entry
(c) A small follow on PR turning fallback_adapter: Optional[Adapter] into fallback_adapters: list[Adapter] = [] (with backward-compat shim so the existing single-fallback usage still works)

Pick any combination — (a) and (b) are tiny, (c) is the more substantive direction decision. 
Seconding @jukiewiczm's interest from November — there's outside demand to land this.

immu4989:
@TomeHirata , checking back here. Of the three follow ups (regression test,`inputs.copy()` fix, `fallback_adapters` list), any you'd like me to send first? Happy to ship whichever, or step aside if the design's moved.

immu4989:
@TomeHirata ,quick FYI in case useful. While running the parallel-bootstrap probe from earlier, I ended up building out the broader optimizer vs AgentDojo measurement and shipped a v0.1 finding this week.

The headline that's relevant to this PR's area: BootstrapFewShot Pareto-dominates MIPROv2 at small N (workspace suite, N=5) — higher utility AND higher security on direct attacks. Not what I expected given MIPROv2's heavier optimization budget. Scaling up to all 4 AgentDojo suites + GEPA in v0.2.

Write up: https://imranahamed.substack.com/p/does-dspy-prompt-optimization-weaken
Repo: https://github.com/immu4989/dspy-security-bench

If the v0.2 result holds at scale, the BootstrapFewShot vs. MIPROv2 finding might be a useful data point for any future optimizer-recommendations work. Happy to share data privately if useful before then.

Re #8050 itself: still happy to take a swing at any of the three follow-ups (regression test, `inputs.copy()` fix, or `fallback_adapters` list) when you have bandwidth.

PR #8056, opened 2025-04-11
Move provide_traceback into global settings

### LINKED ISSUE #1665: Preserving Tracebacks
There are a few places where getting to root errors is difficult because of the retry attempts and how errors are being preserved. I noticed for example that it was easier to go in and remove the try except blocks to debug the code than to keep them there. Im wondering if in the next version this could be worked on a bit. For example, maybe a inspect_errors() method? that shows the exact errors and full tracebacks?

Thanks!
-Devin

TomeHirata:
Resolve #1665


chenmoneygithub (approved):
LGTM!

PR #8057, opened 2025-04-11
Add max size for global history

### LINKED ISSUE #8045: [Feature] Make default value of settings.disable_history=True

Currently, settings.disable_history equals False by default which means that, in a production application, global variable `dspy.clients.base_lm.GLOBAL_HISTORY` increases indefinitely, meaning an OOM is inevitable, eventually. If we don't want to set settings.disable_history to True, I think the parameter should at least be better documented and the potential for OOM be made more clear. As a user, there is no way to be aware that history is stored indefinitely: I just spent a lot of time figuring out why my app was crashing only to find out GLOBAL_HISTORY was causing an OOM. 

Another possibility would be to set a max size to `GLOBAL_HISTORY`.

Relevant line of code: https://github.com/stanfordnlp/dspy/blob/53045805e6867fae0c15ef2f9bb541cd97822721/dspy/dsp/utils/settings.py#L22

### Additional Context

_No response_

TomeHirata:
Resolve #8045

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,11 +1,12 @@
 import datetime
 import uuid
 from abc import ABC
+from collections import deque
 
 from dspy.dsp.utils import settings
 from dspy.utils.callback import with_callbacks
 
-GLOBAL_HISTORY = []
+GLOBAL_HISTORY = deque(maxlen=100)
 
 
 class BaseLM(ABC):
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -5,7 +5,7 @@ from abc import ABC
 from dspy.dsp.utils import settings
 from dspy.utils.callback import with_callbacks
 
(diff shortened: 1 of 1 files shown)


dspy/clients/base_lm.py
```diff
@@ -5,6 +5,7 @@
 from dspy.dsp.utils import settings
 from dspy.utils.callback import with_callbacks
 
+MAX_HISTORY_SIZE = 100
```
chenmoneygithub, line 8: Shall we make this configurable for users? 100 is usually sufficient for me, but there is one time that I need to maintain a long history for trajectory lookup when building `dspy.Refine`.

okhat: I agree, but will merge now so we don't accumulate too many PRs. Making this configurable is a P2 or P3...

okhat: I also increased the default to 10_000 so that request (configurable count) is much less urgent now

PR #8076, opened 2025-04-17
Add tests for Unbatchify util

assadyousuf:
Adding tests for Unbatchify util. Seemed to be the only Util missing a test

### DIFF
--- a/tests/utils/test_unbatchify.py
+++ b/tests/utils/test_unbatchify.py
@@ -1,0 +1,75 @@
+import time
+from unittest.mock import MagicMock
+from concurrent.futures import Future
+
+from dspy.utils.unbatchify import Unbatchify
+
+
+def simple_batch_processor(batch):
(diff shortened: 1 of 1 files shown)


tests/utils/test_unbatchify.py
```diff
@@ -0,0 +1,75 @@
...
+from dspy.utils.unbatchify import Unbatchify
+
+
+def simple_batch_processor(batch):
+    """A simple batch function that adds 1 to each item."""
+    return [item + 1 for item in batch]
+
+
+def submit(self, input_item: any) -> Future:
+    """Submits an item for processing and returns a Future."""
```
assadyousuf, line 14: Needed to monkey patch Unbatchify._call_ because its blocking. This is test-only method that provides a non-blocking way to add an item to the queue

assadyousuf:
@okhat Would love to get some eyes on this when you get a chance

PR #8077, opened 2025-04-17
SIMBA Improvements

klopsahlong:
This PR makes the following updates:

- **Support for teacher and prompt model.** The teacher model will be used to generate 1/N trajectories for each example, so that we are still targeting rule generation improvements based on task model failure modes.
- **Don't append_demo if demo score < 10th percentile**. This will help us avoid adding in poor demos.
- **Support for metric metadata**. Allow for additional metadata to be passed back in a dspy.Prediction object, in addition to the score. Note that the one downside here is that users must know that the score should be called 'score'. To address this for now, we've added in an error message telling users to add in score to their dspy.Prediction object.
- **Fix for Optional fields**. Fixes a bug in the `parse_value` function that parses strings like "152" as numbers when they are of annotation type Optional(None, str)

Handled here, but merged sooner in another commit:
- **Fixes max recursion depth exceeded error**. Previously this error showed up when we tried to save the optimized program. This was due to one of the candidate_programs attached to the best_program being the best_program itself. This was fixed by making best_program a deep copy the max program in candidate_programs.


dspy/clients/lm.py
```diff
@@ -35,8 +35,8 @@ def __init__(
         self,
         model: str,
         model_type: Literal["chat", "text"] = "chat",
-        temperature: float = 0.0,
-        max_tokens: int = 1000,
+        temperature: Optional[float] = None,
+        max_tokens: Optional[int] = None,
```
klopsahlong, line 39: Updating defaults, which were throwing errors for reasoning models. Now, instead of defaulting to a temp of 0.0 and max_tokens 1000 (and erroring out automatically for o3mini), we are setting temp and max_tokens based on whether the model is a reasoning model or not. If the user has intentionally set one of the values to something the reasoning model can't handle (i.e. temperature=0.7), then we will still throw an error.

dspy/teleprompt/simba.py
```diff
@@ -41,6 +44,8 @@ def __init__(
         self.num_candidates = num_candidates
         self.max_steps = max_steps
         self.max_demos = max_demos
```
klopsahlong, line 46: Adding support for prompt / teacher models

dspy/teleprompt/simba.py
```diff
@@ -310,7 +316,7 @@ def register_new_program(prog: dspy.Module, score_list: list[float]):
                 trial_logs[idx_prog-1]["train_score"] = avg_score
         
         best_idx = scores.index(max(scores)) if scores else 0
-        best_program = candidate_programs[best_idx]
```
klopsahlong, line 313: Fixing max recursion depth error

dspy/teleprompt/simba_utils.py
```diff
@@ -3,20 +3,37 @@
...
+def prepare_models_for_resampling(program: dspy.Module, n: int, teacher_settings: Optional[Dict] = None):
     lm = program.get_lm() or dspy.settings.lm
-    temps = [lm.kwargs["temperature"]] + [0.5 + i * (0.5 / n) for i in range(n)]
-    temps = list(dict.fromkeys(temps))[:n]
-    return [lm.copy(temperature=t) for t in temps]
 
+    # Check to see if our model is a reasoning model, which means temp must stay as 1.0
+    model_family = lm.model.split("/")[-1].lower() if "/" in lm.model else lm.model.lower()
+    model_pattern = re.match(r"^o([13])(?:-mini)?", model_family)
+
```
klopsahlong, line 19: This function has been updated to add support for teacher model (used for 1 of the N trajectories)

dspy/teleprompt/simba_utils.py
```diff
@@ -28,30 +45,46 @@ def wrapped_program(example):
                 print(e)
             trace = dspy.settings.trace.copy()
```
klopsahlong, line 47: Updated to handle additional metric metadata in addition to the score. To do this, we check if the output from the metric is a float or int (in which case we use it as the score) or a dspy.Prediction object

dspy/teleprompt/simba_utils.py
```diff
@@ -116,12 +149,16 @@ def append_a_rule(bucket, system, **kwargs):
         worse_program_outputs=dict(bad["prediction"] or {}),
         worse_reward_value=bad["score"],
         better_reward_value=good["score"],
```
klopsahlong, line 151: again, adding additional metric metadata here

dspy/teleprompt/simba_utils.py
```diff
@@ -28,30 +45,46 @@ def wrapped_program(example):
...
-
 def append_a_demo(demo_input_field_maxlen):
     def append_a_demo_(bucket, system, **kwargs):
         predictor2name, name2predictor = kwargs["predictor2name"], kwargs["name2predictor"]
+        batch_10p_score = kwargs["batch_10p_score"]
 
-        trace = bucket[0]["trace"]
+        good = bucket[0]
+        trace = good["trace"]
         name2demo = {}
```
klopsahlong, line 83: Double checking that the demo we're appending is not at or below the 10th percentile of scores

dspy/adapters/utils.py
```diff
@@ -130,8 +131,19 @@ def find_enum_member(enum, identifier):
...
 
+def _strip_optional(ann):
+    """If ann is Union[..., NoneType] return the non‑None part, else ann."""
+    if get_origin(ann) is Union and NoneType in get_args(ann):
+        # keep the first non‑None member (there will be only one in Optional[T])
+        return next(a for a in get_args(ann) if a is not NoneType)
+    return ann
 
 def parse_value(value, annotation):
+    annotation = _strip_optional(annotation)
```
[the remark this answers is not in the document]
klopsahlong: More info on the failure case this fixes:

Previously, this was failing for fields of Union(str, None) with values of type 'str' that could also be parsed as ints. Ex: "9812750". 

The problem is in this sequence:
value = "9812750" (string)
annotation = typing.Optional[str]
candidate = json_repair.loads("9812750") → 9812750 (parses as integer, not str)
TypeAdapter(typing.Optional[str]).validate_python(9812750) → Fails with pydantic.ValidationError since the value is neither a str nor None
Exception handler is triggered: except pydantic.ValidationError as e:
Then we hit this line : issubclass(annotation, Type), which throws the error `issubclass() arg 1 must be a class` because `typing.Optional[str]` is not a class, since it's a type annotation/Union.

This fix involves first parsing the Optional field to get the expected non-Null type, and parse the value according to this. Now pydantic handles the type coercion correctly from str -> str instead of int-> int when the non-null annotation type is 'str'

Issue #8075, opened 2025-04-16
[Bug] "openai>=0.28.1,<=1.61.0" version restriction


softwareentrepreneer:
### What happened?

Couldn't install DSPy due to the <=1.61.0 restriction, is there a strong reason for this upper bound?

### Steps to reproduce

pip install openai-agents dspy

### DSPy version

2.6.17

okhat:
There was a LiteLLM issue requiring the pin, but it may be fixed now. Could you overrule that restriction?

softwareentrepreneer:
I am maintaining some packages as well, so overruling version constraints isn't ideal in my setup.
Does removing the upper limit break anything in DSPy?

To my knowledge:
`litellm 1.66.1` requires `openai >=1.68.2`
`openai-agents 0.0.11` requires `openai>=1.66.5`
so keeping the `openai<=1.61.0` restriction in DSPy makes it harder for people to adopt the library in a project setting,
especially when other dependencies require newer OpenAI versions.

Would you consider removing it in the next release?

cc-c4:
I'm also running into this issue; this restriction prevents me from using OpenAI's latest features such as the Responses API for vector search and other things. My workaround has been to install dspy first and then upgrade openai (dspy works fine with `openai==1.74.0`) but this is not ideal for CI/CD setups.

PR #8079, opened 2025-04-17
Remove openai<=1.61.0 due to incompatibility with litellm and openai-agents

Hangzhi:
https://github.com/stanfordnlp/dspy/issues/8075


okhat:
OK I'm fixing this in #8084 and I made you a co-author of it, @Hangzhi

PR #8082, opened 2025-04-18
Support default args in dspy.Tool

chenmoneygithub:
Before this PR default arg information is lost when we convert functions to `dspy.Tool`. with this PR we are resolving the issue.


jason-faro:
@chenmoneygithub This now causes a `TypeError: 'str' object does not support item assignment` if the assignment to "Any" happens here: https://github.com/stanfordnlp/dspy/blob/main/dspy/primitives/tool.py#L122

PR #8095, opened 2025-04-23
Support composite arg type parsing in `dspy.Tool`

chenmoneygithub:
Currently our arg parsing cannot handle composite type like `list[MyPydanticModel]`, so we make this PR to resolve the issue. 

The solution is creating a pydantic wrapper with one single dummy field `value` of the arg type, and utilize pydantic `model_validate()` to convert the LLM-generated args (json format) into the right type.

### DIFF
--- a/tests/primitives/test_tool.py
+++ b/tests/primitives/test_tool.py
@@ -159,3 +159,17 @@ def test_tool_with_default_args_without_type_hints():
     tool = Tool(foo)
     assert tool.args["x"]["default"] == 100
     assert not hasattr(tool.args["x"], "type")
+
+
+def test_tool_call_parses_args():
+    tool = Tool(dummy_with_pydantic)
+
+    args = {
+        "model": {
+            "field1": "hello",
+            "field2": 123,
+        }
+    }
+
+    result = tool(**args)
+    assert result == "hello 123"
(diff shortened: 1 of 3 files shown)


tests/primitives/test_tool.py
```diff
@@ -159,3 +159,17 @@ def foo(x=100):
     tool = Tool(foo)
     assert tool.args["x"]["default"] == 100
     assert not hasattr(tool.args["x"], "type")
+
+
+def test_tool_call_parses_args():
```
TomeHirata, line 164: nit: Can we also add a unit test for the nested type like `list[DummyModel]`

tests/primitives/test_tool.py
```diff
@@ -159,3 +159,17 @@ def foo(x=100):
...
+    tool = Tool(dummy_with_pydantic)
+
+    args = {
+        "model": {
+            "field1": "hello",
+            "field2": 123,
+        }
+    }
+
+    result = tool(**args)
```
TomeHirata, line 174: Q: What are the types we officially support as tool-callable for ReAct? Python primitives + Pydantic?

tests/primitives/test_tool.py
chenmoneygithub: good call!

tests/primitives/test_tool.py
chenmoneygithub: yes, everything supported by typing + pydantic

TomeHirata (approved):
LGTM

PR #8096, opened 2025-04-23
Allow overwrite max_iter in ReAct

TomeHirata:
Allow passing `max_iters` in the forward method to overwrite the default.


chenmoneygithub (approved):
LGTM!

PR #8073, opened 2025-04-16
Remove community retrievers

TomeHirata:
## Context
As part of the strategic shift of DSPy toward a reliable framework, we will stop supporting various community retrievers within the DSPy library since 2.7.0 due to the challenges of maintaining them reliably and the inactive status of their development.

Resolves #8072

## Changes
Removed the following retrievers and related doc pages. Also, the retrieve dir is unified with retrievers dir.
- AzureAISearchRM
- ChromadbRM
- ClarifaiRM
- FalkordbRM
- FaissRM
- LancedbRM
- MilvusRM
- MyScaleRM
- Neo4jRM
- QdrantRM
- RAGatouilleRM
- SnowflakeRM
- WatsonDiscovery
- YouRM

PR #8100, opened 2025-04-23
refactor: Restore QdrantRM

Anush008:
## Description

This PR restores the `QdrantRM` retriever, which was previously commented out in [#1937](https://github.com/stanfordnlp/dspy/pull/1937) due to its reliance on deprecated `dsp/*` module imports.

To resolve this, all necessary imports have been made inline. This change restores `QdrantRM`'s original functionality without issues.

### DIFF
--- a/dspy/retrieve/qdrant_rm.py
+++ b/dspy/retrieve/qdrant_rm.py
@@ -1,116 +1,185 @@
-# from collections import defaultdict
-# from typing import Optional, Union
+import abc
+from collections import defaultdict
+from typing import List, Optional, Union
 
-# import dspy
-# from dspy.dsp.modules.sentence_vectorizer import BaseSentenceVectorizer, FastEmbedVectorizer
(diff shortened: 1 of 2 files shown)


okhat:
Thank you @Anush008 ! Q: Do you think this code should live inside DSPy? It seems like it can just be an external piece of code.

Anush008:
I believe a library would be an overkill since there are several retriever implementations in https://github.com/stanfordnlp/dspy/tree/main/dspy/retrieve already. This would be another addition(restoration).

okhat:
Oh all of these are getting deleted, see then recent PR: #8073

Anush008:
Oh. Makes sense then. Thanks.

Anush008:
Hey @okhat.
We'll move the retriever to a separate `qdrant-dspy` package.

Anush008:
https://github.com/qdrant/dspy-qdrant

PR #8103, opened 2025-04-24
Add callback to async methods

TomeHirata:
While async methods are supported for major parts of DSPy, they don't call callback handlers. This PR enables callbacks to the async methods. To implement the callback wrapper in a general way both for sync and async methods, a generator function is used for common logic.

image

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -90,6 +90,7 @@ class BaseLM(ABC):
 
         return outputs
 
+    @with_callbacks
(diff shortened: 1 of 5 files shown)


okhat:
Thank you @TomeHirata ! LGTM but curious what @chenmoneygithub thinks. Feel free to merge @chenmoneygithub

chenmoneygithub (commented):
I am a little concerned with the generator usage, will do some testing on my side. Otherwise LGTM.

Issue #8101, opened 2025-04-23
[Bug] Predictor does not return logprobs


tikoehle:
### What happened?

```logprobs = True``` should return the log probabilities of the output tokens but they are missing.

This has been implemented in https://github.com/stanfordnlp/dspy/pull/1895 and looks like it worked in the past.

There is also a more recent feature request for it: https://github.com/stanfordnlp/dspy/issues/7967

### Steps to reproduce

Reproducible with latest code (2.6.18) from main branch.

```
>>> import dspy
>>> kwargs = {"logprobs": True, "temperature": 0.5, "max_tokens": 100}
>>> lm = dspy.LM("openai/gpt-4o-mini", api_key=api_key, **kwargs)
>>> dspy.configure(lm=lm)
>>> pred = dspy.Predict("question -> answer")
>>> result = pred(question="Where is the Eiffel Tower located?")
>>> print(result)
Prediction(
    answer='The Eiffel Tower is located in Paris, France.'
)
```
=> 'logprobs' are missing in Prediction.

### Root Cause

In dspy/clients/base_lm.py:

```diff
$ git diff dspy/clients/base_lm.py
diff --git a/dspy/clients/base_lm.py b/dspy/clients/base_lm.py
index c0a18123..c909850f 100644
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -49,7 +49,7 @@ class BaseLM(ABC):
         self.history = []
 
     def _process_lm_response(self, response, prompt, messages, **kwargs):
-        if kwargs.get("logprobs"):
+        if self.kwargs.get("logprobs"):
             outputs = [
                 {
                     "text": c.message.content if hasattr(c, "message") else c["text"],
```

=> Now 'logprobs' are there, but in OpenAI different format, such as containing ' ChatCompletionTokenLogprob'.

```
>>> kwargs = {"logprobs": True, "temperature": 0.5, "max_tokens": 100}
>>> lm = dspy.LM("openai/gpt-4o-mini", api_key=api_key, **kwargs)
>>> dspy.configure(lm=lm)
>>> 
>>> pred = dspy.Predict("question -> answer")
>>> result = pred(question="Where is the Eiffel Tower located?")
>>> print(result)
Prediction(
    answer='The Eiffel Tower is located in Paris, France.',
[...]

Issue #7967, opened 2025-03-16
Token Probability


rubenskx:
### What feature would you like to see?

Hi Devs,
I would like to know if there is any provision to view the confidence score of classification tasks. Also if its' any accurate. Something similar to `logprobs` in OpenAI API.

Thanks! 

### Would you like to contribute?

- [ ] Yes, I'd like to help implement this.
- [x] No, I just want to request it.

### Additional Context

OpenAI: https://cookbook.openai.com/examples/using_logprobs

chenmoneygithub:
`logprobs ` is only meant for token confidence, not the task confidence. You can add `confidence` to the DSPy signature to let the LM decide the confidence score, or send the LM's answer to a more powerful LM for judge. But in general, task-level confidence is not something reliable now.

PR #8104, opened 2025-04-24
Fix: Make sure Base LM builds on self.kwargs

tikoehle:
Fix for https://github.com/stanfordnlp/dspy/issues/8101
logprobs = True should return the log probabilities of the output tokens but they are missing.

This has been implemented in #1895 and looks like it worked in the past.
There is also a more recent feature request for it: #7967

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -49,32 +49,12 @@ class BaseLM(ABC):
         self.history = []
 
     def _process_lm_response(self, response, prompt, messages, **kwargs):
-        if self.kwargs.get("logprobs"):
-            # If logprobs is set, we need to extract the logprobs from the response
-            # and format them according to the OpenAI API response format.
-            # This is a workaround for the fact that the OpenAI API does not return
-            # logprobs in the same format as the response.
-            # We need to check if the response is a list of choices or a single choice.
+        merged_kwargs = {**self.kwargs, **kwargs}
+        if merged_kwargs.get("logprobs"):
             outputs = [
                 {
                     "text": c.message.content if hasattr(c, "message") else c["text"],
-                    "logprobs": {
-                        "content": [
-                            {
-                                "token": token_logprob.token,
-                                "bytes": token_logprob.bytes,
-                                "logprob": token_logprob.logprob,
-                                "top_logprobs": [
-                                    {
-                                        "token": top_logprob.token,
-                                        "bytes": top_logprob.bytes,
-                                        "logprob": top_logprob.logprob,
-                                    } for top_logprob in (token_logprob.top_logprobs or [])
-                                ],
-                            } for token_logprob in (c.logprobs.content if hasattr(c.logprobs, "content") else [])
-                        ],
-                        "refusal": c.logprobs.refusal if hasattr(c.logprobs, "refusal") else None,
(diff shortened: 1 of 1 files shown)


dspy/clients/base_lm.py
```diff
@@ -49,11 +49,32 @@ def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, c
...
+                                    {
+                                        "token": top_logprob.token,
+                                        "bytes": top_logprob.bytes,
+                                        "logprob": top_logprob.logprob,
+                                    } for top_logprob in (token_logprob.top_logprobs or [])
+                                ],
+                            } for token_logprob in (c.logprobs.content if hasattr(c.logprobs, "content") else [])
+                        ],
+                        "refusal": c.logprobs.refusal if hasattr(c.logprobs, "refusal") else None,
+                    } if hasattr(c, "logprobs") else c["logprobs"],
```
okhat, line 77: What motivates this very large blurb, compared to the 1-liner that was there before?

dspy/clients/base_lm.py
```diff
@@ -49,11 +49,32 @@ def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, c
         self.history = []
 
     def _process_lm_response(self, response, prompt, messages, **kwargs):
-        if kwargs.get("logprobs"):
+        if self.kwargs.get("logprobs"):
```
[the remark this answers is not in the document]
tikoehle: Yes you are right, we could merge the kwargs to ensure consider both sources
```
        merged_kwargs = {**self.kwargs, **kwargs}
        if merged_kwargs.get("logprobs"):
```

dspy/clients/base_lm.py
tikoehle: With the 1-liner I got additional outputs, like: ChoiceLogprobs and ChatCompletionTokenLogprob:
```
logprobs=ChoiceLogprobs(content=[ChatCompletionTokenLogprob(token='[[', bytes=[91, 91], ....
```
My intention was to extract the logprobs from the response so that they look like the OpenAI API response format. 
```
logprobs={'content': [{'token': '[[', 'bytes': [91, 91],  ...
```
But I am not 100% sure where ChoiceLogprobs and ChatCompletionTokenLogprob come from and if there is a better solution.

okhat:
@tikoehle I'm happy to merge this if you address the two comments above! Let's stick the 1-liner for now, feel free to extract the content in user code.

dspy/clients/base_lm.py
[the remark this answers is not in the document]
tikoehle: Done.

dspy/clients/base_lm.py
tikoehle: I keep the 1 liner. Done

tikoehle:
### Documentation
The log probabilities of the output tokens are returned in the following format and can be extracted in user code.
```
Prediction(
    answer='The Eiffel Tower is located in Paris, France.',
    logprobs=ChoiceLogprobs(content=[ChatCompletionTokenLogprob(token='[[', bytes=[91, 91], logprob=-1.2664456789934775e-06, top_logprobs=[]), ...
```

PR #8105, opened 2025-04-24
Add async callback 

chenmoneygithub:
Based on #8103, but instead of using the generator, we break down the code.

### DIFF
--- a/dspy/utils/callback.py
+++ b/dspy/utils/callback.py
@@ -316,7 +316,7 @@ def with_callbacks(fn):
     
     if inspect.iscoroutinefunction(fn):
         @functools.wraps(fn)
-        async def wapper(instance, *args, **kwargs):
+        async def wrapper(instance, *args, **kwargs):
             generator = wrapper_generator(instance, *args, **kwargs)
             next(generator)
             try:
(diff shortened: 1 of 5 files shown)


dspy/utils/callback.py
```diff
@@ -256,60 +256,96 @@ def on_evaluate_end(
...
+            except Exception as e:
+                exception = e
+                raise exception
+            finally:
+                ACTIVE_CALL_ID.set(parent_call_id)
+                _execute_end_callbacks(instance, fn, call_id, results, exception, callbacks)
+
+        return async_wrapper
+
+    else:
```
TomeHirata, line 320: nit: we could remove this else condition

dspy/utils/callback.py
```diff
@@ -256,60 +256,96 @@ def on_evaluate_end(
...
-
-        # If no callbacks are provided, just call the function
-        if not callbacks:
-            return fn(instance, *args, **kwargs)
-
-        # Generate call ID as the unique identifier for the call, this is useful for instrumentation.
-        call_id = uuid.uuid4().hex
+    """Decorator to add callback functionality to instance methods."""
 
+    def _execute_start_callbacks(instance, fn, call_id, callbacks, args, kwargs):
```
TomeHirata, line 261: nit: since they are internal methods within with_callbacks, we don't need to add "_" prefix.

TomeHirata (approved):
Left some comments, otherwise LGTM

PR #8106, opened 2025-04-24
Support async in dspy.Tool

chenmoneygithub:
Support async path in dspy.Tool

### DIFF
--- a/dspy/primitives/tool.py
+++ b/dspy/primitives/tool.py
@@ -1,3 +1,4 @@
+import asyncio
 import inspect
 from typing import Any, Callable, Optional, get_origin, get_type_hints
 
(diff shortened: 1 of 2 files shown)


dspy/primitives/tool.py
```diff
@@ -147,19 +164,15 @@ def _parse_args(self, **kwargs):
...
-
-        parsed_kwargs = self._parse_args(**kwargs)
-        return self.func(**parsed_kwargs)
+        parsed_kwargs = self._validate_and_parse_args(**kwargs)
+        result = self.func(**parsed_kwargs)
+        if asyncio.iscoroutine(result):
+            raise ValueError("You are calling `__call__` on an async tool, please use `acall` instead.")
+        return result
+
+    async def acall(self, **kwargs):
```
TomeHirata, line 173: Q: do we plan to support async tools in ReAct?

dspy/primitives/tool.py
```diff
@@ -147,19 +164,15 @@ def _parse_args(self, **kwargs):
...
-                if type_str is not None and type_str != "Any":
-                    validate(instance=instance, schema=self.args[k])
-            except ValidationError as e:
-                raise ValueError(f"Arg {k} is invalid: {e.message}")
-
-        parsed_kwargs = self._parse_args(**kwargs)
-        return self.func(**parsed_kwargs)
+        parsed_kwargs = self._validate_and_parse_args(**kwargs)
+        result = self.func(**parsed_kwargs)
+        if asyncio.iscoroutine(result):
```
TomeHirata, line 169: Shall we help auto-convert them using syncify? I think it's helpful if ReAct cal handle both sync tool and async tool together.

dspy/primitives/tool.py
chenmoneygithub: yes in a followup PR!

dspy/primitives/tool.py
chenmoneygithub: we will have an async path in ReAct in the next PR!

PR #8107, opened 2025-04-25
Add 4-bit quantization support for local fine-tuning

asad-aali:
This PR adds optional 4-bit quantization via `BitsAndBytesConfig` in `train_sft_locally` within `lm_local.py`. When `use_quantization=True` is passed in `train_kwargs`, the model is loaded in 4-bit (NF4) with double quantization, enabling memory-efficient fine-tuning of LLMs on limited hardware.


okhat:
Oh wow, super cool, thanks @asad-aali !! Tagging @Ziems

PR #8110, opened 2025-04-25
Async fixes

chenmoneygithub:
There was a problem that we didn't add `pytest-asyncio` to the deps list, which caused async tests to not run on github CI tests so we didn't detect a few issues. This PR adds the dependency and fix corresponding code and test.


pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "rich>=13.7.1",
     "numpy>=1.26.0,<2.2; python_version == '3.9'",
     "numpy>=1.26.0; python_version >= '3.10'",
+    "pytest-asyncio>=0.26.0",
```
okhat, line 48: I can merge if this is urgent(?), but should this be a dev-only dependency? It doesn't seem important for users

chenmoneygithub: ah good catch! moved

PR #8125, opened 2025-04-26
Update simba.py

brando90:
added max errors to simba and self.max_errors too for init


brando90:
we need to add import sys at the top.

brando90:
ok fixed it

brando90:
changing in favour of this PR from my fork to avoid confusion: https://github.com/stanfordnlp/dspy/pull/8126

PR #8126, opened 2025-04-26
Enhance SIMBA class by adding max_errors parameter to Parallel runner…

brando90:
```
class SIMBA(Teleprompter):
    def __init__(
        self,
        *,
        metric: Callable,
        bsize=32,
        num_candidates=6,
        max_steps=8,
        max_demos=4,
        demo_input_field_maxlen=100_000,
        num_threads=None,
        max_errors=sys.maxsize,
        ```
        
        I assume you want 10 by default, change it if you want to 10

### DIFF
--- a/dspy/teleprompt/simba.py
+++ b/dspy/teleprompt/simba.py
@@ -1,6 +1,7 @@
 import dspy
 import random
 import logging
+import sys
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


brando90:
@omkar-sh Is it possible to include this? 🙏

PR #8128, opened 2025-04-27
Intregrate MCP into ReAcT Module [WIP]

ThanabordeeN:
IntegrateTools into to facilitate the conversion of MCP tools into DSPy Tools.

In Progress:
Ensure the converted MCP tools align DSPy forward non-async functionality.

Struggling Seeking a method to maintain forward non-async functionality.


dspy/utils/MCPTools.py
```diff
@@ -0,0 +1,215 @@
...
+        # Try object attributes
+        if hasattr(tool_info, 'name') and hasattr(tool_info, 'description'):
+            return (
+                tool_info.name,
+                tool_info.description,
+                getattr(tool_info, 'inputSchema', None)
+            )
+            
+        # Try dict format
+        if isinstance(tool_info, dict):
```
TomeHirata, line 83: Why do we need to support types other than `mcp.types.Tool`?

ThanabordeeN: Okay, I will utilize the provided types in MCP.

TomeHirata: I got that point, I mean, when we receive MCP tool info as dict or string?

ThanabordeeN: Got it. The check over there is unnecessary, I'm going to remove it

okhat:
Thanks so much @ThanabordeeN ! We just merged #8130 with co-authorship by you!

PR #8129, opened 2025-04-27
Adding unit tests for dspy.retrievers.Embeddings

Krishn1412:
Added three unit tests for the embeddings.

1) test_embeddings_basic_search: Verifies that the retriever returns the correct top k relevant passages and their indices for a single query.

2) test_embeddings_forward_batch: Ensures the retriever handles batch queries correctly, returning the top k relevant passages and indices for each query.

3) test_normalization: Confirms that the embeddings are correctly normalized to have a norm close to 1 after processing.

### DIFF
--- a/tests/retrievers/test_embeddings.py
+++ b/tests/retrievers/test_embeddings.py
@@ -1,0 +1,85 @@
+import numpy as np
+import pytest
+
+from dspy.retrievers.embeddings import Embeddings
+
+@pytest.fixture
+def dummy_corpus():
+    return [
+        "The cat sat on the mat.",
+        "The dog barked at the mailman.",
+        "The sun rises in the east.",
+        "The quick brown fox jumps over the lazy dog.",
+        "An apple a day keeps the doctor away."
+    ]
+
+@pytest.fixture
+def dummy_embedder():
+    def embed(texts):
+        # Simple dummy embedder: convert each character's ASCII values and sum for each string
+        embeddings = []
+        for text in texts:
+            vec = np.array([sum(ord(c) for c in text[i::10]) for i in range(10)], dtype=np.float32)
+            embeddings.append(vec)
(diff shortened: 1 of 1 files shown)


tests/retrievers/test_embeddings.py
```diff
@@ -0,0 +1,85 @@
...
+    assert isinstance(prediction.indices, list)
+
+    assert len(prediction.passages) == 2
+    assert len(prediction.indices) == 2
+
+    for passage in prediction.passages:
+        assert isinstance(passage, str)
+        assert passage in dummy_corpus
+
+def test_embeddings_forward_batch(dummy_corpus, dummy_embedder):
```
chenmoneygithub, line 50: We can remove the test and the one below, since they are testing the private methods.

tests/retrievers/test_embeddings.py
```diff
@@ -0,0 +1,85 @@
...
+def dummy_embedder():
+    def embed(texts):
+        # Simple dummy embedder: convert each character's ASCII values and sum for each string
+        embeddings = []
+        for text in texts:
+            vec = np.array([sum(ord(c) for c in text[i::10]) for i in range(10)], dtype=np.float32)
+            embeddings.append(vec)
+        return np.stack(embeddings)
+    return embed
+
```
chenmoneygithub, line 26: Let's add a test case to test calling the embedding at multithreading setup. Behind the scene we are queuing the embedding queries to batch then call embedding models in batch, which could have issues (low chance) on high concurrency.

chenmoneygithub (commented):
Thanks for the PR! Left some comments

Krishn1412:
Hey @chenmoneygithub , can you take a look now? TIA.

chenmoneygithub (approved):
LGTM! I will do some minor cleanup and merge. Thanks for the contribution again!

PR #8130, opened 2025-04-28
Add `dspy.Tool.from_mcp_tool`

TomeHirata:
This PR introduces a util method to convert an MCP tool into a DSPy async tool. Taking over https://github.com/stanfordnlp/dspy/pull/8128, and combined with the follow-up change of async tool execution mentioned [here](https://github.com/stanfordnlp/dspy/pull/8106#discussion_r2059439157), we will enable DSPy users to use MCP tools with dspy.ReAct.
A follow-up PR will be filed for the tutorial update.


dspy/utils/__init__.py
```diff
@@ -27,4 +28,5 @@ def download(url):
     "dummy_rm",
     "StatusMessage",
     "StatusMessageProvider",
+    "convert_mcp_tool",
```
okhat, line 31: Thinking out loud: should this be `dspy.Tool.from_mcp(...)`?

PR #8131, opened 2025-04-28
Increasing test coverage for JsonAdapter

Y-1huadb:
This PR adds three additional test cases for the JsonAdapter:

+ Handling Different Input and Output Field Types: The test ensures that the JSONAdapter correctly processes various input and output field types, including nested Pydantic models and complex data structures. It verifies that the adapter handles both simple and nested fields in the expected manner when making requests to the model.
+ Passing Signature Information to the Model: This test verifies that the request sent to the language model (LM) includes the correct signature information. It checks that the input and output fields, along with their descriptions and types, are correctly embedded in the request, ensuring the LM receives the signature as intended.
+ Raising Explicit Errors on Failure: The test ensures that when parsing errors or mismatches between expected and actual field keys occur, the JSONAdapter raises a clear and explicit error. Specifically, if the model's output does not align with the expected fields defined in the signature, a ValueError is raised with an appropriate message. This helps prevent silent failures and provides clear feedback when something goes wrong during parsing.

All tests could be passed locally.

If there are any mistakes in my attempt, I sincerely apologize and appreciate you taking the time to review my pull request. I have great respect for your work. Thanks.


tests/adapters/test_json_adapter.py
```diff
@@ -105,3 +107,114 @@ async def test_json_adapter_async_call():
...
+
+    response_format = call_kwargs.get("response_format")
+    assert response_format is not None
+
+    assert "output1" in response_format.model_fields
+    
+    # Check that the output field "output2" is also in the response format
+    assert "output2" in response_format.model_fields
+
+    output2_field = response_format.model_fields["output2"]
```
chenmoneygithub, line 173: We should be able to directly compare the entire output against the expected output.

tests/adapters/test_json_adapter.py
```diff
@@ -105,3 +107,114 @@ async def test_json_adapter_async_call():
...
+    _, call_kwargs = mock_completion.call_args
+
+    request_data = call_kwargs.get("messages")[1]
+
+    assert request_data is not None
+
+    input_message_content = request_data.get('content')
+    assert input_message_content is not None
+
+    input1_json_start = input_message_content.find('[[ ## input1 ## ]]')
```
chenmoneygithub, line 150: We can use `assert "[[ ## input1 ## ]]" in input_message_content`. 

We should be able to write an `expected_prompt`, and directly compare it against the `input_message_content`.

Y-1huadb: Hi, @chenmoneygithub ,
Thanks for your suggestion! I'm currently trying to modify the code based on your comments.

However, I've run into a problem that I haven't been able to resolve. Here's the relevant part of the code:
```
def test_json_adapter_passes_different_input_and_output_fields():
    from litellm.utils import ModelResponse, Message, Choices
    class User(pydantic.BaseModel):
        id: int = pydantic.Field(description="Id of the user who asks the question")
        name: str = pydantic.Field(description="Name of the user who asks the question")
        email: str = pydantic.Field(description="Email of the user who asks the question")
    
    class Answer(pydantic.BaseModel):
        analysis: str = pydantic.Field(description="Analysis of this question")
        result: str = pydantic.Field(description="Result of this question")

    class TestSignature(dspy.Signature):
        user: User = dspy.InputField(desc="The user who asks the question")
        question: str = dspy.InputField(desc="Question the user asks")
        answer: Answer = dspy.OutputField(desc="Answer to this question")

    program = dspy.Predict(TestSignature)

    dspy.configure(lm=dspy.LM(model="openai/gpt4o"), adapter=dspy.JSONAdapter())

    with mock.patch("litellm.completion") as mock_completion:
        mock_completion.return_value = ModelResponse(
            choices=[
                Choices(
                    message=Message(
                        content="{'answer': {'analysis': 'Paris is the captial of France', 'result': 'Paris'}}"
                    )
                )
            ],
            model="openai/gpt4o"
        )
        program(user={'id': 5, 'name': 'name_test', 'email': 'email_test'}, question="What is the capital of France?")
        mock_completion.assert_called_once()
```
When I run the test, I get the following error:
```
FAILED test.py::test_json_adapter_passes_different_input_and_output_fields - AssertionError: Expected 'completion' to have been called once. Called 0 times.
```
It seems like litellm.completion is not being called as expected. Do you have any suggestions on how I can fix this issue?

chenmoneygithub: Try turning off the cache by `dspy.configure(lm=dspy.LM(model="openai/gpt4o", cache=False), adapter=dspy.JSONAdapter())`

Y-1huadb: Thank you for your suggestions. I've updated the code based on your comments. Please let me know if you feel the number of assertions is too high or if the structure could be improved further. Thanks again for your time and input!

chenmoneygithub (approved):
Thanks for fixing! Overall LGTM, I will take over the cleanup and merge it. Thanks again for the conbribution!

Y-1huadb:
I'm glad I could help. Thanks for handling the cleanup and merge!

PR #8165, opened 2025-05-03
Updating MIPROv2 Auto Settings

klopsahlong:
This PR updates the MIPROv2 auto settings based on intuition we've formed on ideal run settings from past experiments. 

The new auto settings are calculated by pinning the number of candidates we want to explore for each model in `light`, `medium`, `heavy`, modes to `n=6`, `n=12`, `n=18` respectively. To calculate the number of trials needed for proper exploration of each candidate, we use the following formula: `trials = max(c*M*log(n), c=2, 3/2*n)`, where M is the number of modules in a program. In non zero-shot settings, when both few-shot examples and instructions are being optimized, we set the number of candidate few-shot examples to `n`, and the number of candidate instructions to `n/2`.

The impacts of this change will generally be to ensure that `light` mode explores enough candidates to make an impact on performance, and to generally decrease the # of candidates explored in `heavy` mode, which we found was often unnecessarily high before.


dspy/propose/grounded_proposer.py
```diff
@@ -349,7 +349,7 @@ def propose_instructions_for_program(
 
         # Create an instruction for each predictor 
         for pred_i, predictor in enumerate(program.predictors()):
-            for demo_set_i in range(num_demos):
+            for demo_set_i in range(num_demos)[:min(N, num_demos)]:
```
klopsahlong, line 352: only create up to N new instructions, not up to `num_demos` instructions (since now num_demos != N)

dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -31,9 +31,9 @@
 MIN_MINIBATCH_SIZE = 50
 
 AUTO_RUN_SETTINGS = {
-    "light": {"num_trials": 7, "val_size": 100},
-    "medium": {"num_trials": 25, "val_size": 300},
-    "heavy": {"num_trials": 50, "val_size": 1000},
+    "light": {"n": 6, "val_size": 100},
+    "medium": {"n": 12, "val_size": 300},
+    "heavy": {"n": 18, "val_size": 1000},
```
klopsahlong, line 36: we now pin auto run settings to the number of candidates we want to evaluate in each setting, not the number of trials we're willing to spend

PR #8166, opened 2025-05-03
MIPROv2 Updates

klopsahlong:
1. Updating MIPROv2 to have the following behavior:
 - Auto is on by default (DONE)
 - All auto-dependent kwargs are None by default
 - If Auto is ON, raise error if any kwarg is set
 - If Auto is OFF, raise error if any kwarg is not set
2. Changes the `log` base 2 in the `num_trials` calculation, as was originally intended.
3. Adds a time out to the cost user message to proceed after 20 seconds if no input is received

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -10,7 +10,6 @@ import time
 import numpy as np
 import optuna
 from optuna.distributions import CategoricalDistribution
-import math
 import dspy
 from dspy.evaluate.evaluate import Evaluate
 from dspy.propose import GroundedProposer
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -3,11 +3,14 @@ import random
 import textwrap
 from collections import defaultdict
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple
+import select
+import sys
+import time
 
 import numpy as np
 import optuna
 from optuna.distributions import CategoricalDistribution
-
+import math
 import dspy
 from dspy.evaluate.evaluate import Evaluate
 from dspy.propose import GroundedProposer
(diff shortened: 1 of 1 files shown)


dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -53,10 +56,8 @@ def __init__(
         teacher_settings: Dict = {},
         max_bootstrapped_demos: int = 4,
         max_labeled_demos: int = 4,
-        auto: Optional[Literal["light", "medium", "heavy"]] = "medium",
-        num_candidates: int = 10,
-        num_fewshot_candidates: Optional[int] = None,
-        num_instruct_candidates: Optional[int] = None,
+        auto: Optional[Literal["light", "medium", "heavy"]] = "light",
+        num_candidates: Optional[int] = None,
```
klopsahlong, line 60: Setting `auto` back to `light` as default & making `num_candidates=None` by default. I also removed setting `num_fewshot & num_instruct_candidates` from `init`, since this overcomplicates the logic of the error message below and was not available to users before anyways.

dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -204,6 +219,15 @@ def _set_random_seeds(self, seed):
         self.rng = random.Random(seed)
         np.random.seed(seed)
 
+    def _set_num_trials_from_num_candidates(self, program, zeroshot_opt, num_candidates):
+        num_vars = len(program.predictors())
+        if not zeroshot_opt:
+            num_vars *= 2  # Account for few-shot examples + instruction variables
+        # Trials = MAX(c*M*log(N), c=2, 3/2*N)
+        num_trials = int(max(2 * num_vars * np.log2(num_candidates), 1.5 * num_candidates))
+
```
klopsahlong, line 228: Note: this both refactors the num_trials calculator as a helper function, so that we can use it in the error message as well, AND changes `np.log` --> `np.log2`.

PR #8170, opened 2025-05-05
Add MariaDB vector store support

kirillermolov:
In PR a retriever based on MariaDB vector database is implemented.


kirillermolov:
oh, didn't noticed that no more support for community retrievers https://github.com/stanfordnlp/dspy/pull/8073

PR #8171, opened 2025-05-05
Add GRPO Optimizer to DSPy

Ziems:
Authored by @LakshyAAAgrawal @dilarasoylu @Harryllh


glesperance:
GRPO landing in DSPy.... Okkkkay 🔥

dspy/adapters/base.py
```diff
@@ -28,8 +28,15 @@ def _call_post_process(self, outputs: list[dict[str, Any]], signature: Type[Sign
 
             if isinstance(output, dict):
                 output, output_logprobs = output["text"], output["logprobs"]
-
-            value = self.parse(signature, output)
+            
+            try:
+                value = self.parse(signature, output)
+            except ValueError as e:
+                if len(e.args) == 3 and isinstance(e.args[1], int) and isinstance(e.args[2], int) and e.args[2] > 0:
```
chenmoneygithub, line 35: curious - Is this error formatting related to GRPO?

dspy/adapters/json_adapter.py
```diff
@@ -62,10 +63,13 @@ def __call__(
             lm_kwargs["response_format"] = structured_output_model
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
         except Exception as e:
-            logger.warning(f"Failed to use structured output format. Falling back to JSON mode. Error: {e}")
+            if dspy.settings.log_json_fallbacks:
```
chenmoneygithub, line 66: Also this one - is this related to GRPO? Instead of adding a global setting, we can switch to logger.debug to decrease the level, but I want to clarify the reason for this change.

dspy/adapters/json_adapter.py
```diff
@@ -62,10 +63,13 @@ def __call__(
...
+                logger.warning(f"Failed to use structured output format. Falling back to JSON mode. Error: {e}")
             try:
                 lm_kwargs["response_format"] = {"type": "json_object"}
                 return super().__call__(lm, lm_kwargs, signature, demos, inputs)
+            except ValueError as ve:
+                raise ve
```
chenmoneygithub, line 72: same here

dspy/clients/provider.py
```diff
@@ -35,10 +35,47 @@ def status(self):
         raise NotImplementedError
 
 
+
+class ReinforceJob:
```
chenmoneygithub, line 39: Do we need this new abstract class? Openai has the same abstraction for SFT and RL, and from the perspective of DSPy, RL and SFT should expose same set of model management APIs. The special method here is `step`, which seems to be a special term in Arbor to issue one GRPO step. I would suggest remove this abstract class and stick with TrainingJob + customization.

dspy/clients/provider.py
```diff
@@ -35,10 +35,47 @@ def status(self):
...
+        raise NotImplementedError
+
+    def status(self):
+        raise NotImplementedError
+
+
 class Provider:
     def __init__(self):
         self.finetunable = False
+        self.reinforceable = False
```
chenmoneygithub, line 76: nit: `self.support_rl_finetuning`

dspy/clients/utils_finetune.py
```diff
@@ -23,6 +18,31 @@ class TrainingStatus(str, Enum):
...
 
 
+class TrainDataFormat(str, Enum):
+    CHAT = "chat"
+    COMPLETION = "completion"
+    GRPO_CHAT = "grpo_chat"
+
+
+class Message(TypedDict):
+    role: Union[Literal["user"], Literal["assistant"], Literal["system"]]
```
chenmoneygithub, line 28: nit: `Literal["user", "assistant", "system"`]

dspy/clients/utils_finetune.py
```diff
@@ -23,6 +18,31 @@ class TrainingStatus(str, Enum):
...
+    content: str
+
+
+class GRPOChatData(TypedDict):
+    messages: List[Message]
+    completion: MessageAssistant
+    reward: float
+
+
+GRPOGroup = List[GRPOChatData]
```
chenmoneygithub, line 43: We can delete this extra class and just use `List[GRPOChatData]`

dspy/dsp/utils/settings.py
```diff
@@ -14,7 +14,6 @@
     bypass_suggest=False,
     assert_failures=0,
     suggest_failures=0,
-    experimental=False,
```
chenmoneygithub, line 17: is this intentional?

dspy/teleprompt/bootstrap_finetune.py
```diff
@@ -213,22 +225,74 @@ def bootstrap_trace_data(
...
+            return prediction.format_reward or format_failure_score
         return metric(example, prediction, trace) if metric else True
 
     def wrapped_program(**kwargs):
         with dspy.context(trace=[]):
-            return program(**kwargs), dspy.settings.trace.copy()
+            try:
+                return program(**kwargs), dspy.settings.trace.copy()
+            except ValueError as ve:
+                if len(ve.args) in [4, 6] and "Failed to parse response as per signature from original completion with input" in ve.args[0] and capture_failed_parses:
```
chenmoneygithub, line 244: This except block looks pretty hacky, is the motivation getting the failed predictions on parse errors? If so, we can make a custom exception `dspy.exception.ParseError` and capture it here.

chenmoneygithub (commented):
Left some comments for clarification, fantastic work!

dspy/adapters/base.py
LakshyAAAgrawal: Yes, we need a mechanism to communicate the raw output text from the model to the GRPO optimizer, when the errors are raised. I decided to use the ValueError args to capture the relevant things (inputs, outputs, raw text) from different locations in the stack.

dspy/dsp/utils/settings.py
Ziems: Yes Omar wants the experimental flag removed from everything

dspy/adapters/json_adapter.py
LakshyAAAgrawal: This was a temporary change since a lot of these logging messages were getting distracting during grpo debugging. Totally okay to change this to logger.debug.

However, I do believe that this shouldn't be a logger.debug message, but a logger.warning, since it is an error the user needs to know about?

dspy/adapters/json_adapter.py
LakshyAAAgrawal: This is necessary, since the other except block captures all errors, and instead raises a RuntimeError. We need to pass in the ValueError as it is to transfer all the args.

dspy/clients/utils_finetune.py
LakshyAAAgrawal: I would like to keep GRPOGroup for better readability. A lot of places in the dspy.GRPO code use nested lists. So being able to distringuish between what is a GRPOGroup, or what is just a list of chat data felt necessary to me (and certainly helped a lot during writing the code).

dspy/teleprompt/bootstrap_finetune.py
LakshyAAAgrawal: Yes, that sounds much better! I wanted to make minimal changes to the rest of the dspy, hence opted to include all as args, but totally agree this is hacky, and defining a custom exception is better.

dspy/clients/provider.py
dilarasoylu: @chenmoneygithub I would vote for keeping these separate for the time being if it's okay. The underlying difference is that one is an offline job and the other one is an online job. More generally, I think we should factor the finetuning interface to more closely match the interface we came up for the ReinforceJob. We can handle this as part of this PR, but it will require touching many other files, so I was thinking of handling this separately. Let me know if you prefer adding these changes now -- happy to elaborate more.

dspy/clients/provider.py
dilarasoylu: Yup, sounds good!

dspy/adapters/json_adapter.py
chenmoneygithub: If this is for development purpose then it's fine, I will keep this open to make sure we don't ship this settings change.

dspy/clients/provider.py
chenmoneygithub: Yes! don't worry about the change, I can handle the engineering specs after merging this PR. The comment is like a reminder for myself.

dspy/adapters/json_adapter.py
LakshyAAAgrawal: Now that there is a parseerror, this should be changed to capture and passthrough the parse error

dspy/adapters/json_adapter.py
LakshyAAAgrawal: I would say that it is best to just delete dspy.settings.log_json_fallbacks, and just log the warning as was happening before

dspy/clients/utils_finetune.py
chenmoneygithub: The problem is `GRPOGroup` doesn't have a clear meaning by the name, while `List[GRPOChatData]` is clear at its purpose. So when developers bump into the code, they need to look up the definition of GRPOGroup and keep remembering that it means `List[GRPOChatData]`, which is nonobvious and creates a maintenance burden.

No need to change the code tho! These are small engineering decisions, once you finish the work (i am still seeing new commits to the PR), I can handle these in one go.

dspy/clients/lm_local_arbor.py
```diff
@@ -0,0 +1,514 @@
...
+
+        self.lm = lm
+        self.train_kwargs = train_kwargs
+        self.provider_job_id = None
+        self.checkpoints = {}
+        self.last_checkpoint = None
+
+    def initialize(self):
+        # TODO(GRPO Team): Set provider job ID
+        num_generations = self.train_kwargs.get("num_generations")
```
LakshyAAAgrawal, line 88: I am wondering why every argument was explicitly extracted, and passed separately?

chenmoneygithub (approved):
Approved! 

We have a few followups, but can all be in separate PRs:
- Unit test on data generation so that we can verify GRPO data generation is correct automatically on every adapter/LM/other critical path's change.
- Unify the ReinforceJob and TrainingJob, these should be the same interface from the perspective of DSPy user, like what OpenAI does.
- Data generation utils should not reside in bootstrap_finetune.py, but in a utils.py file.
- Implementation of dspy.GRPO and arbor provider can be simplified for better scalability.
- Style fix for the added files.
- A high-level explanation in API documentation explaning what dspy.GRPO does, and what's a good use case for it over other optimizers.

PR #8183, opened 2025-05-07
Make it easier to do sync streaming

chenmoneygithub:
Introduce a bool flag `async_streaming` to `dspy.streamify` so that users can switch easily between sync and async streaming.

### DIFF
--- a/docs/docs/tutorials/streaming/index.md
+++ b/docs/docs/tutorials/streaming/index.md
@@ -393,3 +393,57 @@ Final output:  Prediction(
     sum='9'
 )
 ```
+
+## Synchronous Streaming
+
+By default calling a streamified DSPy program produces an async generator. In order to get back
+a sync generator, you can set the flag `async_streaming=False`:
+
+```
+import os
+
+import dspy
+
+os.environ["OPENAI_API_KEY"] = "your_api_key"
+
+dspy.configure(lm=dspy.LM("openai/gpt-4o-mini"))
+
+predict = dspy.Predict("question->answer")
+
+# Enable streaming for the 'answer' field
+stream_predict = dspy.streamify(
+    predict,
+    stream_listeners=[dspy.streaming.StreamListener(signature_field_name="answer")],
+)
+```
+
+To consume the streamed output:
+
+```python
+import os
+
+import dspy
+
+os.environ["OPENAI_API_KEY"] = "your_api_key"
+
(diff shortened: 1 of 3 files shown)


docs/docs/tutorials/streaming/index.md
```diff
@@ -393,3 +393,57 @@ Final output:  Prediction(
     sum='9'
 )
 ```
+
+## Synchronous Streaming
+
+By default calling a streamified DSPy program produces an async generator. In order to get back
+a sync generator, you can set the flag `async_streaming=False`:
+
+```
```
TomeHirata, line 402:
```suggestion
```python
```

TomeHirata: Actually is this block necessary? It seems to be duplicated with the following example.

tests/streaming/test_streaming.py
```diff
@@ -249,11 +249,11 @@ def __call__(self, x: str, **kwargs):
             dspy.streaming.StreamListener(signature_field_name="judgement"),
         ],
         include_final_prediction_in_output_stream=False,
+        async_streaming=False,
```
TomeHirata, line 252: Is this test run in the CI pipeline? It seems to require `OPENAI_API_KEY`

docs/docs/tutorials/streaming/index.md
chenmoneygithub: ah thank you for catching it

tests/streaming/test_streaming.py
chenmoneygithub: it's not running in CI, but written for local checking. streaming behavior is a bit unreliable to verify with mock, and we should find a way to set up API key for github actions.

TomeHirata: I see, is it difficult to mock the LiteLLM call with a dummy generator?

TomeHirata (approved):
Approving to unblock, let's discuss how to run these tests in CI offline

tests/streaming/test_streaming.py
chenmoneygithub: Mocking streaming is messy but feasible, the more concerning part is the assumption on stream tokens can go wrong, which could lead to false positive test. But I agree with you that we can add another test based on mock. Merging this one now, will follow up with the test shortly.

PR #8190, opened 2025-05-07
Allow state management of tools

TomeHirata:
This PR allows tools to maintain states during the lifespan of each `ReAct.forward` without affecting other executions by making the copies of tool objects. Users can create a tool class that mutates instance variables. 

```python
class Counter():
    def __init__(self):
        self.count = 0

    def __call__(self, i: int):
        self.count += i
        return self.count

counter = Counter()

react = dspy.ReAct("max -> sum:int", tools=[counter])
```

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -1,4 +1,5 @@
 import logging
+from copy import deepcopy
 from typing import Any, Callable, Literal
 
 from litellm import ContextWindowExceededError
@@ -76,6 +77,7 @@ class ReAct(Module):
     def forward(self, **input_args):
         trajectory = {}
         max_iters = input_args.pop("max_iters", self.max_iters)
+        tools = deepcopy(self.tools)
         for idx in range(max_iters):
             try:
                 pred = self._call_with_potential_trajectory_truncation(self.react, trajectory, **input_args)
@@ -88,7 +90,7 @@ class ReAct(Module):
             trajectory[f"tool_args_{idx}"] = pred.next_tool_args
 
             try:
-                trajectory[f"observation_{idx}"] = self.tools[pred.next_tool_name](**pred.next_tool_args)
+                trajectory[f"observation_{idx}"] = tools[pred.next_tool_name](**pred.next_tool_args)
             except Exception as err:
                 trajectory[f"observation_{idx}"] = f"Execution error in {pred.next_tool_name}: {_fmt_exc(err)}"
 
@@ -101,6 +103,7 @@ class ReAct(Module):
     async def aforward(self, **input_args):
         trajectory = {}
         max_iters = input_args.pop("max_iters", self.max_iters)
+        tools = deepcopy(self.tools)
         for idx in range(max_iters):
             try:
                 pred = await self._async_call_with_potential_trajectory_truncation(self.react, trajectory, **input_args)
@@ -113,7 +116,7 @@ class ReAct(Module):
             trajectory[f"tool_args_{idx}"] = pred.next_tool_args
 
             try:
-                trajectory[f"observation_{idx}"] = await self.tools[pred.next_tool_name].acall(**pred.next_tool_args)
+                trajectory[f"observation_{idx}"] = await tools[pred.next_tool_name].acall(**pred.next_tool_args)
             except Exception as err:
                 trajectory[f"observation_{idx}"] = f"Execution error in {pred.next_tool_name}: {_fmt_exc(err)}"
 
@@ -211,9 +214,4 @@ TOPIC 04: Default behavior when the trajectory gets too long.
 TOPIC 05: Adding more structure around how the instruction is formatted.
     * Concretely, it's now a string, so an optimizer can and does rewrite it freely.
     * An alternative would be to add more structure, such that a certain template is fixed but values are variable?
-
-
-TOPIC 06: Idiomatically allowing tools that maintain state across iterations, but not across different `forward` calls.
-    * So the tool would be newly initialized at the start of each `forward` call, but maintain state across iterations.
-    * This is pretty useful for allowing the agent to keep notes or count certain things, etc.
 """
--- a/tests/predict/test_react.py
+++ b/tests/predict/test_react.py
@@ -332,3 +332,61 @@ async def test_async_error_retry():
     for i in range(2):
         obs = traj[f"observation_{i}"]
         assert re.search(r"\btool error\b", obs), f"unexpected observation_{i!r}: {obs}"
+
+
+def test_tool_call_state_management():
+    # Create a tool class that maintains state
+    class Counter():
+        def __init__(self):
+            self.count = 0
+
+        def __call__(self, i: int):
+            self.count += i
+            return self.count
+    
+    counter = Counter()
+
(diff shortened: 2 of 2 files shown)


dspy/predict/react.py
```diff
@@ -76,6 +77,7 @@ def _format_trajectory(self, trajectory: dict[str, Any]):
     def forward(self, **input_args):
         trajectory = {}
         max_iters = input_args.pop("max_iters", self.max_iters)
+        tools = deepcopy(self.tools)
```
chenmoneygithub, line 80: we may want to wrap this in a try-except block and fall back to shallow copy in cases that deepcopy fails

chenmoneygithub (commented):
One minor comment

dspy/predict/react.py
```diff
@@ -162,6 +165,12 @@ def truncate_trajectory(self, trajectory):
             trajectory.pop(key)
 
         return trajectory
+    
+    def _copy_tools(self, tools):
+        try:
```
okhat, line 170: I don't think this is deal. Now if just one tool can't be deep copied, you get silent corruption.

First improvement would be to return a _list_ of deepcopied/copied items, with per-item try/except.

Second improvement would be to only try to deep copy tools that are not functions but are Callable objects. And then print a warning if it's not deepcopy-able.

Third improvement is to change the API slightly: NO deep copy in the first place at all. Instead, _instantiate_ / construct objects if the tool is a *Class* with `__call__`. I think this is the only way to ensure that semantics aren't mixed up. If the user wants the `forward` function to create a copy, it should just be by giving it the _class_ and it can create an object by `Class()`.

TomeHirata: Good call for minimizing the scope of copying, let me change the logic to copy tools one by one. For the third point, how do you think we can determine the correct arguments and keyword arguments for instantiation?

TomeHirata: The 1st and 2nd points are addressed

dspy/predict/react.py
```diff
@@ -162,6 +166,19 @@ def truncate_trajectory(self, trajectory):
...
+    def _copy_tools(self, tools):
+        results = tools.copy()
+        for tool_name, tool in tools.items():
+            if inspect.isfunction(tool.func):
+                results[tool_name] = tool
+            else:
+                try:
+                    results[tool_name] = deepcopy(tool)
+                except Exception as e:
+                    logger.warning(f"Failed to deepcopy tool: {tool!r}. Consider making your tool deep-copyable "
```
chenmoneygithub, line 179: I feel this could be distracting for people passing callable class as tools, which is not deep copyable but meanwhile their class is stateless. 

And for thinking - I think our goal here is making sure invoking callable classes in one forward call doesn't affect other forward calls, so instead of silently deep copying behind the scene, maybe we can expose a call time arg `tools`, where users can override the tools to use? In this way, it's clear to users that each call to `__call__`/`forward` starts with a brand new class instance.

TomeHirata: As we discussed offline, let's add the argument for additional runtime tools

dspy/predict/react.py
```diff
@@ -45,14 +43,12 @@ def __init__(self, signature, tools: list[Callable], max_iters=5):
...
-        for idx, tool in enumerate(tools.values()):
-            instr.append(f"({idx + 1}) {tool}")
-
         react_signature = (
             dspy.Signature({**signature.input_fields}, "\n".join(instr))
             .append("trajectory", dspy.InputField(), type_=str)
+            .append("tools", dspy.InputField(desc="Tools you select from when selecting the next_tool_name and its next_tool_args"), type_=list[str])
             .append("next_thought", dspy.OutputField(), type_=str)
-            .append("next_tool_name", dspy.OutputField(), type_=Literal[tuple(tools.keys())])
+            .append("next_tool_name", dspy.OutputField(), type_=str)
```
TomeHirata, line 51: this part feels a bit downgraded, but at the same time, I wouldn't dynamically update the signature at runtime.

dspy/predict/react.py
```diff
@@ -45,14 +43,12 @@ def __init__(self, signature, tools: list[Callable], max_iters=5):
             args={},
         )
 
-        for idx, tool in enumerate(tools.values()):
-            instr.append(f"({idx + 1}) {tool}")
-
         react_signature = (
             dspy.Signature({**signature.input_fields}, "\n".join(instr))
             .append("trajectory", dspy.InputField(), type_=str)
+            .append("tools", dspy.InputField(desc="Tools you select from when selecting the next_tool_name and its next_tool_args"), type_=list[str])
```
chenmoneygithub, line 49: We could maybe include more details for this field, for example:

```
dspy.InputField(
    desc=(
        "List of tools you can choose from to gather additional information for achieving the final goal. "
        "Each tool includes its name, description, and a JSON Schema describing its input parameters."
    ),
    type_=list[str],
)
```

dspy/predict/react.py
chenmoneygithub: This feels a little scary because it's hard to tell if this is going to cause regression or not.

@okhat From your experience, do you think LM can remember only to choose from provided tool without the `Literal` constrant?

dspy/predict/react.py
```diff
@@ -159,6 +162,20 @@ def truncate_trajectory(self, trajectory):
...
+        trajectory_signature = dspy.Signature(f"{', '.join(trajectory.keys())} -> x")
+        return adapter.format_user_message_content(trajectory_signature, trajectory)
+    
+    def _convert_tools(self, tools: Optional[list[Callable]]) -> dict[str, Tool]:
+        """Convert the tools to a dictionary of name -> tool."""
+        tools = [t if isinstance(t, Tool) else Tool(t) for t in tools or []]
+        return {tool.name: tool for tool in tools}
+    
+    def _format_tools_string(self, tools: dict[str, Tool]) -> list[str]:
+        """Format the tools into a list of string."""
```
chenmoneygithub, line 177: nit: plural form, strings

dspy/predict/react.py
```diff
@@ -95,12 +92,18 @@ def forward(self, **input_args):
         extract = self._call_with_potential_trajectory_truncation(self.extract, trajectory, **input_args)
         return dspy.Prediction(trajectory=trajectory, **extract)
 
-    async def aforward(self, **input_args):
+    async def aforward(self, additional_tools: Optional[list[Callable]] = None, **input_args):
         trajectory = {}
         max_iters = input_args.pop("max_iters", self.max_iters)
+        tools = self.tools | self._convert_tools(additional_tools)
```
chenmoneygithub, line 98: should we raise an error on duplicated tool instead of silent overriding since we call this `additional_tools`?

chenmoneygithub (commented):
A few minor comments, otherwise LGTM!

dspy/predict/react.py
TomeHirata: Yeah, tool names are passed to the LLM in the same call, so LLMs should have enough context. We could dynamically construct the literal at runtime, but I think it would increase the complexity so want to avoid it unless it causes regression.

dspy/predict/react.py
TomeHirata: Sounds good, added the validation

PR #8199, opened 2025-05-08
Per module lm history

chenmoneygithub:
Make it possible to do `module.inspect_history`, which gives the LM history of the module. In nesting case, the  same LM history message will be available to all the ancestor callers.

Example code:

```
import dspy


class MyProgram(dspy.Module):
    def __init__(self):
        super().__init__()
        self.cot = dspy.ChainOfThought("question -> answer")

    def forward(self, question: str, **kwargs) -> str:
        return self.cot(question=question)


dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini", cache=False), adapter=dspy.JSONAdapter())
program = MyProgram()
program(question="What is the capital of France?")

program.inspect_history()

```

It will give:

```
[2025-05-08T14:21:06.581570]

System message:

Your input fields are:
1. `question` (str)
Your output fields are:
1. `reasoning` (str)
2. `answer` (str)

Inputs will have the following structure:

[[ ## question ## ]]
{question}

Outputs will be a JSON object with the following fields.

{
  "reasoning": "{reasoning}",
  "answer": "{answer}"
}


User message:

[[ ## question ## ]]
What is the capital of France?

Respond with a JSON object in the following order of fields: `reasoning`, then `answer`.


Response:

{"reasoning":"The capital of France is a well-known fact, commonly taught in geography. Paris is recognized globally as the capital city, serving as the political, economic, and cultural center of the country.","answer":"Paris"}
```

You can also access the history by `program.cot.inspect_history`.


dspy/predict/predict.py
```diff
@@ -17,10 +17,10 @@
 
 class Predict(Module, Parameter):
     def __init__(self, signature, callbacks=None, **config):
+        super().__init__(callbacks=callbacks)
```
okhat, line 20: Btw typical custom DSPy programs don't call `super().__init__(callbacks=callbacks)`.

Will that be an issue?

dspy/primitives/program.py
```diff
@@ -20,26 +21,38 @@ def _base_init(self):
     def __init__(self, callbacks=None):
         self.callbacks = callbacks or []
         self._compiled = False
+        # LM calling history of the module.
+        self.history = []
```
okhat, line 25: Cool! Does this work if there are multiple LMs used in the module for different sub-modules?

dspy/predict/predict.py
chenmoneygithub: Nothing will crash, but they will lose some perks from `dspy.Module`. We can follow what PyTorch does - recommending calling `super().__init__()` when customizing modules: https://docs.pytorch.org/tutorials/beginner/examples_nn/polynomial_module.html#pytorch-custom-nn-modules

PR #8201, opened 2025-05-09
Remove `LM.forward()`'s callback

TomeHirata:
Reverts stanfordnlp/dspy#8169


TomeHirata:
cc: @okhat Sorry, I fogot to share with you the conclusion here. I and @chenmoneygithub agreed that calling the lm callback twice and also found that we can show raw response using `mlflow.autolog` instead of `mlflow.dspy.autolog`.

PR #8204, opened 2025-05-10
Fix reference to undefined dsp.passages2text

kojiwakayama:
Fixes a missing import bug in GenerateInstructionGivenAttempts (instruction_proposal.py), where dsp.passages2text was referenced but not defined. This breaks the COPRO optimizer (copro_optimizer.py). Replaced with a descriptive InputField definition to restore functionality without changing behavior.


okhat:
This is great thank you @kojiwakayama , but looks like we have a conflict?

okhat:
I'd love to merge this but not urgent

isaacbmiller:
This file was removed in #8185

PR #8210, opened 2025-05-13
support for custom input types in DSPy signatures

arnavsinghvi11:
adds support for defining custom input types in signatures with unified formatting of input messages in BaseAdapter

also compatible with existing types (dspy.Image, dspy.History), refactors some hard-coded functions of dspy.Image to use the CustomType-BaseAdapter integration


chenmoneygithub:
Per our discussions offline, moving away from special case handling for DSPy type is agreed on, but we need to simplify this implementation for scalability. I will push a few commits this week.

arnavsinghvi11:
Closed via #8217

PR #8212, opened 2025-05-14
Add AdapterParseError to dspy

chenmoneygithub:
We have been throwing wildcard exception in Adapter, but in certain cases we need to be able to access the raw LM response together with intermediate parse result. To make it happen, we are introducing `dspy.utils.exceptions.AdapterParseError`, which captures:

- Adapter information
- Signature
- Raw LM response
- Intermediate parsed result (could have partial output fields from the signature)
- Predefined error message

A sample error looks like:

```
E           dspy.utils.exceptions.AdapterParseError: Adapter JSONAdapter failed to parse the LM response. 
E           
E           LM Response: {'answer1': 'Paris'} 
E           
E           Expected to find output fields in the LM response: [answer] 
E           
E           Actual output fields parsed from the LM response: []
```

And users can access the corresponding information by:

```
except AdapterParseError as e:
    lm_response = e.lm_response
    parsed_result = e.parsed_result
```

Note: We don't capture the input, because it should already be available at the code capturing the exception.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -16,6 +16,7 @@ from dspy.adapters.utils import (
 from dspy.clients.lm import LM
 from dspy.signatures.signature import Signature
 from dspy.utils.callback import BaseCallback
+from dspy.utils.exceptions import AdapterParseError
 
 field_header_pattern = re.compile(r"\[\[ ## (\w+) ## \]\]")
 
@@ -168,11 +169,19 @@ class ChatAdapter(Adapter):
                 try:
                     fields[k] = parse_value(v, signature.output_fields[k].annotation)
                 except Exception as e:
-                    raise ValueError(
-                        f"Error parsing field {k}: {e}.\n\n\t\tOn attempting to parse the value\n```\n{v}\n```"
+                    raise AdapterParseError(
+                        adapter_name="ChatAdapter",
+                        signature=signature,
+                        lm_response=completion,
+                        message=f"Failed to parse field {k} with value {v} from the LM response. Error message: {e}",
                     )
         if fields.keys() != signature.output_fields.keys():
-            raise ValueError(f"Expected {signature.output_fields.keys()} but got {fields.keys()}")
+            raise AdapterParseError(
+                adapter_name="ChatAdapter",
+                signature=signature,
+                lm_response=completion,
+                parsed_result=fields,
+            )
 
         return fields
 
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -18,6 +18,7 @@ from dspy.adapters.utils import (
 )
 from dspy.clients.lm import LM
 from dspy.signatures.signature import Signature, SignatureMeta
+from dspy.utils.exceptions import AdapterParseError
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


dspy/adapters/chat_adapter.py
```diff
@@ -168,11 +169,19 @@ def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
                 try:
                     fields[k] = parse_value(v, signature.output_fields[k].annotation)
                 except Exception as e:
-                    raise ValueError(
-                        f"Error parsing field {k}: {e}.\n\n\t\tOn attempting to parse the value\n```\n{v}\n```"
+                    raise AdapterParseError(
+                        adapter_name="ChatAdapter",
```
okhat, line 173: Should this be hard-coded like this? Can't it be inferred?

dspy/adapters/chat_adapter.py
```diff
@@ -168,11 +169,19 @@ def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
...
+                        signature=signature,
+                        lm_response=completion,
+                        message=f"Failed to parse field {k} with value {v} from the LM response. Error message: {e}",
```
okhat, line 176: Why did we change the message

dspy/adapters/chat_adapter.py
```diff
@@ -168,11 +169,19 @@ def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
...
                     )
         if fields.keys() != signature.output_fields.keys():
-            raise ValueError(f"Expected {signature.output_fields.keys()} but got {fields.keys()}")
+            raise AdapterParseError(
+                adapter_name="ChatAdapter",
+                signature=signature,
+                lm_response=completion,
+                parsed_result=fields,
```
okhat, line 183: What happened to the error message?

dspy/adapters/json_adapter.py
```diff
@@ -124,7 +129,12 @@ def parse(self, signature: Type[Signature], completion: str) -> dict[str, Any]:
         fields = json_repair.loads(completion)
 
         if not isinstance(fields, dict):
-            raise ValueError(f"Expected a JSON object but parsed a {type(fields)}")
+            raise AdapterParseError(
+                adapter_name="JSONAdapter",
+                signature=signature,
+                lm_response=completion,
+                message="LM response cannot be serialized to a JSON object.",
```
okhat, line 136: Again why are we changing the error msg

dspy/adapters/chat_adapter.py
chenmoneygithub: Yea it's the equivalent because this code only exists inside `ChatAdapter`.

dspy/adapters/chat_adapter.py
chenmoneygithub: It's generated inside the `AdapterParseError` dynamically based on the attributes. Sometimes we don't have the partial parsing, and sometimes we want to provide a custom message.

dspy/adapters/json_adapter.py
chenmoneygithub: Similar to the above one, this rewriting is for better rendering with the new error message format, and distinguish itself from parse error from other situations.

PR #8217, opened 2025-05-15
Provide a standard base class for creating custom Signature field type

chenmoneygithub:
Modifying from #8210 by @arnavsinghvi11.

`dspy.BaseType` will be the base class for custom types for signature usage, including `dspy.Image`, `dspy.Audio` and so on.

`dspy.History` does not subclass from `dspy.BaseType` because it needs to modify the signature and get formatted as multiturn messages.


dspy/adapters/types/base_type.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Returns:
+        A list of messages with the content split into a list of content blocks around custom types content.
+    """
+    for message in messages:
+        if message["role"] != "user":
+            # Custom type messages are only in user messages
+            continue
+
+        pattern = r"<<CUSTOM-TYPE-START-IDENTIFIER>>(.*?)<<CUSTOM-TYPE-END-IDENTIFIER>>"
```
TomeHirata, line 65: Shall we make `<CUSTOM-TYPE-START-IDENTIFIER>` a const?

dspy/adapters/types/base_type.py
```diff
@@ -0,0 +1,100 @@
...
+        result = []
+        last_end = 0
+        content = message["content"]
```
TomeHirata, line 68: Is it a fair assumption that content here is always content part list rather than a single string?

dspy/adapters/types/image.py
```diff
@@ -27,6 +28,17 @@ class Image(pydantic.BaseModel):
         "extra": "forbid",
     }
 
+    def format(self) -> Union[list[dict[str, Any]], str]:
+        try:
+            url = self.url
+            if isinstance(url, str) and "<DSPY_IMAGE_START>" in url and "<DSPY_IMAGE_END>" in url:
```
TomeHirata, line 34: nit: should we make `<DSPY_IMAGE_START>` and `<DSPY_IMAGE_END>` a const?

tests/adapters/test_chat_adapter.py
```diff
@@ -94,3 +95,93 @@ async def test_chat_adapter_async_call():
...
+    # The message should have 3 chunks of types: text, image_url, text
+    assert len(user_message_content) == 3
+    assert user_message_content[0]["type"] == "text"
+    assert user_message_content[2]["type"] == "text"
+
+    # Assert that the image is formatted correctly
+    expected_image_content = {"type": "image_url", "image_url": {"url": "https://example.com/image.jpg"}}
+    assert expected_image_content in user_message_content
+
+    # Test formatting with few-shot examples
```
TomeHirata, line 124: nit: shall we parameterize or split the test for clear test cases?

dspy/adapters/types/base_type.py
chenmoneygithub: With current adapter design, yes, we format user inputs into a string, then split around custom type. Before this PR, we only split around the image tag.

dspy/adapters/types/image.py
```diff
@@ -27,6 +28,17 @@ class Image(pydantic.BaseModel):
...
+                url = url.split("<DSPY_IMAGE_START>", 1)[-1].split("<DSPY_IMAGE_END>", 1)[0]
```
TomeHirata, line 35: nit: we can use re.search

dspy/adapters/types/image.py
chenmoneygithub: This is a condition check for legacy code's compatibility, which we can remove in 1-2 version, so i would prefer not exposing a file-level constant.

tests/adapters/test_chat_adapter.py
chenmoneygithub: Splitting sounds good to me

dspy/adapters/types/image.py
chenmoneygithub: I read the existing dspy.Image's code again, and i think we can safely delete this code chunk.

dspy/adapters/types/base_type.py
chenmoneygithub: yes!

dspy/adapters/types/image.py
TomeHirata: Yeah, I was just checking how `DSPY_IMAGE_START` is used after this code change and reached the same conclusion 😄

dspy/adapters/types/base_type.py
TomeHirata: Thanks, can we add a comment for the context?

chenmoneygithub: sg, added!

PR #8222, opened 2025-05-15
Add CodeAct module

TomeHirata:
This PR add a new module called CodeAct, which adopts the [CodeAct architecture](https://arxiv.org/abs/2402.01030) to manipulate the Python interpreter with predefined tools. 

```python
from dspy.predict import CodeAct
def factorial(n):
    if n == 1:
        return 1
    return n * factorial(n-1)

act = CodeAct("n->factorial", tools=[factorial])
act(n=5) # 120
```


dspy/predict/code_act.py
```diff
@@ -0,0 +1,116 @@
...
+        trajectory = {}
+        max_iters = kwargs.pop("max_iters", self.max_iters)
+        for idx in range(max_iters):
+            code_data = self.codeact(trajectory=trajectory, **kwargs)
+            output = None
+            code, error = self._parse_code(code_data)
+            if not error:
+                output, error = self._execute_code(code)
+
+                trajectory[f"generated_code_{idx}"] = code
```
okhat, line 106: Shouldn't this be moved before _execute_code is called? That way the exception is raised and handled after this code is preserved

dspy/predict/react.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, signature, tools: list[Callable], max_iters=5):
 
         for idx, tool in enumerate(tools.values()):
             instr.append(f"({idx + 1}) {tool}")
+        instr.append("You should pass the tool argument in JSON format")
```
okhat, line 50: Maybe revise to "When providing next_tool_args, the value inside the field must be in JSON format".

okhat: Also I'm confused about line 37, which has

```
                "When selecting the next_tool_name and its next_tool_args, the tool must be one of:\n",
```

I thought that was removed? Am I thinking of a different PR?

TomeHirata: Line 37 will be removed in https://github.com/stanfordnlp/dspy/pull/8190

dspy/predict/code_act.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Args:
+            signature (Union[str, Type[Signature]]): The signature of the module.
+            tools (list[Callable]): The tool callables to be used. CodeAct only accepts functions and not callable objects.
+            max_iters (int): The maximum number of iterations to generate the answer.
+        
+        Example:
+
+            ```python
+            from dspy.predict import CodeAct
```
chenmoneygithub, line 33: Can we follow up with a code tutorial demonstrating a use case of CodeAct?

dspy/predict/code_act.py
```diff
@@ -0,0 +1,121 @@
...
+            ```
+        """
+        self.signature = ensure_signature(signature)
+        self.max_iters = max_iters
+
+        tools = [t if isinstance(t, Tool) else Tool(t) for t in tools]
+        if any(
+            not inspect.isfunction(tool.func) for tool in tools
+        ):
+            raise ValueError("CodeAct only accepts functions and not callable objects.")
```
chenmoneygithub, line 50: curious - why do we have this constraint?

dspy/predict/code_act.py
```diff
@@ -0,0 +1,121 @@
...
+        tools = {tool.name: tool for tool in tools}
+
+        instructions = self._build_instructions(self.signature, tools)
+
+        codeact_signature = (
+            dspy.Signature({**self.signature.input_fields}, "\n".join(instructions))
+            .append("trajectory", dspy.InputField(), type_=str)
+            .append("generated_code", dspy.OutputField(desc="python code that answers the question"), type_=str)
```
chenmoneygithub, line 58: Instead of "python code that answers the question", probably say "python code that when executed, produces output relevant to answering the question"

dspy/predict/code_act.py
```diff
@@ -0,0 +1,121 @@
...
+        instructions = [f"{signature.instructions}\n"] if signature.instructions else []
+        inputs = ", ".join([f"`{k}`" for k in signature.input_fields.keys()])
+        outputs = ", ".join([f"`{k}`" for k in signature.output_fields.keys()])
+
+        instructions.append(
+            f"You are an intelligent agent. For each episode, you will receive the fields {inputs} as input.\n"
+            f"Your goal is to generate executable Python code that collects any necessary information for producing {outputs}.\n"
+            "For each iteration, you will generate a code snippet that either solves the task or progresses towards the solution.\n"
+            "Ensure any output you wish to extract from the code is printed to the console. The code should be enclosed in a fenced code block.\n"
+            f"When all information for producing the outputs ({outputs}) are available to be extracted, mark `finished=True` besides the final Python code.\n"
```
chenmoneygithub, line 83: should that be "besides the final Python code" or `finished=True` and `generated_code is not None` be mutual exclusive?

dspy/predict/code_act.py
```diff
@@ -0,0 +1,121 @@
...
+
+        for idx, tool in enumerate(tools.values()):
+            instructions.append(f"({idx + 1}) {tool}")
+        
+        return instructions
+
+    def forward(self, **kwargs):
+        # Define the tool funcitons in the interpreter
+        for tool in self.tools.values():
+            self.interpreter(inspect.getsource(tool.func))
```
chenmoneygithub, line 95: shall we move it to __init__, and only run this line for call-time tools? (the new change in ReAct)

tests/predict/test_code_act.py
```diff
@@ -0,0 +1,147 @@
...
+    }
+    assert program.interpreter.deno_process is None
+
+
+class CustomTool:
+    def __call__(self, a: float, b: float) -> float:
+        return a + b
+
+@skip_if_deno_not_available
+def test_codeact_tool_validation():
```
chenmoneygithub, line 145: We may also want to test the input flow - use a mock litellm.completion to capture the prompt, and validate our tool + code information are correctly included in the prompt.

chenmoneygithub (commented):
Awesome work!

dspy/predict/code_act.py
TomeHirata: Since it is hard to define the callable object in the python interpreter. Unlike functions, we need to define class and what parameters are passed to the object initialization, which is not visible.

dspy/predict/code_act.py
TomeHirata: Since we have an extractor at last, I designed it in a way that the code passed with `finished=True` is executed to minimize the interaction count.

dspy/predict/code_act.py
TomeHirata: Since python interpreter is shutdown every `forward`, this needs to happen in `forward` too.

dspy/predict/code_act.py
TomeHirata: Yup, I will file a follow up PR for documentation

chenmoneygithub:
I encountered an error when the function arg is of pydantic type:

```
import dspy

lm_4o_mini = dspy.LM("openai/gpt-4o-mini")
dspy.configure(
    lm=lm_4o_mini,
)


from dspy.datasets import DataLoader

kwargs = dict(fields=("claim", "supporting_facts", "hpqa_id", "num_hops"), input_keys=("claim",))
hover = DataLoader().from_huggingface(dataset_name="hover-nlp/hover", split="train", trust_remote_code=True, **kwargs)

hpqa_ids = set()
filtered_hover = []
for x in hover:
    if x["num_hops"] == 3 and x["hpqa_id"] not in hpqa_ids:
        hpqa_ids.add(x["hpqa_id"])
        filtered_hover.append(
            dspy.Example(claim=x.claim, titles=list(set([y["key"] for y in x.supporting_facts]))).with_inputs("claim")
        )
hover = filtered_hover

trainset, devset, testset = hover[:100], hover[100:150], hover[650:]

example = trainset[0]

print("Claim:", example.claim)
print("Pages that must be retrieved:", example.titles)

DOCS = {}


class SearchInput(BaseModel):
    query: str


def search(query: str, k: int) -> list[str]:
    results = dspy.ColBERTv2(url="http://20.102.90.50:2017/wiki17_abstracts")(query, k=k)
    results = [x["text"] for x in results]

    for result in results:
        title, text = result.split(" | ", 1)
        DOCS[title] = text

    return results


def search_wikipedia(query: SearchInput) -> list[str]:
    """Returns top-5 results and then the titles of the top-5 to top-30 results."""

    topK = search(query.query, 30)
    titles, topK = [f"`{x.split(' | ')[0]}`" for x in topK[5:30]], topK[:5]
    return topK + [f"Other retrieved pages have titles: {', '.join(titles)}."]


def lookup_wikipedia(title: str) -> str:
    """Returns the text of the Wikipedia page, if it exists."""

    if title in DOCS:
        return DOCS[title]

    results = [x for x in search(title, 10) if x.startswith(title + " | ")]
    if not results:
        return f"No Wikipedia page found for title: {title}"
    return results[0]


instructions = "Find all Wikipedia titles relevant to verifying (or refuting) the claim."
signature = dspy.Signature("claim -> titles: list[str]", instructions)

tools = [dspy.Tool(search_wikipedia), dspy.Tool(lookup_wikipedia)]

codeact = dspy.CodeAct(signature, tools=tools, max_iters=20)

output = codeact(claim="David Gregory was born in 1625.")
```

Error:

```
dspy.primitives.python_interpreter.InterpreterError: NameError: ["name 'SearchInput' is not defined"]
```

Looks like we also need to register the custom types to the interpreter.

TomeHirata:
@chenmoneygithub Thanks for trying this out. That is a known limitation where Tool functions can only depend on builtin methods. In other words, tools should be self-contained. We could re-register all variables in `locals()`, but I worry that this causes unnecessary latency or failures of the re-registration due to external packages or definition ordering. Since LLM can generate code rather than json to pass args to the tools, I don't think this is the blocking limitation. The [langgraph-codeact](https://github.com/langchain-ai/langgraph-codeact/tree/main) also has this limitation.

PR #8232, opened 2025-05-16
feat(dspy): custom type resolution in Signatures

isaacbmiller:
## 📝 Changes Description
Added support for custom type resolution in Signatures. Allows users to reference their own types in string-based signature definitions.

Added custom_types parameter to make_signature. Uses caller's frame to detect custom types.

## ⚠️ Warnings


okhat:
WHOA soo cool @isaacbmiller !!!

PR #8237, opened 2025-05-17
fix(tests): use temp file for dataset

estsauver:
Uses temporary files for test data and removes the extraneous prints.


okhat:
Thank you @estsauver !!

PR #8238, opened 2025-05-18
Revert uv.lock changes

estsauver:
I *think* there's an error here where the entries are not being merged symmetrically. Before the fix, the function only handled keys already present in usage_entry2. Any key that existed solely in usage_entry1 was ignored.

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -43,6 +43,8 @@ class UsageTracker:
                 else:
                     result[k] = result[k] or 0
                     result[k] += v if v else 0
+            else:
+                result[k] = dict(v) if isinstance(v, dict) else v
         return result
 
     def add_usage(self, lm: str, usage_entry: dict):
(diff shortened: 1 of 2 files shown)


dspy/utils/usage_tracker.py
```diff
@@ -43,6 +43,8 @@ def _merge_usage_entries(self, usage_entry1, usage_entry2) -> dict[str, dict[str
                 else:
                     result[k] = result[k] or 0
                     result[k] += v if v else 0
+            else:
+                result[k] = dict(v) if isinstance(v, dict) else v
```
chenmoneygithub, line 47: probably use the following code to reduce nesting:

```
        result = dict(usage_entry2)
        for k, v in usage_entry1.items():
            current_v = result.get(k)
            if isinstance(v, dict):
            else:
                result[k] = current_v or 0
                result[k] += v if v else 0
        return result
```

chenmoneygithub (commented):
Nice catch, and thanks for the PR! Proposed a simpler way to reduce nesting level

dspy/utils/usage_tracker.py
estsauver: Nice edit

chenmoneygithub (approved):

PR #8239, opened 2025-05-18
List root types

estsauver:
[feat(json_adapter): Enhance parsing for single-field outputs & Pydantic v1/v2 compat

This commit introduces several improvements to the JSONAdapter:

1.  **Improved Parsing for Single Output Fields:**
    The `parse()` method now correctly handles scenarios where a Language Model
    (LM) returns a raw JSON array when the DSPy Signature expects only a single output field of a list type.
    Previously, this could lead to parsing errors if the signature didn't wrap the
    single output in a JSON object. 

2.  **Pydantic v1/v2 Compatibility for Structured Outputs:**
    The `_get_structured_outputs_response_format()` method has been updated
    to dynamically use Pydantic's `ConfigDict` for v2.x and the older
    `BaseConfig` for v1.x. This ensures that the structured output model
    generation (used with OpenAI's function calling/tool use features)
    is compatible across different Pydantic versions by correctly setting
    `extra="forbid"`.

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import traceback
 from typing import Any, Dict, Type, get_origin
 
 import json_repair
@@ -60,7 +59,6 @@ class JSONAdapter(ChatAdapter):
             lm_kwargs["response_format"] = structured_output_model
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
         except Exception as e:
-            logger.debug(traceback.format_exc(e))
             logger.warning("Failed to use structured output format, falling back to JSON mode.")
             try:
                 lm_kwargs["response_format"] = {"type": "json_object"}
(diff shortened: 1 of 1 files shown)


dspy/adapters/json_adapter.py
```diff
@@ -122,19 +118,22 @@ def format_assistant_message_content(
...
         fields = json_repair.loads(completion)
-
         if not isinstance(fields, dict):
-            raise AdapterParseError(
-                adapter_name="JSONAdapter",
-                signature=signature,
-                lm_response=completion,
-                message="LM response cannot be serialized to a JSON object.",
-            )
+            if isinstance(fields, list) and len(signature.output_fields) == 1:
```
okhat, line 127: Isn't this only applicable if the (single) output field has type `list` ?

okhat: Or are you saying basically let's just rely on the validation below to check anyway?

estsauver: I think so, maybe it would be better to run the check on output fields directly instead of on the "fields" variable.

okhat: OK so I think this is a great PR but it may need to be more general

estsauver: Just trying to understand what you're proposing: The idea would be to do a check on the output_fields  and see what type it is and try to parse it as that type?

PR #8242, opened 2025-05-20
Support `dspy.Tool` as input field type and `dspy.ToolCall` as output field type

chenmoneygithub:
The main goal is making `dspy.Tool` a valid input type with special format so that users can easily use `dspy.Tool` without using `dspy.ReAct`. `dspy.ToolCall` is just a thing wrapper over tool name and tool args to simplify the output field definition involving tool calls. 

Sample usage:

```
import dspy

dspy.configure(
    lm=dspy.LM("openai/gpt-4o-mini", cache=False),
    # lm=dspy.LM("anthropic/claude-3-5-sonnet-20240620", cache=False),
)


    question: str = dspy.InputField(description="The question to answer")
    tools: list[dspy.Tool] = dspy.InputField(description="The tools to use")
    answer: str = dspy.OutputField(description="The answer to the question, if no more tool calls are needed")
    tool_call: dspy.ToolCall = dspy.OutputField(description="The tool call information, including name and arguments")


def get_weather(city: str) -> str:
    """Get the weather for a city"""
    return f"The weather in {city} is sunny"


tools = [dspy.Tool(get_weather)]
tools_dict = {t.name: t for t in tools}

predict = dspy.Predict(MySignature)

result = predict(question="What is the weather in Paris?", tools=tools)
if result.tool_call:
    print(f"Executing tool: {result.tool_call.name}")
    print(f"Tool result: {tools_dict[result.tool_call.name](**result.tool_call.args)}")

dspy.inspect_history()
```

With sample output (not including history):

```
Executing tool: get_weather
Tool result: The weather in Paris is sunny
```


chenmoneygithub:
@okhat @TomeHirata  I am going to add some adapter change to this PR, will ping you for review after that is done.

dspy/adapters/types/tool.py
```diff
@@ -181,6 +200,44 @@ def __str__(self):
         return f"{self.name}{desc} {arg_desc}"
 
 
+class ToolCalls(BaseType):
```
TomeHirata, line 203: q: Why do we make `ToolCalls` a first citizen rather than `ToolCall`? Other frameworks basically define ToolCall first ([langchain](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.tool.ToolCall.html)) and then treat tool call response as `list[ToolCall]`.

chenmoneygithub: Yes you spotted the weirdo, and this is a design after multiple edits. I started with `dspy.Tool`, but then realized we are leaving the output field in free forms. For example, users might do:

```
class QAWithMultipleToolCall(dspy.Signature):
    tool_call_1: dspy.ToolCall = dspy.OutputField()
    tool_call_2: dspy.ToolCall = dspy.OutputField()
```

while we expect them to do:
```
class QAWithToolCall(dspy.Signature):
    tool_calls: list[dspy.ToolCall] = dspy.OutputField()
```

The bad thing about `QAWithMultipleToolCall` is when we use the native function calling (see JSONAdapter in this PR), we have no clue how to write the tool calls back to the output field. With a regulation that the output field must be `dspy.ToolCalls`, in native function calling use case, we can locate the output field and write the tool call info there. 

For some more context, there is a caveat that OAI models are bad at outputting dict, so the current JSONAdapter is broken with tool calling. For a quick test, you can try using JSONAdapter + ReAct, in my testing that doesn't really work. Native function calling resolves this issue.

dspy/adapters/base.py
```diff
@@ -20,18 +27,78 @@ def __init_subclass__(cls, **kwargs) -> None:
...
-            if output_logprobs is not None:
+                text = output["text"]
+                output_logprobs = output.get("logprobs")
+                tool_calls = output.get("tool_calls")
+
+            if text:
+                value = self.parse(signature, text)
+            else:
+                value = {}
+                for field_name in signature.output_fields.keys():
```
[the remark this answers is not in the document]
chenmoneygithub: Output handling is a bit different from input handling, which we can possibly generalize.

However I would be cautious about doing it because we don't know yet if generalization makes sense here - for the `ToolCalls` case, we are reading from the LM response and write back to the output field that of type `dspy.ToolCalls` if native function calling is used, which may be completely different from the second output field we introduce.

dspy/adapters/base.py
```diff
@@ -297,6 +366,22 @@ def _get_history_field_name(self, signature: Type[Signature]) -> bool:
...
+            # Look for annotation `list[dspy.Tool]` or `dspy.Tool`
+            origin = get_origin(field.annotation)
+            if origin is list and field.annotation.__args__[0] == Tool:
+                return name
+            if field.annotation == Tool:
+                return name
+        return None
+
+    def _get_tool_call_output_field_name(self, signature: Type[Signature]) -> bool:
+        for name, field in signature.output_fields.items():
```
TomeHirata, line 380: What happens when there are multiple fields with type `ToolCalls`?

dspy/adapters/types/tool.py
TomeHirata: Thanks for the detailed response. I agree this is tricky. We could have a validation logic in `dspy.Signature` and disallow the usage of a single `dspy.ToolCall` field. Then, when we use the native function calling, how do we deal with the multiple tool call case? To my understanding, the native function calling does not provide any semantic grouping.

```
class QAWithToolCall(dspy.Signature):
    tier1_tool_calls: list[dspy.ToolCall] = dspy.OutputField()
    tier2_tool_calls: list[dspy.ToolCall] = dspy.OutputField()
```

dspy/predict/react.py
```diff
@@ -4,8 +4,8 @@
 from litellm import ContextWindowExceededError
 
 import dspy
+from dspy.adapters.types.tool import Tool
 from dspy.primitives.program import Module
-from dspy.primitives.tool import Tool
```
TomeHirata, line 8: q: do we have isort running on your local? Shall we include in our CI?

chenmoneygithub: yea I have isort locally, and I remember there are a few files we don't want to sort, we can check after merging #7885

dspy/adapters/base.py
chenmoneygithub: Then only one field will be populated with value. We can raise a warning when there are multiple ToolCalls field, I kinda doubt if users will do that though. We made it `ToolCalls` as an indicator that it has multiple `ToolCall`s.

arnavsinghvi11: had the same question for multimodal. in a single call (diff from `n`), i don't believe the models can produce multiple multimodal outputs with 2+ dspy.OutputFields (it'll raise the output [exception](https://github.com/stanfordnlp/dspy/blob/0e6856868a38efad5bb3728ce5d271776ef361cb/dspy/utils/exceptions.py#L27) we see sometimes)  

could be different for ToolCalls tho, but might be safer to give that warning globally (maybe for the select ones in types?)

TomeHirata: I agree with Arnav. For cases where users define multiple output fields with a type that can only have a single field, such as tool calls or multi-modal, can we raise an exception when the invalid signature is created rather than warning since this is a wrong usage of signature?

TomeHirata: @chenmoneygithub Since this PR is merged, can you follow up with another PR on this?

paul-tharun:
@okhat @chenmoneygithub any specific reason why the tools are passed in user message rather than system message? 

Passing them in system message makes more sense as history would be handled cleanly and as tools do not change generally.

PR #8246, opened 2025-05-20
Run tests with extra dependencies separately

TomeHirata:
Currently, the ci pipeline installs all dependencies and runs all tests at a time. But we want to confirm if core tests are passed successfully with minimum dependencies. This PR separate the test steps into two; core dependencies and extra dependencies. This is a prerequisite for the upcoming separation of large dependencies.


TomeHirata:
@chenmoneygithub This separation is necessary to test the core tests with the core dependencies of DSPy. While currently there is only one extra test, I'll move dataset and optimizer related tests into this step after removing pandas or optuna.

.github/workflows/run_tests.yml
```diff
@@ -83,6 +83,10 @@ jobs:
           args: check --fix-only
       - name: Run tests with pytest
         run: uv run -p .venv pytest tests/
+      - name: Install optional depencies
```
chenmoneygithub, line 86: typo: dependencies

.github/workflows/run_tests.yml
```diff
@@ -83,6 +83,10 @@ jobs:
...
+        run: uv sync -p .venv --extra dev --extra mcp
```
chenmoneygithub, line 87: nonblocking for this PR - should we create another group `full_test` which include all these extra deps?

TomeHirata: Let me create one quickly

.github/workflows/run_tests.yml
TomeHirata: Oh, good catch.

pyproject.toml
```diff
@@ -62,9 +62,21 @@ dev = [
     "build>=1.0.3",
     "litellm>=1.60.3; sys_platform == 'win32'",
     "litellm[proxy]>=1.60.3; sys_platform != 'win32'",
-    "mcp>=1.5.0; python_version >= '3.10'",
 ]
 mcp = ["mcp; python_version >= '3.10'"]
+full_test = [
```
chenmoneygithub, line 67: Sorry this one is on me, maybe we should make this as extras to the `dev` group? 

So something like:

```
extended_test = [
  "mcp; python_version >= '3.10'",
  "pandas",
  "datasets",
]
```

Then in the action we do `uv sync -p .venv --extra dev --extra extended_test`?

TomeHirata: I added a group called `test_extras` for adding all extras

chenmoneygithub (approved):
LGTM!

PR #8247, opened 2025-05-20
Support predicted output in DSPy

chenmoneygithub:
Support the predicted output feature: https://platform.openai.com/docs/guides/predicted-outputs. The implementation is based on litellm's predicted output: https://github.com/chenmoneygithub/dspy

To avoid the edge case that user input field has `prediction`, I am adding a check on the format of `prediction`, and only when it's a valid predicted output format, I am piping it to the LM call kwargs.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -108,6 +108,17 @@ class Predict(Module, Parameter):
         if (temperature is None or temperature <= 0.15) and num_generations > 1:
             config["temperature"] = 0.7
 
+        if "prediction" in kwargs:
+            if (
+                isinstance(kwargs["prediction"], dict)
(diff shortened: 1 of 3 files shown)


dspy/predict/predict.py
```diff
@@ -108,6 +108,17 @@ def _forward_preprocess(self, **kwargs):
         if (temperature is None or temperature <= 0.15) and num_generations > 1:
             config["temperature"] = 0.7
 
+        if "prediction" in kwargs:
+            if (
```
okhat, line 112: Happy to merge for now but:

1) Should we just check (or maybe _also_ check?) whether `prediction` is a key in the signature?

2) I suppose the cleanest way for this would have been to make it a new type and handle it in adapters 😅  But this would start to feel like too many types...

PR #8251, opened 2025-05-21
Pin json-repair for python 3.9 support

TomeHirata:
json-repair dropped the support for python 3.9 since 0.45.0. So we should pin json-repair<0.45.0 until we drop the support for python 3.9
https://github.com/mangiucugna/json_repair/issues/124

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "magicattr>=0.1.6",
     "litellm>=1.60.3",
     "diskcache>=5.6.0",
-    "json-repair>=0.30.0",
+    "json-repair>=0.30.0,<0.45.0",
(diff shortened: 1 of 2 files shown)


TomeHirata:
0.45.0 is yanked

chenmoneygithub (commented):
Thanks Tomu! Maybe we should pin different version based on python type, like what we are doing for numpy:

```
    "numpy>=1.26.0,<2.2; python_version == '3.9'",
    "numpy>=1.26.0; python_version >= '3.10'",
```

TomeHirata:
This PR is not needed since json_repair 0.45.0 is yanked

TomeHirata:
This PR is not needed since json_repair 0.45.0 is yanked

PR #8274, opened 2025-05-26
Remove pandas and datasets from core dependencies

TomeHirata:
Since DSPy 3.0, DSPy does not include dependencies for optimizers such as `pandas`, or `datasets` as its core dependencies. This is for making DSPy production-ready, allowing users to use DSPy programs without installing large dependencies during inference time. This PR reduces the file volume size of DSPy dependencies from 284 MB to 115MB.


dspy/evaluate/evaluate.py
```diff
@@ -178,12 +179,13 @@ def process_item(example):
         logger.info(f"Average Metric: {ncorrect} / {ntotal} ({round(100 * ncorrect / ntotal, 1)}%)")
 
         if display_table:
-            # Rename the 'correct' column to the name of the metric object
-            metric_name = metric.__name__ if isinstance(metric, types.FunctionType) else metric.__class__.__name__
-            # Construct a pandas DataFrame from the results
-            result_df = self._construct_result_table(results, metric_name)
+            if importlib.util.find_spec("pandas") is not None:
```
chenmoneygithub, line 182: shall we log a warning here to inform people that displaying table fails because pandas is not found?

pyproject.toml
```diff
@@ -66,6 +68,9 @@ dev = [
 ]
 test_extras = [
     "mcp; python_version >= '3.10'",
+    "datasets>=2.14.6",
```
chenmoneygithub, line 71: open for discussion - do you think it's better to have these go to `dev` or `test_extras`? My first impression is they belong to `dev` better, but not very sure

chenmoneygithub (commented):
Left 2 minor comments, otherwise LGTM!

pyproject.toml
TomeHirata: I think `test_extras` should be a combination of all extra dependencies that are installed when running `pytest`. To my understanding, `dev` usually only contains packages that help development such as linter, test libraries. So we could include them in `dev`, but including them in `test_extras` is a must

Issue #8285, opened 2025-05-27
[Feature] Allow pickling imported modules by value when saving whole programs


erandeutsch:
### What feature would you like to see?

Currently, a saved dspy program cannot be reliably loaded from a different directory if it uses an imported signature from another file.

By default, Cloudpickle does not serialize imported modules, and refers to them by reference. Due to this, a dspy program which uses an imported module which is saved in one directory, and loaded in a different directory, will not function properly. This leads to issues when transfering saved dspy programs across computers and repositories.

This has led to issues such as #8279 and #8198.

Currently, the only consistent way to use a saved program which uses an imported signature across directories is to add the module to the sys path. This isn't a great fix, and requires the script loading the dspy program to know which modules are being used beforehand.

Fortunately, cloudpickle supports pickling imported modules using `cloudpickle.register_pickle_by_value(my_module)`. We can add the option to pass a list of modules to dspy's `.save` function, and then it will register the modules for serialization before dumping the dspy program using cloudpickle.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Additional Context

A demonstration of this issue: https://github.com/erandeutsch/dspy-importing-module-bug

I have tested implementing cloudpickle.register_pickle_by_value in the dspy save function, and it resolves this issue; I will submit a pull request.

erandeutsch:
This feature is now integrated in [dspy 2.6.25](https://github.com/stanfordnlp/dspy/releases/tag/2.6.25)! For more info, see the new [Serializing Imported Modules](https://dspy.ai/tutorials/saving/#:~:text=modules_to_serialize) section in the Saving and Loading tutorial.

PR #8286, opened 2025-05-27
Support custom imported module serialization with cloudpickle

erandeutsch:
## Description 
This pull request implements the ability to pickle imported modules by value when saving whole programs.

This aims to solve issue #8285; a saved dspy program cannot be reliably loaded from a different directory if it uses an imported signature from another file. 

## Changes
A new optional `modules_to_serialize` parameter is added to the `save` method, allowing users to specify custom imported modules for serialization with cloudpickle. `modules_to_serialize` is a list of modules, which are then registered for serialization by value using `cloudpickle.register_pickle_by_value`.

For example, saving a module which uses an imported module:
```python
module.save("testModule", save_program=True, modules_to_serialize=[ExtractInfo])
```

Using this, modules can be loaded using `dspy.load` in a separate directory without any problems related to importing modules. This improves the ability to share dspy programs across computers and repositories, without any import issues.

### DIFF
--- a/docs/docs/tutorials/saving/index.md
+++ b/docs/docs/tutorials/saving/index.md
@@ -7,7 +7,7 @@ This guide demonstrates how to save and load your DSPy program. At a high level,
 
 ## State-only Saving
 
-State represents the DSPy program's internal state, including the signature, demos (few-shot examples), and other informaiton like
+State represents the DSPy program's internal state, including the signature, demos (few-shot examples), and other information like
 the `lm` to use for each `dspy.Predict` in the program. It also includes configurable attributes of other DSPy modules like
 `k` for `dspy.retrievers.Retriever`. To save the state of a program, use the `save` method and set `save_program=False`. You can
 choose to save the state to a JSON file or a pickle file. We recommend saving the state to a JSON file because it is safer and readable.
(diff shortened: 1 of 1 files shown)


dspy/primitives/module.py
```diff
@@ -198,6 +205,10 @@ def save(self, path, save_program=False):
                 path.mkdir(parents=True)
 
             try:
+                if modules_to_serialize is not None:
```
chenmoneygithub, line 208: nit: remove this if condition check and use `modules_to_serialize = modules_to_serialize or []` to reduce the nesting level.

dspy/primitives/module.py
```diff
@@ -172,6 +172,10 @@ def save(self, path, save_program=False):
         - `save_program=True`: Save the whole module to a directory via cloudpickle, which contains both the state and
             architecture of the model.
 
+        If save_program=True and modules_to_serialize are provided, it will register those modules for serialization
```
chenmoneygithub, line 175: nit: wrap variable name and code by backtick, `save_program=True` and `modules_to_serialize`...

chenmoneygithub (commented):
Thanks for the PR! Left 2 minor comments, otherwise LGTM!

dspy/primitives/module.py
erandeutsch: Thanks, I replaced the condition check with your suggestion.

dspy/primitives/module.py
erandeutsch: Thanks, I added the proper style formatting for the parameter names.

erandeutsch:
I added a section to the "Saving and Loading your DSPy program" tutorial explaining how to use the new `modules_to_serialize` parameter in the `save` method for serializing imported modules.

https://github.com/stanfordnlp/dspy/blob/c243a386b4b7da0c88192129105e4b5de4841c73/docs/docs/tutorials/saving/index.md?plain=1#L99-L129

chenmoneygithub (approved):
LGTM!

PR #8292, opened 2025-05-29
Add langchain tool support

TomeHirata:
Similar to how we support MCP, we support [Langchain Tools](https://python.langchain.com/docs/concepts/tools/) by providing a conversion method from Langchain Tools to DSPy tool. I will follow up with a tutorial for the langchain tool usage with ReAct.

```python
import dspy
from langchain_community.tools.yahoo_finance_news import YahooFinanceNewsTool
lm = dspy.LM(model='openai/gpt-4o-mini', max_tokens=10000, temperature = 1.0)
dspy.configure(lm=lm, cache=False)
react = dspy.ReAct("query -> result", tools=[dspy.Tool.from_langchain(YahooFinanceNewsTool())])
res = await react.acall(query="Any interesting news happen today?")
res.result

-> "Interesting news today includes discussions about Apple's competitive challenges against its peers in the tech sector and Elon Musk's announcement regarding the end of his role with DOGE, hinting at a shift in focus back to Tesla and its operations."
```

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -16,7 +16,7 @@ class ReAct(Module):
         """
         `tools` is either a list of functions, callable classes, or `dspy.Tool` instances.
         """
-
+        super().__init__()
         self.signature = signature = ensure_signature(signature)
         self.max_iters = max_iters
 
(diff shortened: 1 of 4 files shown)


okhat:
LGTM we should merge this right after we finish releasing 2.6.25 correctly. Once that's out, anyone can merge. Don't wait for me.

PR #8293, opened 2025-05-29
Support for Service Principal Auth for Databricks Retrieve with DSPy 

willsmithDB:
### Changes

- Added support for service principal authentication in the DatabricksRM class
- Added two new optional parameters to the constructor:
  - databricks_client_id: Databricks service principal ID
  - databricks_client_secret: Databricks service principal secret
- Updated authentication logic to use service principal credentials when provided
- Added appropriate error handling and validation
- Added clear logging statements to indicate which authentication method is being used
- Updated documentation with new parameter descriptions

### Implementation Details
The implementation prioritizes service principal authentication when both client credentials and tokens are provided. This follows the standard pattern where more specific authentication methods take precedence over more general ones.

### Authentication now follows this logic:

- If client ID and secret are provided, use service principal authentication
- Otherwise, fall back to token-based authentication (existing behavior)

### Testing
Tested with:

- Service principal authentication using client ID and secret
- Token-based authentication (ensuring backward compatibility)
- Various combinations of missing credentials to verify error handling

### Documentation
Updated the docstring to include descriptions of the new parameters and their usage.

### DIFF
--- a/dspy/retrieve/databricks_rm.py
+++ b/dspy/retrieve/databricks_rm.py
@@ -133,9 +133,13 @@ class DatabricksRM(dspy.Retrieve):
         self.databricks_endpoint = (
             databricks_endpoint if databricks_endpoint is not None else os.environ.get("DATABRICKS_HOST")
         )
-        self.databricks_client_id = databricks_client_id if databricks_client_id is not None else os.environ.get("DATABRICKS_CLIENT_ID")
+        self.databricks_client_id = (
+            databricks_client_id if databricks_client_id is not None else os.environ.get("DATABRICKS_CLIENT_ID")
+        )
         self.databricks_client_secret = (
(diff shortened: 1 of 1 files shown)


willsmithDB:
Absolutely, thank you so much for reviewing! 

Are there standards for contributing tests that you would recommend for third party retrievers? 

### SP Flow
SPFlow

### PAT Flow
PATFlow

### Invalid PAT Flow
InvalidPAT

Cheers!

chenmoneygithub:
@willsmithDB  We are actually going to remove most of the 3rd party retreivers from DSPy due to the toughness of testing and maintaining. DatabricksRM will remain for a bit while longer, and probably get moved to Databricks internal repo later. Regarding testing, since it's hard to write unit test and run on CI, we just ask for proof that the change works.

chenmoneygithub (approved):
LGTM!

PR #8294, opened 2025-05-29
Add tutorial of DSPy cache

chenmoneygithub:
Add a tutorial of DSPy cache, including:

- usage guide
- high-level mechanism explanation
- customization guide

### DIFF
--- a/docs/docs/tutorials/cache/index.md
+++ b/docs/docs/tutorials/cache/index.md
@@ -1,0 +1,185 @@
+# Use and Customize DSPy Cache
+
+In this tutorial, we will explore the design of DSPy's caching mechanism and demonstrate how to effectively use and customize it.
+
+## DSPy Cache Structure
+
+DSPy's caching system is architected in three distinct layers:
+
+1.  **In-memory cache**: Implemented using `cachetools.LRUCache`, this layer provides fast access to frequently used data.
+2.  **On-disk cache**: Leveraging `diskcache.FanoutCache`, this layer offers persistent storage for cached items.
+3.  **Prompt cache (Server-side cache)**: This layer is managed by the LLM service provider (e.g., OpenAI, Anthropic).
+
+While DSPy does not directly control the server-side prompt cache, it offers users the flexibility to enable, disable, and customize the in-memory and on-disk caches to suit their specific requirements.
+
+## Using DSPy Cache
+
+By default, both in-memory and on-disk caching are automatically enabled in DSPy. No specific action is required to start using the cache. When a cache hit occurs, you will observe a significant reduction in the module call's execution time. Furthermore, if usage tracking is enabled, the usage metrics for a cached call will be `None`.
+
+Consider the following example:
+
+```python
+import dspy
+import os
+import time
+
+os.environ["OPENAI_API_KEY"] = "{your_openai_key}"
+
+dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini"), track_usage=True)
+
+predict = dspy.Predict("question->answer")
+
+start = time.time()
+result1 = predict(question="Who is the GOAT of basketball?")
+print(f"Time elapse: {time.time() - start: 2f}\n\nTotal usage: {result1.get_lm_usage()}")
+
+start = time.time()
+result2 = predict(question="Who is the GOAT of basketball?")
+print(f"Time elapse: {time.time() - start: 2f}\n\nTotal usage: {result2.get_lm_usage()}")
+```
+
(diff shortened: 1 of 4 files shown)


docs/docs/tutorials/cache/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini"))
+
+class CustomCache(dspy.clients.Cache):
+
+    def cache_key(self, request: Dict[str, Any], ignored_args_for_cache_key: Optional[list[str]] = None) -> str:
+        messages = request.get("messages", [])
+        return sha256(ujson.dumps(messages, sort_keys=True).encode()).hexdigest()
+
+dspy.cache = CustomCache(enable_disk_cache=True, enable_memory_cache=True, disk_cache_dir=dspy.clients.DISK_CACHE_DIR)
```
TomeHirata, line 171: q: should we make this configurable through `dspy.settings.configure` for consistency?

docs/docs/tutorials/cache/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Disabling/Enabling DSPy Cache
+
+There are scenarios where you might need to disable caching, either entirely or selectively for in-memory or on-disk caches. For instance:
+
+- You require different responses for identical LM requests.
+- You lack disk write permissions and need to disable the on-disk cache.
+- You have limited memory resources and wish to disable the in-memory cache.
+
+DSPy provides the `dspy.configure_cache()` utility function for this purpose. You can use the corresponding flags to control the enabled/disabled state of each cache type:
```
TomeHirata, line 59: Same for `dspy.configure_cache()`. Wouldn't it be better if we enable users to configure cache setting through `dspy.settings.configure`?

TomeHirata:
The tutorial looks great, left some comments/questions on the cache configuration interfaces.

docs/docs/tutorials/cache/index.md
chenmoneygithub: `cache` is more like a global singleton, while `dspy.settings` are supposed to be configured for individual working threads. and btw, I feel that `dspy.settings` has been abused as a data carrier and has become too big, we may want to delete some fields?

docs/docs/tutorials/index.md
```diff
@@ -34,6 +34,7 @@ Welcome to DSPy tutorials! We've organized our tutorials into three main categor
     - [Use MCP in DSPy](/tutorials/mcp/)
     - [Output Refinement](/tutorials/output_refinement/best-of-n-and-refine/)
     - [Saving and Loading](/tutorials/saving/)
+    - [Cache](/tutorials/cache/)
```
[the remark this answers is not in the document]
chenmoneygithub: it's in the range of "Development", but I agree that [Tools, Development, and Deployment](https://dspy.ai/tutorials/core_development/) is a bit unclear. Probably "Development and Productization"?

docs/docs/tutorials/cache/index.md
TomeHirata: For my understanding, can I ask if we make cache a global setting since we believe different threads should benefit from a common cache? I'm not sure about the exact usecase, but how should users do when they want to use a separate cache strategy only inside a context? 

For dspy.settings, I agree data carrier shouldn't be accessed by `dspy.configure`. Let's discuss offline about the better separation.

docs/docs/tutorials/index.md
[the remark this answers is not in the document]
TomeHirata: Oh I see, maybe it's only me but I felt "development" a bit different from "developer guide". If this section means the "developer guide", I agree async or streaming can be included. But then, other sections, such as pages under [Build AI Programs with DSPy](https://dspy.ai/tutorials/build_ai_program/) are also kinds of developer guides. Thinking out loud, I think most content here are advanced features, so we could name it "advanced" as Autogen does (https://microsoft.github.io/autogen/stable//user-guide/agentchat-user-guide/index.html). Anyway, let's discuss offline about the tutorial organization.

TomeHirata (approved):

docs/docs/tutorials/cache/index.md
chenmoneygithub: Think about dspy.Evaluate - when we launch evaluation we are spinning up multiple working threads, but we still want to reuse the cache to save budget and speed up future queries. In most cases cache is something users shouldn't need to be aware of, while `dspy.settings` are meant to be configurable by users.

PR #8296, opened 2025-05-30
[WIP] add more permission capabilities to PythonInterpreter

arnavsinghvi11:
improving the PythonInterpreter with higher [permissions](https://docs.deno.com/runtime/fundamentals/security/#permissions)

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -2,7 +2,8 @@ import json
 import os
 import subprocess
 from types import TracebackType
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Iterable, Union
+from os import PathLike
 
 
 class InterpreterError(RuntimeError):
@@ -24,15 +25,73 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(self, deno_command: Optional[List[str]] = None, enable_read: Iterable[Union[PathLike, str]] = None, enable_env_vars: bool | Iterable[str] = False, enable_network_access: bool | Iterable[str] = False, enable_write: bool | Iterable[Union[PathLike, str]] = False) -> None:
         if isinstance(deno_command, dict):
             deno_command = None  # no-op, just a guard in case someone passes a dict
-        self.deno_command = deno_command or ["deno", "run", "--allow-read", self._get_runner_path()]
+
+        self.enable_read = enable_read
+        self.enable_env_vars = enable_env_vars
+        self.enable_network_access = enable_network_access
+        self.enable_write = enable_write
+        #TODO later on add enable_run (--allow-run) by proxying subprocess.run through Deno.run() to fix 'emscripten does not support processes' error
+
+        if deno_command:
+            self.deno_command = list(deno_command)
+        else:
+            permissions = {
+                'enable_env_vars': 'env',
+                'enable_network_access': 'net',
+                'enable_write': 'write',
+            }
+            args = ['deno', 'run', '--allow-read']
+            for attr, perm in permissions.items():
+                val = getattr(self, attr)
+                if not val:
+                    continue
+                if val is True:
+                    args.append(f'--allow-{perm}')
+                else:
+                    args.append(f"--allow-{perm}={','.join(str(x) for x in val)}")
+            args.append(self._get_runner_path())
+            self.deno_command = args
+
         self.deno_process = None
 
     def _get_runner_path(self) -> str:
         current_dir = os.path.dirname(os.path.abspath(__file__))
         return os.path.join(current_dir, "runner.js")
+
+    def _mount_files(self):
+        if getattr(self, "_mounted_files", False):
+            return
+        paths_to_mount = []
+        if self.enable_read and not isinstance(self.enable_read, bool):
+            paths_to_mount.extend(self.enable_read)
+        if self.enable_write and not isinstance(self.enable_write, bool):
+            paths_to_mount.extend(self.enable_write)
+        if not paths_to_mount:
+            return
+        for path in paths_to_mount:
+            if path in (None, ""):
+                continue
+            if not os.path.exists(path):
+                if self.enable_write and path in self.enable_write:
+                    open(path, "a").close()  
+                else:
+                    continue
+            mount_msg = json.dumps({"mount_file": str(path), "virtual_path": str(path)})
+            self.deno_process.stdin.write(mount_msg + "\n")
+            self.deno_process.stdin.flush()
+        self._mounted_files = True
+
+    def _sync_files(self):
+        if not self.enable_write or isinstance(self.enable_write, bool):
+            return
+        for path in self.enable_write:
+            sync_msg = json.dumps({"sync_file": str(path)})
+            self.deno_process.stdin.write(sync_msg + "\n")
+            self.deno_process.stdin.flush()
+
 
     def _ensure_deno_process(self) -> None:
         if self.deno_process is None or self.deno_process.poll() is not None:
@@ -43,6 +102,7 @@ class PythonInterpreter:
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     text=True,
+                    env = os.environ.copy()
                 )
             except FileNotFoundError as e:
                 install_instructions = (
@@ -87,6 +147,7 @@ class PythonInterpreter:
         variables = variables or {}
         code = self._inject_variables(code, variables)
         self._ensure_deno_process()
+        self._mount_files()
 
         # Send the code as JSON
         input_data = json.dumps({"code": code})
@@ -127,6 +188,7 @@ class PythonInterpreter:
                 raise InterpreterError(f"{error_type}: {result.get('errorArgs') or error_msg}")
 
         # If there's no error or got `FinalAnswer`, return the "output" field
+        self._sync_files()
         return result.get("output", None)
 
     def __enter__(self):
--- a/dspy/primitives/runner.js
+++ b/dspy/primitives/runner.js
@@ -5,6 +5,18 @@ import { readLines } from "https://deno.land/std@0.186.0/io/mod.ts";
 
 const pyodide = await pyodideModule.loadPyodide();
 
+
+try {
+  for (const [key, value] of Object.entries(Deno.env.toObject())) {
+    pyodide.runPython(`
+import os
+os.environ["${key}"] = """${value}"""
+    `);
+  }
+} catch (e) {
+
+}
+
 for await (const line of readLines(Deno.stdin)) {
   let input;
   try {
@@ -16,6 +28,43 @@ for await (const line of readLines(Deno.stdin)) {
     }));
     continue;
   }
+
+  if (input.mount_file) {
+      const hostPath = input.mount_file;
+      const virtualPath = input.virtual_path || hostPath;
+      try {
+          const contents = await Deno.readFile(hostPath);
+          const dirs = virtualPath.split('/').slice(1, -1);
+          let cur = '';
+          for (const d of dirs) {
+              cur += '/' + d;
+              try {
+                  pyodide.FS.mkdir(cur);
+              } catch (e) {
+                  if (!e.message.includes('File exists')) {
+                      console.log("[DEBUG] Error creating directory in Pyodide file system:", cur, "|", e.message);
+                  }
+              }
+          }
+          pyodide.FS.writeFile(virtualPath, contents);
+      } catch (e) {
+          console.log(JSON.stringify({error: "Failed to mount file: " + e.message}));
+      }
+      continue;      
+  }
+
+  if (input.sync_file) {
+      const virtualPath = input.sync_file;
+      const hostPath = virtualPath;
+      try {
+          const contents = pyodide.FS.readFile(virtualPath);
+          await Deno.writeFile(hostPath, contents);
+      } catch (e) {
+          console.log("[DEBUG] Failed to sync file:", hostPath, "|", e.message);
+      }
+      continue;
+  }
+
 
   // Expecting an object like { "code": "...", ... }
   if (typeof input !== 'object' || input === null) {
--- a/tests/primitives/test_python_interpreter.py
+++ b/tests/primitives/test_python_interpreter.py
@@ -1,8 +1,7 @@
 import random
 import shutil
-
+import os
 import pytest
-
 from dspy.primitives.python_interpreter import InterpreterError, PythonInterpreter
 
 # This test suite requires deno to be installed. Please install deno following https://docs.deno.com/runtime/getting_started/installation/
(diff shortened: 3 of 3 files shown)


tests/primitives/test_python_interpreter.py
```diff
@@ -59,5 +58,101 @@ def test_final_answer_trick():
...
+        assert result == "", "Environment variables should be inaccessible without allow-env"
+
+    with PythonInterpreter(enable_env_vars=True) as interpreter:
+        code = "import os\nresult = os.getenv('FOO_TEST_ENV')\nresult"
+        result = interpreter.execute(code)
+        assert result == "test_value", "Environment variables should be accessible with allow-env"
+
+
+def test_read_file_access_control():
+    testfile_path = os.path.abspath("tests/primitives/test_temp_file.txt")
```
chenmoneygithub, line 79: nit: use `tmp_path` fixture so that it's automatically cleaned up. Examole: https://github.com/stanfordnlp/dspy/blob/9287208b94bcf4a0295e9f8b3d9a1c5f18418029/tests/utils/test_saving.py#L38

tests/primitives/test_python_interpreter.py
```diff
@@ -59,5 +58,101 @@ def test_final_answer_trick():
...
+            f"    data = str(e)\n"
+            f"data"
+        )
+        result = interpreter.execute(code)
+        assert ("PermissionDenied" in result or "denied" in result.lower() or "no such file" in result.lower()), "Test file should not be accessible without enable_read"
+
+    os.remove(testfile_path)
+
+def test_enable_write_flag():
+    testfile_path = os.path.abspath("tests/primitives/test_temp_output.txt")
```
chenmoneygithub, line 108: same here

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(self, deno_command: Optional[List[str]] = None, enable_read: Iterable[Union[PathLike, str]] = None, enable_env_vars: bool | Iterable[str] = False, enable_network_access: bool | Iterable[str] = False, enable_write: bool | Iterable[Union[PathLike, str]] = False) -> None:
         if isinstance(deno_command, dict):
             deno_command = None  # no-op, just a guard in case someone passes a dict
-        self.deno_command = deno_command or ["deno", "run", "--allow-read", self._get_runner_path()]
+
+        self.enable_read = enable_read
```
chenmoneygithub, line 32: I looked at the code, seems only enable_read and enable_write are being used? If so, we can delete the other two flags.

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(self, deno_command: Optional[List[str]] = None, enable_read: Iterable[Union[PathLike, str]] = None, enable_env_vars: bool | Iterable[str] = False, enable_network_access: bool | Iterable[str] = False, enable_write: bool | Iterable[Union[PathLike, str]] = False) -> None:
```
chenmoneygithub, line 28: Do we need to support bool type for these flags? From the code `list` seems to be sufficient. We can default to None.

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
...
+        if self.enable_read and not isinstance(self.enable_read, bool):
+            paths_to_mount.extend(self.enable_read)
+        if self.enable_write and not isinstance(self.enable_write, bool):
+            paths_to_mount.extend(self.enable_write)
+        if not paths_to_mount:
+            return
+        for path in paths_to_mount:
+            if path in (None, ""):
+                continue
+            if not os.path.exists(path):
```
chenmoneygithub, line 77: Should we just error out here? is there a benefit of creating an empty file and mounting to pyodide?

dspy/primitives/python_interpreter.py
```diff
@@ -127,6 +193,7 @@ def execute(
                 raise InterpreterError(f"{error_type}: {result.get('errorArgs') or error_msg}")
 
         # If there's no error or got `FinalAnswer`, return the "output" field
+        self._sync_files()
```
chenmoneygithub, line 196: I am not sure if that's the right behavior to always sync the modified file back to the filesystem. for example, I am running CodeAct in parallel, and all of them rely on an original data called `student_profiles.csv`. Each CodeAct call can make internal changes as intermediate steps, but I don't want to write those back.

chenmoneygithub (commented):
Nice extension to PythonInterpreter! Left some comments. 

Btw, I am working on a pure python-based code runner, as Pyodide is not really designed for this purpose, which makes our code unnecessarily complex.

dspy/primitives/python_interpreter.py
arnavsinghvi11: [`enable_env_vars`](https://github.com/stanfordnlp/dspy/pull/8296/files#diff-8504ca30f66c4f55d651624e5af84c70b72c67f9a700e88784f35e29ef403ba5R64) and [`enable_network_access`](https://github.com/stanfordnlp/dspy/pull/8296/files#diff-8504ca30f66c4f55d651624e5af84c70b72c67f9a700e88784f35e29ef403ba5R139) are also supported 

write and read need more explicit host-sandbox integration so more changes while env and net access work more natively (env does need this [addition](https://github.com/stanfordnlp/dspy/pull/8296/files#diff-3b4155772fb433a605308cbf7790629fa336d884064c025e0a8037534a4846abR110) tho)

dspy/primitives/python_interpreter.py
arnavsinghvi11: yea changed to raise error for read access, and creating file only for writing

dspy/primitives/python_interpreter.py
arnavsinghvi11: ah in that case `student_profiles.csv` wouldn't be included in `enable_write`, only in `enable_read`. the sync here is only for files specified to be written too. any internal changes are permitted but won't be saved back once the interpreter closes

tests/primitives/test_python_interpreter.py
```diff
@@ -59,5 +58,98 @@ def test_final_answer_trick():
...
+            interpreter.execute(code)
+
+    with PythonInterpreter(enable_network_access=True) as interpreter:
+        code = (
+            "import js\n"
+            f"resp = await js.fetch({repr(test_url)})\n"
+            "resp.status"
+        )
+        result = interpreter.execute(code)
+        assert int(result) == 200, "Network access is permitted with --allow-net"
```
TomeHirata, line 155: nit: need blank line at the end of this file

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
...
+            self.deno_command = args
+
         self.deno_process = None
 
     def _get_runner_path(self) -> str:
         current_dir = os.path.dirname(os.path.abspath(__file__))
         return os.path.join(current_dir, "runner.js")
 
+    def _mount_files(self):
+        if getattr(self, "_mounted_files", False):
```
TomeHirata, line 65: Let's add `_mounted_files` in `__init__` to avoid adding new fields after initialization.

dspy/primitives/runner.js
```diff
@@ -5,6 +5,18 @@ import { readLines } from "https://deno.land/std@0.186.0/io/mod.ts";
...
+
+try {
+  for (const [key, value] of Object.entries(Deno.env.toObject())) {
+    pyodide.runPython(`
+import os
+os.environ["${key}"] = """${value}"""
+    `);
+  }
+} catch (e) {
+
```
TomeHirata, line 17: Let's log the error at least

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
...
+            return
+        for path in paths_to_mount:
+            if path in (None, ""):
+                continue
+            if not os.path.exists(path):
+                if self.enable_write and path in self.enable_write:
+                    open(path, "a").close()
+                else:
+                    raise FileNotFoundError(f"Cannot mount non-existent file: {path}")
+            virtual_path = f"/sandbox/{os.path.basename(path)}"
```
TomeHirata, line 82: q: does `/sandbox` dir exist without initialization?

dspy/primitives/python_interpreter.py
chenmoneygithub: how is it used tho? In the code I see `self.enable_env_vars = enable_env_vars`, but `self.enable_env_vars` is never used?

dspy/primitives/python_interpreter.py
arnavsinghvi11: yea its created in the pyodide [file sys](https://pyodide.org/en/stable/usage/file-system.html#mounting-a-file-system)

dspy/primitives/python_interpreter.py
chenmoneygithub: This is still dangerous IMO, we are mixing "granting the write permission" with "force a write", which are not necessarily the same thing. Also is this thread-safe? let's say I launch a dspy.Evaluate with ProgramOfThough and rely on some file, does the write cause race conditions?

dspy/primitives/python_interpreter.py
chenmoneygithub: is this resolved?

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,79 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(self, deno_command: Optional[List[str]] = None, enable_read: Iterable[Union[PathLike, str]] = None, enable_env_vars: Iterable[str] = None, enable_network_access: Iterable[str] = None, enable_write: Iterable[Union[PathLike, str]] = None) -> None:
```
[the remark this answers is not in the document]
chenmoneygithub: This is marked as resolved, but seems unresolved?

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,90 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(self, deno_command: Optional[List[str]] = None, enable_read_paths: Optional[Union[bool, Iterable[Union[PathLike, str]]]] = None, enable_env_vars: Optional[Union[bool, Iterable[str]]] = None, enable_network_access: Optional[Union[bool, Iterable[str]]] = None, enable_write_paths: Optional[Union[bool, Iterable[Union[PathLike, str]]]] = None) -> None:
```
TomeHirata: Can we add a docstring for these arguments?

dspy/primitives/python_interpreter.py
TomeHirata: I saw the doc, but didn't see it mentioning `/sandbox`. Does the pyodide file system allow the creation of files under a nonexistent directory?

dspy/primitives/python_interpreter.py
TomeHirata: BTW, ProgramOfThought or CodeAct initializes PythonInterpreter inside the object, how do we expect users to configure these settings?

dspy/primitives/python_interpreter.py
arnavsinghvi11: yup this is allowed

dspy/primitives/python_interpreter.py
arnavsinghvi11: updated

dspy/predict/code_act.py
```diff
@@ -67,8 +67,7 @@ def factorial(n):
         self.codeact = dspy.Predict(codeact_signature)
         self.extractor = dspy.ChainOfThought(extract_signature)
         # It will raises exception when dspy cannot find available deno instance by now.
-        self.interpreter = PythonInterpreter()
-
+        self.interpreter = interpreter if interpreter is not None else PythonInterpreter()
```
chenmoneygithub: you can do:

```
self.interpreter = interpreter or PythonInterpreter()
```

dspy/predict/program_of_thought.py
```diff
@@ -59,7 +60,7 @@ def __init__(self, signature: Union[str, Type[Signature]], max_iters=3):
             ),
         )
         # It will raises exception when dspy cannot find available deno instance by now.
-        self.interpreter = PythonInterpreter()
+        self.interpreter = interpreter if interpreter is not None else PythonInterpreter()
```
chenmoneygithub: same here

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
     ```
     """
 
-    def __init__(self, deno_command: Optional[List[str]] = None) -> None:
+    def __init__(
+        self,
+        deno_command: Optional[List[str]] = None,
+        enable_read_paths: Optional[Iterable[Union[PathLike, str]]] = None,
```
chenmoneygithub: actually can we just do `List[...]`? do we expect users to use other iterables?

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
...
+        enable_env_vars: Optional[Iterable[str]] = None,
+        enable_network_access: Optional[Iterable[str]] = None,
+        enable_write_paths: Optional[Iterable[Union[PathLike, str]]] = None,
```
chenmoneygithub: nit: keep `enable_read_paths` and `enable_write_paths` next to each other.

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
...
         if isinstance(deno_command, dict):
             deno_command = None  # no-op, just a guard in case someone passes a dict
-        self.deno_command = deno_command or ["deno", "run", "--allow-read", self._get_runner_path()]
+
+        self.enable_read_paths = enable_read_paths
+        self.enable_env_vars = enable_env_vars
+        self.enable_network_access = enable_network_access
+        self.enable_write_paths = enable_write_paths
+        self.sync_files = sync_files
+        #TODO later on add enable_run (--allow-run) by proxying subprocess.run through Deno.run() to fix 'emscripten does not support processes' error
```
chenmoneygithub: nit: space after `#`

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
...
+            enable_env_vars: Environment variable names to allow in the sandbox.
+            enable_network_access: Domains or IPs to allow network access in the sandbox.
+            enable_write_paths: Files or directories to allow writing to in the sandbox.
+            sync_files: If set, syncs changes within the sandbox back to original files after execution.
+        """
         if isinstance(deno_command, dict):
             deno_command = None  # no-op, just a guard in case someone passes a dict
-        self.deno_command = deno_command or ["deno", "run", "--allow-read", self._get_runner_path()]
+
+        self.enable_read_paths = enable_read_paths
```
chenmoneygithub: we can simplify the implementation by: `self.enable_read_paths = enable_read_paths or []`, then you don't need to check `if self.enable_read_paths` in the following code.

dspy/primitives/python_interpreter.py
```diff
@@ -43,6 +127,7 @@ def _ensure_deno_process(self) -> None:
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     text=True,
+                    env = os.environ.copy()
```
chenmoneygithub: nit: no space around `=` in args

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
...
+    ) -> None:
+        """
+        Args:
+            deno_command: command list to launch Deno.
+            enable_read_paths: Files or directories to allow reading from in the sandbox.
+            enable_env_vars: Environment variable names to allow in the sandbox.
+            enable_network_access: Domains or IPs to allow network access in the sandbox.
+            enable_write_paths: Files or directories to allow writing to in the sandbox.
+            sync_files: If set, syncs changes within the sandbox back to original files after execution.
+        """
```
chenmoneygithub, line 45: Not necessarily in this PR, but we should add some code examples to explain how to use these news args (also when should users use these args)

chenmoneygithub (commented):
Some minor comments on the coding style, otherwise LGTM!

dspy/primitives/python_interpreter.py
```diff
@@ -24,16 +25,99 @@ class PythonInterpreter:
...
+            return
+        paths_to_mount = []
+        if self.enable_read_paths:
+            paths_to_mount.extend(self.enable_read_paths)
+        if self.enable_write_paths:
+            paths_to_mount.extend(self.enable_write_paths)
+        if not paths_to_mount:
+            return
+        for path in paths_to_mount:
+            if path in (None, ""):
```
TomeHirata: nit: can we do `if not path:`?

PR #8303, opened 2025-06-02
Add developer guide for CodeAct

TomeHirata:
Add developer guide for CodeAct

### DIFF
--- a/docs/docs/api/modules/CodeAct.md
+++ b/docs/docs/api/modules/CodeAct.md
@@ -31,3 +31,107 @@
         separate_signature: false
         inherited_members: true
 <!-- END_API_REF -->
+
+# CodeAct
+
+CodeAct is a DSPy module that combines code generation with tool execution to solve problems. It generates Python code snippets that use provided tools and the Python standard library to accomplish tasks.
+
+## Basic Usage
+
+Here's a simple example of using CodeAct:
+
+```python
+import dspy
+from dspy.predict import CodeAct
+
+# Define a simple tool function
+def factorial(n: int) -> int:
+    """Calculate the factorial of a number."""
+    if n == 1:
+        return 1
+    return n * factorial(n-1)
+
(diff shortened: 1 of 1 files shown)


docs/docs/api/modules/CodeAct.md
```diff
@@ -31,3 +31,107 @@
...
+    return 
+
+def secret_function():
+    print("external")
+
+def passed_function():
+    secret_function()
+
+dspy.CodeAct("question -> answer", tools=[age, secret_function])
+```
```
chenmoneygithub, line 123: helper function is also not supported, worth calling out:

```
def foo():
    ...

def bar():
    foo()
```

we cannot simply register `bar ` as a tool.

TomeHirata: Can't we do this?  Like the last example, I think this should work.
```
def foo():
    ...

def bar():
    foo()

dspy.CodeAct(..., tools=[foo, bar])
```

chenmoneygithub: I think this is a better example:

```
def _helper1(self, ...)
    ...

def _helper2(self, ...)
    ...


def my_tool(self, ...)
    _helper1()
    _helper2()
```

It' hard to force people to do `tools =[my_tool, _helper1, _helper2]`

TomeHirata: Yes, that's why I explicitly added that limitation in this doc page. You need to pass all the necessary functions to CodeAct. It is not intuitive without this guide, but I believe users can see the guide and include all helper functions.

chenmoneygithub: oh I missed the `passed_function` from your code above. We should be good to go!

TomeHirata: Sorry for the confusion, I've updated the guide to use child_function and parent_function for consistency.

PR #8314, opened 2025-06-03
Fix inspect_history bug in base_lm.py

dsblank:
This was changed two weeks ago, but the global history inspect_history only takes `n`


okhat:
Thanks so much @dsblank ! Closing in favor of #8316 which has a slightly more general fix

PR #8320, opened 2025-06-03
Fix the custom type extraction in dspy.BaseType

chenmoneygithub:
Following up with #8318, this PR handles extracting custom field type from annotations like `list[MyPydanticType]`.

The unit tests passed on both on python 3.10 and 3.12. The original error results from the diff between python 3.10 and python 3.12

On python 3.12:

```
>>> isinstance(list[str], type)
False
```

but on python 3.10:

```
>>> isinstance(list[str], type)
True
```


chenmoneygithub:
Signed off offline by @okhat

PR #8327, opened 2025-06-04
Updated the DatabricksRM class to use Databricks service principals.

willsmithDB:
- Updated the base class to include the optional parameters
- Updated print statements for which auth method is invoked.
- Updated docstrings.
- Updated logic for databricks token to use oauth token with SP when using the REST API with client secret and client id
- Updated logic on the _query_via_databricks_sdk to use the SP credentials if they exist otherwise will fallback to PAT.

Screenshots of REST API with SP tests:

image

Screenshot 2025-06-04 at 10 06 33 AM

Screenshot 2025-06-04 at 10 07 49 AM

image


chenmoneygithub (commented):
Thanks for the PR! 

I feel this is going too complex with supporting all Databricks auth methods. We should move this implementation into dastabricks-ai-bridge IMO. I will discuss with the team and make a decision.

willsmithDB:
@chenmoneygithub , 

Pardon me for the extra PR. This was supposed to be included with my other PR to add the same SP support for both methods that were already present (SDK and REST API). It's the same mechanism, but inadvertently developed for the SDK before the API when they should've been added together if that makes sense. 

Thank you for your thoughts on this!

willsmithDB:
@chenmoneygithub 

I hope that you are doing well! I believe that the direction has been to gradually move away from community retrievers. 

I'll work with our team on this, but may I ask that this change gets considered for the time being as this is the completion of a prior PR (https://github.com/stanfordnlp/dspy/pull/8293) and auth using DSPy for external systems may be broken with service principles. 

Thank you!

PR #8330, opened 2025-06-04
Update mlflow port in math tutorial

beydogan:
The port 5000 is being used by some OS program on MacOS, so using this causes some 403 errors as below:

> mlflow.exceptions.MlflowException: API request to endpoint /api/2.0/mlflow/experiments/get-by-name failed with error code 403 != 200. Response body: ''

It is a bit hard to track and wasting time for no good reason. 5001 instead of 5000 should work too.


chenmoneygithub:
@TomeHirata  Do we have similar reports on the MLflow side?

TomeHirata:
The default port for MLflow server is 5000. Whether another port is used or not depends on each developer's environment. In general, any server implementation use the default port in its documentation, so we should stick to 5000.

beydogan:
Ah, i was not aware it is the default for mlflow. I'll close it

PR #8334, opened 2025-06-06
Add docstring for dspy.ReAct

chenmoneygithub:
Add docstring for dspy.ReAct.

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -1,5 +1,5 @@
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type
+from typing import TYPE_CHECKING, Any, Callable, Literal, Type
 
 from litellm import ContextWindowExceededError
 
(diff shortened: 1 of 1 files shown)


dspy/predict/react.py
```diff
@@ -10,11 +10,44 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
+    from dspy.signatures.signature import Signature
+
 
 class ReAct(Module):
-    def __init__(self, signature, tools: list[Callable], max_iters=5):
+    def __init__(self, signature: Type["Signature"], tools: list[Callable], max_iters: Optional[int] = 5):
```
TomeHirata: nit: no need to make the type of max_iters Optional since it has default value.

dspy/predict/react.py
```diff
@@ -10,11 +10,44 @@
...
+
+        dspy.configure(lm=dspy.LM("openai/gpt-4o-mini"))
+
+        def get_weather(city: str) -> str:
+            return f"The weather in {city} is sunny."
+
+        react = dspy.ReAct(signature="question->answer", tools=[get_weather])
+
+        pred = react(question="What is the weather in Tokyo?")
+
```
okhat: Hmm why isn’t this example 3 lines. All the imports (esp Tool) and API key stuff is unnecessary

chenmoneygithub: Usually I prefer to have the code example fully runnable instead of assuming users already have the prerequisite knowledge, unless the full code is extremely long. For our new users, it may take a while to figure out all the setups.

PR #8335, opened 2025-06-06
Extends lm.inspect_history to print usage and reasoning content

nguyenhoan1988:
Enhanced the `pretty_print_history` function in `lm.inspect_history` to include:

_Usage Details_: Prints response usage information when available
_Reasoning Content_: Displays reasoning content from the response message if present.


arnavsinghvi11:
Hi @nguyenhoan1988 , thanks for the PR! could this instead be supported like how we do image & audio in inspect_history rather than keeping a separate `verbose` flag? Also, we don't need to introduce new colors - all outputs should be green

PR #8356, opened 2025-06-09
fix: stream listerner strip out space in the chunk

YGao119:
issue: https://github.com/stanfordnlp/dspy/issues/8355


chenmoneygithub:
@YGao119 Thanks for the PR! 

The fix is incorrect, we don't really call `strip()` on streaming tokens: https://github.com/stanfordnlp/dspy/blob/main/dspy/streaming/streaming_listener.py#L141, the code you are modifying is how we decide if a field starts to appear in the stream.

YGao119:
Can you take a look at the issue https://github.com/stanfordnlp/dspy/issues/8355, it strips the rhs white space in the first non-field-start token, which I think should not be expected behavior.

PR #8326, opened 2025-06-04
Fix Enum prompts and add XML adapter

okhat:
## Summary
- show enum values in adapter prompts
- add lightweight XMLAdapter for simple XML prompts

## Testing
- `pytest -q` *(fails: ModuleNotFoundError: No module named 'datamodel_code_generator')*

------
https://chatgpt.com/codex/tasks/task_e_683fadf314608329b8b5237911da3f7b

PR #8358, opened 2025-06-10
Add xml adapter

chenmoneygithub:
Forking from #8326 because I don't have rebase permission in the PR. 

Merged main branch, then made some minor edits and added unit test


Goooyi:
xml seems nice since it align with claude's prompt guides lines, anyone see performance difference between chatAdapter and XMLadapter?

PR #8372, opened 2025-06-13
Fix: Unsafe Code Loading from User Input Detected in dspy/signatures/signature.py

kira-offgrid:
**Context and Purpose:**

This PR automatically remediates a security vulnerability:
- **Description:** Untrusted user input in `importlib.import_module()` function allows an attacker to load arbitrary code. Avoid dynamic values in `importlib.import_module()` or use a whitelist to prevent running untrusted code.
- **Rule ID:** python.lang.security.audit.non-literal-import.non-literal-import
- **Severity:** MEDIUM
- **File:** dspy/signatures/signature.py
- **Lines Affected:** 528 - 528

This change is necessary to protect the application from potential security risks associated with this vulnerability.

**Solution Implemented:**

The automated remediation process has applied the necessary changes to the affected code in `dspy/signatures/signature.py` to resolve the identified issue.

Please review the changes to ensure they are correct and integrate as expected.

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -17,6 +17,29 @@ For compatibility with the legacy dsp format, you can use the signature_to_templ
 
 import ast
 import importlib
+
+
+# Whitelist of allowed modules for secure importing
+ALLOWED_MODULES = {
+    'dspy.primitives',
+    'dspy.predict',
+    'dspy.retrieve',
+    'dspy.signatures',
+    'dspy.teleprompt',
+    'dspy.utils',
+    'typing',
+    'collections',
+    'functools',
+    'inspect',
+    'copy'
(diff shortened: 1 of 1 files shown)


chenmoneygithub:
@kira-offgrid  Thanks for the PR! This could actually break signature construction when using third party packages.

I don't feel this is a valid security issue though. I can imagine two scenarios that attack can happen:

- Attach a deployed DSPy program. When DSPy program is deployed, the signature should be fixed, so the attacker doesn't have access to it.
- Share a saved DSPy program via pickle with other people, while some signature contains malicious code. This is technically possible, but the vulnerability comes from pickle instead of the signature. The signature just provides one more way to attack due to pickle along with tons of other ways.

PR #8375, opened 2025-06-13
Implement CoT no-op for reasoning models

FireMasterK:
- Add an optional parameter to allow specifying whether a model is a reasoning model or not. If not specified, we to determine it using `litellm.supports_reasoning`.
- I had to upgrade to upgrade the LiteLLM library to have access to the function, as it wasn't present in the version used earlier.

Additional context: https://x.com/DSPyOSS/status/1931886536221724848

### DIFF
--- a/dspy/predict/chain_of_thought.py
+++ b/dspy/predict/chain_of_thought.py
@@ -84,3 +84,17 @@ class ChainOfThought(Module):
             _ = self.predict
             state = self.__dict__.copy()
         return state
+
+    def named_parameters(self):
+        """Override to ensure the predict property is cached and included in named parameters."""
+        # Force evaluation of the cached_property if not already done
+        # This ensures it gets stored in __dict__ and picked up by the base implementation
+        if "predict" not in self.__dict__:
+            try:
+                _ = self.predict  # This triggers the cached_property
+            except Exception:
+                # If accessing predict fails for any reason, continue without it
+                pass
+
+        # Now call the base implementation which will include the cached predict
+        return super().named_parameters()
--- a/dspy/streaming/streaming_listener.py
+++ b/dspy/streaming/streaming_listener.py
@@ -207,7 +207,10 @@ def find_predictor_for_stream_listeners(program: "Module", stream_listeners: Lis
         if listener.predict:
             predict_id_to_listener[id(listener.predict)].append(listener)
             continue
-        if listener.signature_field_name not in field_name_to_named_predictor:
+        if (
+            listener.signature_field_name not in field_name_to_named_predictor
+            or field_name_to_named_predictor[listener.signature_field_name] is None
+        ):
             raise ValueError(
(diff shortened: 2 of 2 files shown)


okhat:
This is AMAZING. Thank you @FireMasterK !!

I think we'll need to think (no pun intended) about reasoning models like Qwen3 that expose their reasoning in the string... What does LiteLLM do for that. I guess it says `False`?

Also I guess we should still return a `reasoning` field, but set it to something that indicates that the reasoning is not available.

FireMasterK:
> I think we'll need to think (no pun intended) about reasoning models like Qwen3 that expose their reasoning in the string... What does LiteLLM do for that. I guess it says `False`?

I don't think they support telling whether a model displays its CoT or not, which is unfortunate.

It also seems like LiteLLM doesn't have enough mappings for reasoning models (eg. Qwen3 models are not there), which means we can't solely rely on LiteLLM telling if a model supports reasoning or not.
Source: https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json

> Also I guess we should still return a `reasoning` field, but set it to something that indicates that the reasoning is not available.

Yes, but some model providers providers summarize the CoT, what should we do about these? I have added the parsing of `reasoning_content` from non-streaming responses. I'm not sure what should be done for reasoning content. Should I create a new class like `dspy.streaming.ReasoningStreamListener`?

dspy/predict/chain_of_thought.py
```diff
@@ -27,15 +28,42 @@ def __init__(
         """
         super().__init__()
         signature = ensure_signature(signature)
+        self.predict_reasoning = dspy.Predict(signature, **config)
```
FireMasterK: Adding this field causes a few tests to fail, any idea on how I could fix these:
```
FAILED tests/teleprompt/test_copro_optimizer.py::test_signature_optimizer_optimization_process - dspy.utils.exceptions.AdapterParseError: LM response cannot be serialized to a JSON object.
FAILED tests/teleprompt/test_copro_optimizer.py::test_signature_optimizer_statistics_tracking - dspy.utils.exceptions.AdapterParseError: LM response cannot be serialized to a JSON object.
FAILED tests/teleprompt/test_copro_optimizer.py::test_optimization_and_output_verification - dspy.utils.exceptions.AdapterParseError: LM response cannot be serialized to a JSON object.
FAILED tests/teleprompt/test_copro_optimizer.py::test_statistics_tracking_during_optimization - dspy.utils.exceptions.AdapterParseError: LM response cannot be serialized to a JSON object.
```

FireMasterK:
I fixed the serialization issues, so all tests now pass! Let me know what I should do about https://github.com/stanfordnlp/dspy/pull/8375#issuecomment-2973454795.

I'm happy to create another PR or add them in this PR. Whichever you prefer.

FireMasterK:
Closing this because of #8822

PR #8380, opened 2025-06-14
Show warning if forward is directly called

TomeHirata:
Show a warning if users directly call forward method of dspy Module


okhat:
Wow this is so cool!! Thank you @TomeHirata !

PR #8392, opened 2025-06-16
Remove dspy.Program

TomeHirata:
We previously had `dspy.Program` that is an alias of `dspy.Module`. We will consolidate into `dspy.Module` for more clarity.


dspy/primitives/base_module.py
```diff
@@ -0,0 +1,270 @@
+import copy
```
TomeHirata: This is the content of previous `dspy/primitives/module.py`. Git just messed up the diff

tests/primitives/test_base_module.py
```diff
@@ -0,0 +1,513 @@
+import asyncio
```
TomeHirata: same for this file

dspy/primitives/base_module.py
chenmoneygithub: this will delete the commit history, could you try revert this file and program.py, then use `mv` command to generate a new commit? that should correctly reflect the file moving.

TomeHirata: I temporarily used module**s**  and will file a follow-up PR to rename

chenmoneygithub (approved):
LGTM!

PR #8382, opened 2025-06-16
Fix ChainOfThoughtWithHint

Hangzhi:
## Why
Resolve https://github.com/stanfordnlp/dspy/issues/8205. The community reported that `ChainOfThoughtWithHint` is broken. By manual verification, it's indeed broken with basic usage as below 

```
class BasicQA(dspy.Signature):
    """Answer questions with short factoid answers."""
    question = dspy.InputField()
    answer = dspy.OutputField(desc="often between 1 and 5 words")

#Pass signature to ChainOfThought module
generate_answer = dspy.ChainOfThoughtWithHint(BasicQA)

# Call the predictor on a particular input alongside a hint.
question='What is the color of the sky?'
hint = "It's what you often see during a sunny day."
pred = generate_answer(question=question, hint=hint)

print(f"Question: {question}")
print(f"Predicted Answer: {pred.answer}")
```

## What 
1. Fix the module  
2. Uncomment the unit test for the ChainOfThoughtWithHint module 

## Testing 
1. Manual Verification with the `BasicQA` above 
2. Unit test

PR #8395, opened 2025-06-16
Clean up ChainOfThoughtWithHint in DSPy 3.0 

Hangzhi:
## Why 
Per suggestion from https://github.com/stanfordnlp/dspy/pull/8382#issuecomment-2975037457, clean up ChainOfThoughtWithHint in DSPy 3.0. 

## What 
Clean up `ChainOfThoughtWithHint` in DSpy 3.0. DSPy 3.0 is till in pre-release. This PR should be able to catch the 3.0 release train.


docs/docs/index.md
```diff
@@ -358,6 +358,7 @@ Given a few tens or hundreds of representative _inputs_ of your task and a _metr
         For a complete RAG example that you can run, start this [tutorial](/tutorials/rag/). It improves the quality of a RAG system over a subset of StackExchange communities by 10% relative gain.
 
     === "Optimizing weights for Classification"
+        <!-- TODO(Hangzhi): Clean example without ChainOfThoughtWithHint. The example is not runnable in the latest DSPy -->
```
chenmoneygithub: This should go with the same PR to avoid linking to dead public API.

chenmoneygithub (commented):
Thanks for the PR! We need to clean up all references to ChainOfThoughtwithHint in one go to avoid linking to dead APIs. Feel free to leave it to us if you don't have bandwidth though.

docs/docs/index.md
Hangzhi: Sounds good. I will try replace the example with ChainOfThoughtWithHint.

chenmoneygithub (approved):
LGTM! 

I pushed a commit to remove the leftover usage, no action item left. Thanks again!

PR #8396, opened 2025-06-16
Move litellm cache testing under tests/client and remove unnecessary tests

chenmoneygithub:
We keep it for now because litellm cache is still useful in certain scenarios. 

We removed the cross-session test because it's a built-in feature of `diskcache.Cache`.


tests/clients/test_litellm_cache.py
```diff
@@ -77,25 +58,6 @@ def test_lm_calls_are_cached_across_lm_instances(litellm_test_server, temporary_
     assert len(request_logs) == 3
 
 
-def test_lm_calls_are_cached_across_interpreter_sessions(litellm_test_server, temporary_populated_cache_dir):
```
TomeHirata, line 80: Q: did you move these test cases somewhere? It seems this PR only contains the removal diff.

chenmoneygithub: Yea I am removing this test case. This test case was written to test that litellm cache can be used across session, but this is unnecessary because it's essentially just `diskcache.Cache` which supports this setup natively.

TomeHirata: I see, then can you update the title and description of the PR?

chenmoneygithub: done!

TomeHirata (approved):
LGTM

PR #8397, opened 2025-06-16
Remove dspy.settings related to dspy.Assertion

chenmoneygithub:
We have officially removed dspy.Assertion, so this PR is to remove the stale configs.


TomeHirata (approved):
LGTM

PR #8400, opened 2025-06-17
Move internal_dspy outside dspy module

TomeHirata:
The dspy/.internal_dspy dir has been used for publishing dspy-ai, but this does not need to exist under the dspy bundle. This PR move the directory outside dspy module.


chenmoneygithub:
@okhat @TomeHirata  Shall we actually stop publishing to dspy-ai from 3.0? We can publish a dummy wheel to pypi, and raise an error to direct users to `pip install dspy` from 3.0. `dspy-ai` before 3.0 will continue to work.

TomeHirata:
@chenmoneygithub I think we've discussed internally and @okhat mentioned we would still publish the wheel to `dspy-ai` for now.

okhat:
Btw we really don’t want more top level folders…

TomeHirata:
@okhat How about having them under .github? I believe this type of dev/infra related codes shouldn't be included in the Python package bundle.

okhat:
Yes for moving to .github! Feel free to merge after that patch. Thank you @TomeHirata

PR #8402, opened 2025-06-17
Move dsp/metrics into evaluate/metrics

TomeHirata:
We had dsp/utils/metrics and evaluate/metrics separately and re-export some metrics under dsp/utils/metrics in evaluate/metrics. To have a clear separation of concerns, this PR moves functions in dsp/utils/metrics into evaluate/metrics.


dspy/evaluate/auto_evaluation.py
```diff
@@ -1,48 +1,50 @@
-import dspy
+from dspy.predict.chain_of_thought import ChainOfThought
```
TomeHirata, line 1: The change here is for avoiding circular imports

PR #8403, opened 2025-06-17
Rename models/ to clients/ in API doc

TomeHirata:
dspy.LM and dspy.Embedder resides under clients, but it has been categorized as Models in our documentation. This PR changes the name to be consistent.


chenmoneygithub:
@TomeHirata We actually did this by intention. The API reference has some special name setup, you can leave it as it is for now.

TomeHirata:
@chenmoneygithub Can you elaborate why we need to use "Models" here?

PR #8408, opened 2025-06-18
Simplify the RAG tutorial

chenmoneygithub:
The current RAG tutorial https://dspy.ai/tutorials/rag/ is too long, which was written at the time we lacked all the developer guides and tutorials, so the old RAG tutorial was essentially a DSPy overview through RAG. With our current documentation, we can focus on explaining RAG functionalities.

The screenshot shows the current RAG tutorial:
image

which includes many DSPy fundamentals like how to build programs, and how to use `dspy.Evaluate`, which all have their dedicated sections now.

### DIFF
--- a/docs/docs/tutorials/rag/index.ipynb
+++ b/docs/docs/tutorials/rag/index.ipynb
@@ -16,7 +16,7 @@
    "metadata": {},
    "source": [
     "\n",
-    "## Configuring the DSPy environment.\n",
+    "## Configuring the DSPy Environment\n",
     "\n",
     "Let's tell DSPy that we will use OpenAI's `gpt-4o-mini` in our modules. To authenticate, DSPy will look into your `OPENAI_API_KEY`. You can easily swap this out for [other providers or local models](https://github.com/stanfordnlp/dspy/blob/main/examples/migration.ipynb).\n"
    ]
@@ -67,13 +67,16 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
     "import dspy\n",
+    "import os\n",
     "\n",
-    "lm = dspy.LM('openai/gpt-4o-mini')\n",
+    "os.environ[\"OPENAI_API_KEY\"] = \"{your_openai_api_key}\"\n",
+    "\n",
+    "lm = dspy.LM(\"openai/gpt-4o-mini\")\n",
     "dspy.configure(lm=lm)"
    ]
   },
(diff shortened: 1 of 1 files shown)


docs/docs/tutorials/rag/index.ipynb
```diff
@@ -1456,7 +1463,7 @@
     "\n",
     "## What's next?\n",
     "\n",
-    "Improving from around 42% to approximately 61% on this task, in terms of `SemanticF1`, was pretty easy.\n",
+    "Improving from around 54% to approximately 61% on this task, in terms of `SemanticF1`, was pretty easy through DSPy optimzier.\n",
```
TomeHirata, line 1466: Just curious, what made the baseline score higher?

chenmoneygithub: Haha good catch! The original tutorial has 3 stages:

- Base baseline: no retriever, just use CoT.
- Baseline: retriever + CoT
- Optimized: retriever + CoT + optimizer

it's written in that way because that used to be our landing tutorial, which teaches multiple things of DSPy ranging from the basic module's usage to applying optimizer. Now we can leave our baseline as an unoptimized RAG.

docs/docs/tutorials/rag/index.ipynb
```diff
@@ -67,67 +67,160 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Exploring some basic DSPy Modules.\n",
+    "## Basic Retrieval-Augmented Generation (RAG)\n",
+    "\n",
+    "First, let's download the corpus data that we will use for RAG search. An older version of this tutorial used the full (650,000 document) corpus. To make this very fast and cheap to run, we've downsampled the corpus to just 28,000 documents."
```
TomeHirata, line 89: nit: we might not need to mention "An older version of this tutorial used the full (650,000 document) corpus." since I think many users will see the page for the first time

docs/docs/tutorials/rag/index.ipynb
```diff
@@ -191,78 +293,38 @@
...
-    "\n",
-    "Let's load a dataset of questions and their (pretty long) gold answers. Since we started this notebook with the goal of building **a system for answering Tech questions**, we obtained a bunch of StackExchange-based questions and their correct answers from the [RAG-QA Arena](https://arxiv.org/abs/2407.13998) dataset.\n",
-    "\n"
+    "Let's first evaluate the RAG we built. In order to evaluate a RAG application, we need a dataset along with an evaluation metric. For this purpose, we will use the [RAG-QA dataset](https://arxiv.org/abs/2407.13998) as the evaluation dataset, and [SemanticF1](https://karmake2.github.io/files/Publications/2022/SEM_F1.pdf) as the evaluation metric. Let's download the dataset and import the evaluation metric."
```
TomeHirata, line 312: just comment: I like the DSPy tutorials that are research/science oriented. But we should probably provide a separate doc for non-DS or non-Eng backgroup people 😄

TomeHirata (approved):
LGTM

okhat:
Thank you @chenmoneygithub ! Not convinced this is much shorter in a way that justifies reducing the depth of the tutorial.

I think a lot of people want a few end-to-end tutorials that go from nothing to full mastery. This could be one of those.

Also in terms of the diff: +901 −894 seems like mostly similar length (I know it's shorter because most of this are inspect_history outputs).

PR #8414, opened 2025-06-18
change the default model to llama-4 for databricks provider

chenmoneygithub:
llama 3.1 is an old model, we should ping users to a newer one for robust experience. Things move too fast in GenAI, cheers!

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -49,7 +49,7 @@ Instead of wrangling prompts or training jobs, DSPy (Declarative Self-improving 
 
         ```python linenums="1"
         import dspy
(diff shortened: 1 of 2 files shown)


docs/docs/index.md
```diff
@@ -49,7 +49,11 @@ Instead of wrangling prompts or training jobs, DSPy (Declarative Self-improving
 
         ```python linenums="1"
         import dspy
-        lm = dspy.LM('databricks/databricks-meta-llama-3-1-70b-instruct')
```
TomeHirata, line 52: q: how did the previous example work?

TomeHirata (approved):
LGTM

PR #8423, opened 2025-06-19
Remove unsupported options of Plugin 'material/search'

TomeHirata:
Fix the following warnings caused by the unsupported options
```
WARNING -  Config value 'plugins': Plugin 'material/search' option 'indexing': Unsupported option
WARNING -  Config value 'plugins': Plugin 'material/search' option 'prebuild_index': Unsupported option
WARNING -  Config value 'plugins': Plugin 'material/search' option 'min_search_length': Unsupported option
WARNING -  Config value 'plugins': Plugin 'material/search' option 'properties': Unrecognised configuration name: properties
WARNING -  Config value 'plugins': Plugin 'material/search' option 'search_index_only': Unrecognised configuration name: search_index_only
```

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -184,15 +184,6 @@ plugins:
     - search:
         lang: en
         separator: '[\s\-\.]+'
-        min_search_length: 2
-        prebuild_index: true
-        indexing: 'full'
-        properties:
-            - title
-            - tags
-            - content
-            - sections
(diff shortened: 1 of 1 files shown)


okhat:
@krypticmouse is reviewing this. We think something is off, since these plugins were added originally for a reason.

@krypticmouse suspects that maybe we removed some dependencies at some point that broke these.

PR #8431, opened 2025-06-20
fix bedrock streaming token tracking

### LINKED ISSUE #8416: [Bug] Token Tracking Issue - Bedrock with Streamified Predict
### What happened?

Token tracking is not working as expected when using a streamified module using Amazon Bedrock. `completion_tokens` is populated as expected, but all other fields have a value of 0. The one exception is `total_tokens`, but that just has the same value as `completion_tokens` since all other fields are 0.  

I have spent a lot of time trying to track down the issue on this. I've walked through the LiteLLM code in the debugger and I can see that all usage fields are being set, but at some point in the chunk iterator it seems that they are getting stripped. The LiteLLM code is way too dense and indirect for me to follow exactly where that is happening.  

The interesting thing is that while walking through LiteLLM code all of the usage fields looked right, total_tokens was the sum of prompt and completion tokens, which tells me that it's getting re-computed at some point.  

### To Recap
Token usage is getting set properly in LiteLLM, but at some point between LiteLLM and DSPy all usage is set to 0 except for `completion_tokens`

I have tried updating to the newest version of LiteLLM and that didn't make a difference.
LiteLLM version tested: `1.71.1`, `1.72.6.post1`

### Steps to reproduce

I have tried many different parameters for the streamify call. The example below uses `async_streaming=False` because the results are the same either way and the example is less cumbersome without async streaming.  

Here is a minimal example that will reproduce the error:
```python
import os
import dspy

# If no AWS profile is configured you can use access key and secret key instead
os.environ["AWS_PROFILE"] = "YOUR_AWS_PROFILE"
# os.environ["AWS_ACCESS_KEY_ID"] = "YOUR_AWS_ACCESS_KEY_ID"
# os.environ["AWS_SECRET_ACCESS_KEY"] = "YOUR_AWS_SECRET_ACCESS_KEY"
# os.environ["AWS_SESSION_TOKEN"] = "YOUR_AWS_SESSION_TOKEN" <-- Optional
os.environ["AWS_REGION"] = "us-east-1"


class QAPredictor(dspy.Signature):
    """Answer questions with helpful, accurate responses."""

    question = dspy.InputField()
    answer = dspy.OutputField()


predictor = dspy.Predict(QAPredictor)
lm = dspy.LM("bedrock/us.amazon.nova-lite-v1:0")
dspy.configure(lm=lm, track_usage=True)
streaming_predictor = dspy.streamify(
    predictor,
    async_streaming=False,  # <--- Doesn't matter if this is True or False fot token tracking
    include_final_prediction_in_output_stream=True,
)


def generate_response(question):
    for chunk in streaming_predictor(question=question):
        if isinstance(chunk, dspy.Prediction):
            usage = chunk.get_lm_usage()
            print(usage)


if __name__ == "__main__":
    generate_response(question="What is the capital of Maine?")
```

### DSPy version

3.0.0b1, also tested with 2.6.27

BTripp1986:
Fixes token tracking when using streamified module when using AWS Bedrock as a provider.

closes #8416

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -269,6 +269,9 @@ def _get_stream_completion_fn(
     stream = cast(MemoryObjectSendStream, stream)
     caller_predict_id = id(caller_predict) if caller_predict else None
 
+    if dspy.settings.track_usage:
+        request["stream_options"] = {"include_usage": True}
+
     async def stream_completion(request: Dict[str, Any], cache_kwargs: Dict[str, Any]):
         response = await litellm.acompletion(
             cache=cache_kwargs,


okhat:
Hey @BTripp1986 I saw on Discord that you said this is resolved. Do we keep or close this PR?

BTripp1986:
> Hey @BTripp1986 I saw on Discord that you said this is resolved. Do we keep or close this PR?

@okhat This PR is unrelated to the stream listener issue I mentioned in Discord. This PR addresses an actual issue with token tracking when using AWS Bedrock(and probably other providers as well) when using a async streamified dspy.module.  

You can refer to #8416 for more details on this issue and how to reproduce.

okhat:
Thank you @BTripp1986 . OK, I want to merge this but I'm worried about adding a new keyword argument to providers that don't need it. It may break every other provider that isn't Bedrock?

BTripp1986:
@okhat I understand and share your concern. I believe that this bug might also be present with other providers, but I cannot confirm. OpenAI is working, but I do not have access to other providers than OAI and Bedrock.  

To be clear, the keyword is not a Bedrock specific keyword, this is a LiteLLM defined kwarg. If anyone has access to other providers the code in the linked issue may reproduce the error if tried with a provider other than OpenAI.  

I've walked through the LiteLLM code quite a bit and it does a lot of input cleaning depending on provider so I don't foresee any issues from the additional kwarg even if it isn't needed for a given provider, but you need something more concrete than my assumption.  

Do you have any suggestions for a safer way to approach this, or a way that we can verify that it isn't going to cause any issues for other providers? We could add to this conditional to check if the provider is Bedrock.

Would something like this be better?
```python
if dspy.settings.track_usage and request["model"].startswith("bedrock/"):
    request["stream_options"] = {"include_usage": True}
```

okhat:
@chenmoneygithub Let's check this for databricks (and OAI) too and if that too is OK, let's merge. Seems like it's just a LiteLLM argument.

chenmoneygithub:
Working all right on my testing script:

```
import asyncio

import dspy

# lm = dspy.LM("databricks/databricks-llama-4-maverick", cache=False)
lm = dspy.LM("gemini/gemini-2.5-flash-preview-04-17", cache=False)
# lm = dspy.LM("anthropic/claude-3-5-sonnet-20240620", cache=False)
# lm = dspy.LM("openai/gpt-4o-mini", cache=False)
dspy.settings.configure(lm=lm, track_usage=True)


class MyModule(dspy.Module):
    def __init__(self):
        super().__init__()

        self.predict1 = dspy.Predict("question->answer")
        self.predict2 = dspy.Predict("question, answer->judgement, score")

    def forward(self, question: str, **kwargs):
        judgement = self.predict2(question=question, answer=answer)
        return judgement


predict = MyModule()
stream_listeners = [
    dspy.streaming.StreamListener(signature_field_name="answer"),
]
stream_predict = dspy.streamify(
    predict,
    include_final_prediction_in_output_stream=True,
)


async def main():
    output = stream_predict(question="why did a chicken cross the kitchen?")

    return_value = None
    async for value in output:
        if isinstance(value, dspy.streaming.StatusMessage):
            print(value.message)
        elif isinstance(value, dspy.Prediction):
            print(value)
            return_value = value
        elif isinstance(value, dspy.streaming.StreamResponse):
            print(value)
        else:
            print(value.choices[0].delta.content)
    return return_value


geez = asyncio.run(main())

print(geez.get_lm_usage())

```

PR #8435, opened 2025-06-23
done

Hangzhi:


chenmoneygithub:
@dilarasoylu Do you remember the context for us always requiring the `pred.lm` to be set? https://github.com/stanfordnlp/dspy/blob/734eff216155207dfea89765a46e935ee1482794/dspy/teleprompt/bootstrap_finetune.py#L69

dilarasoylu:
> @dilarasoylu Do you remember the context for us always requiring the `pred.lm` to be set? https://github.com/stanfordnlp/dspy/blob/734eff216155207dfea89765a46e935ee1482794/dspy/teleprompt/bootstrap_finetune.py#L69

Oh, yes, we always want to ensure that predictor level LMs are set for the finetune optimizers, otherwise we might end up modifying a globally set LM as part of the finetune process, which might be confusing for the users. We use the 'set_lm(...)' method on the program to be finetuned before calling any optimizer that modifies the weights in order to ensure that predictor level LMs are set.

Hangzhi:
@chenmoneygithub @dilarasoylu 
Thanks for the suggestions. I updated the example code and `set_lm` works.

docs/docs/index.md
```diff
@@ -403,17 +403,20 @@ Given a few tens or hundreds of representative _inputs_ of your task and a _metr
...
         classify = dspy.ChainOfThought(signature)
+        classify.set_lm(lm)
 
         # Optimize via BootstrapFinetune.
         optimizer = dspy.BootstrapFinetune(metric=(lambda x, y, trace=None: x.label == y.label), num_threads=24)
         optimized = optimizer.compile(classify, trainset=trainset)
 
         optimized(text="What does a pending cash withdrawal mean?")
+        
+        # For a complete fine-tuning tutorial, see: https://dspy.ai/tutorials/classification_finetuning/
```
Hangzhi, line 419: @chenmoneygithub let me know if there are better way refer to another doc page.

okhat:
@Hangzhi This looks great to me, and I was about to merge. But why are we catching all exceptions from the teacher? I think the original behavior should have been to let it be handled by dspy.settings.max_errors essentially (through dspy.Evaluate). This allows more fine-grained control and alerts the user if the program seems broken.

Hangzhi:
@okhat That makes sense. Thanks for the clarification. 
I removed the call exception catching logic.

PR #8438, opened 2025-06-23
Update signatures.md

Skybladev2:
Add missing print in sentiment example. Without it there is no output.


okhat:
I think people can hopefully recognize this.

Skybladev2:
Yes, but this print/no print pattern is not consistent across the docs.
I spent several hours trying to figure out why 
```python
lm("Say this is a test!", temperature=0.7)  # => ['This is a test!']
lm(messages=[{"role": "user", "content": "Say this is a test!"}])  # => ['This is a test!']
```
shows nothing. Only when I found the next section
```python
# Define a module (ChainOfThought) and assign it a signature (return an answer, given a question).
qa = dspy.ChainOfThought('question -> answer')

# Run with the default LM configured with `dspy.configure` above.
response = qa(question="How many floors are in the castle David Gregory inherited?")
print(response.answer)
```
I realized that I should print `lm` object.

PR #8441, opened 2025-06-23
Docs Fix: Update tutorial to consistently use ⁠mlflow.dspy.autolog()

mznmel:
The original tutorial code was inconsistent regarding autolog methods, mentioning both ⁠mlflow.dspy.autolog() and ⁠mlflow.autolog(), and the code example didn't use either. I updated the documentation and code example to use ⁠mlflow.dspy.autolog().

### DIFF
--- a/docs/docs/tutorials/observability/index.md
+++ b/docs/docs/tutorials/observability/index.md
@@ -105,7 +105,7 @@ If you don't specify a different port via `--port` flag, you MLflow server will 
 Now let's change our code snippet to enable MLflow tracing. We need to:
 
 - Tell MLflow where the server is hosted.
-- Apply `mlflow.autolog()` so that DSPy tracing is automatically captured.
+- Apply `mlflow.dspy.autolog()` so that DSPy tracing is automatically captured.
 
 The full code is as below, now let's run it again!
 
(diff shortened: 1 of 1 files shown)


okhat:
Thanks so much @mznmel !

chenmoneygithub:
@TomeHirata  Is the error that litellm generates multiple trace entities fixed on the mlflow side? 

Meanwhile, there is a special case that when streaming is enabled, `dspy.LM` combines the stream tokens into one response. So in this setup, tracing `LM.forward()` makes sense to me, wdyt?

TomeHirata:
@chenmoneygithub The fix for the issue has been filed for the LiteLLM repo a while ago (https://github.com/BerriAI/litellm/pull/8202), but we haven't heard back from them. Can we ping them directly?

TomeHirata:
fyi: We're talking to the LiteLLM team. Please keep this PR on hold for a while.

MaximeRivest:
@TomeHirata  has this progress with litellm? (I am going through docs related PR as part of an effort to improve the docs)

TomeHirata:
Can we actually use `⁠mlflow.autolog` everywhere so that litellm spans are included?

PR #8450, opened 2025-06-24
Add a tutorial for generating a code example from doc

TomeHirata:
This PR introduces a tutorial for learning the usage of an unfamiliar library using DSPy


TomeHirata:
Thank you for the suggestion, yeah, crawling and interaction can enhance the AI system for sure! Can we add these to "Next Steps" instead of implementing the enhanced functionalities in this tutorial itself?

chenmoneygithub:
@TomeHirata Web crawling as a followup sounds good! Can we add interaction to this PR? That should be relatively simple, while making it more user-involved.

TomeHirata:
Interaction learning has been added

PR #8456, opened 2025-06-25
Drop Python 3.9 support

TomeHirata:


dspy/adapters/chat_adapter.py
```diff
@@ -1,6 +1,6 @@
 import re
 import textwrap
-from typing import Any, Dict, NamedTuple, Optional, Type
```
TomeHirata, line 3: ruff changes the recommended style based on the supported python version

TomeHirata:
hmm let me fix CoT

Issue #8446, opened 2025-06-23
[Bug] `str | None` fails in in-line signatures and when using `ChainOfThought`


hrolfurinn:
### What happened?

`str | None` works in class-based signatures when used with e.g. `Predict`, but not `ChainOfThought`. 

They don't work in in-line definitions.

The pipe syntax is preferred over `Optional` and `Union`

https://peps.python.org/pep-0604/

### Steps to reproduce

```
from typing import Optional, Union
import dspy

dspy.configure(lm=dspy.LM(model="openai/gpt-4o-mini"))

input = "Say Hello world"

# inline style using Optional[T]:
print(
    "Optional inline Predict:",
    dspy.Predict(dspy.Signature("input: Optional[str] -> output"))(input=input).output,
)
print(
    "Optional inline Chain of Thought:",
    dspy.ChainOfThought(dspy.Signature("input: Optional[str] -> output"))(input=input).output,
)

# inline style using T | None
try:
    dspy.Signature("input: str | None -> output")
except ValueError as e:
    print("Inline pipe syntax fails:", e)

# inline style using Union[str, None] syntax:
print(
    "Union inline Predict:",
    dspy.Predict(dspy.Signature("input: Union[str, None] -> output"))(input=input).output,
)
print(
    "Union inline Chain of Thought:",
    dspy.ChainOfThought(dspy.Signature("input: Union[str, None] -> output"))(input=input).output,
)


# class style using Optional[T]
class OptionalSyntaxSignature(dspy.Signature):
    input: Optional[str] = dspy.InputField()
    output: str = dspy.OutputField()


print("Class Optional Predict works:", dspy.Predict(OptionalSyntaxSignature)(input=input).output)
print(
    "Class Optional Chain of Thought works:",
    dspy.ChainOfThought(OptionalSyntaxSignature)(input=input).output,
)


# class style using T | None
class PipeSyntaxSignature(dspy.Signature):
    input: str | None = dspy.InputField()
    output: str = dspy.OutputField()


print("Class pipe syntax Predict works:", dspy.Predict(PipeSyntaxSignature)(input=input).output)
try:
    dspy.ChainOfThought(PipeSyntaxSignature)(input=input)
except Exception as e:
    print("Class pipe syntax Chain of Thought fails:", e)


# class style using Union[T,None]
class UnionSyntaxSignature(dspy.Signature):
    input: Union[str, None] = dspy.InputField()
    output: str = dspy.OutputField()


[...]

PR #8465, opened 2025-06-26
Fix PEP 604 union types when using Class Signatures for ChainOfThought

erandeutsch:
## Description

This pull request fixes a bug where [PEP 604](https://peps.python.org/pep-0604/)-style union types in Class-based DSPy Signatures work when using `dspy.Predict`, but fail when using `dspy.ChainOfThought`.

This is due to PEP 604 Union Types being defined as a [types.UnionType](https://docs.python.org/3/library/types.html#types.UnionType), which was not included in the list of types to check in `make_signature` in `signature.py`. For reference, the original union type is defined as a [typing.Union](https://docs.python.org/3/library/typing.html#typing.Union)

## Example

```python
class SayTheRightType(dspy.Signature):
    input: str = dspy.InputField()
    output: str | int = dspy.OutputField()


input1 = "Say Hello World"
output = dspy.ChainOfThought(SayTheRightType)(input=input1).output
print(f"Output: {output} with type {type(output)}")
# Output: Hello World with type <class 'str'>

input1 = "Say the number 42"
output = dspy.ChainOfThought(SayTheRightType)(input=input1).output
print(f"Output: {output} with type {type(output)}")
# Output: 42 with type <class 'int'>

```

## Additional Info

This bug fixes one of two issues in issue #8446 related to support of union types.
The other issue was fixed in pull request #8449 where support for PEP 604 union types was added to inline signatures. 

PEP 604 union types seems to be fully supported in signatures now, I will make another pull request over the next few days to add more testing for PEP 604 union types to confirm.

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -430,7 +430,7 @@ def make_signature(
         # program of thought and teleprompters, so we just silently default to string.
         if type_ is None:
             type_ = str
-        if not isinstance(type_, (type, typing._GenericAlias, types.GenericAlias, typing._SpecialForm)):
+        if not isinstance(type_, (type, typing._GenericAlias, types.GenericAlias, typing._SpecialForm, types.UnionType)):
             raise ValueError(f"Field types must be types, but received: {type_} of type {type(type_)}.")
         if not isinstance(field, FieldInfo):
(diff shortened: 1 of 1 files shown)


okhat:
We are so lucky to have your contributions, @erandeutsch , thank you!!

PR #8467, opened 2025-06-26
[WIP] MCP code executor to execute LLM-generated code flexibly in CodeAct and ProgramOfThought

chenmoneygithub:
Very dirty code now, please don't look at it😶‍🌫️


arthurcolle:
I would like to recommend using a newer cheaper model

gpt-4.1-nano is better for examples

PR #8472, opened 2025-06-28
[WIP] React use `dspy.ToolCalls`

chenmoneygithub:
Refactor dspy.ReAct to use the prebuilt types. 

I am observing some unstable behavior, still doing experiments.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -56,11 +56,10 @@ class Adapter:
 
                 lm_kwargs["tools"] = litellm_tools
 
-                import pdb
-
-                pdb.set_trace()
-
                 signature_for_native_function_calling = signature.delete(tool_call_output_field_name)
+                signature_for_native_function_calling = signature_for_native_function_calling.delete(
+                    tool_call_input_field_name
+                )
 
                 return signature_for_native_function_calling
 
@@ -91,10 +90,6 @@ class Adapter:
                 value = {}
                 for field_name in signature.output_fields.keys():
                     value[field_name] = None
-
-            import pdb
-
-            pdb.set_trace()
             if tool_calls and tool_call_output_field_name:
                 tool_calls = [
                     {
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -56,6 +56,10 @@ class Adapter:
 
                 lm_kwargs["tools"] = litellm_tools
 
+                import pdb
+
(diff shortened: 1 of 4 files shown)


ryanh-ai:
@chenmoneygithub not 100% sure, but the regression you see may be due to how the message history and tool call history/results are passed back to the model provider.  It looks like the trajectory right now is still formatted as a string, but you may get better results if you build and pass a full message stack back to the model provider with the right content types for tool call and tool call results.  Have you tried this?

chenmoneygithub:
@ryanh-ai Thanks for the suggestion! I assume you mean formatting the trajectory as multiturn message instead of a big json? I have tried it, but doesn't produce meaningful improvements.

I have kinda of spotted the problem, it's the LM is doing a worse job of understanding nested type requirements, like `dspy.ToolCalls` essentially does something like `output.toolcalls.tool_calls[0].args`, while the original ReAct uses `output.next_tool_call_args`.

ryanh-ai:
I mean passing tool results back as tool result content blocks as part of user the, same for assistant turns text content, etc - perhaps that is what your experiment was, but wanted to be clear.  Here is the page in LiteLLM:

https://docs.litellm.ai/docs/completion/function_call

I know some model providers use the structured message schema and content types in the way they format what the LLM sees.

chenmoneygithub:
@ryanh-ai Thanks! So you mean native function calling. We have noticed that native function calling yields a very poor quality, if you are interested in helping us improve DSPy, I would like to get your experiment result on how it works. Thank you!

ryanh-ai:
Sounds good! I have not done it with DSPy but was trying to hence coming across this thread.  I have done it outside DSPy on a couple of providers.

Let me see if I can implement in DSPy when I find some time for a test.

ryanh-ai:
Started working on coding this up over the weekend.... will share back when I have some results!

quancore:
Cna I ask what it the progress here for getting paralel tool calling?

baughmann:
Random two cents regarding the original description @chenmoneygithub provided: you mentioned experimenting with native tool calls. One of my favorite things about DSPy is that full control is delegated to the client. This gives maximum flexibility to users and enables faster iterations, at least when compared to native calls in vLLM where I'd need to alter the tool call parser and/or the chat template and restart.

kargarisaac:
how is the situation with this pr?

PR #8475, opened 2025-06-30
Add testing for PEP604 union types

erandeutsch:
## Description

This pull request adds testing for the support of [PEP 604](https://peps.python.org/pep-0604/)-style union types added in pull requests #8449 and #8465.

These tests verify that PEP 604 union types (e.g., `int | str`) are now fully supported in both inline and class-based signatures! 🎉

## Tests Added
 -  **test_pep604_union_type_inline**  
  Verifies that inline signature parsing correctly recognizes PEP 604 union types such as `str | None` and `None | int` for input, and `int | str` for output.

- **test_pep604_union_type_inline_equivalence**  
  Ensures that signatures using PEP 604 union types are equivalent to those using `Optional` and `Union` from `typing`.

- **test_pep604_union_type_inline_nested**  
  Checks parsing of nested union types (e.g., `str | (int | float) | None`).

- **test_pep604_union_type_class_nested**  
  Validates that nested union types in class-based signatures are recognized and handled as expected.

- **test_pep604_union_type_class_equivalence**  
  Confirms equivalence between class-based signatures using PEP 604 and those using `Optional`/`Union`, and that the annotation types are correct (`UnionType`).

- **test_pep604_union_type_insert**  
  Ensures that inserting a new field with a union type into a signature does not cause errors, verifying that this doesn't crash. This is for testing the bug fixed in #8465

- **test_pep604_union_type_with_custom_types**  
  Validates union types with custom types, similar to `test_basic_custom_type`, but when PEP 604 union types are involved.

### DIFF
--- a/tests/signatures/test_signature.py
+++ b/tests/signatures/test_signature.py
@@ -1,3 +1,4 @@
+from types import UnionType
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pydantic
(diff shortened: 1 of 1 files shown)


tests/signatures/test_signature.py
```diff
@@ -436,3 +437,139 @@ def test_custom_type_from_different_module():
     path_obj = Path("/test/path")
     pred = dspy.Predict(test_signature)(input=path_obj)
     assert pred.output == "/test/path"
+
+
+def test_pep604_union_type_inline():
+    sig = Signature(
+        "input_opt_right: str | None, input_opt_left: None | int -> output_union: int | str"
```
chenmoneygithub: the variable name `input_opt_right` and `input_opt_left` get me confused for a while, I would suggest using the simple `x1, x2` or `input1, input2` and add comment at assert statement to clarify what we are testing here.

tests/signatures/test_signature.py
```diff
@@ -436,3 +437,139 @@ def test_custom_type_from_different_module():
...
+    )
+
+    assert "input_opt_right" in sig.input_fields
+    input_opt_right_annotation = sig.input_fields["input_opt_right"].annotation
+    assert input_opt_right_annotation == Optional[str] or (
```
chenmoneygithub: question - Why do we need the `or` check here? is that because of compatibility with multiple python versions?

tests/signatures/test_signature.py
```diff
@@ -436,3 +437,139 @@ def test_custom_type_from_different_module():
...
+    assert sig1.input_fields["input"].annotation == sig2.input_fields["input"].annotation
+    assert sig1.output_fields["output"].annotation == sig2.output_fields["output"].annotation
+
+
+def test_pep604_union_type_inline_nested():
+    sig = Signature(
+        "input: str | (int | float) | None -> output: str"
+    )
+    assert "input" in sig.input_fields
+    input_ann = sig.input_fields["input"].annotation
```
chenmoneygithub: nit: `input_ann` -> `input_annotation` for clarity.

tests/signatures/test_signature.py
```diff
@@ -436,3 +437,139 @@ def test_custom_type_from_different_module():
...
+    assert isinstance(Sig1.output_fields["output"].annotation, UnionType)
+
+
+def test_pep604_union_type_insert():
+    class PEP604Signature(Signature):
+        input: str | None = InputField()
+        output: int | str = OutputField()
+
+    # This test ensures that inserting a field into a signature with PEP 604 UnionType works
+    # Previously, this would crash due to make_signature not handling UnionType correctly.
```
chenmoneygithub: We can delete this "Previously" comment, because the reader in the future won't have this context.

chenmoneygithub (commented):
Thanks for the PR! 

Pretty solid work, left some minor comments.

tests/signatures/test_signature.py
erandeutsch: Thanks, I renamed the variables to input1 and input2, and added a comment to clarify.

tests/signatures/test_signature.py
erandeutsch: Thanks, I renamed it to input_annotation

tests/signatures/test_signature.py
erandeutsch: Thanks, I removed this comment.

tests/signatures/test_signature.py
erandeutsch: I did an `or` check with Optional to match the behaviour from a previous test case, `test_typed_signatures_unions_and_optionals`, which had the same `or` logic and the following comment:
>     # Depending on the environment, it might resolve to Union[str, None] or Optional[str], either is correct.

This might be out of date and not needed anymore, now that Python 3.9 support has been dropped, but I made sure to do the `or` check just in case there is an environment where this might break.

TomeHirata (approved):
LGTM

tests/signatures/test_signature.py
```diff
@@ -436,3 +437,140 @@ def test_custom_type_from_different_module():
...
+    assert sig.input_fields["input"].annotation == Union[CustomType, None]
+    assert sig.output_fields["output"].annotation == Union[int, str]
+
+    lm = DummyLM([{"output": "processed"}])
+    dspy.settings.configure(lm=lm)
+
+    custom_obj = CustomType(value="test")
+    pred = dspy.Predict(sig)(input=custom_obj)
+    assert pred.output == "processed"
+
```
TomeHirata: nit: unnecessary last line

erandeutsch: Thanks, I removed the extra newline.

PR #8477, opened 2025-07-01
Fix typo: "reuires" → "requires" in program_of_thought.py

LukasMurdock:
Saw it in the API docs here: https://dspy.ai/api/modules/ProgramOfThought/

Thank you for your work


chenmoneygithub (approved):
Thank you for the PR!

PR #8478, opened 2025-07-01
Add token streaming support for XMLAdapter

chenmoneygithub:
Support XMLAdapter in `StreamListener`

## Note

This PR is generated by [OpenAI codex](https://openai.com/codex/), I am testing how deep it can understand DSPy codebase with this specific use case.

It passes on my tests:

```
import asyncio

import dspy

# lm = dspy.LM("gemini/gemini-2.5-flash-preview-04-17", cache=False)
lm = dspy.LM("databricks/databricks-claude-3-7-sonnet", cache=False)
# lm = dspy.LM("openai/gpt-4o-mini", cache=False)
dspy.settings.configure(lm=lm, adapter=dspy.XMLAdapter())


class MyModule(dspy.Module):
    def __init__(self):
        super().__init__()

        self.predict1 = dspy.Predict("question->answer")
        self.predict2 = dspy.Predict("question, answer->judgement, score")

    def forward(self, question: str, **kwargs):
        judgement = self.predict2(question=question, answer=answer)
        return judgement


predict = MyModule()
stream_listeners = [
    dspy.streaming.StreamListener(signature_field_name="answer"),
]
stream_predict = dspy.streamify(
    predict,
)


async def main():
    with dspy.settings.context(adapter=dspy.XMLAdapter()):
        output = stream_predict(question="why did a chicken cross the kitchen?")

        return_value = None
        async for value in output:
            if isinstance(value, dspy.streaming.StatusMessage):
                print(value.message)
            elif isinstance(value, dspy.Prediction):
                print(value)
            elif isinstance(value, dspy.streaming.StreamResponse):
                print(value)
            else:
                print(value.choices[0].delta.content)
    return return_value


import time

start = time.time()
asyncio.run(main())
end = time.time()


print(f"Time taken: {end - start} seconds")

dspy.inspect_history()
```

### DIFF
--- a/dspy/streaming/streaming_listener.py
+++ b/dspy/streaming/streaming_listener.py
@@ -7,6 +7,7 @@ from litellm import ModelResponseStream
 
 from dspy.adapters.chat_adapter import ChatAdapter
 from dspy.adapters.json_adapter import JSONAdapter
+from dspy.adapters.xml_adapter import XMLAdapter
 from dspy.dsp.utils.settings import settings
 from dspy.streaming.messages import StreamResponse
 
@@ -51,6 +52,9 @@ class StreamListener:
         self.chat_adapter_start_identifier = f"[[ ## {self.signature_field_name} ## ]]"
         self.chat_adapter_end_identifier = re.compile(r"\[\[ ## (\w+) ## \]\]")
 
+        self.xml_adapter_start_identifier = f"<{self.signature_field_name}>"
+        self.xml_adapter_end_identifier = re.compile(rf"</{self.signature_field_name}>")
+
     def _buffered_message_end_with_start_identifier(self, concat_message: str, start_identifier: str) -> str:
         for i in range(len(concat_message)):
             if start_identifier.startswith(concat_message[len(concat_message) - i - 1 :]):
@@ -68,9 +72,14 @@ class StreamListener:
             end_identifier = self.chat_adapter_end_identifier
 
             start_indicator = "["
+        elif isinstance(settings.adapter, XMLAdapter):
+            start_identifier = self.xml_adapter_start_identifier
+            end_identifier = self.xml_adapter_end_identifier
+
+            start_indicator = "<"
         else:
             raise ValueError(
-                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter or "
+                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter, XMLAdapter or "
                 "JSONAdapter for streaming purposes."
             )
 
@@ -178,9 +187,14 @@ class StreamListener:
         elif isinstance(settings.adapter, ChatAdapter) or settings.adapter is None:
             boundary_index = last_tokens.find("[[")
             return last_tokens[:boundary_index]
+        elif isinstance(settings.adapter, XMLAdapter):
+            boundary_index = last_tokens.find(f"</{self.signature_field_name}>")
+            if boundary_index == -1:
+                boundary_index = len(last_tokens)
+            return last_tokens[:boundary_index]
         else:
             raise ValueError(
-                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter or "
+                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter, XMLAdapter or "
                 "JSONAdapter for streaming purposes."
             )
 
--- a/dspy/streaming/streaming_listener.py
+++ b/dspy/streaming/streaming_listener.py
@@ -14,6 +14,8 @@ from dspy.streaming.messages import StreamResponse
 if TYPE_CHECKING:
     from dspy.primitives.module import Module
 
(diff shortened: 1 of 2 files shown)


dspy/streaming/streaming_listener.py
```diff
@@ -63,14 +67,19 @@ def receive(self, chunk: ModelResponseStream):
             end_identifier = self.json_adapter_end_identifier
 
             start_indicator = '"'
+        elif isinstance(settings.adapter, XMLAdapter):
+            start_identifier = self.xml_adapter_start_identifier
+            end_identifier = self.xml_adapter_end_identifier
+
+            start_indicator = "<"
```
TomeHirata: Can we also make start_indicator an instance variable like we do for start_identifier and end_identifier?

dspy/streaming/streaming_listener.py
```diff
@@ -63,14 +67,19 @@ def receive(self, chunk: ModelResponseStream):
...
         elif isinstance(settings.adapter, ChatAdapter) or settings.adapter is None:
             start_identifier = self.chat_adapter_start_identifier
             end_identifier = self.chat_adapter_end_identifier
 
             start_indicator = "["
         else:
             raise ValueError(
-                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter or "
+                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter, XMLAdapter or "
```
TomeHirata: nit: Can we define const for a list of adapters that support streaming?

dspy/streaming/streaming_listener.py
TomeHirata: As I mentioned above, maybe it's better to move the logic into each adapter class and remove if-else branch
```python
start_identifier = settings.adapter.get_start_identifier(self.signature_field_name)
end_identifier= settings.adapter.get_end_identifier(self.signature_field_name)
start_indicator= settings.adapter.get_start_indicator()
```

dspy/streaming/streaming_listener.py
```diff
@@ -51,6 +52,9 @@ def __init__(
         self.chat_adapter_start_identifier = f"[[ ## {self.signature_field_name} ## ]]"
         self.chat_adapter_end_identifier = re.compile(r"\[\[ ## (\w+) ## \]\]")
 
+        self.xml_adapter_start_identifier = f"<{self.signature_field_name}>"
+        self.xml_adapter_end_identifier = re.compile(rf"</{self.signature_field_name}>")
+
```
TomeHirata, line 68: Since we already have three adapters, shall we define a mapping for adapter-specific parameters? Alternatively, we can move the identifier into each adapter class for better encapsulation

dspy/streaming/streaming_listener.py
```diff
@@ -175,12 +184,21 @@ def flush(self) -> str:
...
             return last_tokens[:boundary_index]
+        elif isinstance(settings.adapter, XMLAdapter):
+            boundary_index = last_tokens.find(f"</{self.signature_field_name}>")
+            if boundary_index == -1:
+                boundary_index = len(last_tokens)
+            return last_tokens[:boundary_index]
         elif isinstance(settings.adapter, ChatAdapter) or settings.adapter is None:
             boundary_index = last_tokens.find("[[")
             return last_tokens[:boundary_index]
+            boundary_index = last_tokens.find(f"</{self.signature_field_name}>")
```
TomeHirata: q: why do we need this change for `ChatAdapter`?

dspy/streaming/streaming_listener.py
```diff
@@ -175,12 +184,21 @@ def flush(self) -> str:
...
+            if boundary_index == -1:
+                boundary_index = len(last_tokens)
+            return last_tokens[:boundary_index]
         else:
             raise ValueError(
-                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter or "
+                f"Unsupported adapter for streaming: {settings.adapter}, please use either ChatAdapter, XMLAdapter or "
```
TomeHirata: ditto

dspy/streaming/streaming_listener.py
chenmoneygithub: good catch, it's a codex mess

dspy/streaming/streaming_listener.py
chenmoneygithub: done!

dspy/streaming/streaming_listener.py
chenmoneygithub: sg!

dspy/streaming/streaming_listener.py
chenmoneygithub: > shall we define a mapping for adapter-specific parameters? 
Yes, good call!

> we can move the identifier into each adapter class for better encapsulation
This is a little subtle, the `start_identifier`/`end_identifier` is not a feature of the adapter, but a feature of the stream listener. The boundary is a bit vague though - adapter can fully function without `start_identifier`/`end_identifier`, and different listeners could introduce different `start_identifier`/`end_identifier` for the same adapter, so it belongs more to the stream listeners.

TomeHirata (approved):
LGTM

PR #8479, opened 2025-07-02
Fix native function calling in adapters

chenmoneygithub:
A few changes:

- In the base adapter, we need to explicitly set the fields to None if they don't present in the response for consistency. This happens often with native function calling, because either the content or tool_calls will be missing.
- In the JSONAdapter, we fall back to json mode if native function calling is not used while ToolCalls are in the output fields. We do this because structured output performs very poorly on fields with type like `dict[str, Any]`, which is the `args` in `dspy.ToolCalls`.
- Add unit tests to cover the native function calling response parsing.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -57,6 +57,9 @@ class Adapter:
                 lm_kwargs["tools"] = litellm_tools
 
                 signature_for_native_function_calling = signature.delete(tool_call_output_field_name)
+                signature_for_native_function_calling = signature_for_native_function_calling.delete(
+                    tool_call_input_field_name
+                )
 
                 return signature_for_native_function_calling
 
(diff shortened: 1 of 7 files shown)


dspy/adapters/base.py
```diff
@@ -57,19 +57,23 @@ def _call_preprocess(
                 lm_kwargs["tools"] = litellm_tools
 
                 signature_for_native_function_calling = signature.delete(tool_call_output_field_name)
+                signature_for_native_function_calling = signature_for_native_function_calling.delete(
```
TomeHirata, line 60: We can chain Signature.delete
```suggestion
                signature_for_native_function_calling = signature.delete(tool_call_output_field_name).delete(tool_call_input_field_name)
```

dspy/adapters/base.py
```diff
@@ -82,10 +86,14 @@ def _call_postprocess(
                 tool_calls = output.get("tool_calls")
 
             if text:
-                value = self.parse(signature, text)
+                value = self.parse(processed_signature, text)
+                for field_name in original_signature.output_fields.keys():
+                    if field_name not in value:
+                        # We need to set the field not present in the processed signature to None for consistency.
```
TomeHirata, line 92: Are we safe to set None? How should we handle the case where the missing field is non-null?

PR #8485, opened 2025-07-03
Support MIPROv2 user confirmation in windows

### LINKED ISSUE #8483: [Bug] Windows OSError in MIPRO Optimizer v2 User Confirmation
### What happened?

Prepared by Grok

An OSError: [WinError 10038] An operation was attempted on something that is not a socket occurs during the user confirmation step in the MIPRO optimizer v2 compilation process.

When running a script that compiles a QAModule using the MIPRO optimizer v2, the program prompts for user input (Do you wish to continue? (y/n)). However, an OSError is raised in the _get_user_confirmation method when attempting to check for input using select.select.

`Traceback (most recent call last):
  File "[path]/miprov2_simple.py", line 64, in <module>
    compiled_qa = optimizer.compile(QAModule(), trainset=trainset, num_trials=1, minibatch_size=1)
  File "[path]/langwatch/dspy/__init__.py", line 596, in compile
    return super().compile(student, **kwargs)
  File "[path]/dspy/teleprompt/mipro_optimizer_v2.py", line 160, in compile
    if not self._get_user_confirmation(
  File "[path]/dspy/teleprompt/mipro_optimizer_v2.py", line 390, in _get_user_confirmation
    if select.select([sys.stdin], [], [], 0.1)[0]:
OSError: [WinError 10038] An operation was attempted on something that is not a socket`

Investigate the use of select.select in _get_user_confirmation (line 390 of mipro_optimizer_v2.py) for Windows compatibility.

Consider alternative methods for non-blocking input checking that are cross-platform compatible, such as msvcrt.kbhit() for Windows.

### Steps to reproduce

1. Set up a Python script using DSPy with the MIPRO optimizer v2.
2. Configure the optimizer to compile a module (e.g., QAModule) with a training set.
3. Run the script on a Windows machine.
4. Encounter the prompt for user confirmation, leading to the error.

### DSPy version

2.6.27

TomeHirata:
Resolve #8483

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -397,7 +397,7 @@ class MIPROv2(Teleprompter):
             if sys.platform == "win32":
                 import msvcrt
                 if msvcrt.kbhit():
-                    user_input = msvcrt.getch().decode('utf-8').strip().lower()
+                    user_input = msvcrt.getch().decode("utf-8").strip().lower()
                     print(user_input)  # Echo the input
                     return user_input == "y"
             else:


dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -395,9 +394,17 @@ def _get_user_confirmation(
         # Wait for input with timeout
         start_time = time.time()
         while time.time() - start_time < 20:
-            if select.select([sys.stdin], [], [], 0.1)[0]:
-                user_input = sys.stdin.readline().strip().lower()
-                return user_input == "y"
+            if sys.platform == "win32":
+                import msvcrt
+                if msvcrt.kbhit():
+                    user_input = msvcrt.getch().decode('utf-8').strip().lower()
```
chenmoneygithub: nit: double quotes

dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -395,9 +394,17 @@ def _get_user_confirmation(
...
+                    print(user_input)  # Echo the input
```
chenmoneygithub, line 401: qq - is this print necessary?

chenmoneygithub (approved):
Nice fix! 

Approved with 2 minor comments.

dspy/teleprompt/mipro_optimizer_v2.py
TomeHirata: It seems like so, `msvcrt.getch()` reads a single character without echoing it to the console, so the behavior would  be different from Mac without print

PR #8486, opened 2025-07-03
Run unit tests with real LLM calls

TomeHirata:
This PR enables running unit tests that require real LLM calls, which have been skipped in CI.
The model for testing is configurable through `LLM_MODEL` env variable, and we use `ollama/llama3.2:3b` in the branch build to balance quality and latency. The Ollama model pulling only introduces ~13s latency, so this PR just enables real LLM tests in the branch build instead of nightly tests.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -48,6 +48,11 @@ jobs:
   test:
     name: Run Tests
     runs-on: ubuntu-latest
+    services:
+      ollama:
+        image: ollama/ollama:latest
+        ports:
+          - 11434:11434
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
@@ -81,11 +86,13 @@ jobs:
         uses: chartboost/ruff-action@v1
         with:
           args: check --fix-only
-      - name: Set LLM model
+      - name: Pull LLM
         run: |
-          echo "LLM_MODEL=${{ secrets.LLM_MODEL }}" >> $GITHUB_ENV
-          echo "DATABRICKS_API_BASE=${{ secrets.DATABRICKS_API_BASE }}" >> $GITHUB_ENV
-          echo "DATABRICKS_API_KEY=${{ secrets.DATABRICKS_API_KEY }}" >> $GITHUB_ENV
+          timeout 60 bash -c 'until curl -f http://localhost:11434/api/version; do sleep 2; done'
+          curl -X POST http://localhost:11434/api/pull \
+            -H "Content-Type: application/json" \
+            -d '{"name": "llama3.2:1b"}'
+          echo "LLM_MODEL=ollama/llama3.2:1b" >> $GITHUB_ENV
       - name: Run tests with pytest
         run: uv run -p .venv pytest tests/
       - name: Install optional dependencies
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -48,11 +48,6 @@ jobs:
   test:
     name: Run Tests
     runs-on: ubuntu-latest
-    services:
-      ollama:
-        image: ollama/ollama:latest
-        ports:
-          - 11434:11434
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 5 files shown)


tests/streaming/test_streaming.py
```diff
@@ -232,22 +229,22 @@ def __call__(self, x: str, **kwargs):
...
+async def test_streaming_handles_space_correctly(llm_model):
     my_program = dspy.Predict("question->answer")
     program = dspy.streamify(
         my_program, stream_listeners=[dspy.streaming.StreamListener(signature_field_name="answer")]
     )
 
     async def gpt_4o_mini_stream(*args, **kwargs):
         yield ModelResponseStream(
-            model="gpt-4o-mini", choices=[StreamingChoices(delta=Delta(content="[[ ## answer ## ]]\n"))]
+            model=llm_model, choices=[StreamingChoices(delta=Delta(content="[[ ## answer ## ]]\n"))]
```
chenmoneygithub: We can leave this as it is, the streaming was recorded from `gpt-4o-mini`

chenmoneygithub:
@TomeHirata  This is awesome! One thing I want to discuss.

I compared the time cost of running test before and after this PR, seems we are seeing a significant increase from 2m => 4m:

Before:
image

After:
image

4m seems to be all right, but that's the outcome of only converting 6 test cases into using Ollama, which means it could become much slower over time. I am thinking about the following:

1. Use pytest markers to split tests requiring actual LM into its own group, like `@pytest.mark.real_lm_call`. These tests will be skipped in normal github tests, and only run in the dedicated action `Run Tests with Real LM`. We may not need to test across all python versions, because essentially it's an supplementary to the normal test to test against actual LM behavior. 
2. Prefer mocking over real LM testing in unit tests. Only when mocking goes too complex like optimizer, or unreliable over time like streaming, we switch to using the real LM.

Let me know what you think!

TomeHirata:
Hi, @chenmoneygithub. I agree with both points. I've split LLM call tests into a separate job. I completely agree with #2. We should limit the usage of LLM call in unit tests because of latency and potential flakiness.

.github/workflows/run_tests.yml
```diff
@@ -82,11 +82,49 @@ jobs:
...
+            **/pyproject.toml
+            **/uv.lock
+      - name: Create and activate virtual environment
+        run: |
+          uv venv .venv
+          echo "${{ github.workspace }}/.venv/bin" >> $GITHUB_PATH
+      - name: Install dependencies
+        run: |
+          uv sync --dev -p .venv --extra dev
+          uv pip list
```
chenmoneygithub, line 118: Is this line intentional?

tests/conftest.py
```diff
@@ -49,3 +50,11 @@ def pytest_collection_modifyitems(config, items):
         for item in items:
             if flag in item.keywords:
                 item.add_marker(skip_mark)
+
+
+@pytest.fixture
+def llm_model():
```
chenmoneygithub: I feel a bit unnatural on this fixture name, because `llm` and `model` are duplicating each other. Maybe something like `def test_lm()`?

chenmoneygithub (commented):
The setup looks good!

.github/workflows/run_tests.yml
TomeHirata: Yup, this is for showing the list of installed packages. We have the same setup for "Run Tests" job

tests/streaming/test_streaming.py
TomeHirata: Does this test verify any point of the actual 4o-mini model? I only see hardcoded "How are you doing". Maybe we can change the internal function name.

tests/conftest.py
TomeHirata: The LM class has model parameter. So used `llm_model`, but yeah, let me change it if this is confusing. Fixture name shouldn't start with `test_`, so how about using `lm_name`?

tests/streaming/test_streaming.py
chenmoneygithub: ohh I mean the streaming tokens were captured by an actual call to gpt-4o-mini. This field is just an indicator, we don't need it to be dynamic.

tests/conftest.py
chenmoneygithub: `lm_name` sg, or maybe `lm_for_test` to enhance it's for testing.

tests/streaming/test_streaming.py
TomeHirata: Good call, this test does not interact with LLM. Sorry, I'm confused by the usage of "gpt-4o-mini", let me revert the change

tests/conftest.py
TomeHirata: Updated!

PR #8498, opened 2025-07-06
Fix: 403 Unauthorized Access

fswair:
Added Authorization header for http requests.
Example now can run as-is.


okhat:
Thanks @fswair ! Is this for private repos

fswair:
> Thanks @fswair ! Is this for private repos

Not actually. Public repositories too. I've tried to run the example and i encounter 403 unauthorization request, then added Bearer and fixed! So, for both.

PR #8499, opened 2025-07-06
Updated Cheatsheet

vacmar01:
I believe this cheatsheet is a great resource for LLMs to learn the newest DSPy syntax.

To bring it up to date, I removed all deprecated stuff from it and added a little snippet for the new-ish SIMBA optimizer. I also refactored the LLM-as-a-judge example to use type annotations, since this makes sense there, in my opinion.


okhat:
Amazing thank you @vacmar01 !!

Issue #8454, opened 2025-06-24
[Bug] Stream listener for dspy.ReAct not working well


YGao119:
### What happened?

I have a module A in which the predictor is a `dspy.ReAct`, s.t.

```
class A(dspy.Module):
      def __init__(self, *, system_prompt: str, tools: list[Callable], max_iters: int = 2):
             super().__init__()
             self.agent = dspy.ReAct(...)

       async def aforward(self, ...) -> dspy.Prediction:
              pred = await self.agent.acall(...)

              return dspy.Prediction(response=pred.response)
```

In each of the tools provided to the ReAct, there are also `dspy.Module` calls.

ISSUE:

I `streamify` the module A, and provide stream listener for its output field (` StreamListener( signature_field_name="response")`. 

There are 2 predictors inside ReAct (pred and extract). The listener will listen to `extract` in this case. The issue is sometimes
1.  i am not getting any `StreamResponse` only the final `dspy.Prediction`
2.  I see  'ModelResponseStream' object has no attribute 'predict_id' and with some print statements, I see that the `caller_predict` is None when making the `extract` module call. 

### Steps to reproduce

I understand this issue is very specific and might be very hard to reproduce. Just wondering if I'm missing anything here or it might be a limitation of the current implementation for streaming.

### DSPy version

2.6.27

YGao119:
seems to related to this issue: https://github.com/stanfordnlp/dspy/issues/8197

chenmoneygithub:
@YGao119 Thanks for reporting the issue!

> i am not getting any StreamResponse only the final dspy.Prediction

That most likely means your request hit the cache. Try turning off the cache by `dspy.LM(..., cache=False)`.

> I see 'ModelResponseStream' object has no attribute 'predict_id' and with some print statements,

The `predict_id` does work in multithreading setup. Could you share a reproducible code? The current information is not sufficient for debugging.

hung-phan:
The setup is like this @chenmoneygithub 

```
class RecommendationAgent(dspy.Module):
    def __init__(self, *, system_prompt: str, tools: list[Callable], max_iters: int = 2):
        super().__init__()

        self.agent = dspy.ReAct(
            signature=RecommendationAgentSignature.with_instructions(system_prompt),
            tools=tools,
            max_iters=max_iters,
        )

    async def aforward(self, conversational_history: str, user_message: str, context: str) -> dspy.Prediction:
        pred = await self.agent.acall(
[...]

PR #8500, opened 2025-07-06
feat: add test for nested stream and check value has predict_id

hung-phan:
This PR is to add test for nested streaming where we have an agent streaming its response and its tools to run some dspy programs.

- I added the test to validate the use case
- I add check the predict_id before attempt to response to client

Related issue: https://github.com/stanfordnlp/dspy/issues/8454

cc @chenmoneygithub, @okhat


hung-phan:
I close this CR for now. Even check predict_id, it is not a good solution. The issue is for some framework like langgraph where they run in a different context and cause the issue. For now i need to reset the context before running the tool

PR #8502, opened 2025-07-07
Add AI2ARC dataset

TomeHirata:
Add a dataset for https://huggingface.co/datasets/allenai/ai2_arc

### DIFF
--- a/dspy/datasets/ai2_arc.py
+++ b/dspy/datasets/ai2_arc.py
@@ -32,9 +32,9 @@ class AI2ARC:
         except Exception as e:
             raise RuntimeError(f"Failed to load {config_name} from {dataset_name}: {e}")
 
-        official_train = self._process_split(hf_dataset["train"]) if "train" in hf_dataset else []
-        official_dev = self._process_split(hf_dataset["validation"]) if "validation" in hf_dataset else []
-        official_test = self._process_split(hf_dataset["test"]) if "test" in hf_dataset else []
+        official_train = self._process_split(hf_dataset["train"])
+        official_dev = self._process_split(hf_dataset["validation"])
+        official_test = self._process_split(hf_dataset["test"])
 
         self.train = [dspy.Example(**x).with_inputs("question", "choices") for x in official_train]
         self.dev = [dspy.Example(**x).with_inputs("question", "choices") for x in official_dev]
@@ -91,8 +91,8 @@ def ai2_arc_metric(gold, pred, trace=None):
     Returns:
         bool: True if prediction matches gold answer
     """
-    pred_answer = str(pred.answer).strip().upper()
-    gold_answer = str(gold.answer).strip().upper()
+    pred_answer = _parse_arc_answer(pred.answer)
+    gold_answer = _parse_arc_answer(gold.answer)
 
     if len(pred_answer) == 1 and pred_answer in ["A", "B", "C", "D"]:
         return pred_answer == gold_answer
@@ -104,7 +104,7 @@ def ai2_arc_metric(gold, pred, trace=None):
     return False
 
 
-def parse_arc_answer(answer_text):
+def _parse_arc_answer(answer_text):
     """Parse answer from model output to extract the letter choice.
     
     Args:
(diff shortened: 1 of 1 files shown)


dspy/datasets/ai2_arc.py
```diff
@@ -0,0 +1,147 @@
...
+    # If no pattern found, return the first letter if it's A, B, C, or D
+    first_char = answer_text.upper()[0] if answer_text else ""
+    if first_char in ["A", "B", "C", "D"]:
+        return first_char
+
+    return answer_text
+
+
+# Convenience functions for loading specific subsets
+def ARC_Challenge():
```
chenmoneygithub: minor nit: lowercase for common word: `ARC_challenge` 

I am not sure if we need these 2 utils, no strong opinion though.

dspy/datasets/ai2_arc.py
```diff
@@ -0,0 +1,147 @@
...
+    
+    Args:
+        gold: Gold example with 'answer' field
+        pred: Predicted example with 'answer' field
+        trace: Optional trace (unused)
+        
+    Returns:
+        bool: True if prediction matches gold answer
+    """
+    pred_answer = str(pred.answer).strip().upper()
```
chenmoneygithub: Shall we call `parse_arc_answer` inside `ai2_arc_metric` so that users can do this in one hop?

dspy/datasets/ai2_arc.py
```diff
@@ -0,0 +1,147 @@
...
+
+        dataset_name = "allenai/ai2_arc"
+        config_name = f"ARC-{subset.title()}"
+
+        try:
+            hf_dataset = load_dataset(dataset_name, config_name)
+        except Exception as e:
+            raise RuntimeError(f"Failed to load {config_name} from {dataset_name}: {e}")
+
+        official_train = self._process_split(hf_dataset["train"]) if "train" in hf_dataset else []
```
chenmoneygithub: do we need the if-else check? I think this is deterministic 🤔

dspy/datasets/ai2_arc.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(self, subset: Literal["challenge", "easy"] = "challenge"):
+        if subset not in ["challenge", "easy"]:
+            raise ValueError("subset must be either 'challenge' or 'easy'")
+
+        self.subset = subset
+
+        from datasets import load_dataset
+
+        dataset_name = "allenai/ai2_arc"
```
chenmoneygithub, line 27: we can inline `dataset_name` and `config_name`: `load_dataset("allenai/ai2_arc", f"ARC-{subset.title()}")`  since they are not supposed to be accessed explicitly.

chenmoneygithub (commented):
This is pretty interesting, I am down to checking in this one. 

I will defer the decision to @okhat  if we want to expand the dataset collection.

arnavsinghvi11:
to my knowledge, the datasets we have in the repo only exist to maintain some past example notebooks/tests (though maybe we should actually just remove them given that we've ported these into tutorials - e.g. I don't see any instances of `dspy.datasets.colors`).

it's probably just cleaner to refactor out these datasets + their references in tutorials (so just adding ai2arc for demonstrative purposes in a tutorial) and only keep Dataset and DataLoader in dspy/datasets/

okhat:
yeah it’s rare that we add datasets anymore 😅

dspy/datasets/ai2_arc.py
TomeHirata: I included them to use them in the exception message too

dspy/datasets/ai2_arc.py
TomeHirata: Let's remove them :)

dspy/datasets/ai2_arc.py
TomeHirata: Good call!

TomeHirata:
Oh, thank you for the context. So we won't add more datasets and will deprecate existing datasets someday?

PR #8504, opened 2025-07-08
Add async support for dspy.Evaluate

chenmoneygithub:
We are introducing:
- `_execute_with_multithreading`  for running evaluation in multithreading
- `_execute_with_event_loop` for running evaluation in async concurrency.

The weird part is although technically async eval should run faster than multithreading because evaluation is IO bound task, but I am not noticing it consistently. 

Testing script is pasted below:

```
import asyncio

import dspy
from dspy.datasets.gsm8k import GSM8K

dspy.configure(lm=dspy.LM("openai/gpt-4o-mini", cache=False))


# Load math questions from the GSM8K dataset.
gsm8k = GSM8K()
gsm8k_trainset, gsm8k_devset = gsm8k.train[:50], gsm8k.dev[:100]


cot = dspy.ChainOfThought("question->answer")


def my_metric(args, pred):
    return 1.0 if pred.answer == args.answer else 0.0


evaluator = dspy.Evaluate(devset=gsm8k_devset, num_threads=50, display_table=False)


import time

start_time = time.time()
result = evaluator(cot, metric=my_metric)
end_time = time.time()
print(f"Time taken with multithreading: {end_time - start_time} seconds")
print(result)


async def main():
    return await evaluator.acall(cot, metric=my_metric)


start_time = time.time()
result = asyncio.run(main())
end_time = time.time()
print(f"Time taken with async: {end_time - start_time} seconds")
print(result)
```

70% time async runs faster than multithreading insignificantly, and 30% time it's the reverse. Two potential theories:
- Bottleneck happens on provider side, like rate limiting.
- Behind the scene litellm acompletion is not true async, I need to dig the code a bit.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -6,7 +6,6 @@ from typing import TYPE_CHECKING, Any, Callable
 if TYPE_CHECKING:
     import pandas as pd
 
-import asyncio
 
 import tqdm
 
@@ -161,7 +160,7 @@ class Evaluate:
         if callback_metadata:
             logger.debug(f"Evaluate.acall is called with callback metadata: {callback_metadata}")
         tqdm.tqdm._instances.clear()
-        results = await self._execute_with_event_loop(program, metric, devset, num_threads)
+        results = await self._execute_with_event_loop(program, metric, devset, num_threads, display_progress)
         return self._process_evaluate_result(devset, results, metric, display_table)
 
     def _resolve_call_args(self, metric, devset, num_threads, display_progress, display_table):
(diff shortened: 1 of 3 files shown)


dspy/evaluate/evaluate.py
```diff
@@ -126,71 +129,132 @@ def __call__(
...
+                vals = [r[-1] for r in results if r is not None]
+                nresults = sum(vals)
+                ntotal = len(vals)
+                pct = round(100 * nresults / ntotal, 1) if ntotal else 0
+                pbar.set_description(f"Average Metric: {nresults:.2f} / {ntotal} ({pct}%)")
+                pbar.update(1)
+                queue.task_done()
+
+        workers = [asyncio.create_task(worker()) for _ in range(num_threads)]
+        await asyncio.gather(*workers)
```
TomeHirata: q: Does `acyncio.gather` use multiple threads with this setup?

dspy/evaluate/evaluate.py
```diff
@@ -126,71 +129,132 @@ def __call__(
...
+
+    async def _execute_with_event_loop(
+        self,
+        program: "dspy.Module",
+        metric: Callable,
+        devset: List["dspy.Example"],
+        num_threads: int,
+    ):
+        queue = asyncio.Queue()
+        results = [None for _ in range(len(devset))]
```
TomeHirata: nit:
```suggestion
        results = [None] * len(devset)
````

okhat:
Wow so cool!

okhat:
Should this be added to dspy.Evaluate or deeper, like dspy.ParallelExecutor

dspy/evaluate/evaluate.py
chenmoneygithub: It doesn't use multiple threads, but does put those async workers to run asynchronously in the same event loop.

dspy/evaluate/evaluate.py
chenmoneygithub: done!

chenmoneygithub:
> Should this be added to dspy.Evaluate or deeper, like dspy.ParallelExecutor

The `_execute_with_event_loop` is a simplified version of `dspy.ParallelExecutor` and just focuses on getting the dspy.Evaluate to work, so I would keep it within `dspy.Evaluate`. 

I am not quite sure if we should have an async version of `dspy.ParallelExecutor`, because async doesn't have thread management, and it's pretty straightforward for the users to use `asyncio.gather()` to utilize the event loop. Then the only benefit will be the progress bar tracking 🤔

dspy/evaluate/evaluate.py
TomeHirata: Yeah, that's my understanding. Might it be misleading if we accept `num_threads` in async eval call?

chenmoneygithub: I missed this message. Yes it's a little confusing, basically `num_threads` means number of workers semantically. I want to avoid having `num_threads` for multithreading and `num_workers` for async setup, so I was reusing the same arg. I didn't rename it for backward compatibility.

TomeHirata: Make sense, I think it's fine to keep `num_threads` as long as we explain the behavior clearly in the docstring/api doc.

enigma:
Would this allow to also run optimizers on `aforward`-based Modules?

metavee:
Hi @chenmoneygithub, I was somewhat curious about this and tried to reproduce some of the measurements.

I made a few tweaks to the script, in a gist [here](https://gist.github.com/metavee/4a707ce7128061f880200c780798c286).

The tweaks are:

- create a custom LM that just wraps the OpenAI SDK. This eliminates the question about whether litellm does async correctly
- switch to the HotPotQA dataset. I was finding with GSM8K the questions vary a lot in difficulty / how many tokens the LLM would return, so it ended up making the data harder to interpret

After running this 20 times, I don't see a significant performance difference between async and threaded.

image

I don't think this is too surprising. I'm guessing any hypothetical overhead from switching threads is dominated by the LLM provider and waiting for the response.

PR #8513, opened 2025-07-10
Use collections for type hints

TomeHirata:
Removed pep585 from the non-check list and fixed violations


chenmoneygithub (approved):
Very nice, LGTM!

PR #8519, opened 2025-07-10
Provide language to dspy.Code

chenmoneygithub:
This is trickier than it looks like, because `language` is not a mutable field, and is by nature a class variable instead of pydantic field. 

Dynamic type hint is strongly discouraged: https://github.com/microsoft/pylance-release/issues/5457, so instead of providing something like `dspy.Code(language="java")` I am taking a different approach `dspy.Code["java"]` to achieve the same. See code below for example:

```
import dspy

dspy.configure(lm=dspy.LM("openai/gpt-4o-mini"))


    """Generate code to answer the question."""

    question: str = dspy.InputField(description="The question to answer")
    code: dspy.Code["java"] = dspy.OutputField(description="The code to execute")


predict = dspy.Predict(MySignature)

result = predict(question="Given an array, find if any of the two numbers sum up to 10")
print(result.code)
dspy.inspect_history()
```

The history is like:

```python
[2025-07-10T16:51:22.370240]

System message:

Your input fields are:
1. `question` (str): The question to answer
Your output fields are:
1. `generated_code` (Code_java): The code to execute
    Type description of Code_java: Code represented in a string, specified in the `code` field. If this is an output field, the code field should follow the markdown code block format, e.g. 
'''python
{code}
''' or 
'''cpp
{code}
'''
Programming language: java

[[ ## question ## ]]
{question}

[[ ## generated_code ## ]]
{generated_code}        # note: the value you produce must adhere to the JSON schema: {"type": "object", "properties": {"code": {"type": "string", "title": "Code"}}, "required": ["code"], "title": "Code_java"}

[[ ## completed ## ]]
        Generate code to answer the question.


User message:

[[ ## question ## ]]
Given an array, find if any of the two numbers sum up to 10

Respond with the corresponding output fields, starting with the field `[[ ## generated_code ## ]]` (must be formatted as a valid Python Code_java), and then ending with the marker for `[[ ## completed ## ]]`.


Response:

[[ ## generated_code ## ]]
```java
public class TwoSum {
    public static boolean hasTwoSum(int[] nums, int target) {
        // Create a HashSet to store numbers we've seen
        java.util.HashSet<Integer> seen = new java.util.HashSet<>();
        
        // Iterate through the array
        for (int num : nums) {
            // Check if the complement (target - num) exists in the set
            if (seen.contains(target - num)) {
                return true;
            }
            // Add the current number to the set
            seen.add(num);
        }
        
        // No pair found
        return false;
    }
    
    public static void main(String[] args) {
        // Example usage
        int[] arr1 = {2, 7, 11, 15};
        int[] arr2 = {3, 5, 1, 9};
        int[] arr3 = {1, 2, 3, 4};
        
        System.out.println("Array [2, 7, 11, 15] has two numbers that sum to 10: " + hasTwoSum(arr1, 10));
        System.out.println("Array [3, 5, 1, 9] has two numbers that sum to 10: " + hasTwoSum(arr2, 10));
        System.out.println("Array [1, 2, 3, 4] has two numbers that sum to 10: " + hasTwoSum(arr3, 10));
    }
}
```

[[ ## completed ## ]]

```


fswair:
It's not pythonic at all

chenmoneygithub:
@fswair  I disagree. 

It’s using `__class_getitem__`, which is part of PEP 560, a rule designed for customizing class behavior with brackets (like generics or DSLs). `__class_getitem__` isn't limited to types, but just a hook for custom class behavior. A close use case is `Literal["java"]`.

fswair:
Not everything that is supported by Python makes sense. Creating a class for every temporary language doesn't make sense. You are modifying Code.__class_getitem__ at runtime, and it always returns something different.

fswair:
Introducing a new dspy.CodeField is a better approach than generating pseudo-generic classes. It's also more user-friendly. While Python supports __class_getitem__, the current usage feels more like retrieving a mapped value.

chenmoneygithub:
@fswair 

> Introducing a new dspy.CodeField is a better approach than generating pseudo-generic classes. It's also more user-friendly.

No, special casing on OutputField is bad, which creates unnecessary learning overhead for users, and opens a can of worm for us.

Basically I am not a fan of discussing what is pythonic and what is not, since Python is keeping changing. If you think this approach raises security concerns or concerning user confusion, I am happy to discuss from there.

okhat:
@fswair What are you proposing as an alternative?

fswair:
> @fswair What are you proposing as an alternative?

A dspy.CodeField bases dspy.OutputField, so in the future, we can add more context to CodeField, by passing values to its parameters

okhat:
@fswair But Code can also be an input field, no?

fswair:
> @fswair But Code can also be an input field, no?

Yeah, probably, it would be. I was pretending it would be an OutputField based on example at the PR's description

PR #8522, opened 2025-07-11
fix UnicodeDecodeError: 'gbk' codec can't decode

nillwyc:
to support utf-8 on loading saved optimizations and dspy.ProgramOfThought


chenmoneygithub:
@arnavsinghvi11  cc our deno expert. Do you think the deno change is safe?

arnavsinghvi11:
LGTM - thanks @nillwyc !

Issue #8469, opened 2025-06-27
[Bug] Multiple LMs are being used in the module


armoucar-neon:
### What happened?

When I'm trying to load an optimized state I saved, I'm getting this error: `Multiple LMs are being used in the module`.

https://github.com/stanfordnlp/dspy/blob/dc9da7a97b263649f8d0a80ae7e5ae89d7baf192/dspy/primitives/module.py#L96-L99

My 2 LMs are the same:

```
{
  "agent.react": {
    "traces": [],
    "train": [],
    "demos": [...],
    "signature": {...},
    "lm": {
      "model": "openai/gpt-4.1-mini-2025-04-14",
      "model_type": "chat",
      "cache": true,
      "cache_in_memory": true,
      "num_retries": 3,
      "finetuning_model": null,
      "launch_kwargs": {},
      "train_kwargs": {},
      "temperature": 0.0,
      "max_tokens": 4000
    }
  },
  "agent.extract.predict": {
    "traces": [],
    "train": [],
    "demos": [...],
    "signature": {...},
    "lm": {
      "model": "openai/gpt-4.1-mini-2025-04-14",
      "model_type": "chat",
      "cache": true,
      "cache_in_memory": true,
      "num_retries": 3,
      "finetuning_model": null,
      "launch_kwargs": {},
      "train_kwargs": {},
      "temperature": 0.0,
      "max_tokens": 4000
    }
  },
  "metadata": {...},
  "custom_metrics": {...}
}
```

```
import json

def to_struct(o):
    return getattr(o, "__dict__", str(o))

lm_a = json.dumps(all_used_lms[0], default=to_struct, sort_keys=True)
lm_b = json.dumps(all_used_lms[1], default=to_struct, sort_keys=True)

lm_a == lm_b
# True
```

But:
```
if len(set(all_used_lms)) == 1
# False
```

Is this because [LM](https://github.com/stanfordnlp/dspy/blob/main/dspy/clients/lm.py#L24) doesn't implement `__eq__`?

### Steps to reproduce

.

### DSPy version

2.6.27

arnavsinghvi11:
Thanks for raising this @armoucar-neon ! yes we likely should have some comparison in `get_lm()`, probably through checking the lm states via `dump_state()`? cc @okhat

nikkindev:
@arnavsinghvi11  opened PR #8530  Let me know of any suggestions for alternatives. Thanks.

nikkindev:
[...]

PR #8530, opened 2025-07-14
Manage duplicate multiple LMs in the same module

nikkindev:
PR addresses  https://github.com/stanfordnlp/dspy/issues/8469

When checking for all used LMs from within `get_lm()` (in *dspy/primitives/module.py*), now the various LMs are compared with each other using `dump_state()` that returns a dictionary of the respective states containing values for the following attributes
```
model
model_type
cache
cache_in_memory
num_retries
finetuning_model
launch_kwargs
train_kwargs
```
Reminder: Only the LMs of instance type `Predict` are considered

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -93,10 +93,19 @@ class Module(BaseModule, metaclass=ProgramMeta):
     def get_lm(self):
         all_used_lms = [param.lm for _, param in self.named_predictors()]
 
-        if len(set(all_used_lms)) == 1:
-            return all_used_lms[0]
+        if len(all_used_lms) == 0:
+            raise ValueError("No LMs are being used in the module. There's no LM to return.")
 
-        raise ValueError("Multiple LMs are being used in the module. There's no unique LM to return.")
+        elif len(all_used_lms) == 1:
+            return all_used_lms.pop()
+
+        else:
+            first_lm = all_used_lms.pop()
+            for lm in all_used_lms:
+                if first_lm.dump_state() != lm.dump_state():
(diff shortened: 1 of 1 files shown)


okhat:
Ah this is interesting, thanks so much @nikkindev (and @arnavsinghvi11 !).

I'm thinking. One clear thought I have is that _directly_ comparing via dumping the state is not ideal. If we want to do this, we'd implement `__eq__` on `dspy.LM` or `dspy.BaseLM` and _there_ we can use dumping the state to compare. That way, here the comparison is done with just `==`.

But at a deeper level, two different objects _ARE_ two different LLMs. They may have different `lm.history` for example! So I kind of think we need to close this PR and the issue. The current behavior _seems_ to me like the intended behavior.

A deeper fix would be to de-duplicate LLMs in DSPy automatically. If you create a new LM that's identical to an existing LM, we just somehow "re-use" the old one. That seems magical and unnecessary tho.

tl;dr I think current behavior is fine. Thoughts?

nikkindev:
@okhat Yes! I was skeptical that only a subset of attributes were considered in `dump_state()` and was wondering if there was a specific reason for that. I also agree that the `__eq__()` implementation would be a better way to do this. If we use `dump_state()` within `__eq__()`, would we also want to add more attributes in `dump_state()`? (including the `lm.history`)

About de-duplication, that is definitely a deeper fix. But then I believe it needs some engineering decisions like do we clear the current history or do we keep them and so on. I have to agree that I haven't looked into the details of all attributes and their meanings in the LM class. I will spend some more time on it.

We can close this issue and PR. I will create a new issue once I tinker with a different solution and feel confident that I have something working. 

Thanks for the comments :)

PR #8534, opened 2025-07-16
Resolve "$def" for nested tool arg type

chenmoneygithub:
Previously when an argument is of type `list[MyPydanticModel]`, then `$def` will be kept inside the json schema. To talk to the LM, we should fully resolve these references.

### DIFF
--- a/tests/adapters/test_tool.py
+++ b/tests/adapters/test_tool.py
@@ -93,7 +93,11 @@ async def async_dummy_with_pydantic(model: DummyModel) -> str:
     return f"{model.field1} {model.field2}"
 
 
-async def async_complex_dummy_function(profile: UserProfile, priority: int, notes: str | None = None) -> dict[str, Any]:
+async def async_complex_dummy_function(
+    profile: UserProfile,
+    priority: int,
+    notes: list[Note] | None = None,
+) -> dict[str, Any]:
     """Process user profile with complex nested structure asynchronously.
 
     Args:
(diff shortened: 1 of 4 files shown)


tests/adapters/test_tool.py
```diff
@@ -88,9 +93,7 @@ async def async_dummy_with_pydantic(model: DummyModel) -> str:
     return f"{model.field1} {model.field2}"
 
 
-async def async_complex_dummy_function(
-    profile: UserProfile, priority: int, notes: str | None = None
-) -> dict[str, Any]:
+async def async_complex_dummy_function(profile: UserProfile, priority: int, notes: str | None = None) -> dict[str, Any]:
```
TomeHirata: Is it intentional to keep notes here an optional str?

tests/adapters/test_tool.py
```diff
@@ -177,6 +180,13 @@ def test_tool_from_function_with_pydantic_nesting():
     assert tool.args["profile"]["properties"]["contact"]["type"] == "object"
     assert tool.args["profile"]["properties"]["contact"]["properties"]["email"]["type"] == "string"
 
+    # Reference should be resolved for nested pydantic models
```
TomeHirata, line 188: nit: it's totally fine as it is, but since the expected json schema is somehow complex, it might be more readable if we define the entire dict and assert `assert tool.args == expected_schema`

tests/adapters/test_tool.py
chenmoneygithub: good catch!

tests/adapters/test_tool.py
chenmoneygithub: This is actually intentional, I tried to use the `expected_schema`, but realized it was too long to be readable. So I moved back to the dedicated fields' check.

PR #8535, opened 2025-07-16
Add flake8 comma rule

TomeHirata:
I've seen inconsistent comma usage among contributors, so adding flake8-comma rule for consistency. I'm open to not having this rule if we want to keep flexibility.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -43,7 +43,7 @@ class Adapter:
                 raise ValueError(
                     f"You provided an output field {tool_call_output_field_name} to receive the tool calls information, "
                     "but did not provide any tools as the input. Please provide a list of tools as the input by adding an "
-                    "input field with type `list[dspy.Tool]`."
+                    "input field with type `list[dspy.Tool]`.",
                 )
 
             if tool_call_output_field_name and litellm.supports_function_calling(model=lm.model):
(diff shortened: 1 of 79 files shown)


chenmoneygithub (commented):
ohh I would not introduce this rule, for example code like this:

```
        system_user_messages = self.format(  # returns a list of dicts with the keys "role" and "content"
            signature=signature, demos=demos, inputs=inputs,
        )
```
is against the official python rule PEP8: https://peps.python.org/pep-0008/#when-to-use-trailing-commas, which says it should be broken down into 3 lines if a trailing comma is used.

TomeHirata:
That's a good call. It's weird that the flake8 rule violates the pep8 recommendation. Let me check how to fix only multi-line trailing commas.

TomeHirata:
Thank you Chen for flagging this. I read the pep8 rule again, and it seems the following example actually does not violate the rule
```python
system_user_messages = self.format( 
    signature=signature, demos=demos, inputs=inputs,
)
```
The rule states, "it does not make sense to have a trailing comma on the same line as the closing delimiter", therefore this would be a violation while it does not appear in this PR.
```python
system_user_messages = self.format(signature=signature, demos=demos, inputs=inputs,)
```

okhat:
Honestly this seems minor enough and changing 79 files will end up just creating a lot of merge conflicts etc.

PR #8552, opened 2025-07-20
Remove user confirmation from MiproV2

TomeHirata:
User confirmation only exists for MiproV2. This PR removes the feature for consistency.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -1,8 +1,5 @@
 import logging
 import random
-import sys
-import textwrap
-import time
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Callable, Literal
 
@@ -112,7 +109,7 @@ class MIPROv2(Teleprompter):
         view_data_batch_size: int = 10,
         tip_aware_proposer: bool = True,
         fewshot_aware_proposer: bool = True,
-        requires_permission_to_run: bool = True,
+        requires_permission_to_run: bool = True, # deprecated
         provide_traceback: bool | None = None,
     ) -> Any:
         effective_max_errors = (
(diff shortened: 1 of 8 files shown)


dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -112,7 +109,7 @@ def compile(
         view_data_batch_size: int = 10,
         tip_aware_proposer: bool = True,
         fewshot_aware_proposer: bool = True,
-        requires_permission_to_run: bool = True,
+        requires_permission_to_run: bool = True, # deprecated
```
TomeHirata, line 112: Unfortunately, removing this argument will result in breaking the users' code that passes `requires_permission_to_run=False`. Do we think this breaking change is acceptable in DSPy 3? We could add `**kwargs` instead, but it would lead to a loss of undefined argument detection.

chenmoneygithub: We can introduce the wildcard `kwargs` and move the deprecated args there, and print a clear warning that it's deprecated.

TomeHirata: Yeah, that's feasible. But I would be cautious about introducing **kwarg as users won't be able to know when they mistakenly pass an undefined argument (or typo). What do you think?

chenmoneygithub: that's a solid concern! IMO it's a tradeoff between safety and unexpected user misbehavior. We do have wild card `kwargs` for a few methods now, e.g., https://github.com/stanfordnlp/dspy/blob/2e60022f9a524dfd406edc04444114f835d017a0/dspy/predict/predict.py#L82, so I won't worry much about it.

TomeHirata: Yeah, we have **kwargs in many modules, shall we remove the requires_permission_to_run parameter and add kwargs then? cc: @klopsahlong @omkar-sh

PR #8557, opened 2025-07-21
Migrate from `set-output` to `$GITHUB_OUTPUT` usage

kurtmckee:
This change resolves deprecation warnings thrown during GitHub workflow runs. [[recent example](https://github.com/stanfordnlp/dspy/actions/runs/16390769890)]

> image

https://github.blog/changelog/2022-10-10-github-actions-deprecating-save-state-and-set-output-commands/

Thanks for your work on dspy!


okhat:
Thanks so much @kurtmckee !

PR #8567, opened 2025-07-23
Adding support for mutliple mcp server connections

anotherdish:
Langchain supports session management and connection to multiple mcp servers (see [here](https://github.com/langchain-ai/langchain-mcp-adapters?tab=readme-ov-file#client-1)), which is a nice feature to have. This PR is heavily inspired/is based on langchain's implementation of their MultiServerMCPClient

### DIFF
--- a/dspy/utils/mcp.py
+++ b/dspy/utils/mcp.py
@@ -1,10 +1,13 @@
 from typing import TYPE_CHECKING, Any
 
 from dspy.adapters.types.tool import Tool, convert_input_schema_to_tool_args
+from langchain_mcp_adapters.sessions import create_session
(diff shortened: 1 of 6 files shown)


chenmoneygithub:
@anotherdish  Thanks for the PR! I am unsure if we want to rely on Langchain API to handle MCP inside DSPy. For multiple MCP connection, can we just run `dspy_tools.append(dspy.Tool.from_mcp_tool(session, tool))` for each of them?

anotherdish:
@chenmoneygithub ya that makes sense. are we interested in having a class to handle the multiple mcp connections (like the MultiServerMCPClient class)? their implementation is nice for its support of automatically creating new client sessions if one doesn't already exist

Issue #8540, opened 2025-07-17
[Bug] Unable to run in WASM (marimo) - cannot install `ujson`


aadityabhatia:
### What happened?

DSPy relies on [ujson](https://pypi.org/project/ujson/), which was put in maintenance mode. According to ujson's homepage-

> Users are encouraged to migrate to [orjson](https://pypi.org/project/orjson/) which is both much faster and less likely to introduce a surprise buffer overflow vulnerability in the future.

[marimo](https://marimo.app) runs on [pyodide](https://pyodide.org/) and does not support `ujson`, but `orjson` works as expected. I am planning to use DSPy within marimo WASM environment and hoping that switching from `ujson` to `orjson` will allow that.

### Steps to reproduce

[marimo snippet that attempts to import dspy and fails](https://marimo.app/#code/JYWwDg9gTgLgBCAhlUEBQaD6mDmBTAOzykRjwBNMB3YGACzgF44AiABgDoBGAFm4CYAtOTwA3NizSIwYJgmSoOAQRkAKAJQY0AAWlgOAYzwAbY2hEAzOJg0AuNHEdxQkWHHIBnMAE8HTqHgwAK5QBFrAVtgEiCB42EzMLNhIwATYLPZOcHocUEEEGmhAA)

Within marimo, click on "Manage packages" in the left menu bar and attempt to install `dspy` or `ujson`. The following error appears-

> Can't find a pure Python 3 wheel for 'ujson'.
See: https://pyodide.org/en/stable/usage/faq.html#why-can-t-micropip-find-a-pure-python-wheel-for-a-package

### DSPy version

2.6.27

okhat:
Oh wow. Sad for ujson to go. I guess we should switch to orjson then :/

aadya940:
@okhat Can I work on this?

0xEval:
+1
Thanks for looking into this. It prevents deploying WASM notebooks that use DSPy.

0xEval:
Unfortunately, it seems `ujson` is not the only culprit here. A trickier problem would be Huggingface's `tokenizer` lib (see: https://github.com/pyodide/pyodide-recipes/issues/199)

PR #8584, opened 2025-07-29
Port to `orjson` from `ujson`

aadya940:
As specified in #8540 , `ujson` is in maintanance mode.


dspy/clients/databricks.py
```diff
@@ -318,7 +318,7 @@ def _save_data_to_local_file(train_data: list[dict[str, Any]], data_format: Trai
             elif data_format == TrainDataFormat.COMPLETION:
                 _validate_completion_data(item)
 
-            f.write(ujson.dumps(item) + "\n")
+            f.write(orjson.dumps(item).decode() + "\n")
```
kurtmckee, line 321: I recommend changing the `open(file_path, "w")`, above, to "wb" mode to eliminate the decodes happening here.

The end result will be:

```suggestion
            f.write(orjson.dumps(item) + b"\n")
```

dspy/clients/utils_finetune.py
```diff
@@ -60,7 +60,7 @@ def get_finetune_directory() -> str:
 def write_lines(file_path, data):
     with open(file_path, "w") as f:
         for item in data:
-            f.write(ujson.dumps(item) + "\n")
+            f.write(orjson.dumps(item).decode() + "\n")
```
kurtmckee, line 63: As above, I recommend eliminating the calls to `.decode()`:

```python
    with open(file_path, "wb") as f:
        for item in data:
            f.write(orjson.dumps(item) + b"\n")
```

This suggestion applies to the other change in this file.

dspy/primitives/base_module.py
```diff
@@ -249,7 +249,7 @@ def load(self, path):
 
         if path.suffix == ".json":
             with open(path, encoding="utf-8") as f:
-                state = ujson.loads(f.read())
+                state = orjson.loads(f.read().encode('utf-8'))
```
kurtmckee, line 252: Last place I'll leave feedback, but in general the suggestion is "don't decode content only to re-encode it immediately:

```python
            with open(path, "rb") as f:
                state = orjson.loads(f.read())
```

Also, these are pathlib objects so this is more ideal:

```python
            state = orjson.loads(path.read_bytes())
```

okhat:
Thanks so much @aadya940 and thanks @kurtmckee for the comments!

@aadya940 Mind addressing the failures (ruff mostly? or maybe some tests?) and checking out the comments (I didn't dive into them)

dspy/clients/databricks.py
chenmoneygithub: `decode()` is beneficial here - one nice benefit of saving state as json is the file is human readable, so that users can copy/paste the demos/instructions to other parts of the pipeline. I have seen many users only use DSPy as the prompt optimization tool, and export the optimized prompt (instruction + demos).

chenmoneygithub:
@aadya940  Thanks for the PR! the most concerning part is if switching to orjson can still load a saved program before this PR. I am running some testing.

dspy/clients/databricks.py
kurtmckee: `.decode()` is not doing the beneficial work that you think it's doing.

It's creating this pipeline:

```python
with open(file_path, "w") as f:
```

Opening the file in text mode means that Python must internally encode the file _when it actually writes the file_. [Python does NOT use UTF-8 for this!](https://docs.python.org/3/library/functions.html#open) It defaults to the system-specific encoding. On Windows in the U.S., for example, this is usually something like ISO-8859-1.

```python
orjson.dumps(item)
```

This produces a bytes object -- perfect for writing to disk!

```python
orjson.dumps(item).decode()
```

This decodes the bytes to UTF-8, and is a performance problem because now Python must re-encode the content before it can be written to disk. What a waste.

```python
orjson.dumps(item).decode() + "\n"
```

Another performance problem -- the entire JSON string must now be copied in memory just so a newline can be added to the end in memory.

```python
f.write(orjson.dumps(item).decode() + "\n")
```

Here's the full pipeline in this one line of code:

* orjson produces a UTF-8-encoded bytes instance.
* The bytes are decoded using UTF-8.
* A new string is constructed in memory, with a newline added to the end.
* Python re-encodes the string to the default system character set, which varies from one system to the next.
* Python writes the bytes instance it just created to disk.

The `.decode()` is not beneficial. It's a waste of cycles.

The most performant option that I'm aware of, which eliminates the decode-with-UTF-8/re-encode-with-??? roundtrip, and eliminates the construct-a-new-string-in-memory is:

```python
with open(file_path, "wb") as f:
    f.write(orjson.dumps(item))
    f.write(b"\n")
```

chenmoneygithub: Yes I am aware of the performance issue. I was worrying about the displayed text when opening the json file is not readable, but it's not really the case since most text editors decode by utf-8 by default.

PR #8587, opened 2025-07-29
Explicitly mark the end of streaming

chenmoneygithub:
It makes users' life easier if there is an explicit signal marking the end of the streaming of one field.


TomeHirata (approved):
LGTM!

PR #8589, opened 2025-07-30
Update OpenAI model list

ZachParent:
## Problem

- I noticed that the list of OpenAI models is out of date. For example, `gpt-4.1` family models are not present
- This causes failures when attempting to finetune, for example using `BootstrapFinetune`

## Solution

- I pulled the currently available models using 

```python
from openai import OpenAI

client = OpenAI()
models = client.models.list()
models_ls = list(models)

open_models = [model for model in models_ls if ":" not in model.id]
open_model_ids = sorted([model.id for model in open_models])
open_model_ids
```

- I updated the list of models in `clients/openai.py` to this list

## Testing

- I would like to test this but I'm not sure how. We could set up a test which queries the `models.list()` endpoint and checks that it matches this list, but this may be brittle and subject to OpenAI changes

## Alternatives

- While I'm confident this list needs updating, I'm not confident this hard-coded list is the best strategy.
- We could instead call the `models.list()` API when checking `is_provider_model(model)`, per the suggestion here
https://github.com/ZachParent/dspy/blob/695d88e5b14a86a40c7c5c5f6b3da841860b56ec/dspy/clients/openai.py#L138-L142


okhat:
Thank you! But this is adding many models that can't be finetuned, like dalle. We should probably switch to something more sustainable here...

ZachParent:
Understood, and agreed! Let me know if you have a direction to look into. I don't see a way in the OpenAI docs to filter models by capability (like finetuning). Maybe a third party provides an API with this list? Otherwise, I'm not sure what's possible beyond hard-coding.

chenmoneygithub (commented):
Thanks for the PR! 

The reason for maintaining this hardcoded list is to make it possible to infer the model provider if there is no prefix provided: https://github.com/stanfordnlp/dspy/blob/8f773dc75e34da21b89480e8bc71df503d5ab5fc/dspy/clients/lm.py#L238. I think we can actually force the users to specify model name in the format `{provider_name}/{model_name}`, the gain of being able to write `gpt-5` instead of `openai/gpt-5` doesn't quite worth the effort of maintaining this hardcode list IMO.

okhat:
@chenmoneygithub Yes I agree, force users to specify

Issue #8590, opened 2025-07-30
[Bug] `trust_remote_code` is not supported anymore.


keviddles:
### What happened?

I am trying to run the [Optimizing prompts for a ReAct agent](https://dspy.ai/#__tabbed_3_1) code from Getting Started III on the homepage.

It fails with "`trust_remote_code` is not supported anymore."

### Steps to reproduce

Running:

```
import dspy
from dspy.datasets import HotPotQA

dspy.configure(lm=dspy.LM("openai/gpt-4o-mini"))

def search_wikipedia(query: str) -> list[str]:
    results = dspy.ColBERTv2(url="http://20.102.90.50:2017/wiki17_abstracts")(query, k=3)
    return [x["text"] for x in results]

trainset = [x.with_inputs('question') for x in HotPotQA(train_seed=2024, train_size=500).train]
react = dspy.ReAct("question -> answer", tools=[search_wikipedia])

tp = dspy.MIPROv2(metric=dspy.evaluate.answer_exact_match, auto="light", num_threads=24)
optimized_react = tp.compile(react, trainset=trainset)
```

Results in:

```
`trust_remote_code` is not supported anymore.
Please check that the Hugging Face dataset 'hotpot_qa' isn't based on a loading script and remove `trust_remote_code`.
If the dataset is based on a loading script, please ask the dataset author to remove it and convert it to a standard format like Parquet.
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[17], line 10
      7     results = dspy.ColBERTv2(url="http://20.102.90.50:2017/wiki17_abstracts")(query, k=3)
      8     return [x["text"] for x in results]
---> 10 trainset = [x.with_inputs('question') for x in HotPotQA(train_seed=2024, train_size=500).train]
     11 react = dspy.ReAct("question -> answer", tools=[search_wikipedia])
     13 tp = dspy.MIPROv2(metric=dspy.evaluate.answer_exact_match, auto="light", num_threads=24)

[...]

PR #8591, opened 2025-07-30
Update docs to specify datasets==3.6.0

keviddles:
https://github.com/stanfordnlp/dspy/issues/8590

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -330,7 +330,7 @@ Given a few tens or hundreds of representative _inputs_ of your task and a _metr
     Examples below rely on HuggingFace/datasets, you can install it by the command below.
 
     ```bash
-    > pip install -U datasets
+    > pip install -U datasets==3.6.0
     ```
 
     === "Optimizing prompts for a ReAct agent"


okhat:
Hmm, what's a more fundamental fix? What was trusting remote code replaced with?

(OR at least let's not pin an exact version. For example pin anything strictly < 4.0)

lhoestq:
Hi ! I updated the hotpot_qa dataset to make it compatible with datasets>=4 here: https://huggingface.co/datasets/hotpotqa/hotpot_qa/discussions/9

it should be all good now :)

I also pinged the authors of PolyAI/banking77 to update their datasets at https://huggingface.co/datasets/PolyAI/banking77/discussions/6

okhat:
Amazing, thank you @lhoestq !!

okhat:
I suppose we should edit our HotPotQA class then to remove that kwarg

lhoestq:
Yes otherwise you'll get some warnings that this kwarg should be removed (it won't error out though since the dataset doesn't have remote code anymore anyways, and it will only show a warning)

Issue #8570, opened 2025-07-25
[Bug] Formatting LLM calls with  dspy.Image's is very slow because of json parsing the image string representation


jmhb0:
### What happened?

I profile image captioning, where the result is already cached. 
For an image with size (1000,1000), running Predict 10 times (cached) takes 19.5 seconds. 
19 of those seconds are caused by calls to `json_repair.loads` in `adapters/types/base_type.py`, the function `split_message_content_for_custom_types`. 
The slowdown is because `loads` is been called on the string representation of the base64 image, which is big.
(Note if we resize the image to (500,500), the 10 calls take 6.9 seconds)

The issue is that in an earlier function, `adapters.base.Adapter.format`, the base64 image is converted to a string and concated with some regular text ... and this later function is now breaking it up, but the parsing is too slow. 

### Steps to reproduce

Here is a script that does profiling .... you just need to change API_KEY.
```
from PIL import Image
import dspy
import cProfile
import pstats
import io
import requests
from io import BytesIO

OPENAI_API_KEY = <>

lm = dspy.LM("openai/gpt-4o-mini", api_key=OPENAI_API_KEY)
dspy.configure(lm=lm, num_threads=1)


def profile_n_cached_caption_calls(n: int, img):

    class CaptionImage(dspy.Signature):
        """Caption an image."""

        image = dspy.InputField()
        caption = dspy.OutputField()

    captioner = dspy.Predict(CaptionImage)

    # run once to make sure the result is cached
    test_run = captioner(image=img)

    pr = cProfile.Profile()
    pr.enable()
    # caption n times but it's all cached
    for _ in range(n):
        result = captioner(image=img)
    pr.disable()
    s = io.StringIO()
    ps = pstats.Stats(pr, stream=s).sort_stats("cumulative")
    ps.print_stats(10)  # print top 10 slowest functions

    print(
        f"Here is a sample captioning result (which should be 'a bird' or something like that): {result.caption}"
    )
    print("main stats")
    print(s.getvalue())


# load this bird image from wikipedia
url = "https://upload.wikimedia.org/wikipedia/commons/thumb/a/a7/RedcrestedTuraco.jpg/500px-RedcrestedTuraco.jpg"
img_pil = Image.open(BytesIO(requests.get(url).content))
target_size = (1000, 1000)
if target_size is not None:
    img_pil = img_pil.resize(target_size)
print(f"Image size: {img_pil.size}")
[...]

PR #8596, opened 2025-08-01
faster json parsing for images in chatadapter

jmhb0:
This fixes the issue I shared in https://github.com/stanfordnlp/dspy/issues/8570

### DIFF
--- a/dspy/adapters/types/base_type.py
+++ b/dspy/adapters/types/base_type.py
@@ -64,7 +64,11 @@ class Type(pydantic.BaseModel):
     def serialize_model(self):
         formatted = self.format()
         if isinstance(formatted, list):
-            return f"{CUSTOM_TYPE_START_IDENTIFIER}{self.format()}{CUSTOM_TYPE_END_IDENTIFIER}"
+            try:
+                str_rep = json.dumps(formatted)
+            except TypeError:
+                str_rep = formatted
+            return f"{CUSTOM_TYPE_START_IDENTIFIER}{str_rep}{CUSTOM_TYPE_END_IDENTIFIER}"
         return formatted
 
 
@@ -115,9 +119,13 @@ def split_message_content_for_custom_types(messages: list[dict[str, Any]]) -> li
             # Parse the JSON inside the block
             custom_type_content = match.group(1).strip()
             try:
-                parsed = json_repair.loads(custom_type_content)
+                try:
+                    parsed = json.loads(custom_type_content)
+                except json.JSONDecodeError:
+                    parsed = json_repair.loads(custom_type_content)
                 for custom_type_content in parsed:
                     result.append(custom_type_content)
+
             except json.JSONDecodeError:
                 # fallback to raw string if it's not valid JSON
                 parsed = {"type": "text", "text": custom_type_content}
--- a/dspy/adapters/types/base_type.py
+++ b/dspy/adapters/types/base_type.py
@@ -116,8 +116,8 @@ def split_message_content_for_custom_types(messages: list[dict[str, Any]]) -> li
             custom_type_content = match.group(1).strip()
             try:
                 try:
-                    custom_type_content = custom_type_content.replace("'", '"')
-                    parsed = json.loads(custom_type_content)
+                    # Replace single quotes with double quotes to make it valid JSON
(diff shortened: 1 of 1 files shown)


chenmoneygithub (approved):
Thanks for the fix, LGTM!

chenmoneygithub (changes requested):
Just found out the unit test is failing, could you fix the failed tests? thanks! 

https://github.com/stanfordnlp/dspy/actions/runs/16664091791/job/47268270362?pr=8596

jmhb0:
Hi @chenmoneygithub, I made a change and now `pytest tests/adapters` passes all tests. 

In the first commit, in `serialize_model`, I changed `str` to `json.dumps` which often works, but if the type is part of a nested  object like in the list of images in `tests/adapters/test_json_adapter.py::test_json_adapter_formats_image_with_nested_images`, we have 
```
class ImageWrapper(pydantic.BaseModel):
        images: list[dspy.Image]
        tag: list[str]
```
The image json is serialized once, and then its string is passed through `serialize_model` again for `ImageWrapper`. So the `"` character - which is json formatting - gets double-escaped. So later, `json.loads` interprets the double-escaped content as a literal string rather than JSON structure, causing parsing to fail (and even json_repair parses it incorrectly). One fix you _could_ do is to manually remove the double-escape while decoding. But this only works for one layer of nesting, and so a more general solution would be more complicated. 

So instead, I leave `serialize_model` as is: keep using `str()` instead of `json.dumps`. Then in decoding, just string replace the `'` with `"` and call `json.loads` instead of `json_repair.loads`. This assumes that in the string representation of these objects, single quotes are only used as JSON delimiters, not as literal characters in data.

chenmoneygithub:
@jmhb0 Thanks! But I am confused why we need this `'` => `"` replacement? I think `json.loads()` work fine with single quotes too.

jmhb0:
@chenmoneygithub yep, no worries. here's an example showing that `'` is a problem for `json.loads`
```
>>> import json
>>> x = {"type": "image_url", "image_url": {"url": "https://example.com/image.jpg"}}
>>> x_as_str = str(x)
>>> x_as_str
{'type': 'image_url', 'image_url': {'url': 'https://example.com/image.jpg'}}
>>> json.loads(x_as_str)
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 338, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 354, in raw_decode
    obj, end = self.scan_once(s, idx)
               ^^^^^^^^^^^^^^^^^^^^^^
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
```
So `loads` only works for double quotes, which you would get by using `jsom.dumps()` instead of `str()`
```
>>> x_as_jsondumps = json.dumps(x)
>>> x_as_jsondumps
'{"type": "image_url", "image_url": {"url": "https://example.com/image.jpg"}}'
>>> json.loads(x_as_jsondumps)
{'type': 'image_url', 'image_url': {'url': 'https://example.com/image.jpg'}}
```
The solution to this before was to use `json_repair.loads()` instead of `json.loads()`, but this is very slow when the image object is the base64 image representation. This slowness is why I made the bug report.

arnavsinghvi11:
+1 to this change @jmhb0 - tested locally and has made runs with images much faster! LGTM from my side (unless @chenmoneygithub has any other concerns)

chenmoneygithub (approved):
I pushed a commit to remove the in-place modification.

Again, very nice catch and fix, thanks for the effort!

jmhb0:
> I pushed a commit to remove the in-place modification.

Yep good point. 

Thanks both for accepting the PR !

PR #8599, opened 2025-08-03
remove history and callbacks from module saving

arnavsinghvi11:
fixes saving program logic, which currently becomes bloated by storing history and callbacks that can become large and are unnecessary for loading


okhat:
Tagging @chenmoneygithub , what do you think?

PR #8602, opened 2025-08-06
Increment python version for pre-commit

TomeHirata:
Since we dropped 3.9 support, we should use 3.10 in the pre-commit config


chenmoneygithub (commented):
LGTM! qq - shall we bump to 3.11 directly?

TomeHirata:
I think we should use the lowest supported version for development in order not to accidentally cause issues for lower versions!

PR #8609, opened 2025-08-07
Update Arbor GRPO Integration

Ziems:
Clean up job_id tracking interface to match `Arbor:main`


udemirezen:
Hi for this framework, when we use even 0.5B parameter model, it consumes too much memory. Is it normal?

PR #8614, opened 2025-08-08
Custom BAMLAdapter to improve structured outputs

### LINKED ISSUE #8613: [Feature]: Improve structured output performance with a custom adapter

DSPy by default uses JSON schema in the prompt it creates for the LM. When using complex nested Pydantic models, the rendering of the schema is very messy and verbose. Example below.

Image

Providing users with a custom adapter that's cleaner, more concise and easier to read (for both the human and the LM), similar to how [BoundaryML](https://boundaryml.com/blog/type-definition-prompting-baml) does it in their language, BAML, works really well across the board. It's also cost-saving because the LLM needs fewer tokens to achieve the better result.

### Additional Context

Structured outputs is an incredibly common use case for LMs these days. DSPy already has excellent primitives to avoid the user writing prompts -- however, the way the signature's underlying data models (typically Pydantic) are translated to JSON schema means that there is a) a lot more verbosity (i.e., more tokens) than are really needed for LMs to understand the schema, and b) the LM gets "lost" in the messiness of JSON schema, where important context about a field can be separated in space from the actual field definition.

Consider this nested Pydantic model for a `Patient`:

```py
class PersonNameAndTitle(BaseModel):
    family: str | None = Field(default=None, alias="surname")
    given: list[str] | None = Field(
        default=None,
        alias="givenName",
        description="First and middle names, as present",
    )
    prefix: str | None = Field(default=None, alias="title of the person, e.g., Dr., Mr., Mrs., Ms.")

class Address(BaseModel):
    line: str | None = Field(default=None, alias="street")
    city: str | None
    state: str | None
    postalCode: str | None
    country: Literal["US"] | None

class Immunization(BaseModel):
    traits: list[str] | None = Field(
        default=None,
        description="Text describing the name and traits of the immunization",
    )
    status: Literal["completed"] | None = Field(
        default=None,
        description="If no traits are present, then the status cannot be determined",
    )
    occurrenceDate: str | None = Field(default=None, description="ISO-8601 format for date")

class Substance(BaseModel):
    category: Literal["environment", "food", "medication", "other"]
    name: str | None
    manifestation: str | None = Field(
        default=None,
        description="Text describing the manifestation of the allergy or intolerance",
    )

class Allergy(BaseModel):
    substance: list[Substance] = Field(description="Substances the patient is allergic to")

class Patient(BaseModel):
    record_id: int | None = Field(default=None)
    name: PersonNameAndTitle | None
    age: int | None
    gender: Literal["male", "female"] | None
    birthDate: str | None = Field(default=None, description="ISO-8601 format for date")
    phone: str | None = Field(default=None, description="Phone number of the patient")
    email: str | None = Field(default=None, description="Email address of the patient")
    maritalStatus: Literal["Married", "Divorced", "Widowed", "NeverMarried"] | None
    address: Address | None = Field(default=None, description="Residence address of the patient")
    allergy: list[Allergy] | None = Field(default=None)
```

Instead of the messy, single-line JSON schema above, we can instead make the representation far more concise as follows:
```
[[ ## schema ## ]]:
{
  record_id: int or null,
  name:   {
    surname: string or null,
    // First and middle names, as present
    givenName: string[] or null,
    title of the person, e.g., Dr., Mr., Mrs., Ms.: string or null,
  } or null,
  age: int or null,
  gender: "male" or "female" or null,
  // ISO-8601 format for date
  birthDate: string or null,
  // Phone number of the patient
  phone: string or null,
  // Email address of the patient
  email: string or null,
  maritalStatus: "Married" or "Divorced" or "Widowed" or "NeverMarried" or null,
  // Residence address of the patient
  address:   {
    street: string or null,
    city: string or null,
    state: string or null,
    postalCode: string or null,
    country: "US" or null,
  } or null,
  allergy: [
    {
      // Substances the patient is allergic to
      substance: [
        {
          category: "environment" or "food" or "medication" or "other",
          name: string or null,
          // Text describing the manifestation of the allergy or intolerance
          manifestation: string or null,
        }
      ],
    }
  ] or null,
}
```

In this approach, which is very similar to how BAML does it, Pydantic field descriptions are converted to comments (beginning with `//`), and they're placed *right next to* the field that they pertain to. This helps the LM not get lost in the context and to better understand the structure of what it's expected to generate.

Code and experiments for the custom adapter proving this hypothesis with actual numbers coming soon!

prrao87:
Closes #8613.

This PR provides a custom adapter named `BAMLAdapter` that subclasses the existing `JSONAdapter` to improve the quality of structured outputs generated by DSPy pipelines. The proposed adapter is inspired by another open source (Apache 2 licensed) project called [BAML](https://github.com/BoundaryML/baml), who created this format to improve upon JSON schema for their users based on the reasons they highlighted [in this blog post](https://boundaryml.com/blog/type-definition-prompting-baml).

The new adapter introduces no breaking changes, is non-intrusive, and is *purely opt-in* for those users (like myself) who need  better structured outputs from DSPy pipelines. So I figured it makes sense to allow the larger community to also benefit.

## Changes

Introduces two new files:
- `adapters/baml_adapter.py`
- `tests/adapters/test_baml_adapter.py`

## Usage

```py
import dspy
from pydantic import BaseModel, Field
from typing import Literal
from dspy.adapters.baml_adapter import BAMLAdapter

# 1. Define Pydantic models
class PatientAddress(BaseModel):
    street: str
    city: str
    country: Literal["US", "CA"]

class PatientDetails(BaseModel):
    name: str = Field(description="Full name of the patient.")
    age: int
    address: PatientAddress | None

# 2. Define a signature using the Pydantic model as an output field
class ExtractPatientInfo(dspy.Signature):
    '''Extract patient information from the clinical note.'''
    clinical_note: str = dspy.InputField()
    patient_info: PatientDetails = dspy.OutputField()

# 3. Configure dspy to use the custom adapter
llm = dspy.OpenAI(model="gpt-4.1-mini")
dspy.configure(lm=llm, adapter=BAMLAdapter())

# 4. Run your program
extractor = dspy.Predict(ExtractPatientInfo)
note = "John Doe, 45 years old, lives at 123 Main St, Anytown. Resident of the US."
result = extractor(clinical_note=note)
print(result.patient_info)

# Expected output:
# PatientDetails(name='John Doe', age=45, address=PatientAddress(street='123 Main St', city='Anytown', country='US'))
```

## Tests

- `BAMLAdapter` produces more readable schemas than JSON schema with simplified type notation (e.g., `name: string`)
- Maintains full compatibility with the parsing behaviour of `JSONAdapter` (no regressions)
- Handles complex nested structures including lists of Pydantic models with proper bracket notation
- Supports all DSPy multimodal types (Images, Tools, Code, History)
- Gracefully handles edge cases like circular references and schema generation errors

## Experiments

I've run extensive experiments on a more complex data model [here](https://github.com/prrao87/dspy-baml-experiments/tree/main/src/dspy). It's clear from the results below that the `BAMLAdapter` performs universally better (or no worse) than JSON schema, regardless of the LM chosen. The numbers below show the percentage of total exact matches (between the model's output and the source of truth data) for every important field in the schema for a dataset of 200 such deeply nested Pydantic models with all sorts of edge cases.

| Model | JSON schema | BAML Adapter |
|-------|-------------------:|-------------------:|
| `google/gemma-3-27b-it` | 88.9% | **94.4%** |
| `mistralai/devstral-medium` | 91.2% |  **93.8%**  |
| `google/gemini-2.0-flash-001` | 90.2% | **94.8%** |
| `openai/gpt-4.1-mini`| 93.8% | **95.3%** |
| `google/gemini-2.5-flash` | 90.6% | **95.4%** |

The main reason the `BAMLAdapter` does better than JSON schema is that it makes better use of the `Field` descriptions in the Pydantic models that the user provides, and it also compresses the schema in a lossless manner by representing the same information with far fewer tokens, allowing the LM to better understand the structure of what it's expected to output. Example transformation below.
image

## Takeaways

- For structured outputs, this adapter is a win-win: it both improves performance and reduces cost (due to fewer tokens being needed to explain the schema to the LM). The difference is more stark, the more complex the schema.
- The performance improvements come *for free*, even before any optimization (all we're really doing is changing the way we represent the schema to the LM)
- Users can opt into the `BAMLAdapter` - if structured outputs are not important, or users have very simple schemas, the benefits aren't that much and they can continue using the `JSONAdapter` as is

I'd be happy to add documentation on this if required, but I hope the rationale, experiments and tests make sense, and that this can be considered to merge so a much larger user community benefits!

---

## Updates

Added a few more changes (and more robust tests) to make the structure adhere more closely to previously existing system messages. Here's how it looked, before and after in my experiments on [this example]https://github.com/prrao87/dspy-baml-experiments/tree/main/src/dspy):

### Before (standard adapter with JSON schema)
```
System message:

Your input fields are:
1. `note` (PatientNote):
Your output fields are:
1. `patient` (Patient):

[[ ## note ## ]]
{note}

[[ ## patient ## ]]
{patient}        # note: the value you produce must adhere to the JSON schema: {"type": "object", "$defs": {"Address": {"type": "object", "properties": {"city": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "City"}, "country": {"anyOf": [{"type": "string", "const": "US"}, {"type": "null"}], "title": "Country"}, "postalCode": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "Postalcode"}, "state": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "State"}, "street": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "title": "Street"}}, "required": ["city", "state", "postalCode", "country"], "title": "Address"}, "Allergy": {"type": "object", "properties": {"substance": {"type": "array", "description": "Substances the patient is allergic to", "items": {"$ref": "#/$defs/Substance"}, "title": "Substance"}}, "required": ["substance"], "title": "Allergy"}, "PersonNameAndTitle": {"type": "object", "properties": {"family": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "surname", "title": "Family"}, "given": {"anyOf": [{"type": "array", "items": {"type": "string"}}, {"type": "null"}], "default": null, "description": "Given names (first and middle names)", "title": "Given"}, "prefix": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "title of the person, e.g., Dr., Mr., Mrs., Ms.", "title": "Prefix"}}, "title": "PersonNameAndTitle"}, "Substance": {"type": "object", "properties": {"category": {"type": "string", "enum": ["environment", "food", "medication", "other"], "title": "Category"}, "manifestation": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "Text describing the manifestation of the allergy or intolerance", "title": "Manifestation"}, "name": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "Name"}}, "required": ["category", "name"], "title": "Substance"}}, "properties": {"address": {"anyOf": [{"$ref": "#/$defs/Address"}, {"type": "null"}], "default": null, "description": "Residence address of the patient"}, "age": {"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "Age"}, "allergy": {"anyOf": [{"type": "array", "items": {"$ref": "#/$defs/Allergy"}}, {"type": "null"}], "default": null, "title": "Allergy"}, "birthDate": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "ISO-8601 format for date", "title": "Birthdate"}, "email": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "Email address of the patient", "title": "Email"}, "gender": {"anyOf": [{"type": "string", "enum": ["male", "female"]}, {"type": "null"}], "title": "Gender"}, "maritalStatus": {"anyOf": [{"type": "string", "enum": ["Married", "Divorced", "Widowed", "NeverMarried"]}, {"type": "null"}], "title": "Maritalstatus"}, "name": {"anyOf": [{"$ref": "#/$defs/PersonNameAndTitle"}, {"type": "null"}]}, "phone": {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, "description": "Phone number of the patient", "title": "Phone"}, "record_id": {"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null, "title": "Record Id"}}, "required": ["name", "age", "gender", "maritalStatus"], "title": "Patient"}

[[ ## completed ## ]]
        - Do not infer any information that is not explicitly mentioned in the text.
        - If you are unsure about any field, leave it as None.
```

### After (with BAML adapter)
```
System message:

Your input fields are:
1. `note` (PatientNote):
Your output fields are:
1. `patient` (Patient):
You must produce a single, valid JSON object that strictly adheres to the following schema. Do not output anything else.

Inputs will have the following structure:

[[ ## note ## ]]
{note}

Outputs will be a JSON object with the following fields.

{
  record_id: int or null,
  name:   {
    # surname
    family: string or null,
    # Given names (first and middle names)
    given: string[] or null,
    # title of the person, e.g., Dr., Mr., Mrs., Ms.
    prefix: string or null,
  } or null,
  age: int or null,
  gender: "male" or "female" or null,
  # ISO-8601 format for date
  birthDate: string or null,
  # Phone number of the patient
  phone: string or null,
  # Email address of the patient
  email: string or null,
  maritalStatus: "Married" or "Divorced" or "Widowed" or "NeverMarried" or null,
  # Residence address of the patient
  address:   {
    # alias: street
    line: string or null,
    city: string or null,
    state: string or null,
    postalCode: string or null,
    country: "US" or null,
  } or null,
  allergy: [
    {
      # Substances the patient is allergic to
      substance: [
        {
          category: "environment" or "food" or "medication" or "other",
          name: string or null,
          # Text describing the manifestation of the allergy or intolerance
          manifestation: string or null,
        }
      ],
    }
  ] or null,
}
        - Do not infer any information that is not explicitly mentioned in the text.
        - If you are unsure about any field, leave it as None.
```
The only part that's changed is the way we format the schema.

## Updated results

The results with the updated schema format have been tested for multiple input and output fields - the system message is identical to the earlier system message, which means that the cleaner formatting of the schema is indeed what causes the improved results. Updated results table below on my sample nested Pydantic schema experiments.

| Model | JSON schema | BAML Adapter |
|-------|-------------------:|-------------------:|
| `google/gemma-3-27b-it` | 88.9% |  **94.4%** |
| `mistralai/devstral-medium` | 91.2% |  **94.4%**  |
| `google/gemini-2.0-flash-001` | 90.2% | **95.5%** |
| `openai/gpt-4.1-mini`| 93.8% | **94.6%** |
| `google/gemini-2.5-flash` | 90.6% | **95.5%** |

### DIFF
--- a/dspy/adapters/baml_adapter.py
+++ b/dspy/adapters/baml_adapter.py
@@ -202,9 +202,8 @@ class BAMLAdapter(JSONAdapter):
     ```
     """
 
-    def format_field_structure(self, signature: type[Signature]) -> str:
-        """Overrides the base method to generate a simplified schema for Pydantic models."""
-
+    def format_field_description(self, signature: type[Signature]) -> str:
+        """Format the field description for the system message."""
         sections = []
 
         # Add input field descriptions
@@ -222,6 +221,18 @@ class BAMLAdapter(JSONAdapter):
                 type_name = getattr(field.annotation, "__name__", str(field.annotation))
                 description = f": {field.description}" if field.description else ":"
                 sections.append(f"{i}. `{name}` ({type_name}){description}")
+
+        return "\n".join(sections)
+
+    def format_field_structure(self, signature: type[Signature]) -> str:
+        """Overrides the base method to generate a simplified schema for Pydantic models."""
+
+        sections = []
+
+        # Add the main instruction
+        sections.append(
+            "You must produce a single, valid JSON object that strictly adheres to the following schema. Do not output anything else."
+        )
 
         # Add structural explanation
         sections.append(
@@ -264,6 +275,14 @@ class BAMLAdapter(JSONAdapter):
                     sections.append(f"Output field `{name}` should be of type: {type_str}")
 
         return "\n".join(sections)
+
+    def format_task_description(self, signature: type[Signature]) -> str:
+        """Format the task description for the system message."""
+        import textwrap
+
+        instructions = textwrap.dedent(signature.instructions)
+        objective = ("\n" + " " * 8).join([""] + instructions.splitlines())
+        return f"In adhering to this structure, your objective is: {objective}"
 
     def format_user_message_content(
         self,
--- a/dspy/adapters/baml_adapter.py
+++ b/dspy/adapters/baml_adapter.py
@@ -2,14 +2,19 @@
 Custom adapter for improving structured outputs using the information from Pydantic models.
 Based on the format used by BAML: https://github.com/BoundaryML/baml
 """
+
 import inspect
 import types
 from typing import Any, Literal, Union, get_args, get_origin
 
+from pydantic import BaseModel
+
 from dspy.adapters.json_adapter import JSONAdapter
 from dspy.adapters.utils import format_field_value as original_format_field_value
 from dspy.signatures.signature import Signature
-from pydantic import BaseModel
+
+# Changing the comment symbol to Python's # rather than other languages' // seems to help
+COMMENT_SYMBOL = "#"
 
 
 def _render_type_str(annotation: Any, _depth: int = 0, indent: int = 0) -> str:
@@ -34,9 +39,7 @@ def _render_type_str(annotation: Any, _depth: int = 0, indent: int = 0) -> str:
     if origin in (types.UnionType, Union):
         non_none_args = [arg for arg in args if arg is not type(None)]
         # Render the non-None part of the union
-        type_render = " or ".join(
-            [_render_type_str(arg, _depth + 1, indent) for arg in non_none_args]
-        )
+        type_render = " or ".join([_render_type_str(arg, _depth + 1, indent) for arg in non_none_args])
         # Add 'or null' if None was part of the union
         if len(non_none_args) < len(args):
             return f"{type_render} or null"
@@ -82,9 +85,7 @@ def _render_type_str(annotation: Any, _depth: int = 0, indent: int = 0) -> str:
     return str(annotation)
 
 
-def _build_simplified_schema(
-    model: type[BaseModel], indent: int = 0, _seen: set[type] | None = None
-) -> str:
+def _build_simplified_schema(model: type[BaseModel], indent: int = 0, _seen: set[type] | None = None) -> str:
     """Builds a simplified, human-readable schema from a Pydantic model.
 
     Args:
@@ -109,12 +110,13 @@ def _build_simplified_schema(
 
         fields = model.model_fields
         if not fields:
-            lines.append(f"{next_indent}// No fields defined")
+            lines.append(f"{next_indent}{COMMENT_SYMBOL} No fields defined")
         for name, field in fields.items():
-            alias = field.alias or name
-
             if field.description:
-                lines.append(f"{next_indent}// {field.description}")
+                lines.append(f"{next_indent}{COMMENT_SYMBOL} {field.description}")
+            elif field.alias and field.alias != name:
+                # If there's an alias but no description, show the alias as a comment
+                lines.append(f"{next_indent}{COMMENT_SYMBOL} alias: {field.alias}")
 
             # Check for a nested Pydantic model
             field_type_to_render = field.annotation
@@ -122,11 +124,7 @@ def _build_simplified_schema(
             # Unpack Optional[T] to get T
             origin = get_origin(field_type_to_render)
             if origin in (types.UnionType, Union):
-                non_none_args = [
-                    arg
-                    for arg in get_args(field_type_to_render)
-                    if arg is not type(None)
-                ]
+                non_none_args = [arg for arg in get_args(field_type_to_render) if arg is not type(None)]
                 if len(non_none_args) == 1:
                     field_type_to_render = non_none_args[0]
 
@@ -135,20 +133,16 @@ def _build_simplified_schema(
             if origin is list:
                 field_type_to_render = get_args(field_type_to_render)[0]
 
-            if inspect.isclass(field_type_to_render) and issubclass(
-                field_type_to_render, BaseModel
-            ):
+            if inspect.isclass(field_type_to_render) and issubclass(field_type_to_render, BaseModel):
                 # Recursively build schema for nested models with circular reference protection
-                nested_schema = _build_simplified_schema(
-                    field_type_to_render, indent + 1, _seen
+                nested_schema = _build_simplified_schema(field_type_to_render, indent + 1, _seen)
+                rendered_type = _render_type_str(field.annotation, indent=indent + 1).replace(
+                    field_type_to_render.__name__, nested_schema
                 )
-                rendered_type = _render_type_str(
-                    field.annotation, indent=indent + 1
-                ).replace(field_type_to_render.__name__, nested_schema)
             else:
                 rendered_type = _render_type_str(field.annotation, indent=indent + 1)
 
-            line = f"{next_indent}{alias}: {rendered_type},"
+            line = f"{next_indent}{name}: {rendered_type},"
 
             lines.append(line)
 
@@ -213,6 +207,9 @@ class BAMLAdapter(JSONAdapter):
 
         instruction = "You must produce a single, valid JSON object that strictly adheres to the following schema. Do not output anything else."
 
+        # Get the first input field name to use as the schema header
+        input_field_name = next(iter(signature.input_fields.keys())) if signature.input_fields else "schema"
+
         output_schemas = []
         for name, field in signature.output_fields.items():
             field_type = field.annotation
@@ -221,9 +218,7 @@ class BAMLAdapter(JSONAdapter):
             # Find the core type if it's wrapped in Optional or Union
             origin = get_origin(field_type)
             if origin in (types.UnionType, Union):
-                non_none_args = [
-                    arg for arg in get_args(field_type) if arg is not type(None)
-                ]
+                non_none_args = [arg for arg in get_args(field_type) if arg is not type(None)]
                 if len(non_none_args) == 1:
                     main_type = non_none_args[0]
 
@@ -235,12 +230,10 @@ class BAMLAdapter(JSONAdapter):
             else:
                 # Handle non-pydantic or primitive types simply
                 type_str = _render_type_str(field_type, indent=0)
-                output_schemas.append(
-                    f"Output field `{name}` should be of type: {type_str}"
-                )
+                output_schemas.append(f"Output field `{name}` should be of type: {type_str}")
 
-        # Assuming a single Pydantic model output field is the common case
-        return f"{instruction}\n\n[[ ## schema ## ]]:\n" + "\n\n".join(output_schemas)
+        # Use the actual input field name instead of hardcoded "schema"
+        return f"{instruction}\n\n[[ ## {input_field_name} ## ]]:\n" + "\n\n".join(output_schemas)
 
     def format_user_message_content(
         self,
@@ -261,9 +254,7 @@ class BAMLAdapter(JSONAdapter):
                     formatted_value = value.model_dump_json(indent=2, by_alias=True)
                 else:
                     # Fallback to the original dspy formatter for other types
-                    formatted_value = original_format_field_value(
-                        field_info=field_info, value=value
-                    )
+                    formatted_value = original_format_field_value(field_info=field_info, value=value)
 
                 messages.append(f"[[ ## {key} ## ]]\n{formatted_value}")
 
--- a/dspy/adapters/baml_adapter.py
+++ b/dspy/adapters/baml_adapter.py
@@ -205,35 +205,65 @@ class BAMLAdapter(JSONAdapter):
     def format_field_structure(self, signature: type[Signature]) -> str:
         """Overrides the base method to generate a simplified schema for Pydantic models."""
 
-        instruction = "You must produce a single, valid JSON object that strictly adheres to the following schema. Do not output anything else."
+        sections = []
 
-        # Get the first input field name to use as the schema header
-        input_field_name = next(iter(signature.input_fields.keys())) if signature.input_fields else "schema"
+        # Add input field descriptions
+        if signature.input_fields:
+            sections.append("Your input fields are:")
+            for i, (name, field) in enumerate(signature.input_fields.items(), 1):
+                type_name = getattr(field.annotation, "__name__", str(field.annotation))
+                description = f": {field.description}" if field.description else ":"
+                sections.append(f"{i}. `{name}` ({type_name}){description}")
 
-        output_schemas = []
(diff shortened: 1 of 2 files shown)


okhat:
This is amazing, thanks @prrao87 and thanks to the BAML contributors. We should merge this, but should we then also accelerate replacing JSONAdapter with this functionality?

prrao87:
Thank you @okhat! I'd be happy to help the DSPy team do the benchmarking and testing of this so we know this works well as the default option across multiple scenarios. Wasn't sure if it might impact existing users. I'd love to help, so I'll be tracking any issues on this topic closely :).

dspy/adapters/baml_adapter.py
```diff
@@ -0,0 +1,312 @@
...
+                    sections.append(f"Output field `{name}` should be of type: {type_str}")
+
+                sections.append("")  # Empty line after each output
+
+        # Add completed section
+        sections.append("[[ ## completed ## ]]")
+
+        return "\n".join(sections)
+
+    def format_task_description(self, signature: type[Signature]) -> str:
```
TomeHirata, line 275: Did we need to implement this method? This is implemented in `ChatAdapter`

TomeHirata:
Hi, @prrao87. Thank you for the great addition! I'd like to be a bit cautious about the naming of the adapter, especially if we want to promote this. While this adapter formats the pydantic types in an original formatting strategy used in BAML, it does not use BAML language at all (see the BAML language syntax below). I think this is more like an improvement for how JSONAdapter formats Python types into a string, so I wonder if we should add an option to JSONAdapter or use a bit more JSON-ish class name to avoid the confusion. Definitely open to discussion as I'm new to BAML. cc: @okhat @chenmoneygithub 

```
# BAML language

class Resume {
  name string
  title string
}

function ExtractResume(resume: string) -> Resume
```

prrao87:
Hi @TomeHirata, I can see why this could lead to the confusion that we use the BAML language, if it's named this way. From my side, I'm convinced that JSON schema is sub par when it comes to getting the LLM to understand the schema (not to mention the BAML schema format string uses far fewer tokens). I have no problem with DSPy renaming it, while keeping the docstrings in place (so that users know where the format came from).
 
 I'm leaning more towards @okhat's idea of potentially replacing the existing JSON schema itself with this format, but keep the string format a separate component so that it can be swapped out in the future (should an even better format arise). It makes sense to have this discussion before documenting this, so you raise a fair point.

dspy/adapters/baml_adapter.py
prrao87: I wanted to keep this implementation as separate as possible from the others, for this initial PR. But a larger refactor/rename might be needed, per your point below.

chenmoneygithub:
@prrao87  @TomeHirata  I played with BAMLAdapter a bit this week, and there are a few issues:

- `BAMLAdapter.__call__` is a ditto of `JSONAdapter.__call__`, which uses structured output, i.e., `response_format` argument by default. When `response_format` is used, it doesn't really matter if the type is formatted via json schema or this baml-variant. This means we have two adapters that use structured output by default, which could lead to user confusion.
- As @TomeHirata  mentioned above, this is not really "BAML", but a simplified version of JSON schema.

With that, introducing a `mode` arg into JSONAdapter makes more sense to me than having a separate BAMLAdapter. Additionally, it's a bit scary to replace the json schema by this baml-variant completely when structured output is not used before we thoroughly test it in multiple scenarios.

prrao87:
@chenmoneygithub, thanks for the feedback! The name `BAMLAdapter` was chosen because it's basically an adaptation of the JSON adapter, but you're right in that there are already folks on X that are assuming that DSPy now somehow integrates the BAML language. So I am on board with the idea of adding a `mode` arg to the adapter and changing the name. Do you think `mode="json"` (for JSON schema) and `mode="jsonish"` for BAML's format makes sense? This explains that the format is JSON-ish, but not really JSON, and it could be documented that it can be used to improve structured outputs.

That said, I'm more and more inclined to test other schema formatting methods that I think could be even better than this `jsonish` format by BAML. Rather than introducing custom adapters each time we find something new like this, it would be amazing if *only* the schema formatting part of the default prompt could be a swappable component. How feasible do you think that would be? There could be several other formats out there that are just as concise, and perform better across a lot of other datasets. Structured outputs is a very common use case and it's only becoming more widespread as people realize LMs are great at this task and can transform their unstructured text into JSON.

I agree that replacing the default JSON schema with this format may not be the wisest idea (because of unintended consequences in specific use cases, and the fact that there could well be much better formats out there), which is why I wanted to separate code clearly from what people are already using. However, I now see that there seems to be a fair bit of code repetition and it doesn't make sense to document this in the current state of things.

To move forward, could we loop in @isaacbmiller as well? He was mentioning to me that his goals are to improve adapters and to "just have better defaults". To get there with this JSON schema conundrum, I think it's best to run a suite of benchmarks with a few different datasets and tasks, and a few classes of models as follows:
- small open source models (<10B parameters)
- medium open source models (10B-30B parameters)
- One recent model from Google/Gemini
- One recent model from OpenAI
- One recent model from Anthropic

This would convince us (and the community) that JSON schema does indeed reduce the accuracy of structured outputs, and replacing it doesn't impact other tasks. The BoundaryML team (who built BAML) have tested this format in the wild for ~2 years and they have done their own benchmarking, which is why I'm more or less sure that there will be no major issues with performance. We just need to check that there are no impacts for other benchmarks that the DSPy team runs.

If there's anywhere I could look at existing benchmarks, or if not, help create them, I'd be happy to contribute.

chenmoneygithub:
@prrao87 Thanks for the detailed response! 

I have a question on your testing - did you explicitly disable the native structured output in the JSONAdapter? Right now the default behavior of BAMLAdapter is using native structured output, i.e., `response_format` argument, which means the simplified schema is not actually used by the LM.

askmuhsin:
image
Just doing some testing, the desc in Signature is no more part of the prompt when using BAMLAdapter is this the expectation. , thank you. 

`dspy.__version__ - '3.0.0'`

prrao87:
@askmuhsin good point - no, that isn't the intention - created an issue for this: #8657. 

Hi @chenmoneygithub - what do you mean "explicitly disable native structured output"? I don't understand how the simplified history above is working so well if `dspy.inspect_history()` clearly shows the simplified schema (as per my example above) and the results are drastically improved.

BTW, I ran another test on a smaller model (`google/gemma-3-4b-it`) and with `BAMLAdapter`, I get a result out of DSPy, but with the regular `JSONAdapter` (that uses JSON schema) my program crashes because the result couldn't be parsed. So this indicates that JSON schema is indeed harder to follow for the smaller models.

I'd like to come to a consensus on how we can move towards making this the default for DSPy, with all the best practices and functionality of before, but just replacing the JSON schema format. Thank you!

dspy/adapters/baml_adapter.py
prrao87: This is addressed in #8663. The redundant method is now removed!

Issue #8616, opened 2025-08-08
[Feature] Datasets dependency is very large and not needed for inference


heyalexchoi:
### What feature would you like to see?

Hi, I've added dspy to a client project and ran into issues deploying it on an aws lambda since the dependencies were so large. `Datasets` dependency in particular is quite large, bringing in pyarrow, and doesn't seem to be used during inference deployment. From the release history I can see `datasets` has been removed already for this exact reason (https://github.com/stanfordnlp/dspy/pull/8274), but then added back in later(https://github.com/stanfordnlp/dspy/commit/8aa0659451a1f44c60b213b7321563a1f677db1e).

IMO it should be left out, and then can be installed separately for optimization as needed.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

heyalexchoi:
I made a PR for this https://github.com/stanfordnlp/dspy/pull/8617

bgourlie:
I just ran into this issue - datasets has been re-added transitively via GEPA: https://github.com/stanfordnlp/dspy/commit/26248897242cfad6efb7119946613c71dc921240#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

PR #8617, opened 2025-08-08
move datasets into offline optional dependency group

heyalexchoi:
More rationale for this change in the Issue: https://github.com/stanfordnlp/dspy/issues/8616

I updated getting started instructions in docs which already seemed to assume that datasets was *not* in the core dependencies (it instructed user to install `datasets` before optimizing). 

I think the optional dependency group will help clarify how datasets is used, and provide space for easier addition of "offline" dependencies (eg: pytorch, diffusers, etc?)


heyalexchoi:
Edit: resolved in https://github.com/stanfordnlp/dspy/pull/8621

PR #8620, opened 2025-08-09
Add gpt-5 as reasoning model

olesyash:
Add openai/gpt-5 as reasoning model that requires max_tokens >=20,000 and temperature 1.0

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -79,7 +79,7 @@ class LM(BaseLM):
         model_family = model.split("/")[-1].lower() if "/" in model else model.lower()
 
         # Match pattern: o[1,3,4] at the start, optionally followed by -mini and anything else
-        model_pattern = re.match(r"^o([134])(?:-mini)?", model_family)
+        model_pattern = re.match(r"^(?:o([1345])|gpt-(5))(?:-mini)?", model_family)
 
         if model_pattern:
             # Handle OpenAI reasoning models (o1, o3)


okhat:
Thanks! Will need gpt5-nano too but we can merge this right away for now

yashgoy41:
@okhat @olesyash Thanks for handling this issue! One follow-up from using DSPy 3.0.1 in an enterprise Databricks setup.

**Issue:** reasoning-model detection in `dspy/clients/lm.py` doesn’t catch vendor-prefixed endpoint names, so DSPy still sends `max_tokens` instead of `max_completion_tokens`. Additionally the temperature is also not required so I run into that problem too.

**Repro**

```py
# Databricks AI Gateway endpoint name
lm = dspy.LM("databricks/se-gpt-5-mini", max_completion_tokens=20_000)
dspy.configure(lm=lm)
lm("hi")
# -> BAD_REQUEST: unsupported parameter 'max_tokens'; use 'max_completion_tokens'
```

**Root cause**

```py
model_pattern = re.match(r"^(?:o([1345])|gpt-(5))(?:-mini)?", model_family)
# ^ anchored at start → doesn’t match "se-gpt-5-..."
```

We (and many orgs) prefix serving endpoints (e.g., `se-gpt-5-mini-*`, `xxx-o4-mini-*`), so the regex misses them.

**Suggested fix**
Allow prefixes and common separators and use `search`:

```py
model_pattern = re.search(r'(?:^|[-_/])(?:o[1345]|gpt-5)(?:-(?:mini|nano))?', model_family)
```

This covers `openai/gpt-5*`, `openai/o[1345]*`, and wrapped names like `se-gpt-5-mini-*` or `db-o4-mini-*`.

Happy to open a PR if that helps!

yashgoy41:
@okhat After digging further, I found there’s a second issue downstream in **LiteLLM’s Databricks adapter** that compounds this.

Even when DSPy passes `max_completion_tokens`, the Databricks transformer (`litellm/llms/databricks/chat/transformation.py`) rewrites it into `max_tokens` by default:

```python
if "max_completion_tokens" in non_default_params and replace_max_completion_tokens_with_max_tokens:
    mapped_params["max_tokens"] = non_default_params["max_completion_tokens"]
    mapped_params.pop("max_completion_tokens", None)
```

Since Databricks’ `/chat/completions` endpoint *only* supports `max_completion_tokens`, the request always fails with:

```
BAD_REQUEST: Unsupported parameter: 'max_tokens'
```

**Repro (with DSPy 3.0.1 + LiteLLM):**

```python
lm = dspy.LM("databricks/se-gpt-5-mini", max_completion_tokens=20_000)
dspy.configure(lm=lm)
lm("hi")
# → error: 'max_tokens' not supported
```

**Temporary fix:**
Remove the `max_tokens` key entirely and do *not* specify `max_completion_tokens` either — this lets Databricks fall back to its default behavior (no explicit token limit). Example:

```python
lm = dspy.LM("databricks/se-gpt-5-mini", temperature=1)
lm.kwargs.pop("max_tokens", None)  # critical line
```

**Proper fixes needed:**

1. In DSPy: broaden regex to catch vendor-prefixed endpoint names (so `max_completion_tokens` is selected when appropriate).
2. In LiteLLM DatabricksConfig: stop rewriting `max_completion_tokens → max_tokens`; instead, promote `max_tokens → max_completion_tokens` for compatibility.

With both in place, Databricks endpoints work fine.

Happy to open a PR to patch the regex in DSPy and/or raise an upstream issue in LiteLLM if that helps.

Gby56:
Has anyone managed to use Azure OpenAI with DSPy + Litellm ?

PR #8626, opened 2025-08-12
Allow BAMLAdapter to be accessed via dspy.BAMLAdapter

BenMcH:
Currently, the BAMLAdapter needs to be imported like `from dspy.adapters.baml_adapter import BAMLAdapter`.

This PR allows a more standard `dspy.BAMLAdapter` to be used :smile:

### DIFF
--- a/dspy/adapters/__init__.py
+++ b/dspy/adapters/__init__.py
@@ -1,5 +1,5 @@
-from dspy.adapters.base import Adapter
 from dspy.adapters.baml_adapter import BAMLAdapter
+from dspy.adapters.base import Adapter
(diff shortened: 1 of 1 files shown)


TomeHirata:
Thanks for the PR! I'd like to keep it on hold until we conclude the naming/direction of BAMLAdapter ([ref](https://github.com/stanfordnlp/dspy/pull/8614#issuecomment-3177551705))

prrao87:
I think we can close this for now? Whether or not BAMLAdapter's functionality should be merged with JSONAdapter's is a larger discussion to be had by the maintainers and this PR is anyway out of date.

PR #8635, opened 2025-08-13
Warn on deprecated requires_permission_to_run

TomeHirata:
## Summary
- warn via logger when deprecated `requires_permission_to_run` is supplied to `MIPROv2.compile`

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -109,13 +109,16 @@ class MIPROv2(Teleprompter):
         view_data_batch_size: int = 10,
         tip_aware_proposer: bool = True,
         fewshot_aware_proposer: bool = True,
-        requires_permission_to_run: bool = False, # deprecated
+        requires_permission_to_run: bool | None = None, # deprecated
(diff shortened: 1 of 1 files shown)


dspy/teleprompt/mipro_optimizer_v2.py
```diff
@@ -112,6 +112,10 @@ def compile(
         requires_permission_to_run: bool = True, # deprecated
         provide_traceback: bool | None = None,
     ) -> Any:
+        if requires_permission_to_run:
```
chenmoneygithub: this will result in warning in the default path, shall we flip the default value to False?

TomeHirata: Good call, let's change the default too.

okhat:
Awesome, but should this actually crash instead? If someone says "Require permission: True", it should just crash I think. (Please feel free to merge directly when ready.)

TomeHirata:
Thanks, actually I think the following behavior is correct as the previous default is True
- User selects requires_permission_to_run=False -> show warning
- Users explicitly selects requires_permission_to_run=True -> raise exception

PR #8636, opened 2025-08-13
Add type hints for UsageTracker

TomeHirata:
Add type hints to UsageTracker for internal consistency

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -2,7 +2,7 @@
 
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import Any
+from typing import Any, Generator
 
 from dspy.dsp.utils.settings import settings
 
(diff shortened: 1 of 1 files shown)


dspy/utils/usage_tracker.py
```diff
@@ -20,16 +20,16 @@ def __init__(self):
         # }
         self.usage_data = defaultdict(list)
 
-    def _flatten_usage_entry(self, usage_entry) -> dict[str, dict[str, Any]]:
+    def _flatten_usage_entry(self, usage_entry: dict[str, Any]) -> dict[str, Any]:
```
chenmoneygithub, line 23: `usage_entry` is not necessarily a dict as i remember, it could be the litellm usage object so I skipped type hints for this method (and the others)

okhat:
Thanks so much @TomeHirata ! Feel free to merge whenever concerns are discussed/resolved!

dspy/utils/usage_tracker.py
TomeHirata: I think it is a dict when `add_usage` is called. We convert the litellm usage object to a dict before calling it. That's probably why the type signature of `add_usage` has been `dict`
https://github.com/stanfordnlp/dspy/blob/34090a255de362647ab0c8eb29d9acdf8307b154/dspy/clients/lm.py#L173

chenmoneygithub: Sorry for causing confusion, here is the more detailed context: internally we have control over what we pass to UsageTracker, but this `UsageTracker` can also be used in custom usage tracking, and the input can be any dict-like types, e.g., litellm usage or openai usage instance. 

Given a second thought, it's not bad to force users to call `dict()` over their dict-like type though, and keeping consistency of using /not using type hints inside the same API is important. I am going ahead approving this PR.

chenmoneygithub (approved):

PR #8650, opened 2025-08-14
Raise explicit error on not meeting reasoning model's requirement 

chenmoneygithub:
Make the error better:

- instead of `assert`, use explicit error raising: git push --set-upstream origin client-style-fix
- include an example to explain what the error message means

And very personally, the old code gets automatically formatted by vscode ruff every time I save the file, which is quite annoying.

### DIFF
--- a/tests/clients/test_lm.py
+++ b/tests/clients/test_lm.py
@@ -219,13 +219,12 @@ def test_reasoning_model_token_parameter():
 
 def test_reasoning_model_requirements():
     # Should raise assertion error if temperature or max_tokens requirements not met
-    with pytest.raises(ValueError) as error:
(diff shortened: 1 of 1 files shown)


tests/clients/test_lm.py
```diff
@@ -219,13 +219,13 @@ def test_reasoning_model_token_parameter():
 
 def test_reasoning_model_requirements():
     # Should raise assertion error if temperature or max_tokens requirements not met
-    with pytest.raises(AssertionError) as exc_info:
+    with pytest.raises(ValueError) as error:
```
TomeHirata: Let's use pytest matcher rather than manually using `assert`
https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

chenmoneygithub: done!

PR #8658, opened 2025-08-14
Add pyupgrade as a pre-commit hook

kurtmckee:
pyupgrade is a useful tool for upgrading old code styles to newer Python idioms.

This PR introduces pyupgrade as a pre-commit hook, runs `pre-commit run -a` twice, and commits the changes.

pyupgrade found the following classes of improvements:

* `typing.Callable`, and several async typing types, are deprecated aliases for `collections.abc.Callable` et al.
* `open(file, "r")` was replaced with `open(file)`, which is equivalent.
* `%`-style string formatting was replaced with f-strings.
* `Union[x, y]` and `Optional[x]` can now be expressed as `x | y` and `x | None`.

The test suite passed locally after these changes, but I'll watch CI as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,3 +26,10 @@ repos:
         args: ["--maxkb=1024"]
       - id: check-merge-conflict
       - id: debug-statements
+
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.21.2
+    hooks:
+      - id: pyupgrade
+        name: Enforce Python 3.10+ idioms
+        args: ["--py310-plus"]
--- a/docs/scripts/generate_api_summary.py
+++ b/docs/scripts/generate_api_summary.py
@@ -60,7 +60,7 @@ def format_nav_section(nav_dict, indent_level=2):
 
 def read_mkdocs_sections(filename: str = "mkdocs.yml"):
     """Read and parse the mkdocs.yml file into sections."""
-    with open(filename, "r") as f:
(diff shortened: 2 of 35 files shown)


okhat:
Wow, `typing.Callable` is deprecated now..?

kurtmckee:
> `typing.Callable` is deprecated now..?

[Yep](https://docs.python.org/3/library/typing.html#typing.Callable), since Python 3.9. I think the difference was that `collections.abc.Callable` didn't support type annotation subscripts initially, so it wasn't possible to annotate the callable's parameters and return value:

```python
def f(function: collections.abc.Callable[[int], bool]) -> bool:
    return function(123)
```

`typing.Callable` was added to support this and backported via `typing_extensions` IIRC, but now `collections.abc.Callable` supports type annotation subscripts.

kurtmckee:
_Rebased on `main`, resolved merge conflicts, and force-pushed._

kurtmckee:
_Rebased on `main`, resolved merge conflicts again, and force-pushed._

kurtmckee:
@okhat would you be willing to take another look at this PR when you have an opportunity? It's okay to reject it, I'm just hoping to avoid the need to resolve merge conflicts with `main`.

kurtmckee:
_Rebased on `main` to resolve merge conflicts._

@okhat @chenmoneygithub @TomeHirata Friendly ping to review this PR! It's okay to reject it, I'm just wanting it to be reviewed and merged, or closed. Thanks!

PR #8662, opened 2025-08-15
feat(optimizers): let users set the seed in LabeledFewShot

jmhb0:
Setting the seed in LabeledFewShot means the caller can try this optimizer for a few seeds and pick the best one. This is useful for systems where LabeledFewShot is a pretty strong baseline (e.g. single Predict module with a small number of classes, where it does not make sense to do bootstrapping.

(A workaround to this is to just shuffle the trainset separately)


okhat:
Yeah just shuffle on the user side...

PR #8663, opened 2025-08-15
Fix minor issues with BAMLAdapter

### LINKED ISSUE #8657: [Bug] BAMLAdapter system message missing description field from signature
### What happened?

Mentioned in #8614 by a user, should be a simple fix.

> It seems that in `BAMLAdapter`, the `desc` in the Signature is no more part of the prompt - this would need to be addressed

### Steps to reproduce

Image

### DSPy version

3.0.0

prrao87:
This PR addresses the following:

- Closes #8657: Signature descriptions are now correctly handled and included in the display for InputField/OutputField (exact same behaviour as JSONAdapter)
  - New test added to check for this
- Addresses @TomeHirata's [comment](https://github.com/stanfordnlp/dspy/pull/8614#discussion_r2268443119) in #8614 about the  redundant `format_task_description` method - this is now reused from JSONAdapter
- Refactors to pretty print optional types to make it simpler for the LM (see an example for the `condition` input field below in the formatted output)
```
Your input fields are:
1. `note` (str):
2. `condition` (Condition or null):
```
I evaluated this on my [nested model benchmark](https://github.com/prrao87/structured-outputs) and it slightly improves performance from the previous version of `BAMLAdapter`.

Although we may end up refactoring `BAMLAdapter` to be a part of `JSONAdapter` going forward, this will at least address the ongoing issue for users like #8657.

### DIFF
--- a/dspy/adapters/baml_adapter.py
+++ b/dspy/adapters/baml_adapter.py
@@ -7,10 +7,11 @@ import inspect
 import types
 from typing import Any, Literal, Union, get_args, get_origin
 
+from pydantic import BaseModel
+
 from dspy.adapters.json_adapter import JSONAdapter
 from dspy.adapters.utils import format_field_value as original_format_field_value
 from dspy.signatures.signature import Signature
-from pydantic import BaseModel
 
 # Changing the comment symbol to Python's # rather than other languages' // seems to help
 COMMENT_SYMBOL = "#"


prrao87:
Closing this as I'd completely missed that there was already a lot of updates in #8654. This PR fixes the user issue #8657, but I think that's very minor and it can be added in once #8654 is merged.

PR #8671, opened 2025-08-16
Add tutorial on adapters

chenmoneygithub:
We have seen confusion from users on what is adapter, and how to pick the right adapter. i am writing this tutorial to fill in the gap.

I am not including XMLAdapter and BAMLAdapter in the tutorial because very few people are using it, and we don't have clear information on where they best fit in.

### DIFF
--- a/docs/docs/learn/programming/adapters.md
+++ b/docs/docs/learn/programming/adapters.md
@@ -1,0 +1,307 @@
+# Understanding DSPy Adapters
+
+## What are Adapters?
+
+Adapters are the bridge between `dspy.Predict` and the actual Language Model (LM). When you call a DSPy module, the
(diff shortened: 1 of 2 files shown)


torchss:
Chen,

unsure if this is the right place to ask this question, or I should make a new issue or join the discord, but is an adapter what I would need to build to add Outlines (https://github.com/dottxt-ai/outlines) support to dspy, if I wanted to control the logits instead of hoping the LLM was "good enough" to output high quality JSON or similar?

Or should I be subclassing `dspy.LM` instead? (in this case are there some examples I could look at?)

Outlines isn't backend agnostic (due to the need to control the logits) and my first implementation would be using vLLM.

PR #8665, opened 2025-08-15
Fix MLFlow GEPA Integration

LakshyAAAgrawal:
Fix #8673

PR #8691, opened 2025-08-21
[Fork of #8665] Fix mlflow gepa

chenmoneygithub:
Use patch solution instead of creating a module wrapper.


okhat:
Patching is scary, but this is probably less scary than the other PR.

PR #8692, opened 2025-08-21
Support OpenAI response api in LM

TomeHirata:
Support https://platform.openai.com/docs/api-reference/responses/create

```python
import dspy
dspy.settings.configure(lm=dspy.LM(
    "openai/gpt-4.1-mini",
    temperature=1.0,
    max_tokens=20_000,
    model_type="responses",
),)
dspy.Predict("question -> answer")(question="1+1")
-> Prediction(
    answer='2'
)
```

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -48,57 +48,6 @@ class BaseLM:
         self.kwargs = dict(temperature=temperature, max_tokens=max_tokens, **kwargs)
         self.history = []
 
-    def _process_completion(self, response, merged_kwargs):
-        """Process the response of OpenAI chat completion API and extract outputs.
-        
-        Args:
-            response: The OpenAI chat completion response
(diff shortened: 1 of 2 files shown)


bradyneal:
@TomeHirata Awesome! Any chance this also supports OpenAI's web_search tool?

TomeHirata:
Hi, @bradyneal. We could support other type of tools but need to consider how to connect it with the DSPy's tool structure. Let me think about it, stay tuned!

bradyneal:
@TomeHirata To be sure, that means DSPy's OpenAI responses API integration doesn't support the OpenAI web_search tool, yet, right?

If so, any update on what I might have to look forward to or what I can use now?

bradyneal:
@TomeHirata It seems like this already works with OpenAI's web_search (LiteLLM probably handles it all)! But it seems like it has a bug when trying to use together with JSONAdapter: https://github.com/stanfordnlp/dspy/issues/8872

PR #8709, opened 2025-08-26
Add doc page to learn tool usage in DSPy

TomeHirata:
Tool call agent is a very common and popular way to implement AI agents. While DSPy supports tool calling in multiple ways, many users don't realize its support due to the lack of documentation. This PR introduces a page for Tools in DSPy to increase the visibility.


prrao87:
THANK YOU! These docs were just what I needed (was very confused by `dspy.Tool` until I read this page thoroughly), and it really helped me understand the differences between manual and ReAct tool calling better. Many thanks @TomeHirata!

PR #8712, opened 2025-08-26
Add experimental annotation

TomeHirata:
Add experimental annotation so that users can know in which APIs they should expect breaking changes

### DIFF
--- a/dspy/utils/annotation.py
+++ b/dspy/utils/annotation.py
@@ -1,0 +1,79 @@
+import inspect
+import re
+import types
+from typing import Callable, ParamSpec, TypeVar
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def experimental(
+    f: Callable[P, R] | None = None,
+    version: str | None = None,
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    """Decorator / decorator creator for marking APIs experimental in the docstring.
+
+    Args:
+        f: The function to be decorated.
+        version: The version in which the API was introduced as experimental.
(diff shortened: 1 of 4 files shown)


dspy/utils/annotation.py
```diff
@@ -0,0 +1,88 @@
+import inspect
+import re
+import types
+from typing import Callable, ParamSpec, TypeVar, overload
+
+P = ParamSpec("P")
```
chenmoneygithub, line 6: qq - is this P, R useful? Personally I find it like noise rather than useful, but maybe that's useful for autocomplete?

dspy/utils/annotation.py
```diff
@@ -0,0 +1,88 @@
...
+    elif isinstance(api, types.MethodType):
+        api_type = "method"
+    else:
+        api_type = str(type(api))
+
+    indent = _get_min_indent_of_docstring(api.__doc__) if api.__doc__ else ""
+
+    version_text = f" (introduced in v{version})" if version else ""
+    notice = (
+        indent + f".. Note:: Experimental: This {api_type} may change or "
```
chenmoneygithub: Maybe remove `..`? Is that some sphinx syntax used in mlflow?

TomeHirata: Oh, yes let's remove it. It's sphinx syntax.

dspy/utils/annotation.py
TomeHirata: This is for keeping the type hint from the original function. I think it's better to keep this annotation in order not to lose type hints.

PR #8725, opened 2025-08-28
8688 csv output

crawftv:
This addresses issue 8688. 
With these changes we add support for saving files locally to csv or json. 
Some code to merge predictions and results is extracted to be reused.


crawftv (commented):
Removed reformatting

Issue #8690, opened 2025-08-20
[Feature] Allow users to provide instruction proposal scheme in GEPA


LakshyAAAgrawal:
### What feature would you like to see?

GEPA allows for different instruction proposal mechanisms, through modifying the propose_new_texts method of GEPAAdapter. Different usecases may require instruction proposal to happen a certain way, for example, for image usecases, GEPA's default instruction proposal prompt generates very long prompts. It should come with several options for users to choose from, or even provide their own implementation.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Additional Context

Reference: https://discord.com/channels/1161519468141355160/1406346534857085009/1406346534857085009

LakshyAAAgrawal:
Proposed signature by @arnavsinghvi11 
```
class ProposeNewInstructionSignature(dspy.Signature):
    """Your task is to **create a comprehensive, clear, and fully self-contained instruction document** that enables people with no prior experience to successfully perform a complex task. The instructions must ensure that users can complete the task accurately using only the document you produce without requiring any external sources.
---
## What You Will Receive
You will be provided with:
- **Original Instructions:**
The previous version of the instruction document.
- **User Examples and Feedback:**
Outputs from users of the previous instruction document, with notes highlighting common mistakes and misunderstandings stemming from incompleteness of the previous guidance document as feedback.
---
## What You Must Produce
Develop a **detailed instruction document** that:
- **Is self-contained.**
All needed definitions, rules, and context are included. Do not refer to any external documents or sources.
- **Addresses all prior confusion.**
Integrate explicit corrections based on user feedback and examples of past errors.
- **Uses clear, precise language.**
- **Includes practical examples.**
Show both correct and incorrect task execution, with explanations.
- **Explains all key terms and concepts.**
- **States all requirements.**
Clearly specify any output formatting requirements, acceptable outputs, and evaluation criteria.
- **Covers edge cases and exceptions.**
- **Includes all required domain knowledge from feedbacks and reference material**
- **Includes repeatable task patterns in how previous users attempted the task correctly**
- **Do not waste space in formatting and headers of the document. Be direct.**
Follow these directions to produce instructions that enable any user—without any prior knowledge of task domain—to complete the task successfully and consistently."""

    
    current_instructions: str = dspy.InputField(
        desc="The current (previous) version of the instruction document that requires improvement."
    )
    
    examples_summary: str = dspy.InputField(
[...]

PR #8737, opened 2025-08-30
Add custom instruction_proposer support to GEPA with multimodal (dspy.Image) handling

andressrg:
Closes https://github.com/stanfordnlp/dspy/issues/8690

# Add support for custom instruction proposers in GEPA with multimodal handling.

As proposed by @LakshyAAAgrawal, this PR makes GEPA’s instruction-proposal step pluggable. Users can supply a custom instruction proposer, enabling GEPA to handle dspy.Image and other custom types.

## Key Features

- **Custom Instruction Proposer Protocol**: New `InstructionProposerProtocol` for implementing custom instruction generation
- **MultiModalInstructionProposer**: Built-in proposer that handles multimodal inputs (text + images) during optimization
- **Structured Image Handling**: Preserves `dspy.Image` objects without serialization in reflection LM calls
- **Backward Compatibility**: Existing GEPA usage continues to work unchanged

### For multi-modal use cases

```python
import dspy
from dspy.teleprompt.gepa.instruction_proposal import MultiModalInstructionProposer

# Configure your language models
lm = dspy.LM(model='gpt-4o-mini')
reflection_lm = dspy.LM(model='gpt-4o', temperature=1.0, max_tokens=32000)

# Create a program that uses images
program = dspy.Predict("image: dspy.Image -> label: str")

# Set up GEPA with custom instruction proposer
gepa = dspy.GEPA(
    metric=your_metric_function,
    max_metric_calls=50,
    reflection_lm=reflection_lm,
    instruction_proposer=MultiModalInstructionProposer()  # Enable multimodal support
)

# Compile with image data
compiled_program = gepa.compile(program, trainset=image_trainset, valset=image_valset)
```

### Advanced: Custom Instruction Proposer

You can implement your own instruction proposer by following the protocol:

```python
from dspy.teleprompt.gepa.instruction_proposal import InstructionProposerProtocol

class MyCustomProposer:
    def __call__(self, current_instruction: str, reflective_dataset: list[dict]) -> str:
        # Your custom logic here
        # reflective_dataset contains examples with inputs, outputs, and feedback
        # May include dspy.Image objects in inputs
        
        # Analyze the feedback and generate improved instruction
        improved_instruction = analyze_and_improve(current_instruction, reflective_dataset)
        return improved_instruction

# Use with GEPA
gepa = dspy.GEPA(
    metric=your_metric,
    max_metric_calls=50,
    reflection_lm=reflection_lm,
    instruction_proposer=MyCustomProposer()
)
```

## Backward Compatibility

This change is fully backward compatible. Existing GEPA code will continue to work without modification:

```python
# This still works exactly as before
gepa = dspy.GEPA(metric=my_metric, max_metric_calls=50, reflection_lm=reflection_lm)
compiled = gepa.compile(program, trainset, valset)
```

## Files Changed

- `dspy/teleprompt/gepa/gepa.py`: Add `instruction_proposer` parameter
- `dspy/teleprompt/gepa/gepa_utils.py`: Integrate custom proposer logic and preserve image objects
- `dspy/teleprompt/gepa/instruction_proposal.py`: New multimodal instruction proposer implementation
- `tests/teleprompt/test_gepa_instruction_proposer.py`

This enables GEPA optimization to work with visual content while maintaining proper image handling throughout the reflection and instruction proposal process.

### DIFF
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -12,6 +12,7 @@ if TYPE_CHECKING:
     from gepa import GEPAResult
 
     from dspy.teleprompt.gepa.gepa_utils import DspyAdapter, DSPyTrace, PredictorFeedbackFn, ScoreWithFeedback
+    from dspy.teleprompt.gepa.instruction_proposal import InstructionProposerProtocol
 
 logger = logging.getLogger(__name__)
 
@@ -282,6 +283,8 @@ class GEPA(Teleprompter):
         track_best_outputs: bool = False,
         # Reproducibility
         seed: int | None = 0,
+        # Custom instruction proposer
+        instruction_proposer: "InstructionProposerProtocol | None" = None,
     ):
         try:
             inspect.signature(metric).bind(None, None, None, None, None)
@@ -312,8 +315,8 @@ class GEPA(Teleprompter):
         # Reflection based configuration
         self.reflection_minibatch_size = reflection_minibatch_size
         self.candidate_selection_strategy = candidate_selection_strategy
-        # self.reflection_lm = reflection_lm
         assert reflection_lm is not None, "GEPA requires a reflection language model to be provided. Typically, you can use `dspy.LM(model='gpt-5', temperature=1.0, max_tokens=32000)` to get a good reflection model. Reflection LM is used by GEPA to reflect on the behavior of the program and propose new instructions, and will benefit from a strong model."
+        self._original_reflection_lm = reflection_lm  # Store original for dspy adapter use
         self.reflection_lm = lambda x: reflection_lm(x)[0]
         self.skip_perfect_score = skip_perfect_score
         self.add_format_failure_as_feedback = add_format_failure_as_feedback
@@ -340,6 +343,8 @@ class GEPA(Teleprompter):
 
         # Reproducibility
         self.seed = seed
+
+        self.custom_instruction_proposer = instruction_proposer
 
     def auto_budget(self, num_preds, num_candidates, valset_size: int, minibatch_size: int = 35, full_eval_steps: int = 5) -> int:
         import numpy as np
@@ -452,6 +457,8 @@ class GEPA(Teleprompter):
             num_threads=self.num_threads,
             add_format_failure_as_feedback=self.add_format_failure_as_feedback,
             rng=rng,
+            reflection_lm=self._original_reflection_lm,
+            custom_instruction_proposer=self.custom_instruction_proposer,
         )
 
         reflection_lm = self.reflection_lm
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -217,6 +217,10 @@ class GEPA(Teleprompter):
             `dspy.teleprompt.gepa.instruction_proposal` for handling visual content. If None (default), 
             GEPA uses its built-in text-optimized proposer (see `gepa.strategies.instruction_proposal.InstructionProposalSignature` 
             for reference implementation).
+            
+            Note: When both instruction_proposer and reflection_lm are set, the instruction_proposer is called 
+            in the reflection_lm context. However, reflection_lm is optional when using a custom instruction_proposer. 
+            Custom instruction proposers can invoke their own LLMs if needed.
         add_format_failure_as_feedback: Whether to add format failures as feedback. Default is False.
         use_merge: Whether to use merge-based optimization. Default is True.
         max_merge_invocations: The maximum number of merge invocations to perform. Default is 5.
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -215,7 +215,7 @@ class GEPA(Teleprompter):
             If provided, GEPA will use this custom proposer instead of its default instruction proposal 
             mechanism to generate improved instructions based on feedback from failed examples. This is 
             particularly useful when you need specialized instruction generation for multimodal inputs 
-            (like dspy.Image) or custom types. Use create_multimodal_proposer() from 
+            (like dspy.Image) or custom types. Use MultiModalInstructionProposer() from 
             dspy.teleprompt.gepa.instruction_proposal for handling visual content. If None (default), 
             GEPA uses its built-in text-optimized proposer.
         add_format_failure_as_feedback: Whether to add format failures as feedback. Default is False.
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -2,17 +2,15 @@ import inspect
 import logging
 import random
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Literal, Optional, Protocol, Union
+from typing import Any, Literal, Optional, Protocol, Union
+
+from gepa import GEPAResult
+from gepa.core.adapter import ProposalFn
 
 from dspy.clients.lm import LM
 from dspy.primitives import Example, Module, Prediction
+from dspy.teleprompt.gepa.gepa_utils import DspyAdapter, DSPyTrace, PredictorFeedbackFn, ScoreWithFeedback
 from dspy.teleprompt.teleprompt import Teleprompter
-
-if TYPE_CHECKING:
-    from gepa import GEPAResult
-    from gepa.core.adapter import ProposalFn
-
-    from dspy.teleprompt.gepa.gepa_utils import DspyAdapter, DSPyTrace, PredictorFeedbackFn, ScoreWithFeedback
 
 logger = logging.getLogger(__name__)
 
@@ -215,9 +213,10 @@ class GEPA(Teleprompter):
             If provided, GEPA will use this custom proposer instead of its default instruction proposal 
             mechanism to generate improved instructions based on feedback from failed examples. This is 
             particularly useful when you need specialized instruction generation for multimodal inputs 
-            (like dspy.Image) or custom types. Use MultiModalInstructionProposer() from 
-            dspy.teleprompt.gepa.instruction_proposal for handling visual content. If None (default), 
-            GEPA uses its built-in text-optimized proposer.
+            (like dspy.Image) or custom types. Use `MultiModalInstructionProposer()` from 
+            `dspy.teleprompt.gepa.instruction_proposal` for handling visual content. If None (default), 
+            GEPA uses its built-in text-optimized proposer (see `gepa.strategies.instruction_proposal.InstructionProposalSignature` 
+            for reference implementation).
         add_format_failure_as_feedback: Whether to add format failures as feedback. Default is False.
         use_merge: Whether to use merge-based optimization. Default is True.
         max_merge_invocations: The maximum number of merge invocations to perform. Default is 5.
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -211,6 +211,13 @@ class GEPA(Teleprompter):
             a strong reflection model. Consider using `dspy.LM(model='gpt-5', temperature=1.0, max_tokens=32000)` 
             for optimal performance.
         skip_perfect_score: Whether to skip examples with perfect scores during reflection. Default is True.
+        instruction_proposer: Optional custom instruction proposer implementing InstructionProposerProtocol. 
+            If provided, GEPA will use this custom proposer instead of its default instruction proposal 
+            mechanism to generate improved instructions based on feedback from failed examples. This is 
+            particularly useful when you need specialized instruction generation for multimodal inputs 
+            (like dspy.Image) or custom types. Use MultiModalProposer(), located at 
+            dspy.teleprompt.gepa.instruction_proposal, for handling visual content. If None (default), 
+            GEPA uses its built-in text-optimized proposer.
         add_format_failure_as_feedback: Whether to add format failures as feedback. Default is False.
         use_merge: Whether to use merge-based optimization. Default is True.
         max_merge_invocations: The maximum number of merge invocations to perform. Default is 5.
@@ -267,6 +274,7 @@ class GEPA(Teleprompter):
         reflection_lm: LM | None = None,
         skip_perfect_score: bool = True,
         add_format_failure_as_feedback: bool = False,
+        instruction_proposer: "InstructionProposerProtocol | None" = None,
         # Merge-based configuration
         use_merge: bool = True,
         max_merge_invocations: int | None = 5,
@@ -283,8 +291,6 @@ class GEPA(Teleprompter):
         track_best_outputs: bool = False,
         # Reproducibility
         seed: int | None = 0,
-        # Custom instruction proposer
-        instruction_proposer: "InstructionProposerProtocol | None" = None,
     ):
         try:
             inspect.signature(metric).bind(None, None, None, None, None)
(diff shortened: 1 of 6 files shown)


dspy/teleprompt/gepa/gepa.py
```diff
@@ -282,6 +283,8 @@ def __init__(
         track_best_outputs: bool = False,
         # Reproducibility
         seed: int | None = 0,
+        # Custom instruction proposer
```
LakshyAAAgrawal: Could you move this up to # Reflection based configuration section (after line 269)

LakshyAAAgrawal: And make a corresponding edit to the docstring.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -1,13 +1,18 @@
...
 from gepa import EvaluationBatch, GEPAAdapter
 
+import dspy
 from dspy.adapters.chat_adapter import ChatAdapter
 from dspy.adapters.types import History
+from dspy.adapters.types.base_type import Type
 from dspy.evaluate import Evaluate
 from dspy.primitives import Example, Prediction
+
+if TYPE_CHECKING:
```
LakshyAAAgrawal: We perhaps don't need the `if` condition here?

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -66,6 +73,35 @@ def __init__(
         self.num_threads = num_threads
         self.add_format_failure_as_feedback = add_format_failure_as_feedback
         self.rng = rng or random.Random(0)
+        self.reflection_lm = reflection_lm
+        self.custom_instruction_proposer = custom_instruction_proposer
+
+        if self.custom_instruction_proposer is not None:
```
LakshyAAAgrawal, line 77: There should be an assert here, that ensures whenever `self.custom_instruction_proposer is not None`, then `assert self.reflection_lm is not None` as well.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -227,29 +268,3 @@ def make_reflective_dataset(self, candidate, eval_batch, components_to_update):
             raise Exception("No valid predictions found for any module.")
 
         return ret_d
-
-    # TODO: The current DSPyAdapter implementation uses the GEPA default propose_new_texts.
```
LakshyAAAgrawal: I recommend retaining this part, as it acts as documentation for future readers.

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -0,0 +1,280 @@
+from typing import Any, Protocol
+
+import dspy
+from dspy.adapters.types.base_type import Type
+
+
+class InstructionProposerProtocol(Protocol):
```
LakshyAAAgrawal: GEPA already has a protocol very similar to this protocol: https://github.com/gepa-ai/gepa/blob/main/src/gepa/core/adapter.py#L29. I am wondering if having a separate protocol supporting very similar (same?) functionality will lead to a drift.

The important of `components_to_update` parameter is that: While currently GEPA only selects one prompt to update at a time, there are many DSPy programs where prompts of 2 different predictors could be so tightly coupled that they need to be generated together in a single LLM call. While this is currently not supported, we should keep the doors open for supporting this (in the very near future)!

LakshyAAAgrawal: I recommend that for now, we can just import GEPA's protocol here, and define a default outer loop, which proposes individual prompts exactly as you do below. In future, somebody could simply override the default loop function, to do the kinds of customization I talk about above.

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -0,0 +1,280 @@
...
+            reflective_dataset: List of examples with inputs, outputs, and feedback
+                               May contain dspy.Image objects or other custom types
+
+        Returns:
+            str: Improved instruction text
+        """
+        ...
+
+
+class GenerateEnhancedMultimodalInstructionFromFeedback(dspy.Signature):
```
LakshyAAAgrawal, line 9: These proposers are AMAZING! Thank you very much

LakshyAAAgrawal (changes requested):

LakshyAAAgrawal:
Dear @andressrg,

Thank you so much for kindly contributing the custom instruction proposal to GEPA. The implementation is great. I have added a few suggestions. Overall I think this is good to merge.

dspy/teleprompt/gepa/gepa.py
jzabroski: `# Reflection based configuration` should also probably say `# Reflection configuration` so that the docstring and code are the same and the code comment functions as a backreference.

LakshyAAAgrawal: Good call @jzabroski , thanks!

dspy/teleprompt/gepa/instruction_proposal.py
andressrg: Great point!

Your proposal is a lot more powerful.

Done.

dspy/teleprompt/gepa/gepa.py
andressrg: For sure!

Updated

dspy/teleprompt/gepa/gepa_utils.py
andressrg: correct. Removed.

dspy/teleprompt/gepa/gepa_utils.py
andressrg: good point. Added 👍

dspy/teleprompt/gepa/instruction_proposal.py
andressrg: Glad you like them 🚀 🚀 🚀

dspy/teleprompt/gepa/gepa_utils.py
andressrg: good point. Added back.

andressrg:
@LakshyAAAgrawal Thank you for the detailed feedback! All comments have been addressed:

- Refactored to use GEPA's native ProposalFn protocol instead of custom InstructionProposerProtocol
- Moved instruction_proposer to Reflection configuration section and updated docstring
- Added assertion for reflection_lm when using custom proposer
- Kept the TODO comment for documentation purposes

The new ProposalFn-based API is indeed much more powerful and extensible. It opens the door for multi-component instruction updates in the future as you mentioned. Great suggestion.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -211,6 +211,13 @@ def metric(
             a strong reflection model. Consider using `dspy.LM(model='gpt-5', temperature=1.0, max_tokens=32000)` 
             for optimal performance.
         skip_perfect_score: Whether to skip examples with perfect scores during reflection. Default is True.
+        instruction_proposer: Optional custom instruction proposer implementing GEPA's ProposalFn protocol. 
+            If provided, GEPA will use this custom proposer instead of its default instruction proposal
```
LakshyAAAgrawal, line 213: Can you include some link to the `default instruction proposal` so that people wanting to implement their own logic have a reference point? Also some kind of link to `MultiModalInstructionProposer() from dspy.teleprompt.gepa.instruction_proposal`.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -10,9 +10,9 @@
 
 if TYPE_CHECKING:
```
LakshyAAAgrawal: TYPE_CHECKING remove as earlier

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -77,29 +75,22 @@ def __init__(
...
-                    dataset_with_feedback = reflective_dataset[name]
-
-                    with dspy.context(lm=self.reflection_lm):
-                        new_texts[name] = proposer(
-                            current_instruction=current_instruction,
-                            reflective_dataset=dataset_with_feedback
-                        )
-
-                return new_texts
+                with dspy.context(lm=self.reflection_lm):
```
LakshyAAAgrawal: Another flexibility we can provide here is, instead of mandating that the reflection_lm be provided and using `dspy.context` here, which essentially forces me to pick one reflection_lm and stick with it through optimization,

we can just let the user provide a custom_instruction_proposer, and call it. The custom_instruction_proposer takes care of what the reflection_lm is. If reflection_lm is provided in addition, then we use it, otherwise it is not mandatory.

LakshyAAAgrawal (changes requested):
Left a few more suggestions, I think this is the final round!

dspy/teleprompt/gepa/gepa.py
andressrg: Sure

Done

dspy/teleprompt/gepa/gepa.py
andressrg: I'm not sure how to add links here. I've just added this:

```
Use `MultiModalInstructionProposer()` from `dspy.teleprompt.gepa.instruction_proposal` 
for handling visual content. If None (default), GEPA uses its built-in text-optimized 
proposer (see `gepa.strategies.instruction_proposal.InstructionProposalSignature` 
for reference implementation).
```

Is this what you were proposing? or is there a way to add hrefs here that I might not be familiar with? 🤔

dspy/teleprompt/gepa/gepa_utils.py
andressrg: I like the idea!

Just made this change:

```
        if self.custom_instruction_proposer is not None:
            # We are only overriding the propose_new_texts method when a custom
            # instruction proposer is provided. Otherwise, we use the GEPA
            # default propose_new_texts.

            def custom_propose_new_texts(
                candidate: dict[str, str],
                reflective_dataset: dict[str, list[dict[str, Any]]],
                components_to_update: list[str]
            ) -> dict[str, str]:
                if self.reflection_lm is not None:
                    with dspy.context(lm=self.reflection_lm):
                        return self.custom_instruction_proposer(
                            candidate=candidate,
                            reflective_dataset=reflective_dataset,
                            components_to_update=components_to_update
                        )
                else:
                    return self.custom_instruction_proposer(
                        candidate=candidate,
                        reflective_dataset=reflective_dataset,
                        components_to_update=components_to_update
                    )

            self.propose_new_texts = custom_propose_new_texts
```

Now, the core `gepa/appi.py` does require the user to provide a reflection_lm:

https://github.com/gepa-ai/gepa/blob/2cf10c79125533af345ffbd48497005dd1f68ee9/src/gepa/api.py#L127

Which means that in order to have true flexibility, the core gepa library needs to be updated as well I believe.

andressrg:
Updated!

Left a question, and one comment, to make sure I understand your feedback correctly.

Thanks

dspy/teleprompt/gepa/gepa.py
LakshyAAAgrawal: I think this is markdown so \[]() should work. I think you can feel free to ignore this as nit for now, otherwise, just include a markdown link to the source location. Won't hold back merge for this!

dspy/teleprompt/gepa/gepa.py
```diff
@@ -309,11 +317,11 @@ def __init__(
         self.max_full_evals = max_full_evals
         self.max_metric_calls = max_metric_calls
 
-        # Reflection based configuration
+        # Reflection configuration
         self.reflection_minibatch_size = reflection_minibatch_size
         self.candidate_selection_strategy = candidate_selection_strategy
-        # self.reflection_lm = reflection_lm
         assert reflection_lm is not None, "GEPA requires a reflection language model to be provided. Typically, you can use `dspy.LM(model='gpt-5', temperature=1.0, max_tokens=32000)` to get a good reflection model. Reflection LM is used by GEPA to reflect on the behavior of the program and propose new instructions, and will benefit from a strong model."
```
LakshyAAAgrawal: This should be modified to say "assert custom_proposer is not None or reflection_lm is not None"

LakshyAAAgrawal (changes requested):

dspy/teleprompt/gepa/gepa_utils.py
LakshyAAAgrawal: Good catch. Let me update that!

LakshyAAAgrawal: Can you take a look at: https://github.com/gepa-ai/gepa/pull/50

I think this should fix it. Can you check?

andressrg: Nice!. Checking now

andressrg: It does fix it 🚀 

Now dspy.GEPA can support reflection_lm=None

LakshyAAAgrawal:
Released gepa v0.0.8 (with the disabled assert). You can update the pyproject.toml to reflect the change, and test against that!

If possible, it would be great to include a unit test.

andressrg:
Great! 

On it.

dspy/teleprompt/gepa/gepa.py
andressrg: sure. Done

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
andressrg: @LakshyAAAgrawal I'm aware my uv (version 0.8.14) updated the revision here.

Let me know if you prefer me to revert it.

andressrg:
@LakshyAAAgrawal Changes ready.

Also added a test to validate the system works as expected if you provide an instruction_proposer and reflection_lm=None.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -210,6 +209,14 @@ def metric(
             a strong reflection model. Consider using `dspy.LM(model='gpt-5', temperature=1.0, max_tokens=32000)` 
             for optimal performance.
         skip_perfect_score: Whether to skip examples with perfect scores during reflection. Default is True.
+        instruction_proposer: Optional custom instruction proposer implementing GEPA's ProposalFn protocol.
```
LakshyAAAgrawal, line 212: Let's add a note here, saying that "if instruction_proposer is provided and reflection_lm is provided, the instruction_proposer is called in context of the reflection_lm. However, it is not mandatory to provide reflection_lm and instruction_proposer can implement custom logic including invocation of custom LLMs." Feel free to improve upon this!

uv.lock
LakshyAAAgrawal: Yes, let's revert it

uv.lock
```diff
@@ -664,7 +664,7 @@ wheels = [
 
 [[package]]
 name = "dspy"
-version = "3.0.0"
+version = "3.0.1"
```
LakshyAAAgrawal: Should this be 3.0.3?

LakshyAAAgrawal: Probably it would be good to merge back/rebase from main and then merge this after.

dspy/teleprompt/gepa/gepa.py
andressrg: sure. On it

andressrg: Added

uv.lock
andressrg: good catch!

Rebased and ran uv lock again.

andressrg:
Feedback addressed.

I like the feedback! We want to land a polished PR 🚀

LakshyAAAgrawal (approved):

LakshyAAAgrawal:
@andressrg, Thank you very very much for this super amazing PR, and then also going through so many rounds of updates. The PR is very high quality, and adds a very useful feature to the dspy.GEPA optimizer.

Thank you very much for your contribution!

CC: @okhat

andressrg:
Thank you very for the guidance and the feedback!

GEPA is amazing and I'm looking forward to all the vision use-cases that can be optimized by leveraging it!

PR #8763, opened 2025-09-03
use mlflow instead of wandb in mlflow

TomeHirata:
This PR needs to be merged after the new gepa version is published

### DIFF
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -288,9 +288,9 @@ class GEPA(Teleprompter):
         # Logging
         log_dir: str = None,
         track_stats: bool = False,
-        use_wandb: bool = False,
-        wandb_api_key: str | None = None,
-        wandb_init_kwargs: dict[str, Any] | None = None,
+        use_mlflow: bool = False,
+        mlflow_tracking_uri: str | None = None,
+        mlflow_experiment_name: str | None = None,
         track_best_outputs: bool = False,
         # Reproducibility
         seed: int | None = 0,
@@ -347,9 +347,9 @@ class GEPA(Teleprompter):
         # Logging configuration
         self.log_dir = log_dir
         self.track_stats = track_stats
-        self.use_wandb = use_wandb
-        self.wandb_api_key = wandb_api_key
-        self.wandb_init_kwargs = wandb_init_kwargs
+        self.use_mlflow = use_mlflow
+        self.mlflow_tracking_uri = mlflow_tracking_uri
+        self.mlflow_experiment_name = mlflow_experiment_name
 
         if track_best_outputs:
             assert track_stats, "track_stats must be True if track_best_outputs is True."
@@ -501,9 +501,9 @@ class GEPA(Teleprompter):
             # Logging
             logger=LoggerAdapter(logger),
             run_dir=self.log_dir,
-            use_wandb=self.use_wandb,
-            wandb_api_key=self.wandb_api_key,
-            wandb_init_kwargs=self.wandb_init_kwargs,
+            use_mlflow=self.use_mlflow,
+            mlflow_tracking_uri=self.mlflow_tracking_uri,
+            mlflow_experiment_name=self.mlflow_experiment_name,
             track_best_outputs=self.track_best_outputs,
             display_progress_bar=True,
             raise_on_exception=True,


LakshyAAAgrawal:
Released gepa==0.0.10. Please update accordingly.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -288,9 +288,9 @@ def __init__(
         # Logging
         log_dir: str = None,
         track_stats: bool = False,
-        use_wandb: bool = False,
-        wandb_api_key: str | None = None,
-        wandb_init_kwargs: dict[str, Any] | None = None,
+        use_mlflow: bool = False,
```
LakshyAAAgrawal: Perhaps we don't even need the parameters anymore?

The users who configure mlflow globally to use with DSPy will get it automatically?

LakshyAAAgrawal:
I think it would be good to also include a unit test running GEPA once with and without MLFlow setup.

LakshyAAAgrawal:
Hi @TomeHirata ,

Thanks for the PR. I am thinking we can keep the wandb related parameters, and ensure that when MLFlow is active in the context, then  gepa.optimize is called with use_mlflow=True, and correct mlflow experiment configuration.

I am aware of some users using this with WandB, so don't want to break their workflows either.

pyproject.toml
```diff
@@ -42,7 +42,7 @@ dependencies = [
     "rich>=13.7.1",
     "numpy>=1.26.0",
     "xxhash>=3.5.0",
-    "gepa[dspy]==0.0.8",
+    "gepa[dspy]>=0.0.10",
```
LakshyAAAgrawal: Please rebase against main. I recommend pinning gepa version to a specific number since GEPA is undergoing rapid changes.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -516,6 +518,7 @@ def feedback_fn(
             use_wandb=self.use_wandb,
             wandb_api_key=self.wandb_api_key,
             wandb_init_kwargs=self.wandb_init_kwargs,
+            use_mlflow=self.use_mlflow,
```
LakshyAAAgrawal, line 533: Don't we need to also pass the `- mlflow_tracking_uri: The tracking URI to use for MLflow.
    - mlflow_experiment_name: The experiment name to use for MLflow.`?

TomeHirata: I actually think we can start with use_mlflow only. Other two parameters can be configured before instantiating dspy.GEPA and we can add them if requested.

LakshyAAAgrawal: What happens if the user has already configured mlflow experiment tracking uri and names etc outside at DSPy level, but does not pass it here?

dspy/teleprompt/gepa/gepa.py
```diff
@@ -310,6 +310,7 @@ def __init__(
         wandb_init_kwargs: dict[str, Any] | None = None,
         track_best_outputs: bool = False,
         warn_on_score_mismatch: bool = True,
+        use_mlflow: bool = False,
```
LakshyAAAgrawal, line 313: Question: If the user has configured MLFlow at DSPy level, why do we even need this flag to be enabled?

Can't we simply check if the user has enabled mlflow, and use that as the flag?

dspy/teleprompt/gepa/gepa.py
LakshyAAAgrawal: Basically, since GEPA will internally log to the mlflow tracking uri, but we are not passing the tracking uri here, what will happen?

TomeHirata: It just works as expected. The metric is logged to the tracking uri + experiment name globally configured.

dspy/teleprompt/gepa/gepa.py
TomeHirata: Configuring mlflow tracking uri globally does not mean the users want to log metrics from GEPA. Users might just want to log traces, so this flag is still necessary.

LakshyAAAgrawal: Then should we rename this to something like "track_gepa_logs_in_mlflow"?

TomeHirata: Hmm, I think it should be consistent with wandb

LakshyAAAgrawal:
Thank you very much @TomeHirata ! LGTM now! Please feel free to merge if you aren't making any other changes.

PR #8766, opened 2025-09-04
simba updates + handling optional fields

klopsahlong:


dspy/adapters/utils.py
```diff
@@ -132,8 +134,19 @@ def find_enum_member(enum, identifier):
...
 
+def _strip_optional(ann):
+    """If ann is Union[..., NoneType] return the non‑None part, else ann."""
+    if get_origin(ann) is Union and NoneType in get_args(ann):
+        # keep the first non‑None member (there will be only one in Optional[T])
+        return next(a for a in get_args(ann) if a is not NoneType)
+    return ann
 
 def parse_value(value, annotation):
+    annotation = _strip_optional(annotation)
```
klopsahlong: This fixes the following failure case:

Previously, this was failing for fields of Union(str, None) with values of type 'str' that could also be parsed as ints. Ex: "9812750". 

The problem is in this sequence:
value = "9812750" (string)
annotation = typing.Optional[str]
candidate = json_repair.loads("9812750") → 9812750 (parses as integer, not str)
TypeAdapter(typing.Optional[str]).validate_python(9812750) → Fails with pydantic.ValidationError since the value is neither a str nor None
Exception handler is triggered: except pydantic.ValidationError as e:
Then we hit this line : issubclass(annotation, Type), which throws the error `issubclass() arg 1 must be a class` because `typing.Optional[str]` is not a class, since it's a type annotation/Union.

This fix involves first parsing the Optional field to get the expected non-Null type, and parse the value according to this. Now pydantic handles the type coercion correctly from str -> str instead of int-> int when the non-null annotation type is 'str'

dspy/teleprompt/simba.py
```diff
@@ -31,6 +31,8 @@ def __init__(
         num_candidates: int = 6,
         max_steps: int = 8,
         max_demos: int = 4,
+        prompt_model: Optional[Any] = None,
+        teacher_settings: Optional[Dict] = None,
```
klopsahlong: Adding support for prompt / teacher models

dspy/teleprompt/simba_utils.py
```diff
@@ -1,58 +1,88 @@
...
-    rollout_ids = [start + i for i in range(n)]
-    return [lm.copy(rollout_id=r, temperature=1.0) for r in rollout_ids]
+
+    start_rollout_id = lm.kwargs.get("rollout_id", 0)
+    rollout_ids = [start_rollout_id + i for i in range(n)]
 
 
+    start_rollout_idx, models = 0, []
+    # If we have a teacher model, use this as the first model
+    if teacher_settings:
```
klopsahlong, line 23: This has been updated to add support for teacher model (used for 1 of the N trajectories)

dspy/teleprompt/simba_utils.py
```diff
@@ -1,58 +1,88 @@
...
+
         try:
-            score = metric(example, prediction)
+            output = metric(example, prediction)
+            if isinstance(output, (int, float)):
+                score = output
+            elif isinstance(output, dspy.Prediction):
+                if not hasattr(output, 'score'):
+                    raise ValueError("dspy.Prediction must contain a 'score' attribute")
+                score = output.score
```
klopsahlong, line 55: Updated to handle additional metric metadata in addition to the score. To do this, we check if the output from the metric is a float or int (in which case we use it as the score) or a dspy.Prediction object, which contains a score + potentially additional meta-data

dspy/teleprompt/simba_utils.py
```diff
@@ -1,58 +1,88 @@
...
     def append_a_demo_(bucket, system, **kwargs):
         predictor2name, name2predictor = kwargs["predictor2name"], kwargs["name2predictor"]
+        batch_10p_score = kwargs["batch_10p_score"]
 
-        trace = bucket[0]["trace"]
+        good = bucket[0]
+        trace = good["trace"]
         name2demo = {}
 
+        if good["score"] <= batch_10p_score:
```
klopsahlong, line 82: Double checking that the demo we're appending is not below the 10th percentile of scores

dspy/teleprompt/simba_utils.py
```diff
@@ -117,12 +147,17 @@ def append_a_rule(bucket, system, **kwargs):
         "worse_program_outputs": dict(bad["prediction"] or {}),
         "worse_reward_value": bad["score"],
         "better_reward_value": good["score"],
+        "worse_reward_info": bad["output_metadata"],
+        "better_reward_info": good["output_metadata"],
         "module_names": module_names,
```
klopsahlong, line 152: adding in the metric meta-data (ex. feedback from a judge) to help come up with a better set of rules

dspy/teleprompt/simba.py
```diff
@@ -31,6 +31,8 @@ def __init__(
         num_candidates: int = 6,
         max_steps: int = 8,
         max_demos: int = 4,
+        prompt_model: Any | None = None,
```
TomeHirata:
```suggestion
        prompt_model: dspy.LM | None = None,
```

TomeHirata: btw, can we add type annotations for other arguments too?

dspy/teleprompt/simba.py
```diff
@@ -62,6 +64,8 @@ def __init__(
         self.num_candidates = num_candidates
         self.max_steps = max_steps
         self.max_demos = max_demos
+        self.prompt_model = prompt_model if prompt_model else dspy.settings.lm
```
TomeHirata:
```suggestion
        self.prompt_model = prompt_model or dspy.settings.lm
```

dspy/teleprompt/simba_utils.py
```diff
@@ -26,33 +38,51 @@ def wrapped_program(example):
...
+            output = metric(example, prediction)
+            if isinstance(output, (int, float)):
+                score = output
+            elif isinstance(output, dspy.Prediction):
+                if not hasattr(output, "score"):
+                    raise ValueError("dspy.Prediction must contain a 'score' attribute")
+                score = output.score
+                # Just extract fields from _store, excluding 'score'
+                output_metadata = {
+                    k: v for k, v in output._store.items() if k != "score"
```
TomeHirata: can't we use `output.items()`?

dspy/teleprompt/simba_utils.py
TomeHirata: We can also use `float(output)`, which might be more intuitive?

dspy/teleprompt/simba_utils.py
```diff
@@ -77,14 +106,15 @@ def append_a_demo_(bucket, system, **kwargs):
 def append_a_rule(bucket, system, **kwargs):
     predictor2name = kwargs["predictor2name"]
     batch_10p_score, batch_90p_score = kwargs["batch_10p_score"], kwargs["batch_90p_score"]
+    prompt_model = kwargs["prompt_model"] or dspy.settings.lm
```
TomeHirata, line 109: q: is it possible that `prompt_model` is not passed? Maybe `kwargs.get("prompt_model")` is safer

dspy/teleprompt/simba_utils.py
```diff
@@ -77,14 +106,15 @@ def append_a_demo_(bucket, system, **kwargs):
...
 
     module_names = [name for name, _ in system.named_predictors()]
     good, bad = bucket[0], bucket[-1]
     example = good["example"]
 
-    if good["score"] < batch_10p_score or bad["score"] > batch_90p_score:
-        logger.info(f"Skipping rule generation as good score {good['score']} is below the 10th percentile "
-                    f"*or* bad score {bad['score']} is above the 90th percentile.")
+    if good["score"] <= batch_10p_score or bad["score"] >= batch_90p_score:
```
TomeHirata, line 115:
```suggestion
    if good <= batch_10p_score or bad >= batch_90p_score:
```

dspy/adapters/utils.py
```diff
@@ -1,3 +1,4 @@
+
```
TomeHirata: nit: blank line

dspy/teleprompt/simba_utils.py
```diff
@@ -26,33 +38,51 @@ def wrapped_program(example):
...
     def append_a_demo_(bucket, system, **kwargs):
         predictor2name, name2predictor = kwargs["predictor2name"], kwargs["name2predictor"]
+        batch_10p_score = kwargs["batch_10p_score"]
 
-        trace = bucket[0]["trace"]
+        good = bucket[0]
+        trace = good["trace"]
         name2demo = {}
 
+        if good["score"] <= batch_10p_score:
```
TomeHirata, line 82: ditto

dspy/adapters/utils.py
TomeHirata: I think we just need to change the condition from `issubclass(annotation, Type)` to `inspect.isclass(annotation) and issubclass(annotation, Type)`

chenmoneygithub: The current approach doesn't handle str | None IIUC, let me file a PR to fix this issue so that we can keep this PR focuses on simba.

chenmoneygithub: https://github.com/stanfordnlp/dspy/pull/8774, which handles the parse issue

dspy/teleprompt/simba_utils.py
```diff
@@ -26,33 +38,51 @@ def wrapped_program(example):
             try:
                 prediction = program(**example.inputs())
             except Exception as e:
-                print(e)
+                logger.info(e)
```
chenmoneygithub: should this be `logger.warning` or `logger.error`?

chenmoneygithub (commented):
Most of my comments overlap with Tomu's. Basically the function looks good, we can merge after addressing the comments.

PR #7732, opened 2025-02-04
Introduce MLflow in tutorials for better understanding for what happens under the hood

B-Step62:
## Overview
MLflow tracing helps DSPy beginners' to learn what happens under the hood and debugging any issues during the tutorials. This PR adds a helpful but non-disturbing nudges in each tutorials.

### How it looks like
Screenshot 2025-02-04 at 14 52 45

### When opened
Screenshot 2025-02-04 at 14 52 52

PR #8772, opened 2025-09-05
Add tracing banner in new tutorials

B-Step62:
Similar change to https://github.com/stanfordnlp/dspy/pull/7732. There are many more tutorials added since the last PR so improving the coverage in a batch.


docs/docs/tutorials/gepa_facilitysupportanalyzer/index.ipynb
B-Step62, line 1: This notebook was somehow condensed into one line unlike any other notebooks. The actual change just adds the same MLflow banner, but it triggers unfolding the line into proper json structure in my local. While it is not great for reviewing this change, I think it is generally a good change.

TomeHirata (approved):
LGTM

PR #8774, opened 2025-09-05
Fix value parsing and add tests

chenmoneygithub:
the current `parse_value` implementation doesn't work well with `Optional[str]` and `str | None`, and this PR fixes it. Also adding unit test for safe guarding.

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -2,6 +2,7 @@ import ast
 import enum
 import inspect
 import json
+import types
 from collections.abc import Mapping
 from typing import Any, Literal, Union, get_args, get_origin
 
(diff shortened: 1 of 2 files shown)


tests/adapters/test_adapter_utils.py
```diff
@@ -0,0 +1,101 @@
+# ruff: noqa: UP007
```
TomeHirata, line 1: can we change the filename to test_utils.py for consistency?

TomeHirata (approved):

tests/adapters/test_adapter_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    # Test invalid literal
+    with pytest.raises(ValueError):
+        parse_value("invalid", Literal["option1", "option2"])
+
+
+def test_parse_value_union():
+    # Test Union with None (Optional)
+    assert parse_value("test", Optional[str]) == "test"
+    assert parse_value("test", str | None) == "test"
+    assert parse_value(None, Optional[str]) is None
```
okhat, line 84: I'm not 100% sure if this covers everything. What if the LLM outputs the string "None".

Testing the function `parse_value` is kind of too late at that point right?

PR #8778, opened 2025-09-08
Allow custom type to be streamed and use native response

TomeHirata:
This PR increases the extensibility of dspy.Type and allows users to define a custom type that can be streamed or use native LM response. This is useful when users want to use provider specific response field such as tool call or citation.

Following methods are introduced into dspy.Type
- is_streamable: Whether the custom type is streamable
- parse_stream_chunk: Parse a stream chunk into the custom type
- use_native_response: Whether to use the native response of the LM such as function calling


okhat:
No concerns on a quick scan. If @chenmoneygithub approves, feel free to merge.

dspy/adapters/base.py
```diff
@@ -64,9 +63,11 @@ def _call_preprocess(
 
                 return signature_for_native_function_calling
 
-        citation_output_field_name = self._get_citation_output_field_name(signature)
-        if citation_output_field_name:
-            signature = signature.delete(citation_output_field_name)
+        # Handle custom types that use native response
+        for name, field in signature.output_fields.items():
+            if isinstance(field.annotation, type) and issubclass(field.annotation, Type):
+                if field.annotation.use_native_response(lm.model):
```
chenmoneygithub: shall we merge this condition check with the line above?

dspy/adapters/base.py
```diff
@@ -114,9 +113,11 @@ def _call_postprocess(
                 ]
                 value[tool_call_output_field_name] = ToolCalls.from_dict_list(tool_calls)
 
-            if citations and citation_output_field_name:
-                citations_obj = Citations.from_dict_list(citations)
-                value[citation_output_field_name] = citations_obj
+            # Parse custom types that does not rely on the adapter parsing
```
chenmoneygithub, line 118: maybe also moving the handling of `tool_calls` into the for loop?

dspy/adapters/types/base_type.py
```diff
@@ -67,6 +68,47 @@ def serialize_model(self):
...
+        Args:
+            chunk: A stream chunk.
+
+        Returns:
+            A custom type object or None if the chunk is not for this custom type.
+        """
+        return None
+
+    @classmethod
+    def use_native_response(cls, model: str) -> bool:
```
chenmoneygithub: I am feeling this doesn't belong to the class. For example, `ToolCalls` can either use native functionality or use custom functionality (so does `Citations` but we just don't recommend the custom path). This is more like a configuration on the adapter level. But i am not 100% sure what's the best way to support this config on adapter, there are a few choices:

- Add an arg to constructor `field_type_using_native_response`, which carries the type that uses native features, e.g., `field_type_using_native_response=[dspy.ToolCalls, dspy.Citations]`. The downside is users can get very confused about this arg.
- For each type we make an attribute like `use_native_tool_call`, `use_native_citation` inside the adapter. The caveat is we may see indefinitely growing attributes.
- We only make nondeterministic types like `ToolCalls` configurable through `use_native_tool_call`, for types we surely know that they come from native features like `Citations` (actually debatable) there will be a static list.

would like to hear your thoughts on this!

dspy/streaming/streaming_listener.py
```diff
@@ -249,10 +249,10 @@ def find_predictor_for_stream_listeners(program: "Module", stream_listeners: lis
                     "predictor to use for streaming. Please specify the predictor to listen to."
                 )
 
-            if field_info.annotation not in [str, Citations]:
+            if not _is_streamable(field_info.annotation):
                 raise ValueError(
-                    f"Stream listener can only be applied to string or Citations output field, but your field {field_name} is of "
-                    f"type {field_info.annotation}."
+                    f"Stream listener can only be applied to string or streamable dspy type output field, "
```
chenmoneygithub: Users may get confused by what "streamable dspy type output field" means, maybe something like `subclass of `dspy.Type` that has `is_streamable() == True`?

chenmoneygithub (commented):
Very clean and cool! Left some comments for discussion.

dspy/adapters/base.py
TomeHirata: Yes, will do in a separate PR

dspy/adapters/types/base_type.py
TomeHirata: I agree it's natural that adapters should own this configuration. But I didn't put the config on adapters at first since users need to implement custom adapters too when they introduce a new custom type. I thought it'd be helpful if users can just introduce a new dspy.Type and use it. But I actually like the `field_type_using_native_response` approach, shall we go with this option?

chenmoneygithub: sorry for the late reply, yes sg!

TomeHirata:
@chenmoneygithub I've moved the native response type config to adapter, can you take a look again?

dspy/streaming/streaming_listener.py
```diff
@@ -217,10 +214,13 @@ def flush(self) -> str:
                 f"{', '.join([a.__name__ for a in ADAPTER_SUPPORT_STREAMING])}"
             )
 
-    def _signature_field_is_citation_type(self) -> bool:
-        """Check if the signature field is a citations field."""
-        from dspy.predict import Predict
-        return isinstance(self.predict, Predict) and getattr(self.predict.signature.output_fields.get(self.signature_field_name, None), "annotation", None) == Citations
+    @property
+    def output_type(self) -> type | None:
```
chenmoneygithub: qq - is this accessed out of the class?

dspy/streaming/streaming_listener.py
```diff
@@ -217,10 +214,13 @@ def flush(self) -> str:
...
+        try:
+            return self.predict.signature.output_fields[self.signature_field_name].annotation
+        except Exception:
```
chenmoneygithub, line 221: when will this throw an exception?

dspy/adapters/base.py
```diff
@@ -18,9 +17,10 @@
 
 
 class Adapter:
-    def __init__(self, callbacks: list[BaseCallback] | None = None, use_native_function_calling: bool = False):
+    def __init__(self, callbacks: list[BaseCallback] | None = None, use_native_function_calling: bool = False, native_response_types: list[type[Type]] | None = None):
         self.callbacks = callbacks or []
         self.use_native_function_calling = use_native_function_calling
+        self.native_response_types = native_response_types or []
```
chenmoneygithub: should we default to `[Citations]`?

chenmoneygithub (commented):
Some minor comments, otherwise LGTM!

dspy/adapters/base.py
TomeHirata: Good point, let's do that.

dspy/streaming/streaming_listener.py
TomeHirata: none, let's make it a private function

dspy/streaming/streaming_listener.py
TomeHirata: It does not fail in normal cases, but since `predict` or `signature_field_name` is not guaranteed to be non-null, it's better to have a safeguard

chenmoneygithub: ic, actually I think `predict` and `signature_field_name` must be configured for stream listener, otherwise it is not doing anything.

chenmoneygithub (commented):
LGTM after https://github.com/stanfordnlp/dspy/pull/8778#discussion_r2350429305 is resolved.

chenmoneygithub (approved):

dspy/streaming/streaming_listener.py
TomeHirata: I know, but typing-wise they are nullable (https://github.com/stanfordnlp/dspy/blob/28a7f786f890d8bcfd9aa83bedf1465c97deb940/dspy/streaming/streaming_listener.py#L24-L29), so I still want to keep the check. We can do if chain, as well.

PR #8779, opened 2025-09-08
feat(adapters): Raise an error for values inside signatures that are not compatible with the DSPyType annotation

isaacbmiller:
Currently, we pass placeholders values into an adapter to get the formatting. E.g.:
```
[ ## question ## ]
{question}
```

This is problematic for custom types. When you pass anything as the value for this field, the adapter just assumes it is a string (or can be represented as a string). The value must always be representable by a string, thus why the `dspy.Type` class has a `format()` method. What the current implementation misses is that sometimes you pass in a value to this field that you need to then coerce into your desired Type.

This is poignant for Images, where a user may pass in a PIL image, a base 64 string, or a URL.  They can pass this value directly into the predictor. We need the adapter to look if the annotation is supposed to be a dspy.Type. If it is, we need to check if the value is a valid init for the Type. If it is, we can coerce it into the Type, then let the adapter call the .format on the value as the proper Type.

Because we pass in both placeholders and the values into the adapter, the adapter needs to know if it should try to coerce into the type or not. You could even make an argument that a dspy.type should come up with it's own placeholder but that is out of scope for this PR.

There are two main features in this PR:
1. Adding the `is_placeholder` field to the FieldInfoWithName class. The reason we add this is because we use this class to represent both placeholder values (e.g. FieldInfoWithName(name="question", info={value="{question}"}). This allows the adapter to know whether or not it needs to try and coerce the type.
2. Add the coercion logic inside all the relevant adapters. This will call the `.model_validate` method on the type with the passed in values as the parameters. All Types have this method.

There are some minor changes as well:
1. We move the `encode_image` logic into the model_validator method for images. You should never be calling this on its own, especially now that we are doing more work behind the scenes to ensure that the user's intent is automatically figured out.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -22,7 +22,6 @@ field_header_pattern = re.compile(r"\[\[ ## (\w+) ## \]\]")
 
 class FieldInfoWithName(NamedTuple):
     name: str
-    is_placeholder: bool
     info: FieldInfo
 
 
@@ -83,11 +82,9 @@ class ChatAdapter(Adapter):
         parts.append("All interactions will be structured in the following way, with the appropriate values filled in.")
 
         def format_signature_fields_for_instructions(fields: dict[str, FieldInfo]):
-            # This is where placeholders are passed in. We use the field_name as the value.
-            # The placeholders are officially generated inside `translate_field_type`
             return self.format_field_with_value(
                 fields_with_values={
-                    FieldInfoWithName(name=field_name, is_placeholder=True, info=field_info): translate_field_type(field_name, field_info)
+                    FieldInfoWithName(name=field_name, info=field_info): translate_field_type(field_name, field_info)
                     for field_name, field_info in fields.items()
                 },
             )
@@ -162,7 +159,7 @@ class ChatAdapter(Adapter):
     ) -> str:
         assistant_message_content = self.format_field_with_value(
             {
-                FieldInfoWithName(name=k, is_placeholder=k in outputs, info=v): outputs.get(k, missing_field_message)
+                FieldInfoWithName(name=k, info=v): outputs.get(k, missing_field_message)
                 for k, v in signature.output_fields.items()
             },
         )
@@ -221,7 +218,7 @@ class ChatAdapter(Adapter):
         """
         output = []
         for field, field_value in fields_with_values.items():
-            formatted_field_value = format_field_value(field_info=field.info, value=field_value, is_placeholder=field.is_placeholder)
+            formatted_field_value = format_field_value(field_info=field.info, value=field_value)
             output.append(f"[[ ## {field.name} ## ]]\n{formatted_field_value}")
 
         return "\n\n".join(output).strip()
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -109,7 +109,7 @@ class JSONAdapter(ChatAdapter):
         def format_signature_fields_for_instructions(fields: dict[str, FieldInfo], role: str):
             return self.format_field_with_value(
                 fields_with_values={
-                    FieldInfoWithName(name=field_name, is_placeholder=True, info=field_info): translate_field_type(field_name, field_info)
+                    FieldInfoWithName(name=field_name, info=field_info): translate_field_type(field_name, field_info)
                     for field_name, field_info in fields.items()
                 },
                 role=role,
@@ -141,7 +141,7 @@ class JSONAdapter(ChatAdapter):
         missing_field_message=None,
     ) -> str:
         fields_with_values = {
-            FieldInfoWithName(name=k, is_placeholder=k in outputs, info=v): outputs.get(k, missing_field_message)
+            FieldInfoWithName(name=k, info=v): outputs.get(k, missing_field_message)
             for k, v in signature.output_fields.items()
         }
         return self.format_field_with_value(fields_with_values, role="assistant")
@@ -192,7 +192,7 @@ class JSONAdapter(ChatAdapter):
         if role == "user":
             output = []
             for field, field_value in fields_with_values.items():
-                formatted_field_value = format_field_value(field_info=field.info, value=field_value, is_placeholder=field.is_placeholder)
+                formatted_field_value = format_field_value(field_info=field.info, value=field_value)
                 output.append(f"[[ ## {field.name} ## ]]\n{formatted_field_value}")
             return "\n\n".join(output).strip()
         else:
--- a/dspy/utils/dummies.py
+++ b/dspy/utils/dummies.py
@@ -101,7 +101,7 @@ class DummyLM(LM):
     def __call__(self, prompt=None, messages=None, **kwargs):
         def format_answer_fields(field_names_and_values: dict[str, Any]):
             fields_with_values = {
-                FieldInfoWithName(name=field_name, info=OutputField(), is_placeholder=False): value
+                FieldInfoWithName(name=field_name, info=OutputField()): value
                 for field_name, value in field_names_and_values.items()
             }
             # The reason why DummyLM needs an adapter is because it needs to know which output format to mimic.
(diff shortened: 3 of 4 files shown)


dspy/adapters/chat_adapter.py
```diff
@@ -44,6 +45,7 @@ def __call__(
                 # On context window exceeded error or already using JSONAdapter, we don't want to retry with a different
                 # adapter.
                 raise e
+            print("fallback to JSONAdapter: ", e)
```
isaacbmiller: remove

tests/adapters/test_json_adapter.py
```diff
@@ -115,6 +116,7 @@ def test_json_adapter_sync_call():
 @pytest.mark.asyncio
 async def test_json_adapter_async_call():
     signature = dspy.make_signature("question->answer")
+    # why is this a chat adapter?
```
isaacbmiller: fix

tests/adapters/test_json_adapter.py
```diff
@@ -106,6 +106,7 @@ class TestSignature(dspy.Signature):
 
 def test_json_adapter_sync_call():
     signature = dspy.make_signature("question->answer")
+    # why is this a chat adapter?
```
isaacbmiller: fix

tests/signatures/test_adapter_image.py
```diff
@@ -126,37 +126,62 @@ def test_basic_image_operations(test_case):
...
+        # "two_step_adapter", 
+        "baml_adapter", 
+        "xml_adapter",
+    ],
+)
 @pytest.mark.parametrize(
     "image_input,description",
     [
         ("pil_image", "PIL Image"),
+        ("dspy_pil_image_with_download", "PIL Image with download=True; This is a slightly adversarial case"),
```
isaacbmiller: change to comment

dspy/utils/dummies.py
```diff
@@ -94,12 +95,24 @@ def _use_example(self, messages):
     @with_callbacks
     def __call__(self, prompt=None, messages=None, **kwargs):
         def format_answer_fields(field_names_and_values: dict[str, Any]):
-            return ChatAdapter().format_field_with_value(
-                fields_with_values={
-                    FieldInfoWithName(name=field_name, info=OutputField()): value
-                    for field_name, value in field_names_and_values.items()
-                }
-            )
+            # This should actually get the current adapter rather than default to chat adapter
```
isaacbmiller: remove comment

dspy/adapters/utils.py
```diff
@@ -32,18 +34,27 @@ def serialize_for_json(value: Any) -> Any:
...
       field_info: Information about the field, including its DSPy field type and annotation.
       value: The value of the field.
+      is_placeholder: Whether the value is a placeholder for a field.
     Returns:
       The formatted value of the field, represented as a string.
     """
     string_value = None
+    value = deepcopy(value)
+
+    # If the annotation is a Type instance, but the value is not a instance of that type, force it to be one.
```
isaacbmiller: change comment to say Type subclass

dspy/adapters/chat_adapter.py
```diff
@@ -82,9 +84,11 @@ def format_field_structure(self, signature: type[Signature]) -> str:
         parts.append("All interactions will be structured in the following way, with the appropriate values filled in.")
 
         def format_signature_fields_for_instructions(fields: dict[str, FieldInfo]):
+            # This is where placeholders are generated. It is weird because we are using the field_name as the value, and that is wrong.
+            # Really we expect that any value going into format_field_with_value is of the annotation type, not we get away with this because we need things as strings.
```
isaacbmiller: not -> but

dspy/adapters/json_adapter.py
```diff
@@ -109,7 +109,7 @@ def format_field_structure(self, signature: type[Signature]) -> str:
         def format_signature_fields_for_instructions(fields: dict[str, FieldInfo], role: str):
             return self.format_field_with_value(
                 fields_with_values={
-                    FieldInfoWithName(name=field_name, info=field_info): translate_field_type(field_name, field_info)
+                    FieldInfoWithName(name=field_name, is_placeholder=True, info=field_info): translate_field_type(field_name, field_info)
```
isaacbmiller: consistent ordering for FieldInfoWithName

dspy/adapters/utils.py
```diff
@@ -32,18 +34,27 @@ def serialize_for_json(value: Any) -> Any:
...
+    if issubclass(field_info.annotation, Type) and not isinstance(value, field_info.annotation) and not is_placeholder:
```
isaacbmiller: @okhat this is the important part of this PR

tests/signatures/test_adapter_image.py
```diff
@@ -126,37 +126,62 @@ def test_basic_image_operations(test_case):
...
+        "baml_adapter": (dspy.BAMLAdapter(), {"probabilities": {"dog": 0.8, "cat": 0.1, "bird": 0.1}}),
+        "xml_adapter": (dspy.XMLAdapter(), {"probabilities": {"dog": 0.8, "cat": 0.1, "bird": 0.1}}),
+        # "two_step_adapter": dspy.TwoStepAdapter(),
+    }
+
+    if adapter_type == "two_step_adapter":
+        pytest.xfail("TwoStepAdapter is not known to support image input")
+
     actual_input = input_map[image_input]
     # TODO(isaacbmiller): Support the cases without direct dspy.Image coercion
```
isaacbmiller: remove comment

chenmoneygithub (commented):
Thanks for the PR!

I am a bit confused about the change, could you share a reproducible code that breaks at the current github code? thank you!

isaacbmiller:
Yup! The error case this fixes is when you pass in what 'should be' a dspy.Type, but isn't.

Notice how in this function, we *do not* call dspy.Image(pil_image)
```python
def test_minimal_image_format_repro():

    lm = dspy.LM("gpt-5-nano", temperature=1.0, max_tokens=16000)
    pred = dspy.Predict("image: dspy.Image -> description: str")

    url = "https://images.dog.ceo/breeds/dane-great/n02109047_8912.jpg"
    response = requests.get(url)
    pil_image = PILImage.open(BytesIO(response.content))
    pred(
        image=pil_image,
    )
    lm.inspect_history(1)
```

Results on main:
```
User message:

[[ ## image ## ]]
<PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=640x392 at 0x10AA57710>

Respond with the corresponding output fields, starting with the field `[[ ## description ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.


Response:

[[ ## description ## ]]
The image is a JPEG (RGB) with resolution 640x392. The content of the scene is not available from the prompt, so I cannot provide a description of what is depicted.
[[ ## completed ## ]]
```

Doesn't error out, just sends the string description of the PIL image.

But on this branch, we get the correct description and the string representation is properly printed:
```
User message:

[[ ## image ## ]]
<data:image/jpeg;base64,<IMAGE BASE 64 ENCODED(104552)>

Respond with the corresponding output fields, starting with the field `[[ ## description ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.


Response:

[[ ## image ## ]]
{image}

[[ ## description ## ]]
A large, muscular blue-gray dog with cropped ears is lounging on the ground in a shaded outdoor area. It wears a red collar and rests with its front legs stretched out in front, near a tree trunk amid dry leaves and low greenery.

[[ ## completed ## ]]
```

Note that it did repeat the image tag but GPT-5-nano was just being silly and it parses fine.

chenmoneygithub:
@isaacbmiller Thanks for the code! 

The current implementation adds lots of code to handle the case where users provide invalid type to certain fields, which will be hard to maintain in a few months. Can we add a check on `dspy.Predict` to verify that every input value follows the declared type, if not just throw an error? We don't need to do silent handling for user errors, and actually silent handling can lead to unexpected issues.

isaacbmiller:
Yeah sure type checking the inputs works too.

PR #8788, opened 2025-09-10
Guide on using native tool calling

chenmoneygithub:
Adding a section to https://dspy.ai/learn/programming/tools/ to guide users on how to use native tool calling.


docs/docs/learn/programming/tools.md
```diff
@@ -75,7 +75,8 @@ class ToolSignature(dspy.Signature):
     """Signature for manual tool handling."""
     question: str = dspy.InputField()
     tools: list[dspy.Tool] = dspy.InputField()
-    outputs: dspy.ToolCalls = dspy.OutputField()
+    tool_calls: dspy.ToolCalls = dspy.OutputField()
+    answer: str = dspy.OutputField()
```
TomeHirata: Isn't it weird to let the Predict return answer before executing tool calls?

docs/docs/learn/programming/tools.md
```diff
@@ -148,6 +149,54 @@ for call in response.outputs.tool_calls:
...
+
+You can enable the [MLflow tracing](https://dspy.ai/tutorials/observability/) to check how native tool
+calling is being used. If you use `JSONAdapter` or `ChatAdapter` with native function calling enabled on the code snippet
+as provided in [the section above](tools.md#basic-setup), you should see native function calling arg `tools` is set like
+the screenshot below:
+
+![native tool calling](../figures/native_tool_call.png)
+
+
+Please note that it's possible that native tool calling produces lower quality than custom tool calling, we encourage you to try both approaches and see which one works better for your use case.
```
TomeHirata: Do we need to repeat this sentence twice?

docs/docs/learn/programming/tools.md
```diff
@@ -148,6 +149,54 @@ for call in response.outputs.tool_calls:
         print(f"Result: {result}")
 ```
 
+## Using Native Tool Calling
```
TomeHirata, line 151: Open to discussion, but I wonder if we should move this section under `Best Practices`? The current flow is ReAct -> ToolOutput -> which to use when, so it might be better to describe this additional information like an appendix?

docs/docs/learn/programming/tools.md
chenmoneygithub: This is the intentional behavior of tool calling, basically the prompt can lead to either a tool call or a direct answer. Without providing the `answer`, our prompt looks a bit strange because the output fields are empty.

docs/docs/learn/programming/tools.md
chenmoneygithub: No strong opinion, but I did this by intention because this is very important. Many people including experienced developers just assume native function calling works better than custom parsing, which could lead to them questioning DSPy's implementation/quality/...

docs/docs/learn/programming/tools.md
chenmoneygithub: Good call! Actually I am not very sure so I just asked claude code. 

My prompt which is neutral:

```
> one question about the layout, do you think the "Using Native Function Calling"
   section should be kept at the current position, or moved under Best Practices?
   both have valid reasons, please help me decide.
```

Truncated Claude Code response:

```
  Why not under Best Practices:

  - Best Practices is typically for optimization tips, coding patterns, and
  general advice
  - Native tool calling is a foundational implementation choice, not a
  "practice" or "tip"
  - It would get buried under Best Practices when it's actually a core concept
  users need to understand early

  The current structure creates a nice flow:
  1. Approach 1: Fully automated (ReAct)
  2. Approach 2: Manual control
  3. Native Tool Calling: Implementation choice for Approach 2
  4. Best Practices: General tips for both approaches

  I'd recommend keeping it where it is! 👍
```

Claude code suggests keeping the current flow.

docs/docs/learn/programming/tools.md
okhat: Ah. I think this is just too weird sadly.

chenmoneygithub: @okhat does it feel strange to include the `answer` output field? I might be playing with OAI's tool call too much, so I feel used to keep both `tool_calls` and `answer` present in the same LM call. Would like to understand this better

docs/docs/learn/programming/tools.md
TomeHirata: Sure, let's keep it as it is then?

docs/docs/learn/programming/tools.md
TomeHirata: I'd remove the duplicated text for now. If people still question it even if we emphasize it in the doc, we can revisit the content.

docs/docs/learn/programming/tools.md
TomeHirata: I see, then I think at least the output field type should be `Optional[str]` so that we don't force LM to always provide an answer. But for the demonstration of the flow, I think we can remove the `answer` field from this example code for simplicity.

chenmoneygithub: sg, removed

docs/docs/learn/programming/tools.md
chenmoneygithub: sg!

docs/docs/learn/programming/tools.md
chenmoneygithub: sure thing, done

PR #8791, opened 2025-09-10
Self contained miprov example

crawftv:
A script with self-contained data and multiple optimizers.


isaacbmiller:
This is similar to existing DSPy tutorials, and in scripts rather than notebooks. Closing for now.

Issue #8804, opened 2025-09-13
[Bug] Unexpected parsing of Optional[str] fields when string has brackets or braces


sontanon:
### What happened?

I was working with a signature that has an `Optional[str]` (equivalently, `str | None`) output field. My use case had the LM returning text such as `"this text has a [placeholder] value"`. This lead to Pydantic validation errors and I was able to trace that the issue in `dspy/adapters/utils.py:parse_value` where there is no corner case for `Optional[str]`. My proposed fix is to add 

```py
# Add this import
from types import UnionType

# ... Continue on line 137
def parse_value(value, annotation):
    if annotation is str:
        return str(value)

    if isinstance(annotation, enum.EnumMeta):
        return find_enum_member(annotation, value)

    origin = get_origin(annotation)

    # Add this check here. Will only return early for `None` or strings
    if (origin in (Union, types.UnionType) 
        and type(None) in get_args(annotation) 
        and str in get_args(annotation)):
        if value is None:
            return None
        if isinstance(value, str):
            return value
    # Rest of the code continues here
```

### Steps to reproduce

Consider a signature that returns an optional string and an associated module:

```py
import dspy

class Foo(dspy.Signature):
    bar: str | None = dspy.OutputField()

predictor = dspy.Predict(Foo)
```

Now, suppose the LM returns valid JSON such as 

```json
{"bar": "this text has a [placeholder] value"}
```

We can mock this using this LM:

```py
from openai.types.chat.chat_completion import ChatCompletion

def create_dummy_lm(content: str) -> dspy.BaseLM:
    dummy_completion = ChatCompletion.model_validate({
        "object": "chat.completion",
        "id": "",
        "created": 0,
        "model": "dummy_model",
        "choices": [
            {
                "finish_reason": "stop",
                "index": 0,
                "message": {
                    "role": "assistant",
                    "content": content
                }
            }
        ],
        "usage": {
            "completion_tokens": 0,
            "prompt_tokens": 0,
            "total_tokens": 0
        }
    })

    class DummyLM(dspy.BaseLM):
        def __init__(self):
            super().__init__(model="dummy_model")

        def forward(self, prompt=None, messages=None, **kwargs):
            return dummy_completion
        
    return DummyLM()
```

Then, the buggy behavior can be seen in this snippet

```py
with dspy.context(lm=create_dummy_lm("{\"bar\": \"this text has a [placeholder] value\"}"):
  p = predictor()
print(test)
```

This will output

```
Prediction(
[...]

PR #8805, opened 2025-09-13
Fix unexpected parsing of Optional[str] fields when string has brackets or braces

sontanon:
This is my proposed fix for #8804


chenmoneygithub (approved):
Thanks for the PR! I made some small edits to get rid of duplicated logic.

PR #8811, opened 2025-09-16
[MIPRO v2] Fix parameter update order in `MIPROv2.compile`

JamesSand:
In the `compile` function of `MIPROv2`, the parameters `max_bootstrapped_demos` and `max_labeled_demos` are intended to override the defaults and update the `zeroshot_opt` value.

However, in the main branch, `zeroshot_opt` is assigned **before** `self.max_bootstrapped_demos` and `self.max_labeled_demos` are updated. This makes the passed-in parameters ineffective.

This PR fixes the issue by moving the update of `self.max_bootstrapped_demos` and `self.max_labeled_demos` **before** assigning `zeroshot_opt`.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -124,6 +124,13 @@ class MIPROv2(Teleprompter):
             if self.max_errors is not None
             else dspy.settings.max_errors
         )
+
+        # Update max demos if specified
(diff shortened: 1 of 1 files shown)


chenmoneygithub (commented):
@klopsahlong qq - do we expect to override the `self.max_bootstrapped_demos` at every call? maybe we should keep the attribute value unchanged?

PR #8814, opened 2025-09-16
Parse doubly-encoded base type in json.parse

TomeHirata:
Currently, DSPy custom types can be encoded twice: once in `Type.serialize_model` and once in `format_field_value`, which makes it impossible to parse the value `json.parse`. This PR fixes the behavior by decoding the extracted string twice.

### DIFF
--- a/dspy/adapters/types/base_type.py
+++ b/dspy/adapters/types/base_type.py
@@ -64,7 +64,7 @@ class Type(pydantic.BaseModel):
     def serialize_model(self):
         formatted = self.format()
         if isinstance(formatted, list):
-            return f"{CUSTOM_TYPE_START_IDENTIFIER}{formatted}{CUSTOM_TYPE_END_IDENTIFIER}"
+            return f"{CUSTOM_TYPE_START_IDENTIFIER}{json.dumps(formatted, ensure_ascii=False)}{CUSTOM_TYPE_END_IDENTIFIER}"
         return formatted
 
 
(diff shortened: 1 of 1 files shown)


dspy/adapters/types/base_type.py
```diff
@@ -141,3 +143,11 @@ def split_message_content_for_custom_types(messages: list[dict[str, Any]]) -> li
         message["content"] = result
 
     return messages
+
+
+def _parse_doubly_quoted_json(json_str: str) -> Any:
+    """
+    Parse a doubly quoted JSON string into a Python dict.
+    BaseType can be json-encoded twice if included in either list or dict.
```
chenmoneygithub: We can change `BaseType` => `Type`.

dspy/adapters/types/base_type.py
```diff
@@ -64,7 +64,7 @@ def extract_custom_type_from_annotation(cls, annotation):
     def serialize_model(self):
         formatted = self.format()
         if isinstance(formatted, list):
-            return f"{CUSTOM_TYPE_START_IDENTIFIER}{formatted}{CUSTOM_TYPE_END_IDENTIFIER}"
+            return f"{CUSTOM_TYPE_START_IDENTIFIER}{json.dumps(formatted, ensure_ascii=False)}{CUSTOM_TYPE_END_IDENTIFIER}"
```
chenmoneygithub: is this `json.dumps` added to replace the `custom_type_content.replace("'", '"')` below?

PR #8818, opened 2025-09-17
Add save/load to Embeddings

### LINKED ISSUE #8807: [Feature] Add .save and .load methods in embeddings.py
I just think it might be better this TODO to be here instead of only hardcoded and hidden in the codebase.

https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/embeddings.py

TomeHirata:
Closes #8807

### DIFF
--- a/dspy/retrievers/embeddings.py
+++ b/dspy/retrievers/embeddings.py
@@ -1,5 +1,5 @@
+import json
 import os
-import pickle
 from typing import Any
 
 import numpy as np
@@ -108,8 +108,8 @@ class Embeddings:
             "has_faiss_index": self.index is not None,
         }
 
-        with open(os.path.join(path, "config.pkl"), "wb") as f:
-            pickle.dump(config, f)
+        with open(os.path.join(path, "config.json"), "w") as f:
+            json.dump(config, f, indent=2)
 
         # Save embeddings
         np.save(os.path.join(path, "corpus_embeddings.npy"), self.corpus_embeddings)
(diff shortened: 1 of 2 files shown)


dspy/retrievers/embeddings.py
```diff
@@ -87,3 +87,92 @@ def _rerank_and_predict(self, q_embeds: np.ndarray, candidate_indices: np.ndarra
...
+
+        # Save configuration and corpus
+        config = {
+            "k": self.k,
+            "normalize": self.normalize,
+            "corpus": self.corpus,
+            "has_faiss_index": self.index is not None,
+        }
+
+        with open(os.path.join(path, "config.pkl"), "wb") as f:
```
chenmoneygithub: Looks like this config is json-serializable, shall we use json instead?

dspy/retrievers/embeddings.py
```diff
@@ -87,3 +87,92 @@ def _rerank_and_predict(self, q_embeds: np.ndarray, candidate_indices: np.ndarra
...
+                import faiss
+                self.index = faiss.read_index(faiss_index_path)
+            except ImportError:
+                # If FAISS is not available, fall back to brute force
+                self.index = None
+        else:
+            self.index = None
+
+        # Reinitialize the search function
+        self.search_fn = Unbatchify(self._batch_forward)
```
chenmoneygithub: do we need to reinitialize the search_fn?

dspy/retrievers/embeddings.py
```diff
@@ -87,3 +87,92 @@ def _rerank_and_predict(self, q_embeds: np.ndarray, candidate_indices: np.ndarra
...
+        if self.index is not None:
+            try:
+                import faiss
+                faiss.write_index(self.index, os.path.join(path, "faiss_index.bin"))
+            except ImportError:
+                # If FAISS is not available, we can't save the index
+                # but we can still save the embeddings for brute force search
+                pass
+
+    def load(self, path: str, embedder):
```
chenmoneygithub, line 127: How do we want this `load` to be called? Do we want users to first create an `Embedding` instance then call `embedding.load()`, or make it a class method that return a loaded embedding?

TomeHirata: This is actually only for consistency with other APIs like module.load. I guess mostly people will use from_saved. Do you think we should make load a classmethod?

dspy/retrievers/embeddings.py
TomeHirata: Yup, this is because `from_saved` bypasses `__init__`

dspy/retrievers/embeddings.py
TomeHirata: Good idea!

dspy/retrievers/embeddings.py
chenmoneygithub: Gotcha, I am asking because of the line `self.search_fn = Unbatchify(self._batch_forward)`, if we do:

```
embedder = dspy.Embeddings(...)
embedder.load(...)
```

do we still need this `self.search_fn = Unbatchify(self._batch_forward)`?

chenmoneygithub:
LGTM after https://github.com/stanfordnlp/dspy/pull/8818#discussion_r2370219637 is resolved, thank you!

dspy/retrievers/embeddings.py
TomeHirata: This was necessary because from_saved bypasses `__init__` and self.load is called inside `from_saved`. Let me move this logic to `from_saved` then.

PR #8821, opened 2025-09-18
perf(adapter): temporarily hash large data when formatting chat prompt for speed+memory performance

jmhb0:
###  Summary
-  Replaces multi‑MB inline payloads (e.g., base64 images) with small hash strings during adapter formatting (few‑shot demos and history), restoring the original data right before sending to the LM. 
- This is much faster for requests with many images (~95% faster in this experiment, but it depends on size and number of images). It also uses less memory. 
- If the maintainers agree with the PR, then I can extend it to the other larger dspy Types like Audio and Document

### Problem
Building prompts in `Adapter` is slow when using large data types like `dspy.Image`. [This gist](https://gist.github.com/jmhb0/0c0d208ff7ef6797723acb726f8a52f2) is a realistic reproducible example: a (1000,1000) image, with 15 few-shot images. It profiles running a prediction for 100 images after warming up the cache: 
```
Total processing time: 53.33 seconds
Examples processed: 100
Average time per example: 0.5333 seconds
Throughput: 1.88 examples/second

Peak RSS memory usage: 6.466 GB
Average RSS memory usage: 3.564 GB
Memory samples collected: 18

Function-level profiling results:
--------------------------------------------------------------------------------
Top 20 functions by TOTAL TIME (self time - actual CPU work):
         1282039 function calls (1208276 primitive calls) in 53.330 seconds

   Ordered by: internal time
   List reduced from 625 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100   32.564    0.326   36.929    0.369 /Users/jamesburgess/dspy/dspy/adapters/types/base_type.py:104(split_message_content_for_custom_types)
     1600   12.177    0.008   12.177    0.008 /Users/jamesburgess/miniconda3/envs/dspy-dev/lib/python3.11/json/encoder.py:205(iterencode)
     1700    4.261    0.003    4.262    0.003 /Users/jamesburgess/miniconda3/envs/dspy-dev/lib/python3.11/json/decoder.py:343(raw_decode)
      100    2.300    0.023    2.300    0.023 {built-in method _hashlib.openssl_sha256}
      100    0.699    0.007    0.699    0.007 {orjson.dumps}
     6100    0.388    0.000    0.388    0.000 {method 'strip' of 'str' objects}
        4    0.154    0.038    0.154    0.038 {built-in method gc.collect}
```
Building 100 prompts takes 53 seconds, which is very slow for just making a prompt.  This is especially noticeable when trying to rerun a cached prediction with a few hundred samples. E.g. in my workflow, I do this as part of dataset preprocessing.

There's also a memory issue: in the Gist, if you look at the `example_RSS_increase.log` file, you'll also see that RSS memory also grows steadily, from 1.192GB to 6.5GB over 100 samples. For a different (real) experiment, this issue eventually led to MemoryError (however the behaviour here did depend on the system: memory kept increasing on one server, but RSS was reclaimed on my laptop, so I'm less sure what's going on here). 

### Cause
In `adapters.base.Adapter`, the `format` function builds a prompt as one big string. So it stringifies large data (dspy types like Image, Audio) and sticks it to the rest of the prompt. Since special data types are sent to the LM as a separate item in the `messages` list, the method searches the string to extract those data - that's the `split_message_content_for_custom_types` function that's taking all the runtime. This is slow because the strings are so big.  

Also, there is high memory demand because each large Image string is copied and concatted together to make the large prompt string. 

### Solution
Temporarily convert large data to a hash before doing the formatting. Then after the `messages` is build, replace the hash with the real data.

Key Changes
- dspy/adapters/utils.py 
      - Introduce a LargePayloadHashManager manager to be used in Adapter.format(). It has one func for replacing large objects inside `inputs` and `demos` with hashes; and one func for restoring the final `messages` with the full data. 
- dspy/adapters/base.py
      - Use the payload manager in `format`
- dspy/adapters/types/image.py
      - ensure that encode_image() accepts hash identifiers and returns them unchanged. This is required to avoid misclassifying hash tokens as invalid inputs.

### results: faster performance
```
Top 20 functions by TOTAL TIME (self time - actual CPU work):
         1478335 function calls (1369172 primitive calls) in 3.349 seconds

   Ordered by: internal time
   List reduced from 642 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100    2.258    0.023    2.258    0.023 {built-in method _hashlib.openssl_sha256}
      100    0.606    0.006    0.606    0.006 {orjson.dumps}
        4    0.120    0.030    0.120    0.030 {built-in method gc.collect}
72400/1800    0.031    0.000    0.062    0.000 /Users/jamesburgess/miniconda3/envs/dspy-dev/lib/python3.11/copy.py:128(deepcopy)
      100    0.020    0.000    2.885    0.029 /Users/jamesburgess/dspy/dspy/clients/cache.py:65(cache_key)
     2103    0.016    0.000    0.017    0.000 {built-in method builtins.next}
```
Runtime reduced from 53.3 seconds to 3.4 seconds, so 94% faster, though this will change as the number and size of images change. 

(BTW,  the major remaining bottleneck is now from `cache.py`, which runs `dumps()` on the whole request and hashes it)

## Extensions 
- If maintainers agree with the PR, I can also extend this to other large dspy Types like Document and Audio. 
- Probably optimizations could be done for cache key optimizations.

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -280,3 +280,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
     else:
         # Neither => enclose in single quotes
         return f"'{s}'"
+
+
+class LargePayloadHashManager:
+    """
+    Used by `format` in adapters.base.py
+    That function formats the input prompt as one string, which means it stringifies
+    large data (dspy types like Image, Audio). Then it splits that prompt string into
+    a list, where special types get their own item. But stringifying large data is slow
+    and memory-intensive. Instead, we replace the large data with a hash token before
+    building the prompt, and restore the original data at the end.
+
+    This class facilitates that with these two methods:
+    - replace_large_data: replace large data with hash tokens for inputs, demos, history
+    - restore_large_data: restore the original data from hash token for LLM messages
+
+    Notes:
+    - Non‑destructive: never mutates inputs; returns new structures mirroring shape.
+    - Threshold-based: strings longer than `LARGE_DATA_THRESHOLD` are hashed
+    - Only Image type is implemented - TODO for audio, etc.
+    """
+
+    # Configurable threshold for what constitutes "large" data
+    LARGE_DATA_THRESHOLD = 1000  # characters
+
+    def __init__(self):
+        """Initialize the manager with fresh hash mappings."""
+        self.hash_to_data = {}  # hash_id -> original_data
+        self.data_to_hash = {}  # original_data -> hash_id
+        self.hash_counter = 0
+
+    def replace_large_data(self, obj: Any) -> Any:
+        """Return a copy of `obj` with large string fields replaced by hash tokens."""
+        if isinstance(obj, Type):
+            # Handle DSPy custom types: Images
+            return self._replace_in_custom_type(obj)
+        elif hasattr(obj, "items") and not isinstance(obj, dict):
+            new_mapping: dict[str, Any] = {}
+            for k, v in obj.items():
+                new_mapping[k] = self.replace_large_data(v)
+            return new_mapping
+        elif isinstance(obj, dict):
+            new_dict = {}
+            for k, v in obj.items():
+                new_dict[k] = self.replace_large_data(v)
+            return new_dict
+        elif isinstance(obj, list):
+            new_list = []
+            for item in obj:
+                new_list.append(self.replace_large_data(item))
+            return new_list
+        else:
+            return obj
+
+    def restore_large_data(self, obj: Any) -> Any:
+        """Return a copy of `obj` with hash tokens restored to original data."""
+        if isinstance(obj, dict):
+            if self._is_message_content_with_large_data(obj):
+                return self._restore_in_message_content(obj)
+            else:
+                return {k: self.restore_large_data(v) for k, v in obj.items()}
+        elif isinstance(obj, list):
+            return [self.restore_large_data(item) for item in obj]
+        elif isinstance(obj, str):
+            if self._is_hash_identifier(obj):
+                return self.hash_to_data.get(obj, obj)
+            else:
+                return obj
+        else:
+            return obj
+
+    def _create_hash_for_data(self, data: str) -> str:
+        """Create a unique hash identifier for large data."""
+        # Check if we've already hashed this exact data (deduplication)
+        if data in self.data_to_hash:
+            return self.data_to_hash[data]
+
+        # Create new hash identifier
+        hash_id = f"__DSPY_LARGE_DATA_HASH_{self.hash_counter}__"
+        self.hash_counter += 1
+
+        # Store bidirectional mapping
+        self.hash_to_data[hash_id] = data
+        self.data_to_hash[data] = hash_id
+
+        return hash_id
+
+    def _is_large_data(self, data: Any) -> bool:
+        """Check if data is large enough to warrant optimization."""
+        if isinstance(data, str):
+            return len(data) > self.LARGE_DATA_THRESHOLD
+        return False
+
+    def _replace_in_custom_type(self, obj: Type) -> Type:
+        """For Image type, replace large payload fields in known custom types; otherwise return `obj`."""
+        if isinstance(obj, History):
+            new_messages = self.replace_large_data(obj.messages)
+            if new_messages is not obj.messages:  # Messages changed
+                return type(obj)(messages=new_messages)
+            else:
+                return obj
+
+        elif hasattr(obj, "url") and self._is_large_data(obj.url):
+            # TODO: it only handles image url, not other types yet
+            hash_url = self._create_hash_for_data(obj.url)
+            return type(obj)(url=hash_url)
+
+        return obj
+
+    def _is_message_content_with_large_data(self, obj: dict) -> bool:
+        """Whether this dict is an image content block with a payload field."""
+        # {"type": "image_url", "image_url": {"url": "hash_id"}}
+        return obj.get("type") == "image_url" and "image_url" in obj and "url" in obj["image_url"]
+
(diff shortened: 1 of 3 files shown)


peekuh:
Unrelated to your PR, but can you direct me to documentation for the dspy Document type? I only see Audio and Image types in the official documentation.

jmhb0:
The type definitions are in https://github.com/stanfordnlp/dspy/tree/main/dspy/adapters/types 

I don't know about official docs, but most of them seem to have good docstrings.

tdb-alcorn:
@okhat or @arnavsinghvi11 - can this PR get on track to be merged? dspy is basically unusable with multimedia prompts, this looks like a reasonable direction to fix

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+        self.hash_to_data = {}  # hash_id -> original_data
+        self.data_to_hash = {}  # original_data -> hash_id
+        self.hash_counter = 0
+
+    def replace_large_data(self, obj: Any) -> Any:
+        """Return a copy of `obj` with large string fields replaced by hash tokens."""
+        if isinstance(obj, DspyType):
+            # Handle DSPy custom types: Images
+            return self._replace_in_custom_type(obj)
+        elif hasattr(obj, "items") and not isinstance(obj, dict):
```
michaelisaac-dev, line 320: What's the difference between this if-case and the one below in line 325?

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+        if isinstance(data, str):
+            return len(data) > self.LARGE_DATA_THRESHOLD
+        return False
+
+    def _replace_in_custom_type(self, obj: DspyType) -> DspyType:
+        """For Image type, replace large payload fields in known custom types; otherwise return `obj`."""
+        if isinstance(obj, History):
+            new_messages = self.replace_large_data(obj.messages)
+            if new_messages is not obj.messages:  # Messages changed
+                return type(obj)(messages=new_messages)
```
michaelisaac-dev, line 382: I think it might be clearer if you change this to `History(messages=new_messages)`

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+
+        return hash_id
+
+    def _is_large_data(self, data: Any) -> bool:
+        """Check if data is large enough to warrant optimization."""
+        if isinstance(data, str):
+            return len(data) > self.LARGE_DATA_THRESHOLD
+        return False
+
+    def _replace_in_custom_type(self, obj: DspyType) -> DspyType:
```
michaelisaac-dev, line 377: I think you might be returning a `pydantic.BaseModel` here

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+        """For Image type, replace large payload fields in known custom types; otherwise return `obj`."""
+        if isinstance(obj, History):
+            new_messages = self.replace_large_data(obj.messages)
+            if new_messages is not obj.messages:  # Messages changed
+                return type(obj)(messages=new_messages)
+            else:
+                return obj
+
+        elif hasattr(obj, "url") and self._is_large_data(obj.url):
```
michaelisaac-dev, line 386: Can we check here for `isinstance(obj, Image)`?

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+            return obj
+
+    def _create_hash_for_data(self, data: str) -> str:
+        """Create a unique hash identifier for large data."""
+        # Check if we've already hashed this exact data (deduplication)
+        if data in self.data_to_hash:
+            return self.data_to_hash[data]
+
+        # Create new hash identifier
+        hash_id = f"__DSPY_LARGE_DATA_HASH_{self.hash_counter}__"
```
michaelisaac-dev, line 362: Would it be better to extract `__DSPY_LARGE_DATA_HASH_` into a constant?

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+    - restore_large_data: restore the original data from hash token for LLM messages
+
+    Notes:
+    - Non‑destructive: never mutates inputs; returns new structures mirroring shape.
+    - Threshold-based: strings longer than `LARGE_DATA_THRESHOLD` are hashed
+    - Only Image type is implemented - TODO for audio, etc.
+    """
+
+    # Configurable threshold for what constitutes "large" data
+    LARGE_DATA_THRESHOLD = 1000  # characters
```
michaelisaac-dev, line 307: Should the threshold be exposed in the constructor?

dspy/adapters/utils.py
```diff
@@ -285,3 +282,130 @@ def _quoted_string_for_literal_type_annotation(s: str) -> str:
...
+        """Return a copy of `obj` with hash tokens restored to original data."""
+        if isinstance(obj, dict):
+            if self._is_message_content_with_large_data(obj):
+                return self._restore_in_message_content(obj)
+            else:
+                return {k: self.restore_large_data(v) for k, v in obj.items()}
+        elif isinstance(obj, list):
+            return [self.restore_large_data(item) for item in obj]
+        elif isinstance(obj, str):
+            if self._is_hash_identifier(obj):
```
michaelisaac-dev, line 348: Is this check necessary?

PR #8822, opened 2025-09-18
[WIP] Allow DSPy to use the native reasoning from models

chenmoneygithub:
This is a followup work of #8764. This problem is harder than it looks like because:

- We need to have ChainOfThought work for both reasoning model and non-reasoning model. And if reasoning model is used, we should avoid having the model generate the `reasoning` field.
- Without explicit `reasoning` field in output fields, `dspy.Predict` should not return `reasoning`. Basically we need to keep the signature faithful.

That basically means we need a way to identify users are specifying "reasoning" in their signature, and it fits into our definition of predefined type. Users are not forbidden from using reasoning of string type, but it won't use the native reasoning.


dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,75 @@
...
+
+        raise ValueError(f"Received invalid value for `dspy.Reasoning`: {data}")
+
+    @classmethod
+    def is_natively_supported(cls, lm, lm_kwargs) -> bool:
+        """Whether the Reasoning type is natively supported by the LM."""
+        if not litellm.supports_reasoning(lm.model):
+            return False
+
+        if "reasoning_effort" in lm_kwargs:
```
TomeHirata, line 54: We don't need to check `lm_kwargs`, instead we should probably overwrite `reasoning_effort` if native reasoning is on

dspy/predict/chain_of_thought.py
```diff
@@ -1,36 +1,30 @@
...
 from dspy.primitives.module import Module
+from dspy.signatures.field import OutputField
 from dspy.signatures.signature import Signature, ensure_signature
 
 
 class ChainOfThought(Module):
     def __init__(
         self,
         signature: str | type[Signature],
-        rationale_field: FieldInfo | None = None,
```
TomeHirata, line 14: should we show deprecation warnings if these arguments are used?

dspy/clients/base_lm.py
```diff
@@ -166,19 +166,23 @@ def update_history(self, entry):
 
     def _process_completion(self, response, merged_kwargs):
         """Process the response of OpenAI chat completion API and extract outputs.
-
```
TomeHirata: Is it auto-formatting? Shall we add lint rule so that linting does not change lines unrelated a PR's scope?

dspy/clients/base_lm.py
```diff
@@ -217,26 +220,41 @@ def _extract_citations_from_response(self, choice):
...
         Returns:
-            List of processed outputs
+            List of processed outputs, which is always of size 1 because the Response API only supports one output.
         """
-        outputs = []
+        text_outputs = []
         tool_calls = []
+        reasoning_contents = []
+
         for output_item in response.output:
```
TomeHirata, line 260: Sorry I'm confused. What's the meaning of multiple items in the output field? I thought it'd be the same as choices of chat completion

dspy/adapters/types/reasoning.py
chenmoneygithub: so `reasoning_effort` has multiple valid values - low, medium, and high. I am setting the default value as `low`, but if users specify the other values we want to keep those unchanged.

dspy/clients/base_lm.py
chenmoneygithub: This is definitely a weird point from Responses API - all these items in `output` belongs to one answer. Basically Responses API doesn't have the `n` arg that controls the number of answer to be generated.

dspy/adapters/types/reasoning.py
arnavsinghvi11: the default `reasoning_effort` on OpenAI reasoning models is`medium`, so I'd be careful setting it differently here

dspy/clients/base_lm.py
TomeHirata: Talked offline. It seems all output items represent one choice.

dspy/adapters/types/base_type.py
```diff
@@ -70,6 +70,11 @@ def serialize_model(self):
             )
         return formatted
 
+    @classmethod
+    def adapt_to_native_lm_feature(cls, lm, lm_kwargs) -> bool:
```
TomeHirata, line 74: Why do we need lm_kwargs here? Isn't the availability determined only by lm?

TomeHirata: Also for the naming, how about `is_native_lm_feature_available`?

dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,78 @@
...
+        raise ValueError(f"Received invalid value for `dspy.Reasoning`: {data}")
+
+    @classmethod
+    def adapt_to_native_lm_feature(cls, lm, lm_kwargs) -> bool:
+        if not litellm.supports_reasoning(lm.model):
+            return False
+
+        if "reasoning_effort" in lm_kwargs:
+            # `lm_kwargs` overrides `lm.kwargs`
+            reasoning_effort = lm_kwargs["reasoning_effort"]
```
TomeHirata, line 56: nit: we can simplify by `reasoning_effort = lm_kwargs.get("reasoning_effort") or lm.kwargs.get("reasoning_effort")`

dspy/predict/chain_of_thought.py
```diff
@@ -1,36 +1,30 @@
...
         signature = ensure_signature(signature)
-        prefix = "Reasoning: Let's think step by step in order to"
-        desc = "${reasoning}"
-        rationale_field_type = rationale_field.annotation if rationale_field else rationale_field_type
-        rationale_field = rationale_field if rationale_field else dspy.OutputField(prefix=prefix, desc=desc)
-        extended_signature = signature.prepend(name="reasoning", field=rationale_field, type_=rationale_field_type)
+
+        from dspy.adapters.types.reasoning import Reasoning
+
+        extended_signature = signature.prepend(name="reasoning", field=OutputField(), type_=Reasoning)
```
TomeHirata, line 33: Should we continue to return string in the reasoning field for backward compatibility?

dspy/clients/lm.py
```diff
@@ -103,12 +103,17 @@ def __init__(
 
         self._warn_zero_temp_rollout(self.kwargs.get("temperature"), self.kwargs.get("rollout_id"))
 
-    def _warn_zero_temp_rollout(self, temperature: float | None, rollout_id):
-        if (
-            not self._warned_zero_temp_rollout
-            and rollout_id is not None
-            and (temperature is None or temperature == 0)
+        # Normalize reasoning_effort to get reasoning summaries (for OpenAI reasoning models which don't expose
```
TomeHirata: Shall we move this logic to _convert_chat_request_to_responses_request?

dspy/clients/base_lm.py
```diff
@@ -217,26 +220,41 @@ def _extract_citations_from_response(self, choice):
...
-
-        if tool_calls:
-            outputs.append({"tool_calls": tool_calls})
-        return outputs
+            elif output_item_type == "reasoning":
+                if getattr(output_item, "content", None) and len(output_item.content) > 0:
+                    for content_item in output_item.content:
+                        reasoning_contents.append(content_item.text)
+                elif getattr(output_item, "summary", None) and len(output_item.summary) > 0:
+                    for summary_item in output_item.summary:
```
TomeHirata, line 272: When is this summary field used?

dspy/adapters/types/reasoning.py
chenmoneygithub: ohh good catch! The structure is right but code is actually buggy - if users explicitly turn off native reasoning, we should respect the setting, so I am not using `get` to capture the value. Updated the code.

dspy/predict/chain_of_thought.py
chenmoneygithub: `dspy.Reasoning` is a str-like type. Basically:

- Printing a dspy.Reasoning prints out the string part, i.e., content.
- Users can directly compare a dspy.Reasoning to a string.
 
We need this specific type to modify the signature inside Adapter. I don't feel this is a perfectly clean solution, but it's the most robust way in my mind. would like to hear your thoughts!

dspy/clients/lm.py
chenmoneygithub: nice catch, done

dspy/clients/base_lm.py
chenmoneygithub: This is another weird spot of Responses API - according to the API documentation, the reasoning content can be contained in `content` or `summary` of the reasoning block: https://platform.openai.com/docs/api-reference/responses/object. While we use the `content` as golden, the API itself suggests that it's possible the reasoning is contained in `summary`.

dspy/clients/base_lm.py
chenmoneygithub: that's my plugin for trailing whitespaces. I didn't expect this PR to be this big at the first place...

dspy/predict/chain_of_thought.py
chenmoneygithub: sg, let me add that

dspy/clients/base_lm.py
TomeHirata: Interesting, so the behavior is not deterministic?

dspy/predict/chain_of_thought.py
TomeHirata: I see, but what's gonna happen when users directly call the string funcitons on the reasoning field like this? Does it continue to work?

```python
cot = dspy.ChainOfThough("question -> answer")
cot(question="where is the capital of Japan?").strip()
```

dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,79 @@
...
+        elif "reasoning_effort" in lm.kwargs:
+            reasoning_effort = lm.kwargs["reasoning_effort"]
+
+        if reasoning_effort is None:
+            # If users explicitly set `reasoning_effort` to None, we don't enable native reasoning
+            return False
+
+        # Turn on the native reasoning
+        lm_kwargs["reasoning_effort"] = "low"
```
TomeHirata: According to @arnavsinghvi11, the OpenAI default is `medium`

dspy/predict/chain_of_thought.py
```diff
@@ -1,36 +1,36 @@
...
 
         Args:
             signature (Type[dspy.Signature]): The signature of the module.
-            rationale_field (Optional[Union[dspy.OutputField, pydantic.fields.FieldInfo]]): The field that will contain the reasoning.
-            rationale_field_type (Type): The type of the rationale field.
             **config: The configuration for the module.
         """
         super().__init__()
         signature = ensure_signature(signature)
-        prefix = "Reasoning: Let's think step by step in order to"
```
TomeHirata, line 29: It seems we will lose this [CoT prompting](https://arxiv.org/abs/2205.11916), is that fine?

chenmoneygithub: it's actually not used today

dspy/adapters/types/reasoning.py
chenmoneygithub: Yes we talked offline a bit about that, basically I won't rely on that default. From Responses API OpenAI is no longer the industry API standard. I am setting it to `low` to avoid too much extra token usage.

dspy/predict/chain_of_thought.py
TomeHirata: Oh, it's not used today because it's the prefix of the output field? Does it mean `ChainOfThought` is a module that just adds a reasoning field to `dspy.Predict` today?

chenmoneygithub: I know this feels odd, but yes exactly - -"

arnavsinghvi11: btw guys lets hold off on merging this until we fully hash out the right design choices for this PR, given that its a decently significant change!

TomeHirata: Agreed, let's discuss offline

dspy/predict/chain_of_thought.py
TomeHirata: @chenmoneygithub for me this is the largest concern, can you take a look?

chenmoneygithub:
@TomeHirata  I want to resume this one, but let me split the PR to have one dedicated on dspy.Reasoning, and a followup one to handle dspy.ChainOfThought. The previous one should be low or no risk, while the latter is more risky.

Btw, litellm still has the issue of litellm.completion doesn't return reasoning summary with gpt-5, while the responses API works fine.

dspy/predict/chain_of_thought.py
```diff
@@ -1,36 +1,36 @@
...
-        desc = "${reasoning}"
-        rationale_field_type = rationale_field.annotation if rationale_field else rationale_field_type
-        rationale_field = rationale_field if rationale_field else dspy.OutputField(prefix=prefix, desc=desc)
-        extended_signature = signature.prepend(name="reasoning", field=rationale_field, type_=rationale_field_type)
+
+        if "rationale_field" in config or "rationale_field_type" in config:
+            logger.warning("`rationale_field` and `rationale_field_type` are deprecated, they are no-op now.")
+
+        from dspy.adapters.types.reasoning import Reasoning
```
TomeHirata, line 31: Can we move this import to the top?

tests/adapters/test_reasoning.py
```diff
@@ -0,0 +1,115 @@
+"""Test that dspy.Reasoning is fully str-like for backward compatibility."""
```
TomeHirata: nit: Let's remove this comment

tests/predict/test_chain_of_thought.py
```diff
@@ -23,3 +26,61 @@ async def test_async_chain_of_thought():
         program = ChainOfThought("question -> answer")
         result = await program.acall(question="What is 1+1?")
         assert result.answer == "2"
+
+
+def test_chain_of_thought_with_native_reasoning():
+    """Test ChainOfThought with native reasoning support where LM returns reasoning natively."""
```
TomeHirata, line 32: could we remove this docstring as it's duplicated with the title?

tests/predict/test_chain_of_thought.py
```diff
@@ -23,3 +26,61 @@ async def test_async_chain_of_thought():
...
+        assert result.reasoning.content == "Step-by-step thinking about the capital of France"
+
+        # Check that the reasoning_effort is automatically set to "low" when the LM supports native reasoning and not
+        # provided in the LM kwargs
+        args, kwargs = mock_completion.call_args
+        assert kwargs["reasoning_effort"] == "low"
+
+
+def test_chain_of_thought_with_manual_reasoning():
+    """Test ChainOfThought with manual reasoning where LM doesn't support native reasoning."""
```
TomeHirata, line 63: ditto

dspy/predict/chain_of_thought.py
chenmoneygithub: Yes I checked it earlier, and it was causing a circular dependency issue if imported at the file top.

PR #8825, opened 2025-09-18
Add ToolCall.execute for smoother tool execution

TomeHirata:
This PR adds a helper function to ToolCall so that users can run tool calls with less boilerplate.

```python
import dspy

class ToolSignature(dspy.Signature):
    """Signature for manual tool handling."""
    outputs: dspy.ToolCalls = dspy.OutputField()

def weather(city: str) -> str:
    """Get weather information for a city."""
    return f"The weather in {city} is sunny"

def calculator(expression: str) -> str:
    """Evaluate a mathematical expression."""
    try:
        result = eval(expression)  # Note: Use safely in production
        return f"The result is {result}"
    except:
        return "Invalid expression"

# Create predictor
predictor = dspy.Predict(ToolSignature)

# Make a prediction
response = predictor(
    question="What's the weather in New York?",
    tools=[dspy.Tool(weather), dspy.Tool(calculator)]
)


# Before
tools = {
    "weather": dspy.Tool(weather),
    "calculator": dspy.Tool(calculator)
}
for call in response.outputs.tool_calls:
    # Execute the tool call
    result = tools[call.name](**call.args)

# After
for call in response.outputs.tool_calls:
    result = call.execute()
```

### DIFF
--- a/docs/docs/learn/programming/tools.md
+++ b/docs/docs/learn/programming/tools.md
@@ -144,14 +144,14 @@ for call in response.outputs.tool_calls:
     
     # Execute individual tool calls with different options:
     
-    # Option 1: Pass tools as a dict (most explicit)
+    # Option 1: Automatic discovery (finds functions in locals/globals)
+    result = call.execute()  # Automatically finds functions by name
+
+    # Option 2: Pass tools as a dict (most explicit)
     result = call.execute(functions={"weather": weather, "calculator": calculator})
     
-    # Option 2: Pass Tool objects as a list
+    # Option 3: Pass Tool objects as a list
     result = call.execute(functions=[dspy.Tool(weather), dspy.Tool(calculator)])
-    
-    # Option 3: Automatic discovery (finds functions in locals/globals)
-    result = call.execute()  # Automatically finds functions by name
     
     print(f"Result: {result}")
 ```
--- a/docs/docs/learn/programming/tools.md
+++ b/docs/docs/learn/programming/tools.md
@@ -106,11 +106,11 @@ response = predictor(
 
 # Execute the tool calls
 for call in response.outputs.tool_calls:
-    if call.name in tools:
-        result = tools[call.name](**call.args)
-        print(f"Tool: {call.name}")
-        print(f"Args: {call.args}")
-        print(f"Result: {result}")
+    # Execute the tool call
+    result = call.execute()
+    print(f"Tool: {call.name}")
+    print(f"Args: {call.args}")
+    print(f"Result: {result}")
 ```
 
 ### Understanding `dspy.Tool`
@@ -134,7 +134,7 @@ print(str(tool))        # Full tool description
 
 ### Understanding `dspy.ToolCalls`
 
-The `dspy.ToolCalls` type represents the output from a model that can make tool calls:
+The `dspy.ToolCalls` type represents the output from a model that can make tool calls. Each individual tool call can be executed using the `execute` method:
 
 ```python
 # After getting a response with tool calls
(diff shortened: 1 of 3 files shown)


docs/docs/learn/programming/tools.md
```diff
@@ -134,18 +134,26 @@ print(str(tool))        # Full tool description
...
     print(f"Tool name: {call.name}")
     print(f"Arguments: {call.args}")
     
-    # Execute the tool
-    if call.name in tools:
-        result = tools[call.name](**call.args)
-        print(f"Result: {result}")
+    # Execute individual tool calls with different options:
+    
+    # Option 1: Pass tools as a dict (most explicit)
```
chenmoneygithub: I like Option 3!  Option 1 and 2 will be much less popular IMO though. My understanding is they handle the cases:

1. users forget to include the tool in runtime, in which case `weather` will be None as well.  
2. Users want to search functions in a tailored namespace, which is cool but I have never seen any use cases.

Can we remove the `functions` arg from `execute`?

TomeHirata: option 3 cannot cover some usecases where the tool name is different from the local variable name too. I like 3 too and this is the main intention, but can we keep the function arg for flexibility? Since it's optional, it doesn't increase friction and I can change the order in this tutorial and explain option 3 first. what do you think?

chenmoneygithub: sorry for the late reply.

I am curious about "where the tool name is different from the local variable name", why is this a valid case?

TomeHirata: Isn't this totally possible? In this case LM recognizes this tool as "weather" while it's registered as "my_tool" in globals.
```python
my_tool = dspy.Tool(name="weather", func: lambda city: f"the weather in {city} is rainy")
```

chenmoneygithub: This is possible, but doesn't seem to be a valid use case IMO. When users want to set `name="weather"`, I don't see why they want to change the tool instance name, and pass it in `functions` later on.

Since this is not breaking and doesn't add too much complexity, I don't want to block on this, and will let you make the call

dspy/adapters/types/tool.py
```diff
@@ -269,6 +269,54 @@ def format(self):
...
+            elif isinstance(functions, dict):
+                func = functions.get(self.name)
+            elif isinstance(functions, list):
+                for tool in functions:
+                    if tool.name == self.name:
+                        func = tool.func
+                        break
+
+            if func is None:
+                raise ValueError(f"Tool function '{self.name}' not found. Pass the tool functions to the `execute` method.")
```
chenmoneygithub: nit: Please pass.

dspy/adapters/types/tool.py
```diff
@@ -269,6 +269,54 @@ def format(self):
...
+
+            try:
+                if self.args:
```
chenmoneygithub: Nonblocking, just I prefer the code below:

```
args = self.args or {}
result = func(**args)
```

tests/adapters/test_tool.py
```diff
@@ -540,3 +540,72 @@ def test_tool_convert_input_schema_to_tool_args_lang_chain():
...
+        result1 = tool_call1.execute()  # Should find local function automatically
+        assert result1 == 25
+
+        tool_call2 = dspy.ToolCalls.ToolCall(name="local_multiply", args={"x": 4, "y": 7})
+        result2 = tool_call2.execute()  # Should find local function automatically
+        assert result2 == 28
+
+        # Test locals take precedence over globals
+        try:
+            globals()["local_add"] = lambda a, b: a + b + 1000
```
chenmoneygithub, line 604: can we define it as a function before `def main()`?

chenmoneygithub (approved):
LGTM with a few nonblocking comments

tests/adapters/test_tool.py
TomeHirata: I think it's still registered in locals in that case since globals are module-level variables.

PR #8831, opened 2025-09-19
Fix the crash when usage tracker is enabled with non-prediction output

TomeHirata:
Currently, dspy modules whose output is not `dspy.Prediction` fail to execute when the usage tracker is enabled. This PR fixes the crash by skipping the usage tracking for this type of modules. While returning objects other than dspy.Prediciton is not expected, this pattern has been found in several places.

Example:
```python
class MyProgram(dspy.Module):
        def __init__(self):
            self.predict = dspy.Predict("question -> answer")

        def forward(self, question: str, **kwargs) -> str:
            return self.predict(question=question).answer

program = MyProgram()
with dspy.context(track_usage=True):
    program(question="What is the capital of France?")
```

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -1,5 +1,6 @@
 import inspect
 import logging
+from typing import Any
 
 import magicattr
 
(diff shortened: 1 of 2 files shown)


TomeHirata:
We shouldn't break program execution due to the failure of observability feature. I've added warning so that users can know why usage is not available.

dspy/primitives/module.py
```diff
@@ -101,7 +101,16 @@ async def acall(self, *args, **kwargs) -> Prediction:
             if settings.track_usage and thread_local_overrides.get().get("usage_tracker") is None:
                 with track_usage() as usage_tracker:
                     output = await self.aforward(*args, **kwargs)
-                    output.set_lm_usage(usage_tracker.get_total_tokens())
+                    tokens = usage_tracker.get_total_tokens()
```
chenmoneygithub, line 93: Maybe move these usage tracking code into a helper function for reuse?

PR #8833, opened 2025-09-20
Allow stream listener to work on any type

chenmoneygithub:
Previously we limit stream listeners to only work with str field and some prebuilt types. This PR lifts up the constraint, and allow streaming on any types. Here is the gist:

- StreamLIstener works the same way as before, i.e., only capture streaming chunks associated with a certain field.
- We don't perform additional handling on the structured output value for each field, but just give back the value. 

To accomodate JSONAdapter, which doesn't have clear boilerplate like `[[ ## answer ## ]]` that splits the fields' streaming in ChatAdapter, we make specific logic chunk based on [jiter](https://github.com/pydantic/jiter) to handle it.


dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +210,87 @@ def receive(self, chunk: ModelResponseStream):
...
-                )
+            if isinstance(settings.adapter, JSONAdapter):
+                # JSONAdapter uses partial json parsing to detect the end of the field we are listening to, instead of
+                # relying on the end_identifier.
+                return self._json_adapter_handle_stream_chunk(token, chunk_message)
+            else:
+                # Other adapters rely on the end_identifier to detect the end of the field we are listening to.
+                return self._default_handle_stream_chunk(token, end_identifier)
+
+    def _json_adapter_handle_stream_chunk(self, token: str, chunk_message: str) -> str:
```
TomeHirata: The return signature should be StreamResponse | None

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +210,87 @@ def receive(self, chunk: ModelResponseStream):
...
+
+        if token:
+            return StreamResponse(
+                self.predict_name,
+                self.signature_field_name,
+                token,
+                is_last_chunk=self.stream_end,
+            )
+
+    def _default_handle_stream_chunk(self, token: str, end_identifier: str) -> str:
```
TomeHirata: ditto

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +210,87 @@ def receive(self, chunk: ModelResponseStream):
...
-                token = token.rstrip()  # Remove the trailing \n\n
-            elif not self._could_form_end_identifier(concat_message, adapter_name):
+
+            if not self._could_form_end_identifier(concat_message, adapter_name):
                 # Buffer cannot form end identifier, safe to flush out the tokens in the buffer.
                 token = self.flush()
             elif self.field_end_queue.qsize() > 10:
-                # Buffer could form end identifier, but we've exceeded max buffer size
-                # Yield the oldest token to prevent unbounded buffering
+                # We keep the last 10 tokens in the buffer to avoid sending the DSPy bolilerplate tokens to users.
```
TomeHirata: Let's note that we buffer only if tokens in the queue can form the boilerplate.

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +210,87 @@ def receive(self, chunk: ModelResponseStream):
...
+            # We add an extra "{" to the beginning of the field_accumulated_tokens, so if we get a balance of curly
+            # brackets, that means the streaming for the entire dspy.Predict is finished.
+            self.stream_end = True
+            last_token = self.flush()
+            right_curly_bracket_index = last_token.rfind("}")
+            token = token + last_token[:right_curly_bracket_index] if token else last_token[:right_curly_bracket_index]
+            return StreamResponse(self.predict_name, self.signature_field_name, token, is_last_chunk=self.stream_end)
+
+        try:
+            parsed = jiter.from_json(
```
TomeHirata, line 254: This is interesting, can't we just count the number of { and }?

dspy/streaming/streaming_listener.py
chenmoneygithub: thanks for catching it!

dspy/streaming/streaming_listener.py
chenmoneygithub: ty!

dspy/streaming/streaming_listener.py
chenmoneygithub: sg!

dspy/streaming/streaming_listener.py
chenmoneygithub: discussed offline, please see the new implementation for a more robust solution.

dspy/streaming/streaming_listener.py
```diff
@@ -49,6 +50,8 @@ def __init__(
         self.cache_hit = False
         self.allow_reuse = allow_reuse
 
+        self.json_adapter_state = {"field_accumulated_tokens": ""}
```
TomeHirata: Do we plan to introduce other keys to `self.json_adapter_state`? or can we flatten the structure?

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+                # relying on the end_identifier.
+                return self._json_adapter_handle_stream_chunk(token, chunk_message)
+            else:
+                # Other adapters rely on the end_identifier to detect the end of the field we are listening to.
+                return self._default_handle_stream_chunk(token, end_identifier)
+
+    def _json_adapter_handle_stream_chunk(self, token: str, chunk_message: str) -> StreamResponse | None:
+        self.json_adapter_state["field_accumulated_tokens"] += chunk_message
+        if self.json_adapter_state["field_accumulated_tokens"].rstrip().endswith("}"):
+            # When the accumulated tokens ends with a curly b   racket, that means the streaming for the predict we are
```
TomeHirata:
```suggestion
            # When the accumulated tokens end with a curly bracket, that means the streaming for the prediction we are
```

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+            if isinstance(settings.adapter, JSONAdapter):
+                # JSONAdapter uses partial json parsing to detect the end of the field we are listening to, instead of
+                # relying on the end_identifier.
+                return self._json_adapter_handle_stream_chunk(token, chunk_message)
+            else:
+                # Other adapters rely on the end_identifier to detect the end of the field we are listening to.
+                return self._default_handle_stream_chunk(token, end_identifier)
+
+    def _json_adapter_handle_stream_chunk(self, token: str, chunk_message: str) -> StreamResponse | None:
+        self.json_adapter_state["field_accumulated_tokens"] += chunk_message
```
TomeHirata: Why do we accumulate `chunk_message` instead of `token`?

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+                    token + last_token[:right_curly_bracket_index] if token else last_token[:right_curly_bracket_index]
+                )
                 return StreamResponse(
-                    self.predict_name,
-                    self.signature_field_name,
-                    token,
-                    is_last_chunk=self.stream_end,
+                    self.predict_name, self.signature_field_name, token, is_last_chunk=self.stream_end
                 )
+            except Exception:
```
TomeHirata: can't we limit this to be `ValueError`?

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+                self.json_adapter_state["field_accumulated_tokens"].encode("utf-8"),
+                partial_mode="trailing-strings",
+            )
+            if len(parsed) > 1:
+                # If partial json parsing finds a second key, that means the streaming for the field we are listening to
+                # is finished.
+                self.stream_end = True
+                last_token = self.flush()
+
+                keys = list(parsed.keys())
```
TomeHirata, line 264: Is `parsed.keys` ordered based on the key order in the json string?

dspy/streaming/streaming_listener.py
```diff
@@ -337,13 +400,3 @@ def find_predictor_for_stream_listeners(
...
     return predict_id_to_listener
-
-
-def _is_streamable(field_type: type | None) -> bool:
-    if field_type is None:
-        return False
-    if field_type is str:
-        return True
-    if issubclass(field_type, Type):
-        return field_type.is_streamable()
```
TomeHirata, line 348: Shall we delete `is_streamable` method of `Type`?

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+                right_curly_bracket_index = last_token.rfind("}")
+                token = (
+                    token + last_token[:right_curly_bracket_index] if token else last_token[:right_curly_bracket_index]
+                )
                 return StreamResponse(
-                    self.predict_name,
-                    self.signature_field_name,
-                    token,
-                    is_last_chunk=self.stream_end,
+                    self.predict_name, self.signature_field_name, token, is_last_chunk=self.stream_end
```
TomeHirata, line 248: So overall we will return a raw string chunk so the deserialization needs to happen on the caller side?

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+
+        if token:
+            return StreamResponse(
+                self.predict_name,
+                self.signature_field_name,
+                token,
+                is_last_chunk=self.stream_end,
+            )
+
+    def _default_handle_stream_chunk(self, token: str, end_identifier: str) -> StreamResponse | None:
```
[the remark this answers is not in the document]
chenmoneygithub: bro

tests/streaming/test_streaming.py
```diff
@@ -1121,7 +1125,349 @@ async def citation_stream(*args, **kwargs):
...
+@pytest.mark.anyio
+async def test_chat_adapter_simple_pydantic_streaming():
+    """Test ChatAdapter streaming with a simple pydantic model."""
+
+    class TestSignature(dspy.Signature):
+        question: str = dspy.InputField()
+        response: SimpleResponse = dspy.OutputField()
+
+    class MyProgram(dspy.Module):
+        def __init__(self):
```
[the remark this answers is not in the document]
chenmoneygithub: no need

dspy/streaming/streaming_listener.py
```diff
@@ -196,30 +203,98 @@ def receive(self, chunk: ModelResponseStream):
...
+                keys = list(parsed.keys())
+                next_field_name = None
+                for key in keys:
+                    if key != self.signature_field_name:
+                        next_field_name = key
+                        break
+
+                last_token_index = last_token.find(next_field_name)
+                token = token + last_token[:last_token_index] if token else last_token[:last_token_index]
+        except ValueError:
```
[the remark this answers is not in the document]
chenmoneygithub: intended

dspy/streaming/streaming_listener.py
chenmoneygithub: `field_accumulated_tokens` is probably a bad name, renaming to `field_accumulated_messages`

chenmoneygithub: Basically token is the one we return, `chunk_message ` is the new message we are receiving.

dspy/streaming/streaming_listener.py
chenmoneygithub: yes, good call!

dspy/streaming/streaming_listener.py
chenmoneygithub: I think so. It shouldn't affect the logic here though, we just need the key name to cut off the extra characters.

dspy/streaming/streaming_listener.py
chenmoneygithub: yes good question, I did think about it, and my take is that's still useful to allow streaming listener to hit this part on certain-type fields like Citation. For custom type that's not streamable, it will use the normal streaming handling.

```
        # Handle custom streamable types
        if self._output_type and issubclass(self._output_type, Type) and self._output_type.is_streamable():
            if parsed_chunk := self._output_type.parse_stream_chunk(chunk):
                return StreamResponse(
                    self.predict_name,
                    self.signature_field_name,
                    parsed_chunk,
                )
```

dspy/streaming/streaming_listener.py
chenmoneygithub: Yes! It should be pretty simple for the caller to accumulate.

dspy/streaming/streaming_listener.py
```diff
@@ -49,6 +50,8 @@ def __init__(
         self.cache_hit = False
         self.allow_reuse = allow_reuse
 
+        self.json_adapter_state = {"field_accumulated_messages": ""}
```
TomeHirata, line 53: Is there any way to isolate json adapter field/method from StreamListener? I feel it's a bit too coupled if we have `json_adapter_state` or `_json_adapter_handle_stream_chunk` in `StreamListener` which should be adapter agnostic as much as possible.

dspy/streaming/streaming_listener.py
TomeHirata: I see, like using json.loads?

dspy/streaming/streaming_listener.py
TomeHirata: Btw, there is a similar comment from the community: https://github.com/stanfordnlp/dspy/issues/8885

chenmoneygithub: Yes I get your point, and thought on the same track earlier. 

Sharing my current thought - this json partial parsing is connected to handling JSON + streaming, not just JSON parsing from normal LM response, so StreamListener is the right place to hold the code. 

A cleaner approach, which I think is the right way, is to have adapter-specific listeners like `JSONStreamListner` and `ChatStreamLisnter` to handle specific adapter stream buffering and handling. However, this would require a significant amount of time. With our current bandwidth I would prefer outsourcing this work to the open source community.

TomeHirata: >adapter-specific listeners like JSONStreamListner and ChatStreamLisnter to handle specific adapter stream buffering and handling

I agree this is the right track. Can we have TODO comments for places we want to refactor later?

chenmoneygithub: for sure, done!

PR #8836, opened 2025-09-20
Fix SFTConfig max_seq_length compatibility with newer TRL versions

### LINKED ISSUE #8762: [Bug] TypeError: SFTConfig.__init__() got an unexpected keyword argument 'max_seq_length'
### What happened?

I am trying to use the BootstrapFinetune optimizer using a local Gemma 3 via SGLang, by following the [tutorial](https://dspy.ai/tutorials/classification_finetuning/) on the DSPy website. Since I want to use Gemma 3, I need a recent version of transformers, so I do not pinpoint any version.

Looking at trl source code, it seems that the keyword `max_seq_length` in the SFTConfig class has been abandonned since the 0.16.0 version (last version is 0.22.2, which is the one installed in my case).

Is there any tutorial somewhere that has been recently tested ?

### Steps to reproduce

My current environment was built as presented below.
```shell
$ uv venv --python=3.11
$ uv pip install dspy
$ uv pip install "sglang[all]"
$ uv pip install transformers accelerate trl peft
```

### DSPy version

3.0.3

fsndzomga:
Fixes compatibility issue with TRL >= 0.16.0 where SFTConfig parameter changed from max_seq_length to max_length.

Added dynamic parameter detection to support both old and new TRL versions.

Fixes #8762

### DIFF
--- a/dspy/clients/lm_local.py
+++ b/dspy/clients/lm_local.py
@@ -253,25 +253,43 @@ def train_sft_locally(model_name, train_data, train_kwargs):
             task_type="CAUSAL_LM",
         )
 
-    sft_config = SFTConfig(
-        output_dir=train_kwargs["output_dir"],
-        num_train_epochs=train_kwargs["num_train_epochs"],
-        per_device_train_batch_size=train_kwargs["per_device_train_batch_size"],
-        gradient_accumulation_steps=train_kwargs["gradient_accumulation_steps"],
-        learning_rate=train_kwargs["learning_rate"],
-        max_grad_norm=2.0,  # note that the current SFTConfig default is 1.0
-        logging_steps=20,
-        warmup_ratio=0.03,
-        lr_scheduler_type="constant",
-        save_steps=10_000,
-        bf16=train_kwargs["bf16"],
-        max_seq_length=train_kwargs["max_seq_length"],
-        packing=train_kwargs["packing"],
-        dataset_kwargs={  # We need to pass dataset_kwargs because we are processing the dataset ourselves
+    # Handle compatibility between different TRL versions
+    # TRL >= 0.16.0 uses 'max_length' instead of 'max_seq_length' in SFTConfig
+    import inspect
+    sft_config_params = inspect.signature(SFTConfig.__init__).parameters
+    
+    # Build config parameters, handling the max_seq_length -> max_length change
+    config_kwargs = {
+        "output_dir": train_kwargs["output_dir"],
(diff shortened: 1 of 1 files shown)


dspy/clients/lm_local.py
```diff
@@ -253,25 +253,43 @@ def tokenize_function(example):
...
             "add_special_tokens": False,  # Special tokens handled by template
             "append_concat_token": False,  # No additional separator needed
         },
-    )
+    }
+
+    # Add the sequence length parameter using the appropriate name for the TRL version
+    if "max_seq_length" in sft_config_params:
+        # Older TRL versions (< 0.16.0)
+        config_kwargs["max_seq_length"] = train_kwargs["max_seq_length"]
```
chenmoneygithub, line 284: It's possible that the user is setting `max_length` instead of `max_seq_length`, then `max_length` will go nowhere. 

Instead of this silent transformation, can we raise an error on detecting mismatch between `train_kwargs` and `SFTConfig` on the name of max length?  The error can be something like "`max_seq_length` is replaced by `max_length` in trl>=0.16.0, but you set ... on trl={the_detected_version}"

chenmoneygithub (commented):
Thanks for the PR!

dspy/clients/lm_local.py
fsndzomga: Good point, thank you for the feedback. After thinking about this, I’ve come to the conclusion that it’s probably better to just let the code fail as it currently does, and let the user either install the correct version of trl or update the current implementation in DSPy to work only with the latest trl API and fail otherwise. In both cases, we should specify which version of trl is recommended and perhaps include a check for it. Otherwise it becomes a maintenance mess if trl API changes again in the future ?

goudtho: @fsndzomga - agree, we should probably update DSPy's SFTConfig to force `>=0.16.0` - as it stands, folks are forced to use `<=0.15.2` which surely isn't the right side of the fence.

fsndzomga: @goudtho Absolutely. If that’s something the core maintainers are open to, I’d be happy to do that.

dilarasoylu: @goudtho @fsndzomga thanks for this, forcing a newer version sounds good to me as well! @chenmoneygithub does this also sound reasonable to you? I think we can merge this PR afterwards.

chenmoneygithub: Yes, let's proceed with that, thank you!

PR #8837, opened 2025-09-21
[WIP] Arbor Interface Update

Ziems:
Hello! I've updated the interface for arbor to make it much more streamlined. No more needing to open a separate terminal!

@chenmoneygithub is there a better place to put this MultiGPUConfig?


okhat:
LGTM

PR #8860, opened 2025-09-26
Fix error handling in dspy parallel

### LINKED ISSUE #8857: batch(...) returns None for failures but leaves failed_examples/exceptions empty in 3.0.3 (breaks SubagentTool.parallel_run)
### What happened?

In 3.0.3, Module.batch(..., return_failed_examples=True, provide_traceback=True) inserts None into results for failures but leaves failed_examples and exceptions empty. This causes SubagentTool.parallel_run to crash when it calls res.model_dump() on None and prevents collecting failure metadata.

### Steps to reproduce

1. Define a module that raises on a specific input.
2. Call batch with return_failed_examples=True and provide_traceback=True.
3. Observe results, failed_examples, exceptions.

Code:
```python
import dspy

class FailingModule(dspy.Module):
    def forward(self, value: int) -> str:
        if value == 42:
            raise ValueError("boom")
        return f"ok-{value}"

module = FailingModule()
examples = [Example(value=v).with_inputs("value") for v in (1, 42, 3)]

results, failed_examples, exceptions = module.batch(
    examples,
    return_failed_examples=True,
    provide_traceback=True,
)

print("results:", results)
print("failed_examples:", failed_examples)
print("exceptions:", exceptions)
```

### DSPy version

3.0.3

TomeHirata:
Fixes #8857

### DIFF
--- a/dspy/predict/parallel.py
+++ b/dspy/predict/parallel.py
@@ -64,7 +64,15 @@ class Parallel:
         # Execute the processing function over the execution pairs
         results = executor.execute(process_pair, exec_pairs)
 
+        # Populate failed examples and exceptions from the executor
         if self.return_failed_examples:
+            for failed_idx in executor.failed_indices:
+                if failed_idx < len(exec_pairs):
+                    _, original_example = exec_pairs[failed_idx]
+                    self.failed_examples.append(original_example)
+                    if exception := executor.exceptions_map.get(failed_idx):
+                        self.exceptions.append(exception)
+
             return results, self.failed_examples, self.exceptions
         else:
             return results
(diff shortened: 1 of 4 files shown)


isaacbmiller:
Current implementation looks good but misses the error reported in the initial issue of model_dump being called on a `None` result.

I am not yet familiar with how `SubagentTool` works

TomeHirata:
I think we should keep `None` inside `results` to keep the length of `results` the same as the `exec_pairs`. The caller can know that a failure has occurred for indices with result=None.

isaacbmiller (approved):
LGTM Assuming we don't need to care about `SubagentTool.parallel_run`

PR #8861, opened 2025-09-26
Remove _compiled field from DSPy modules

TomeHirata:
Remove the `_compiled` field from DSPy modules, which has been used to specify whether a module has been optimized.


okhat:
I think we should actually do the opposite and probably remove that flag from everywhere..

TomeHirata:
@okhat that works too. I'm fine as long as the behavior is consistent, let's remove them entirely then. I filed this PR since we got a memory leak report when running `MIPROV2.compile` several times over the same program - which we need to fix separately.

PR #8862, opened 2025-09-26
bug(lm): Avoid unnecessary cache key computation

### LINKED ISSUE #8856: [Bug] disabled cache still uses ram
### What happened?

Even if I had
`dspy.configure_cache(enable_disk_cache=False, enable_memory_cache=False)`
i.e., disable all cache, cache keys are still always calculated. That also probably leaks memory, in addition to just wasting resources.

I've attached scalene report

[performance_profile.html](https://github.com/user-attachments/files/22547617/performance_profile.html)

### Steps to reproduce

Disable cache
`dspy.configure_cache(enable_disk_cache=False, enable_memory_cache=False)`
Process things. Probably least tokens needed to notice the cache memory usage would be with `dspy.Image.from_url(my_url, download=True)`.
Watch your ram usage explode

### DSPy version

3.0.3

isaacbmiller:
When both memory and disk cache are disabled, we want to avoid creating a cache key that will not be used.

Creating the cache key is expensive for prompts with images in them.

Fixes #8856

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -96,10 +96,8 @@ class Cache:
         return sha256(orjson.dumps(params, option=orjson.OPT_SORT_KEYS)).hexdigest()
 
     def get(self, request: dict[str, Any], ignored_args_for_cache_key: list[str] | None = None) -> Any:
-        use_memory_cache = self.enable_memory_cache
-        use_disk_cache = self.enable_disk_cache
 
-        if not use_memory_cache and not use_disk_cache:
+        if not self.enable_memory_cache and not self.enable_disk_cache:
             return None
 
         try:
@@ -108,13 +106,13 @@ class Cache:
             logger.debug(f"Failed to generate cache key for request: {request}")
             return None
 
-        if use_memory_cache and key in self.memory_cache:
+        if self.enable_memory_cache and key in self.memory_cache:
             with self._lock:
                 response = self.memory_cache[key]
-        elif use_disk_cache and key in self.disk_cache:
+        elif self.enable_disk_cache and key in self.disk_cache:
             # Found on disk but not in memory cache, add to memory cache
             response = self.disk_cache[key]
-            if use_memory_cache:
+            if self.enable_memory_cache:
                 with self._lock:
                     self.memory_cache[key] = response
         else:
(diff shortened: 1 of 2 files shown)


dspy/clients/cache.py
```diff
@@ -96,19 +96,25 @@ def transform_value(value):
         return sha256(orjson.dumps(params, option=orjson.OPT_SORT_KEYS)).hexdigest()
 
     def get(self, request: dict[str, Any], ignored_args_for_cache_key: list[str] | None = None) -> Any:
+        use_memory_cache = self.enable_memory_cache
```
chenmoneygithub: we can remove this renaming

dspy/clients/cache.py
```diff
@@ -128,17 +134,24 @@ def put(
         ignored_args_for_cache_key: list[str] | None = None,
         enable_memory_cache: bool = True,
     ) -> None:
+        write_to_memory_cache = self.enable_memory_cache and enable_memory_cache
```
chenmoneygithub: nit: keep `enable_memory_cache` for consistency.

tests/clients/test_lm.py
```diff
@@ -96,6 +97,35 @@ def test_dspy_cache(litellm_test_server, tmp_path):
...
+                return ModelResponse(
+                    choices=[Choices(message=Message(role="assistant", content="Hi!"))],
+                    usage={"prompt_tokens": 1, "completion_tokens": 1, "total_tokens": 2},
+                    model="dummy",
+                )
+
+            monkeypatch.setattr(litellm, "completion", fake_completion)
+
+            dummy_lm = DummyLM([{"answer": "ignored"}])
+            dummy_lm.forward(messages=[{"role": "user", "content": "Hello"}])
```
chenmoneygithub: let's use dummy_lm() instead of dummy_lm.forward(), because users may copy code from our unit tests.

chenmoneygithub (commented):
Thanks for fixing it!

tests/clients/test_lm.py
isaacbmiller: The reason why I used `DummyLM.forward` was because `DummyLM.__call__` is overriden whereas `LM.__call__` goes to `BaseLM.__call__`

The right answer is to probably rewrite `DummyLM.__call__` to use the forward method inside of it.

WDYT of creating another issue to track this and then rewriting it. If we don't use .forward, then no caching logic gets hit by DummyLM currently IIUC.

isaacbmiller: #8864

chenmoneygithub: Yes a followup sg, I dropped the stamp to unblock this PR

PR #8863, opened 2025-09-26
fix(JSONAdapter): Braces inside output not properly escaped by parser

### LINKED ISSUE #8759: [Bug] JSONAdapter's parse method regex pattern bug
## What happened?

Use dspy.ChainOfThought to execute task signature, as shown in the following example

```python

class Issue(BaseModel):
    issue_type: str
    severity_level: str
    problem_code_snippet: str


class FindIssue(dspy.Signature):
    code = dspy.InputField(description="check code")
    issue_list: list[Issue] = dspy.OutputField(description="issue list")


class FindIssueModule(dspy.Module):
    def __init__(self):
        super().__init__()
        self.find_issue = dspy.ChainOfThought(FindIssue)

    def forward(self, code: str):
        return self.find_issue(code=code)

fim = FindIssueModule()
fim(code="...code...")
```

### llm output

```text
'{\n  "reasoning": "think step by step",\n  "issue_list": [\n    {\n  "issue_type": "some type",\n      "severity_level": "fatal",\n  "problem_code_snippet": "if (user) {"\n    }\n  ]\n}'
```

### error message

 Adapter JSONAdapter failed to parse the LM response.
```text
{
  "issue_type": "some type",
  "severity_level": "fatal",
  "problem_code_snippet": "if (user) {"
    }
  ]
}
```

Expected to find output fields in the LM response: [reasoning, issue_list]

## Problem diagnosis

I checked the code and output and found that the problem lies in the greedy matching behavior of the regular expression and the handling of nested curly braces. When the text contains curly braces within backticks (`), these curly braces are incorrectly treated as part of the JSON structure, interfering with the regular expression matching.

The parsing error is in https://github.com/stanfordnlp/dspy/blob/main/dspy/adapters/json_adapter.py#L150.

### Steps to reproduce

The parsing error is in https://github.com/stanfordnlp/dspy/blob/main/dspy/adapters/json_adapter.py#L150.

### test code

```python
text = '{\n  "reasoning": "think step by step",\n  "issue_list": [\n    {\n  "issue_type": "some type",\n      "severity_level": "fatal",\n  "problem_code_snippet": "if (user) {"\n    }\n  ]\n}'


pattern = r"\{(?:[^{}]|(?R))*\}"
match = regex.search(pattern, text, regex.DOTALL)
if match:
    text = match.group(0)

print(text)
```

output:
```text
{
  "issue_type": "some type",
      "severity_level": "fatal",
  "problem_code_snippet": "if (user) {"
    }
  ]
}
```

### DSPy version

3.0.3

isaacbmiller:
Adds a json extraction function, which will allow us to properly parse brackets that jsonAdapter outputs:

Before, we would use a regex string that naively matches curly braces, but now we have a smarter extractor that takes the first full object.

We always pass this object to json_repair after.

```json
{
  "reasoning": "think step by step",
  "issue_list": [
    {
      "issue_type": "some type",
      "severity_level": "fatal",
      "problem_code_snippet": "if (user) {"
    }
  ]
}
```

We recently added behavior to retry json_repair (#9182), but this used the regex module. We can replace it with a simple bracket based search (unescaped) to mimic this behavior.

This will also allow us to remove the regex dependency in a follow on PR.

Fixes #8759

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -10,6 +10,7 @@ from pydantic.fields import FieldInfo
 from dspy.adapters.chat_adapter import ChatAdapter, FieldInfoWithName
 from dspy.adapters.types.tool import ToolCalls
 from dspy.adapters.utils import (
+    _extract_first_json_object,
     format_field_value,
     get_annotation_name,
     parse_value,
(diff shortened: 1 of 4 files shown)


isaacbmiller:
@Copilot

anubhavshrimal:
+1 to this PR.

This is one of the issues I have faced recently, what does it take to move it to main branch?

Thanks.

hufeng:
Yes, the wait has been too long.

isaacbmiller:
superseded by #9182

PR #8874, opened 2025-09-30
add py.typed

### LINKED ISSUE #8873: [Bug] package is missing py.typed
### What happened?

the api has type annotations, but there is no `py.typed` file so consumers cannot check types when calling the package

### Steps to reproduce

1. Add dspy as a dep to a project
2. copy and paste one of the examples into the project
3. configure the project to use mypy
4. run mypy
~~~
 error: Cannot find implementation or library stub for module named "dspy"  [import-not-found]
~~~

### DSPy version

3.0.3

phelps-sg:
Resolves #8873

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ include = ["dspy", "dspy.*"]
 exclude = ["tests", "tests.*"]
 
 [tool.setuptools.package-data]
-dspy = ["primitives/*.js"]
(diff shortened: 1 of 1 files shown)


phelps-sg:
ah ok thanks.  Shall I raise a new issue and corresponding PR to fix type hints and implement static type checking in github actions?

Issue #8875, opened 2025-09-30
[Bug] XMLAdapter inserting wrong markers


enitrat:
### What happened?

The XMLAdapter inserts a `[[ ## completed ## ]]` instead of having no completion tag _at all_ (or a <completed> tag)
Also, the `User message:` is formatted using ChatAdapter format for the input fields instead of XML format.

Reading the full prompt when using XMLAdapter, we see the wrong tags used:

```text

[2025-09-30T23:08:40.684097]

System message:

Your input fields are:
1. `query` (str):
Your output fields are:
1. `answer` (str):
All interactions will be structured in the following way, with the appropriate values filled in.

<query>
{query}
</query>

<answer>
{answer}
</answer>

[[ ## completed ## ]]
In adhering to this structure, your objective is:
        Given the fields `query`, produce the fields `answer`.


User message:

[[ ## query ## ]]
when was Marie Curie born

Respond with the corresponding output fields wrapped in XML tags`<answer>`, and then end with the `<completed>` tag.


Response:

<query>
when was Marie Curie born
</query>

<answer>
Marie Curie was born on November 7, 1867.
</answer>

[[ ## completed ## ]]
```

Although it says `Respond with the corresponding output fields wrapped in XML tags`<answer>`, and then end with the `<completed>` tag.`, there's no `<completed>` tag, instead there's a `[[ ## completed ## ]]` inherited from the base ChatAdapter.

### Steps to reproduce

```python
from dspy.adapters import XMLAdapter

dspy.configure(lm=dspy.LM("gemini/gemini-flash-latest", max_tokens=30000, cache=False), adapter=XMLAdapter())
prog2 = dspy.Predict("query: str -> answer: str")
prog2(query="when was Marie Curie born")
dspy.inspect_history(n=1)
```

output:

### DSPy version

3.0.3

TomeHirata:
Thanks for the report. Please take a look at my reply [here](https://github.com/stanfordnlp/dspy/pull/8876#issuecomment-3354514729).

chenmoneygithub:
I am actually confused where it grabs the knowledge of `[[ ## completed ## ]]`, as the entire prompt doesn't mention `[[ ## completed ## ]]` at all. 

Speaking of XMLAdapter, the current implementation doesn't really make much difference in the quality, we have a pending PR #8482 that introduces the desired behavior, but unfortunately we haven't yet gained bandwidth to land it.

[...]

PR #8482, opened 2025-07-02
feat(adapter): rewrite XMLAdapter for nested-data support

Bhuvanesh09:
Closes #8481

## TL;DR
* Replaces regex parsing with `xml.etree.ElementTree`.
* Supports nested Pydantic models, repeated tags → `List`, mixed data types.
* Keeps all existing flat-structure behaviour (no breaking changes expected).

---

## Motivation
`XMLAdapter` failed on any hierarchical XML (see #8481).  Users were forced to switch to `JSONAdapter`, losing the readability benefits of XML.  This PR brings feature-parity with `JSONAdapter`.

---

### 1. Parsing & Formatting
| Area | Old | New |
|------|-----|-----|
| Parsing engine | Regex `<(\w+)>(.*?)</\1>` | `ElementTree` traversal |
| Formatting | JSON-in-XML | Canonical nested XML |
| Error handling | Bare exceptions | Explicit `AdapterParseError` with context |

### 2. New helpers
* `_xml_to_dict(element) → Any`
* `_dict_to_xml(data, tag) → str`

### 3. Removed
* `_parse_field_value()` – superseded by full XML mapping.

---

## Backwards compatibility
* Flat structures behave exactly as before (all original tests still pass).
* No API signature changes; only internal behaviour differs.

---

## Example (was failing, now passes)

```python
class Address(BaseModel):
    street: str; city: str

class Person(BaseModel):
    name: str; age: int; address: Address

class Sig(dspy.Signature):
    text: str = dspy.InputField()
    person: Person = dspy.OutputField()

xml_out = """
<person><name>John</name><age>30</age>
  <address><street>Main</street><city>NYC</city></address>
</person>
"""

assert dspy.XMLAdapter(Sig()).parse(xml_out).person.name == "John"
```

### Tests added
- appropriate tests have been added at : tests/adapters/test_xml_adapter.py

### Risks / limitations
- ElementTree does not preserve attribute order; irrelevant for our use-case but worth noting.
- Doesn’t yet support XML attributes `(<tag attr="…">)`

Happy to engage in conversations, and grateful for this chance to contribute to DSPy.

PR #8876, opened 2025-09-30
fix: XML adapter's markers

enitrat:
# Summary

## Issue with the XMLAdapter
- XMLAdapter inherited ChatAdapter’s bracketed completion marker (`[[ ## completed ## ]]`) behavior, leading to prompts that mixed XML tags with ChatAdapter markers. This confused models and produced `[[ ## completed ## ]]` instead of the expected XML-only output (Closes https://github.com/stanfordnlp/dspy/issues/8875)

## Cause of the issue
- XMLAdapter was designed to emit XML tags (e.g., `<answer>...</answer>`), but it only overrode parts of the inherited ChatAdapter class:
  - Overrode: `format_field_with_value` (to XML), `user_message_output_requirements` (wrongly instructing to add a `<completed>` tag), and `parse`.
  - Did not  override all ChatAdapter-provided formatting paths that inject the bracketed style:
    - `format_field_structure` (system structure example) added `[[ ## completed ## ]]` in ChatAdapter.
    - `format_assistant_message_content` appended `[[ ## completed ## ]]` in ChatAdapter
    - `format_user_message_content` for inputs used ChatAdapter’s `[[ ## field ## ]]` headers.

As a result, the final prompt contained the ChatAdapter marker `[[ ## completed ## ]]` and bracketed sections instead of XML fields.

## How I fixed it
- Ensure XMLAdapter formats the entire prompt surface (system, user, assistant) with XML tags only.
- Changes:
  - Override XMLAdapter input message formatting to emit XML tags:
    - `dspy/adapters/xml_adapter.py:44` (`format_user_message_content`) now wraps inputs with `<field>...</field>`.
  - Keep structure and assistant outputs XML-only, (no `[[ ## completed ## ]]`, no `<completed>` tag):

- Tests added:
  - `test_xml_adapter_user_message_contains_xml_input_tags` asserts that `XMLAdapter.format(...)` produces `<question>...</question>` in the user message for an input named `question`, rather than using ChatAdapter's formatting.

## Further discussion

This issue could have been avoided with a better architectural design of the adapters; notably by avoiding inheriting from a base class ChatAdapter that has opinionated formatting of the prompt. Perhaps in a future PR we could take another approach, but this is out of scope for now and would require more discussions.

### DIFF
--- a/dspy/adapters/xml_adapter.py
+++ b/dspy/adapters/xml_adapter.py
@@ -19,10 +19,68 @@ class XMLAdapter(ChatAdapter):
             output.append(f"<{field.name}>\n{formatted}\n</{field.name}>")
         return "\n\n".join(output).strip()
 
+    def format_field_structure(self, signature: type[Signature]) -> str:
+        """
+        XMLAdapter requires input and output fields to be wrapped in XML tags like `<field_name>`.
+        """
+        from dspy.adapters.utils import translate_field_type
+        from pydantic.fields import FieldInfo
+
+        parts = []
+        parts.append("All interactions will be structured in the following way, with the appropriate values filled in.")
+
+        def format_signature_fields_for_instructions(fields: dict[str, FieldInfo]):
+            return self.format_field_with_value(
+                fields_with_values={
+                    FieldInfoWithName(name=field_name, info=field_info): translate_field_type(field_name, field_info)
+                    for field_name, field_info in fields.items()
+                },
+            )
+
+        parts.append(format_signature_fields_for_instructions(signature.input_fields))
+        parts.append(format_signature_fields_for_instructions(signature.output_fields))
+        return "\n\n".join(parts).strip()
+
+    def format_user_message_content(
+        self,
+        signature: type[Signature],
+        inputs: dict[str, Any],
+        prefix: str = "",
+        suffix: str = "",
+        main_request: bool = False,
+    ) -> str:
+        messages = [prefix]
+        for k, v in signature.input_fields.items():
+            if k in inputs:
+                value = inputs.get(k)
+                formatted_field_value = format_field_value(field_info=v, value=value)
+                messages.append(f"<{k}>\n{formatted_field_value}\n</{k}>")
+
+        if main_request:
+            output_requirements = self.user_message_output_requirements(signature)
+            if output_requirements is not None:
+                messages.append(output_requirements)
+
+        messages.append(suffix)
+        return "\n\n".join(messages).strip()
+
+    def format_assistant_message_content(
+        self,
+        signature: type[Signature],
+        outputs: dict[str, Any],
+        missing_field_message=None,
+    ) -> str:
+        return self.format_field_with_value(
+            {
+                FieldInfoWithName(name=k, info=v): outputs.get(k, missing_field_message)
+                for k, v in signature.output_fields.items()
+            },
+        )
+
     def user_message_output_requirements(self, signature: type[Signature]) -> str:
-        message = "Respond with the corresponding output fields wrapped in XML tags"
+        message = "Respond with the corresponding output fields wrapped in XML tags: "
         message += ", then ".join(f"`<{f}>`" for f in signature.output_fields)
-        message += ", and then end with the `<completed>` tag."
+        message += "."
         return message
 
     def parse(self, signature: type[Signature], completion: str) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


TomeHirata:
Thanks @enitrat. I left some comments, can you take a look?

dspy/adapters/xml_adapter.py
```diff
@@ -19,10 +19,68 @@ def format_field_with_value(self, fields_with_values: dict[FieldInfoWithName, An
...
+                    FieldInfoWithName(name=field_name, info=field_info): translate_field_type(field_name, field_info)
+                    for field_name, field_info in fields.items()
+                },
+            )
+
+        parts.append(format_signature_fields_for_instructions(signature.input_fields))
+        parts.append(format_signature_fields_for_instructions(signature.output_fields))
+        return "\n\n".join(parts).strip()
+
+    def format_user_message_content(
```
TomeHirata: Do we need to overwrite `format_user_message_content` and pass the input fields in XML format? Like JSONAdapter, I think it's fine to keep the user message consistent and just modify the response instruction and parsing.

dspy/adapters/xml_adapter.py
```diff
@@ -19,10 +19,68 @@ def format_field_with_value(self, fields_with_values: dict[FieldInfoWithName, An
             output.append(f"<{field.name}>\n{formatted}\n</{field.name}>")
         return "\n\n".join(output).strip()
 
+    def format_field_structure(self, signature: type[Signature]) -> str:
+        """
+        XMLAdapter requires input and output fields to be wrapped in XML tags like `<field_name>`.
+        """
+        from dspy.adapters.utils import translate_field_type
+        from pydantic.fields import FieldInfo
```
TomeHirata: can we import these in the top level?

dspy/adapters/xml_adapter.py
enitrat: My personal opinion (and expectation as a user) would be that everything's passed as XML. JSON is a bit different as it's not very well adapted for that, but XML is better adapted

enitrat:
@TomeHirata Do you agree that the `completed` marker shuold be `<completed/>`, not `<completed>`? as it's a "self closing" tag

edit: after further reading about the purpose of a completion marker, it seems that it's only useful for the `ChatAdapter` but the `XMLAdapter` should not include it.

The `StreamListener` class (dspy/streaming/streaming_listener.py) uses a token buffering mechanism to detect end markers in an LLM response. Notably, it looks for the end marker that matches the regex defined for the given adapter
   - ChatAdapter: `r"\[\[ ## (\w+) ## \]\]"` (matches next field or `[[ ## completed ## ]]`)
   - JSONAdatper: `r"\w*\"(,|\s*})"`
   - XMLAdapter: `rf"</{self.signature_field_name}>"`
   - Only when this pattern is found, it sets `stream_end = True`

As such, a `<completed>` marker is useless for XMLAdapter and should then be removed

dspy/adapters/xml_adapter.py
enitrat: In any case i'd be interesting of hearing the counter argument, as i'm trying to understand better the rationale behind how adapters are structured!

TomeHirata: I'd limit the size of this PR since we'll eventually refactor XMLAdapter in https://github.com/stanfordnlp/dspy/pull/8482. So I'd suggest we just focus on removing `<completed>` tag in this PR, or fix the issue in the other PR.

enitrat: Got it. Reverted these changes.

enitrat:
Hey @TomeHirata, thanks for the answer. I've properly applied the suggestions and restricted the scope of the changes, and updated the PR summary to match the current changes implemented in this PR. Thanks for the review!

dspy/adapters/xml_adapter.py
```diff
@@ -19,10 +21,43 @@ def format_field_with_value(self, fields_with_values: dict[FieldInfoWithName, An
...
+        return self.format_field_with_value(
+            {
+                FieldInfoWithName(name=k, info=v): outputs.get(k, missing_field_message)
+                for k, v in signature.output_fields.items()
+            },
+        )
+
     def user_message_output_requirements(self, signature: type[Signature]) -> str:
-        message = "Respond with the corresponding output fields wrapped in XML tags"
+        message = "Respond with the corresponding output fields wrapped in XML tags, starting with the field "
```
TomeHirata: Do we need this change?

enitrat: ah sorry that slipped through, I originally made the change to be more consistent with the chatadapter. but i can remove it.

```suggestion
        message = "Respond with the corresponding output fields wrapped in XML tags"
```

enitrat:
@TomeHirata should be all good now!

PR #8880, opened 2025-10-01
Fix responses API

chenmoneygithub:
Responses API always return one choice instead of multiple choice. This is a very confusing behavior without clear documentation, but basically all items in the `output` in Response object map to one choice.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -251,6 +251,7 @@ class BaseLM:
             result["tool_calls"] = tool_calls
         if len(reasoning_contents) > 0:
             result["reasoning_content"] = "".join(reasoning_contents)
+        # All `response.output` items map to one answer, so we return a list of size 1.
         return [result]
 
 


dspy/clients/base_lm.py
```diff
@@ -217,26 +217,42 @@ def _extract_citations_from_response(self, choice):
...
         Returns:
-            List of processed outputs
+            List of processed outputs, which is always of size 1 because the Response API only supports one output.
         """
-        outputs = []
+        text_outputs = []
         tool_calls = []
+        reasoning_contents = []
+
         for output_item in response.output:
```
TomeHirata, line 232: nit: Maybe add a comment that even though the output field of the responses API is a list, all output items belong to a single choice

dspy/clients/base_lm.py
```diff
@@ -217,26 +217,42 @@ def _extract_citations_from_response(self, choice):
...
+            output_item_type = output_item.type
+            if output_item_type == "message":
                 for content_item in output_item.content:
-                    outputs.append(content_item.text)
-            elif output_item.type == "function_call":
+                    text_outputs.append(content_item.text)
+            elif output_item_type == "function_call":
                 tool_calls.append(output_item.model_dump())
+            elif output_item_type == "reasoning":
+                if getattr(output_item, "content", None) and len(output_item.content) > 0:
```
TomeHirata, line 240: q: is it possible that `output_item` does not have `content` attribute when the item type is `reasoning`?

dspy/clients/base_lm.py
```diff
@@ -217,26 +217,42 @@ def _extract_citations_from_response(self, choice):
...
+                    for content_item in output_item.content:
+                        reasoning_contents.append(content_item.text)
+                elif getattr(output_item, "summary", None) and len(output_item.summary) > 0:
```
TomeHirata, line 243: ditto

dspy/clients/base_lm.py
chenmoneygithub: technically it's possible that either content or summary is unset, but again OpenAI is doing an odd job documenting Responses API.

dspy/clients/base_lm.py
chenmoneygithub: done!

Issue #8689, opened 2025-08-20
[Feature] Allow metrics to return multiple scores, and allow optimizers like GEPA to support multi-score optimization


LakshyAAAgrawal:
### What feature would you like to see?

Many tasks have multiple objectives which may be complementary, or a trade-off. How can we allow metrics to return such subscores, and have optimizers optimize them?

This will require thinking about:
1. What is the right interface for the user-defined metric to return multiple scores
2. How can the user specify how different sub-scores should be aggregated? Arithmetic/Geometric mean, weighted avg, etc.
3. How should scores across instances be aggregated?

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Additional Context

Relevant discussion: https://discord.com/channels/1161519468141355160/1161519469319946286/1407399169517420718

Upstream GEPA issue: https://github.com/gepa-ai/gepa/issues/2

LakshyAAAgrawal:
https://github.com/gepa-ai/gepa/pull/47 and https://github.com/gepa-ai/gepa/issues/2

MatsErdkamp:
[copy/paste of the Discord response]

Finally got some time to get back around to this. Here's my current thinking.

### Every custom score function should be usable.
If we implement something like this, we might as well go all the way: no half-measures where only weighted sums are supported. Any arbitrary formula should be valid.

### Score should ideally be adapted to just take a function instead of adding a separate score_fn field.
For optimizers that don’t support evaluating the function we could just calculate the outcome and send that scalar to the optimizer.

### Notation
As far as notation goes; I think there are three good sub-options.

**a. String-based formulas with declared axes.**
```
score = dspy.metrics.formula(
    axes={"acc": exact_match, "bleu": bleu_like},
    expr="0.7*acc + 0.3*bleu**2"
)
```

Concise, serializable, config/CLI-friendly.

**b. Function-based single score_fn.**
```
def score_fn(ax):
    acc  = ax.acc(exact_match)
    bleu = ax.bleu(bleu_like)
    return 0.7*acc + 0.3*bleu**2

score = dspy.metrics.custom(score_fn)
```

Pythonic, IDE-friendly, type-safe.

**c. Composable builder-style.**
```
score = (
    dspy.metrics.combo()
    .weighted({"acc": 0.7, "bleu": 0.3})
    .clip(min=-1, max=1)
    .penalize("len_ratio", factor=0.1, fn="abs_log")
    .build(axes={
[...]

PR #8888, opened 2025-10-02
Allow metrics to return multiple scores

MatsErdkamp:
aims to solve [this issue](https://github.com/stanfordnlp/dspy/issues/8689).


MatsErdkamp:
Returned this to draft since the downstream GEPA implementation is almost final. Will make some changes in accordance withthe GEPA upgrade

LakshyAAAgrawal:
@MatsErdkamp can you pull from main (where I have updated gepa version to 0.0.24 for multi objective) and resolve all the merge conflicts?

LakshyAAAgrawal:
@MatsErdkamp, The current implementation touches many different components. I believe there exists a much simpler change, that simply reads a specific field, something like "scores" from the return value of the metric (which is a dspy.Prediction object), and then only dspy.GEPA needs to interpret it. What do you think? Can you redo things such that the changes are minimized?

MatsErdkamp:
The advantage of this system would be that the entire score_fn can be known by the optimizer. While not relevant for GEPA as of today, it could be interesting to provide that knowledge. An LLM could reason over the score function and see what the highest impact change to score would be. You could argue that one can already do this manually via the feedback field, but this implementation would open up the door for future optimizers that actually use knowledge of the score_fn as a core part of their architecture. They could treat score_fn as a white-box objective: it parses the function and creates a gradient landscape, and then dynamically weights which “levers” to optimize next (e.g., factuality vs. brevity vs. format) based on highest expected marginal gains.

I agree that it touches quite a bit more and makes implementation more complex. But I would personally be interested to create an optimizer that functions as described above. If you think it's not worth the tradeoff that's very fair. I could reimplement it in the {"scores"} way. 

@LakshyAAAgrawal Let me know what you think and if this makes sense at all :-)

LakshyAAAgrawal:
@MatsErdkamp, while such an optimizer would indeed be useful, I don't understand how the changes across the codebase here enable that. Even without making these changes, whatever information needs to be surfaced from within the metric can be provided in the dspy.Prediction object returned by the metric?

I will share that one of the biggest reasons python+pytorch became popular for deep learning is that they don't get in the way of the programmer, who just expresses the neural-network forward pass in a very easy to use DSL, and the underlying engine takes care of the rest. If we mandate every user to provide a custom object like "Score" or "Subscore", it becomes really difficult to use it.

I highly encourage exploring optimization directions, and happy to discuss ideas, but I am not sure if this interface change adds anything beyond what dspy.Prediction() can already achieve.

MatsErdkamp:
@LakshyAAAgrawal Fair; any ideas on how to for example implement a score_fn argument in Prediction()? What should the type be? Passing it as a string feels very brittle. 

Basically. Say my score_fn is "quality**2 + leakage"; how can we make sure optimizers can understand this formula? The reason I added subscore() is such that you can write your score_fn with the variable names in place. Also you could hook it up to mlflow for example and get them registered as metrics out of the box, but thats more of a nice-to-have.

That being said; I do agree with you that it might not be worth the complexity; but eager to hear your thinking on how to make score_fn="quality**2 + leakage" or something similar functional. Or maybe you have even better totally different ideas of course!

LakshyAAAgrawal:
Why does not returning the string "Your quality score is X and leakage score is Y. The aggregated score, calculated as quality**2 + leakage is Z. Can you improve the aggregate score"? For example, have a look at https://dspy.ai/tutorials/gepa_papillon/#optimize-papillon-with-dspygepa where we use a similar feedback.

MatsErdkamp:
@LakshyAAAgrawal  Yep — that pattern works well for GEPA today (Papillon is a good proof). My only hesitation is making it the core interface for “objective-aware” optimizers: once the objective lives only in feedback=, it’s always best-effort parsing of a narrative string. If we ever want optimizers to truly operate on the objective (in a more reliable / model-agnostic way), a structured/typed representation is a cleaner contract.

Admittedly, this is a bit ahead of the facts since I’m not aware of any optimizer that actually does this today, so I get your point on keeping things simple. Do you think it’s worth keeping the door open for optimizers to programmatically understand score_fn, or should we just keep this entirely in “LLM-land” via feedback=? 

For now, we could probably allow Prediction() to accept multiple scores. The functionality of it also accepting a scoreFormula could be added later once it might be needed (if ever).

LakshyAAAgrawal:
dspy.Prediction is just a dict, which can map to values of any type. Let's say even if there's a need for more interpretable types, I don't understand how this design will not capture it? I would highly recommend simplifying the PR, and minimizing the code changes required.

PR #8892, opened 2025-10-03
Fix chunk order issue in native field streaming

TomeHirata:
The current streaming listener implementation has an issue that chunk order is modified (by at most 10 chunks) when a native response chunk is included in the string streaming due to the buffering logic for the string output field. This PR fixes the issue by flushing the chunks once when the native response chunk is received.

### DIFF
--- a/dspy/streaming/streaming_listener.py
+++ b/dspy/streaming/streaming_listener.py
@@ -97,8 +97,6 @@ class StreamListener:
 
         try:
             chunk_message = chunk.choices[0].delta.content
-            if chunk_message is None:
-                return
         except Exception:
             return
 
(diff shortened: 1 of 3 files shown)


dspy/streaming/streaming_listener.py
```diff
@@ -112,6 +110,20 @@ def receive(self, chunk: ModelResponseStream):
                     is_last_chunk=self.stream_end,
                 )
 
+        # If we receive an empty chunk but streaming has started, flush the buffer.
+        # LiteLLM does not send completely empty chunks (https://github.com/BerriAI/litellm/blob/main/litellm/litellm_core_utils/model_response_utils.py#L10),
+        # so empty content means it has other native fields such as provider_specific_fields.
+        if not chunk_message:
```
chenmoneygithub, line 116: is this correct? From the log shared internally, the LM can produce citation chunks in between of field chunks, then it could lead to unintended flush?

TomeHirata: Flush is for finding the end token, right? It is true that the native chunk is passed between field chunks, when the citation chunk is passed, the end token (like `[[ ##`) shouldn't be in the queue. So it's a safe time to flush the tokens for the ongoing string field.

chenmoneygithub: I am not entirely sure about the LM streaming order, like how it tangles normal text with native features/events, but if it looks like:

1. text: what
2. text: the
3. citation: <citation>
4. text: jesus?
5. text: [[ ##
6. citation: <citation>
7. text: completed ## ]]

Then at step 6, ` [[ ##` will be yielded by flush call.

TomeHirata: In that example, does it mean `completed ## ]]` is supported by the citation #5? I'm assuming that LM produces the native response and string response in the right order. If it makes mistakes, yeah it won't work well.

PR #8897, opened 2025-10-03
Allow skipping stream buffering

TomeHirata:


dspy/streaming/streaming_listener.py
```diff
@@ -160,6 +160,14 @@ def receive(self, chunk: ModelResponseStream):
                 return
 
         if self.stream_start and chunk_message:
+            if getattr(chunk, "_dspy_skip_buffer", None):
```
chenmoneygithub, line 163: Per our discussion offline, functionality-wise this has no problem, but this line will for sure confuse our contributors except us two, because `_dspy_skip_buffer` is not declared/referenced anywhere in the codebase. This looks similar to [predict_id](https://github.com/stanfordnlp/dspy/blob/a0b2155a9b295585d113cf5f826e9b7c9dad6103/dspy/clients/lm.py#L318) handling, but it's entirely internal logic, while this `_dspy_skip_buffer` requires users to explicitly set it. 

I am wondering if it is a better option to just make buffer size adjustable, and let the users handle boilerplate strings. That will be less ideal for internal use cases, while it keeps DSPy unbiased towards broad audience.

TomeHirata: If users need to deal with the boilerplate and want to disable buffering for all chunks, I think users can just use streamify without StreamListener. This option serves for a different case where users (custom LM) know **some chunks** that don't contain boilerplate.

PR #8903, opened 2025-10-06
Add Learn MCP page

TomeHirata:
MCP has more and more popularity, so we should document how to use MCP with DSPy in the main section in addition to tutorial.

### DIFF
--- a/docs/docs/learn/programming/mcp.md
+++ b/docs/docs/learn/programming/mcp.md
@@ -1,0 +1,187 @@
+---
+sidebar_position: 3
+---
+
+# Model Context Protocol (MCP)
+
+The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) is an open protocol that standardizes how applications provide context to language models. DSPy supports MCP, allowing you to use tools from any MCP server with DSPy agents.
+
+## Installation
+
+Install DSPy with MCP support:
+
+```bash
+pip install -U "dspy[mcp]"
+```
+
+## Overview
+
+MCP enables you to:
+
+- **Use standardized tools** - Connect to any MCP-compatible server
+- **Share tools across stacks** - Use the same tools across different frameworks
+- **Simplify integration** - Convert MCP tools to DSPy tools with one line
+
+As long as you have an `mcp.ClientSession`, you can use MCP tools with DSPy's `ReAct` module or any other tool-based workflow.
+
+## Using MCP with DSPy
+
+### 1. Stdio Server (Local Process)
+
+The most common way to use MCP is with a local server process communicating via stdio:
+
+```python
+import asyncio
+import dspy
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+
+async def main():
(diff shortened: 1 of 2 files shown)


TomeHirata:
The test failure is not related to this PR

docs/docs/learn/programming/mcp.md
```diff
@@ -0,0 +1,187 @@
...
+
+```bash
+pip install -U "dspy[mcp]"
+```
+
+## Overview
+
+MCP enables you to:
+
+- **Use standardized tools** - Connect to any MCP-compatible server
```
chenmoneygithub: nit: period at the end

docs/docs/learn/programming/mcp.md
```diff
@@ -0,0 +1,187 @@
...
+- **Share tools across stacks** - Use the same tools across different frameworks
+- **Simplify integration** - Convert MCP tools to DSPy tools with one line
+
+As long as you have an `mcp.ClientSession`, you can use MCP tools with DSPy's `ReAct` module or any other tool-based workflow.
```
chenmoneygithub: I think users will be confused about "As long as you have an `mcp.ClientSession`", because it's not a well-known thing for many people.

We should maybe add a short paragraph to explain how DSPy manages the connections to the MCP server - we utilize the native MCP session for all connection types, and don't provide extra features for connection handling

docs/docs/learn/programming/mcp.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Using MCP with DSPy
+
+### 1. Stdio Server (Local Process)
+
+The most common way to use MCP is with a local server process communicating via stdio:
```
chenmoneygithub: this is a question not comment - do you think HTTP connection is more commonly used than Stdio now? Stdio is a strange usage especially in production environment IMO

docs/docs/learn/programming/mcp.md
```diff
@@ -0,0 +1,187 @@
...
+                max_iters=5
+            )
+
+            result = await react_agent.acall(task="Check the weather in Tokyo")
+            print(result.result)
+
+asyncio.run(main())
+```
+
+## Creating an MCP Server
```
chenmoneygithub: Do we need this section to be a standalone section? maybe we can move the content under "## Tool Conversion", and say something like "Let's build a simple MCP server with tools and surface the MCP tools to DSPy"

docs/docs/learn/programming/mcp.md
TomeHirata: I think HTTP is getting more common, shall we change the order?

docs/docs/learn/programming/mcp.md
TomeHirata: I think this part can be covered by tutorial, let's just remove this section to simplify

docs/docs/learn/programming/mcp.md
TomeHirata: Updated!

PR #8907, opened 2025-10-06
Make the buffer condition more precise

TomeHirata:
Currently, StreamListener always buffers tokens up to 10 chunks to find the end token. This behavior has an issue that 1) it causes unnecessary delay in token streaming and 2) may change the chunk order for native response chunks. This PR updates the buffer condition to buffer chunks only when it is possible to form the end boilerplate.


kenctrl:
LGTM, as long as the method matches all potential cases. Thanks Tomu!

dspy/streaming/streaming_listener.py
```diff
@@ -73,6 +79,41 @@ def _buffered_message_end_with_start_identifier(self, concat_message: str, start
...
+        we know it cannot form the pattern, so we should yield immediately.
+
+        Args:
+            concat_message: The concatenated buffered message
+            adapter_name: The name of the adapter being used
+
+        Returns:
+            True if the message could potentially form part of the end identifier
+        """
+        if adapter_name not in self.adapter_identifiers:
```
chenmoneygithub: This has been verified in `receive()`, we don't need to double verify since this is a private method.

dspy/streaming/streaming_listener.py
```diff
@@ -163,19 +204,22 @@ def receive(self, chunk: ModelResponseStream):
...
                 # The next field is identified, we can end the stream and flush out all tokens in the buffer.
                 self.stream_end = True
-                last_token = self.flush()
-                token = token + last_token if token else last_token
+                token = self.flush()
                 token = token.rstrip()  # Remove the trailing \n\n
+            elif not self._could_form_end_identifier(concat_message, adapter_name):
+                # Buffer cannot form end identifier, safe to yield the oldest token
+                # Keep at least 1 token in buffer in case next token creates end pattern
+                if self.field_end_queue.qsize() > 1:
```
chenmoneygithub: shall we just call `flush()` here? this won't affect the use case we are tackling, but technically a direct `flush()` can fit here.

tests/streaming/test_streaming.py
```diff
@@ -1102,3 +1092,41 @@ async def citation_stream(*args, **kwargs):
...
+        # XMLAdapter - should return True for partial closing tag
+        ("XMLAdapter", "some text <", True),
+        ("XMLAdapter", "some text </", True),
+        ("XMLAdapter", "some text </result", True),
+        # XMLAdapter - should return False for text that cannot form the pattern
+        ("XMLAdapter", "hello world", False),
+        ("XMLAdapter", "some text", False),
+    ],
+)
+def test_stream_listener_could_form_end_identifier(adapter_name, message, expected):
```
chenmoneygithub: This parameterization will generate lots of tiny test cases, shall we just move it to the test body?

tests/streaming/test_streaming.py
```diff
@@ -511,21 +501,17 @@ async def gpt_4o_mini_stream_2(*args, **kwargs):
...
-        assert all_chunks[2].predict_name == "predict2"
-        assert all_chunks[2].signature_field_name == "judgement"
-        assert all_chunks[2].chunk == "The"
-        assert all_chunks[3].chunk == " answer"
-        assert all_chunks[4].chunk == " is"
-        assert all_chunks[5].chunk == " humorous"
-        assert all_chunks[6].chunk == " and"
-        assert all_chunks[7].chunk == " plays on the very funny and classic joke format."
+        # Verify judgement chunks
+        judgement_chunks = [chunk for chunk in all_chunks if chunk.signature_field_name == "judgement"]
```
chenmoneygithub: open for discussion - do you think we should verify each chunk matches the original stream text as we did in the old file?

TomeHirata: That may be better, let me change

PR #8908, opened 2025-10-06
fix(LM): Change default temperature and max_tokens to be None

isaacbmiller:
Will allow users to just write dspy.LM("gpt-5-nano") but also there is the case where this overrides a user's settings.

I can see an argument against this change, but it seems the direction we are heading is that API users have less explicit control over the temperature.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -89,7 +89,7 @@ class LM(BaseLM):
 
         if model_pattern:
 
-            if (temperature is not None and temperature != 1.0) or (max_tokens is not None and max_tokens < 16000):
+            if (temperature and temperature != 1.0) or (max_tokens and max_tokens < 16000):
                 raise ValueError(
                     "OpenAI's reasoning models require passing temperature=1.0 or None and max_tokens >= 16000 or None to "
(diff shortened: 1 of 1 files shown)


dspy/clients/lm.py
```diff
@@ -88,7 +88,16 @@ def __init__(
         model_pattern = re.match(r"^(?:o[1345]|gpt-5)(?:-(?:mini|nano))?", model_family)
 
         if model_pattern:
-            if max_tokens < 16000 or temperature != 1.0:
+            if temperature == 0 and max_tokens == 4000:
```
chenmoneygithub: Per our offline discussion, let's proceed with `temperature=None` and `max_tokens=None` as the default, which aligns with litellm defaults.

dspy/clients/lm.py
```diff
@@ -88,9 +88,10 @@ def __init__(
         model_pattern = re.match(r"^(?:o[1345]|gpt-5)(?:-(?:mini|nano))?", model_family)
 
         if model_pattern:
-            if max_tokens < 16000 or temperature != 1.0:
+
+            if (temperature is not None and temperature != 1.0) or (max_tokens is not None and max_tokens < 16000):
```
chenmoneygithub:
```suggestion
            if (temperature and temperature != 1.0) or (max_tokens and max_tokens < 16000):
```

PR #8910, opened 2025-10-06
fix(ReAct): Change format so that trajectory demos inside history are used using chat format

isaacbmiller:


isaacbmiller:
superseded by reactv2

PR #8912, opened 2025-10-07
Better tutorial sidebar - do not auto expand all tutorials

chenmoneygithub:
The current tutorial page expands all items by default, making the sections other from the first invisible:
image

It's a good idea to expand, but we can limit each section to have only 3, then use a "more tutorials" button to bring users to the specific section's landing page:
image

Clicking on more tutorials bring users to the landing page:

image

### DIFF
--- a/docs/docs/js/tutorial-nav.js
+++ b/docs/docs/js/tutorial-nav.js
@@ -1,0 +1,77 @@
+document.addEventListener('DOMContentLoaded', function () {
+  // Only run on the tutorials page
+  const isTutorialsPage = window.location.pathname.includes('/tutorials');
+
+  if (isTutorialsPage) {
+    collapseTutorialNav();
+  }
+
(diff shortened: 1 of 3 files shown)


docs/docs/js/tutorial-nav.js
```diff
@@ -0,0 +1,77 @@
+document.addEventListener('DOMContentLoaded', function () {
```
TomeHirata, line 1: I like the behavior! Just wondering if there's an existing plugin/configuration we can use to achieve this behavior.

chenmoneygithub: good question! I searched around and mkdocs doesn't provide an out-of-box feature for it. I also thought about making "More tutorials" an item, but it could be a bit confusing because it's not equivalent to other tutorials.

TomeHirata (approved):

PR #8927, opened 2025-10-10
feat(gepa): add tool description optimization for multi-agent systems

Ju-usc:
## Problem
GEPA optimizes signatures but not tool descriptions, limiting ReAct agent optimization. 

Closes #8706

## Solution

When `optimize_tool_descriptions=True`, GEPA now:
1. Extracts tool descriptions from all nested modules (via `named_sub_modules()`)
2. Includes them in the optimization process alongside signature instructions
3. Returns optimized system with improved tool descriptions

## Changes

- Add `optimize_tool_descriptions` parameter to GEPA (default `False`)
- Extract tool descriptions using `named_sub_modules()` traversal in `compile()`
- Apply optimized descriptions in `DspyAdapter.build_program()`
- Add 4 comprehensive tests covering single-agent and multi-agent scenarios

## Usage Example

```python
import dspy

# Create multi-agent system
class ResearchAssistant(dspy.Module):
    def __init__(self):
        super().__init__()
        search_tool = dspy.Tool(search_fn, name="search", desc="Searches web")
        self.researcher = dspy.ReAct("query -> findings", tools=[search_tool])
        
        def delegate_research(query):
            return self.researcher(query=query).findings
        
        research_tool = dspy.Tool(delegate_research, name="research", desc="Research things")
        calc_tool = dspy.Tool(calc_fn, name="calculator", desc="Does math")
        self.assistant = dspy.ReAct("question -> answer", tools=[research_tool, calc_tool])

# Enable tool optimization
optimizer = dspy.GEPA(
    metric=my_metric,
    reflection_lm=lm,
    auto="light",
    optimize_tool_descriptions=True,  # ← Enable tool optimization
)

# Optimizes ALL tools (calculator, research, search) holistically
optimized = optimizer.compile(ResearchAssistant(), trainset=train, valset=val)
```

## Backward Compatibility

✅ Fully backward compatible - default `optimize_tool_descriptions=False`

## Tests

- All 16 tests pass (4 new + 12 existing GEPA tests)
- Tests cover: adapter functionality, single-agent, multi-agent nested discovery, end-to-end optimization

PR #8928, opened 2025-10-10
feat(gepa): add tool description optimization for multi-agent systems

### LINKED ISSUE #8706: [Feature] Allow GEPA to update tool descriptions and tool error responses

GEPA should be able to optimize the tool descriptions for dspy.Tool and the error messages returned by users corresponding to it.

### Additional Context

Reference: https://discord.com/channels/1161519468141355160/1399482432239046677/1405225791809130587

Ju-usc:
## Summary

Addresses #8706 which requested GEPA to optimize tool descriptions.

When `enable_tool_optimization=True`, GEPA jointly optimizes:
- **ReAct modules**: react instructions, extract instructions, tool descriptions, and tool argument descriptions
- ~~Generic tool-using predictors support is TODO'd out pending DSPy trace lineage improvements.~~

All components are optimized together based on shared execution traces, enabling the reflection LM to see how components work together.

Backward compatible - `enable_tool_optimization=False` (default) preserves existing behavior.

## Issue

Closes #8706

### Core Implementation

- `enable_tool_optimization` parameter on GEPA (default `False`)
- ~~**Type-based detection**: Identifies tool-using predictors via signature field annotations (`dspy.Tool`, `list[dspy.Tool]`, `dict[str, Tool]`)~~ *(TODO)*
- ~~**Trace-based tool extraction**: Discovers actual tools used at runtime from execution traces~~ *(TODO)*
- **`ToolProposer`**: Specialized proposer with dynamic signature generation for each tool and argument
- **ReAct handling**: Discovers ReAct modules via `isinstance()` check, includes both react and extract predictors
- **Routing**: ReAct modules → `ToolProposer`; regular predictors → default/custom proposer
- **Serialization**: Tool modules stored as JSON configs with predictor instructions and tool schemas
- **Application**: Applies optimized descriptions to `dspy.Tool` objects by matching `tool.name`

### Testing

6 tests covering:
- Skip predictors without tool annotations
- ReAct module detection (single, multiple, nested)
- Apply optimized ReAct descriptions
- Selective optimization when LM returns `None`

### Documentation

- `GEPA_Advanced.md` - Tool optimization guide with usage examples
- `overview.md` - Brief introduction linking to advanced guide

### ReAct Agent

```python
import dspy

def search_web(query: str) -> str:
    return f"Search results for: {query}"

search_tool = dspy.Tool(search_web, name="search_web", desc="Search the web")

agent = dspy.ReAct("question -> answer", tools=[search_tool])

gepa = dspy.GEPA(
    metric=my_metric,
    reflection_lm=dspy.LM("openai/gpt-5-mini"),
    enable_tool_optimization=True,
    auto="medium"
)

optimized = gepa.compile(agent, trainset=trainset, valset=valset)
```

## Key Features

- **Joint Optimization**: Predictor instructions and tool descriptions optimized together
- **Selective Updates**: LM returns `None` for unchanged components
- **Multi-Agent Support**: Discovers nested ReAct modules


Ju-usc:
Apologies for accidentally closing #8927 

Thank you for the thorough review, @LakshyAAAgrawal! I'll address your feedback:

1. Since tools are categorically different from prompts, they should use a different reflection meta prompt. The default reflection meta prompt is shown here https://dspy.ai/api/optimizers/GEPA/GEPA_Advanced/#default-implementation, whereas I assume that the tool must use somewhat different meta prompt. Can you implement a propose_new_texts method that mimics the default_proposer shown in the link above for all prompts, but calls to a tool description specific prompt/signature for tool evolution.
2. Can you also add some description to the documentation, explaining that this feature is beneficial for React agents.
3. (This is not a requirement to merge the PR) Would it be possible to add a simple and short tutorial demonstrating the use and performance improvement via tool evolution?

I'll start working on items 1 and 2 and update the PR soon. Please let me know if you have any specific preferences for the tutorial format!

LakshyAAAgrawal:
Thanks a lot! For the tutorial, I think you can follow the current GEPA tutorial format (load a dataset, show an example from the dataset, build a dspy program, evaluate the baseline program on testset, run GEPA with new optimization settings, show the optimized programs' prompts and tool descriptions, and finally evaluate the optimized program).

Hopefully we should be able to see a nice and large gain on agentic tasks with this amazing contribution by you!

Ju-usc:
Hi @LakshyAAAgrawal,

I've implemented the tool-specific proposer as requested! Here's what's included:

**1. Tool-Specific Proposer Implementation** ✅
- Added `GenerateImprovedToolDescriptionFromFeedback` signature with a specialized reflection prompt
- Implemented `ToolProposer` and `SingleComponentToolProposer` following the MultiModalInstructionProposer pattern
- Routing logic in `DspyAdapter` that directs tools to `ToolProposer` and signatures to custom/default proposers
- Fully backward compatible with existing custom instruction proposers

**2. Documentation** ✅
- Added comprehensive section to GEPA_Advanced.md
- Explains when to use tool optimization (ReAct agents, multi-agent systems)
- Includes usage examples for both simple and nested agent architectures
- Documents how to inspect optimized tool descriptions

**Reflection Prompt Design:**
The tool-specific prompt is intentionally open-ended to avoid prescriptive patterns that might lead to local minima. It asks the LM to identify patterns in successful/unsuccessful tool usage and extract domain-specific information, without suggesting specific heuristics.

**Before I create a short tutorial (item #3), would you have any feedback on:**
- The reflection prompt design - is it general enough? Any improvements you'd suggest?
- The implementation approach - does the routing logic make sense?
- The documentation - anything unclear or missing?

Any feedback would be helpful before I invest time in the tutorial. Thank you!

Ju-usc:
wait there is a bug in the implementation working on it to fix. Also test has to be fixed.

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,199 @@ gepa = dspy.GEPA(
...
 )
 ```
+
+## Tool Description Optimization
+
+### What is optimize_tool_descriptions?
+
+The `optimize_tool_descriptions` parameter enables GEPA to optimize tool descriptions in addition to signature instructions. This is particularly valuable for ReAct agents and other tool-using systems, where the quality of tool descriptions directly impacts the agent's ability to select appropriate tools for each task.
+
+Unlike signature instructions that guide reasoning strategies, tool descriptions serve a fundamentally different purpose: they help agents decide **which tool to use** in a given situation. GEPA recognizes this categorical difference and applies a specialized reflection prompt tailored for tool selection decisions.
```
LakshyAAAgrawal: which tool to use, when to use it, and how to use it. All three are captured by the description.

LakshyAAAgrawal: Let's avoid the word "fundamentally". One can imagine that all of tool descriptions can (and many times do) simply included in the system prompt itself.

LakshyAAAgrawal: Please also add a corresponding entry in GEPA Overview, that links to this file/section.

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,199 @@ gepa = dspy.GEPA(
...
+    auto="medium"
+)
+optimized_program = gepa.compile(student, trainset=examples)
+```
+
+### When to Use optimize_tool_descriptions
+
+Consider enabling `optimize_tool_descriptions=True` when:
+
+- **Building ReAct agents**: ReAct agents rely on tool descriptions to make action selection decisions
```
LakshyAAAgrawal: One should consider using this, when they use dspy.Tool anywhere in the DSPy program. Here are a few scenarios for using dspy.Tool:
1. 2. 3.

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,199 @@ gepa = dspy.GEPA(
...
+- **Complex tool sets**: When managing many tools with overlapping capabilities
+- **Domain-specific tools**: Tools requiring specialized knowledge or context for proper usage
+
+### How It Works
+
+When enabled, GEPA:
+
+1. **Discovers all tools**: Traverses your program including nested sub-modules to find all `dspy.Tool` instances
+2. **Categorizes components**: Separates tools (identified by `tool:` prefix) from signature instructions
+3. **Routes components appropriately**:
```
LakshyAAAgrawal: During reflection, routes the components to be modified by the right metaprompt...

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,199 @@ gepa = dspy.GEPA(
...
+gepa = dspy.GEPA(
+    metric=my_metric,
+    reflection_lm=dspy.LM(model="gpt-5", temperature=1.0, max_tokens=32000, api_key=api_key),
+    instruction_proposer=MultiModalInstructionProposer(),  # For signatures
+    optimize_tool_descriptions=True,  # Enables ToolProposer for tools
+    auto="medium"
+)
+```
+
+**Note:** Tool optimization is fully backward compatible. Existing programs without tools, or with `optimize_tool_descriptions=False`, continue to work exactly as before.
```
LakshyAAAgrawal: I don't think we need to inform users about backward compatibility here. It should be implicit that there should be no behaviour changes for any program not containing dspy.Tool.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -273,6 +273,9 @@ def metric(
         warn_on_score_mismatch: GEPA (currently) expects the metric to return the same module-level score when 
             called with and without the pred_name. This flag (defaults to True) determines whether a warning is 
             raised if a mismatch in module-level and predictor-level score is detected.
+        optimize_tool_descriptions: Whether to optimize tool descriptions for modules with tools 
+            (e.g., ReAct agents). When enabled, tool descriptions are included in the optimization 
+            process alongside signature instructions. Default is False.
```
LakshyAAAgrawal: Add a link to GEPA Advanced/Tool section

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -88,42 +94,100 @@ def __init__(
...
                 else:
                     return self.custom_instruction_proposer(
                         candidate=candidate,
                         reflective_dataset=reflective_dataset,
-                        components_to_update=components_to_update
+                        components_to_update=components_to_update,
                     )
 
             self.propose_new_texts = custom_propose_new_texts
+        elif self.optimize_tool_descriptions:
```
LakshyAAAgrawal: Edge case: What should happen when user tries to provide both a custom proposer, and enables `optimize_tool_descriptions`

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -88,42 +94,100 @@ def __init__(
...
+                components_to_update: list[str],
+            ) -> dict[str, str]:
+                """Route components: tools to ToolProposer, signatures to GEPA's default."""
+                tool_components = [c for c in components_to_update if c.startswith("tool:")]
+                sig_components = [c for c in components_to_update if not c.startswith("tool:")]
+                
+                # Handle signature components - replicate proposer's default behavior
+                sig_texts = {}
+                if sig_components:
+                    from gepa.strategies.instruction_proposal import InstructionProposalSignature
```
LakshyAAAgrawal: This is a slight deviation from this PR, but would be a large enhancement (feel free to ignore):
1. Create 2 fields, self.instruction_proposal_signature and self.tool_proposer, which are initialized to the default InstructionProposalSignature and ToolProposerSignature.
2. Take an argument from dspy.GEPA that can override the default signature values.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -284,6 +358,28 @@ def make_reflective_dataset(self, candidate, eval_batch, components_to_update) -
                 continue
             ret_d[pred_name] = items
 
+        # Second pass: Process tools by copying ReAct data with annotation
+        react_module_name = None
+        for name in ret_d.keys():
+            if "react" in name.lower():
```
LakshyAAAgrawal: Is this robust? Might it be better to use isinstance or some other way?

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +310,141 @@ def __call__(
...
+    ```
+
+    This tool is available to the assistant. The following are examples of task inputs provided to the assistant, the assistant's decisions about which tools to use, and feedback on whether those decisions were correct:
+    ```
+    <examples_with_feedback>
+    ```
+
+    Your task is to write a better description for this tool.
+
+    Read the examples carefully and identify patterns in when the tool was used successfully versus when it was misused or overlooked. Identify any domain-specific information about the tool's capabilities or appropriate usage that may not be available to the assistant in the future. The assistant may have developed effective patterns for tool selection - if so, ensure the tool description supports those patterns.
```
LakshyAAAgrawal: Tool use. Also suggest identifying any failure modes of the tool?

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +310,141 @@ def __call__(
...
+
+    Provide the new tool description within ``` blocks."""
+
+    current_tool_description = dspy.InputField(desc="The current description of the tool")
+    examples_with_feedback = dspy.InputField(desc="Examples showing tool usage decisions and feedback on correctness")
```
LakshyAAAgrawal: Since this is a DSPy signature, we can avoid including formatting specifiers like "```", which DSPy will handle. Look at some of the other custom instruction proposer signatures.

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +310,141 @@ def __call__(
...
+
+    improved_tool_description = dspy.OutputField(
+        desc="An improved description that helps with tool selection decisions"
+    )
+
+
+class SingleComponentToolProposer(dspy.Module):
```
LakshyAAAgrawal: The functionality is great, but I am wondering if there's a way to reduce indirection, introduce less classes, and perhaps directly make this a part of propose_new_texts, with max 1 level of indirection.

LakshyAAAgrawal:
Dear @Ju-usc,

This is a great PR. Thanks a lot! I have tried to be overly critical and made too many nits. Feel free to ignore if you disagree with something. Let me know if you'd like me to address anything!

Regarding the meta prompt, overall I think it looks great. However, I suggest that as you build the tutorial, you may find that the reflection prompt needs tweaking, or the content exposed in reflective_dataset for the tool may be lacking or need improvement. This is going to be an empirical exercise, which will guide what works in the reflection meta prompts. ! Looking forward to the tutorial on this too!

You may already have thoughts about what you'd like to show in the tutorial, but if not, you may consider building off (https://kargarisaac.medium.com/building-and-optimizing-multi-agent-rag-systems-with-dspy-and-gepa-2b88b5838ce2) by @kargarisaac.

Ju-usc:
Hi @LakshyAAAgrawal,

Thanks for the detailed review! I tried my best to address all the critical points, though I may have missed something or introduced new issues. Feel free to add any thoughts! I also have a few questions for you.

Implementation:
Built tool optimization by discovering tools through ReAct modules, creating reflective datasets that share full trajectories with each tool, and routing to a specialized tool proposer. Refactored the string-matching approach you flagged (replaced `"react" in name.lower()` with proper module traversal). Also streamlined tests and fixed the custom proposer routing to ensure independence.

Documentation:
Added comprehensive docs covering the tool-specific reflection prompt, implementation flow (Pass 1/Pass 2), usage examples with visualization, and design rationale in code comments. Restructured "When to Use" as a numbered list per your feedback and removed implementation details from user-facing sections.

Key design decision: For reflective dataset, we share/deepcopy the complete ReAct trajectory with each tool (not just tool-specific segments) because tools need full context to understand selection patterns and workflow dependencies. This trades token efficiency for richer optimization.

Future work - would love your thoughts:

For reflective dataset for tools, I documented two approaches to reduce token duplication (see code comments for details): joint optimization with ReAct instruction by a separate proposer, or extending the instruction proposer to include tools. The second seems more elegant. What's your take - should we pursue one of these, or is the current approach good enough?

Testing:
The existing tests cover the core functionality, but I'm not sure if they're robust/clean enough. Could you provide guidance on:
- What test scenarios should we add to make this production-ready?
- Any edge cases or failure modes I should be testing for?
- Test structure/organization improvements?

Ready for re-review!

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,301 @@ gepa = dspy.GEPA(
...
+This means:
+- A tool receives the FULL ReAct trajectory (thoughts, actions, observations) in the "Generated Outputs" field
+- The metric can optionally examine the trajectory and include tool-specific insights in the feedback text
+- The reflection LM sees complete context about how and when the tool was used
+
+**Component Identification & Proposer Routing:**
+
+GEPA discovers tools by traversing ReAct modules and extracting their associated `dspy.Tool` instances. Once identified, GEPA routes components to appropriate proposers:
+- **Signature instructions** → Custom instruction proposer (if provided) OR default GEPA proposer
+- **Tool descriptions** → Built-in `ToolProposer` (always used, not customizable)
```
LakshyAAAgrawal: This is customizable, right? If not, shouldn't we make this customizable too?

LakshyAAAgrawal:
> Key design decision: For reflective dataset, we share/deepcopy the complete ReAct trajectory with each tool (not just tool-specific segments) because tools need full context to understand selection patterns and workflow dependencies. This trades token efficiency for richer optimization.

This is great, and the right approach. I would just like to highlight that there may be one issue: In react, the trajectory often ends up containing repeated prefixes, i.e., let's say the actual trajectory is {system prompt, user message 1, assistant 1, tool 1, assistant 2, tool 2, assistant 3, end}, then the react trajectory will contain a list of 7 invocations, each containing prefix of size 1 through 7. We can avoid showing all prefixes, and just show the final trajectory. Could you please ensure we aren't running into that issue here? Relevant https://github.com/gepa-ai/gepa/issues/97

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,301 @@ gepa = dspy.GEPA(
...
+
+    Review the current description along with examples of the assistant's tool decisions 
+    and the feedback those decisions received.
+
+    Read them together and refine the description.
+    So the agent understands when this tool actually helps, what argument or result matters, 
+    and what misuse the feedback exposed. Keep the tool's voice and only change what the 
+    evidence justifies.
+
+    Return a refined description that helps the assistant quickly recognize good
```
LakshyAAAgrawal: Have you had an opportunity to test this prompt for any workload? I would be glad if you could run this prompt just once with a trajectory on any agentic task, and see the tool description it produces.

Ju-usc: Yes! i will share them in the tutorial!

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
Ju-usc: user can plug custom instruction proposer but it will override with tool proposer and default instruction proposer. i am not fully sure if it's good idea to expose two seperate custom instruction proposer for each instruction and tool. 

one approach i'm thinking is that to write an example in the doc how user can define their own custom proposer that optimize both tool description and module instruction

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
Ju-usc: also thank you for highlighting the issue!  ill definitely fix it for react's reflective dataset. seems like we can just extract the last trajectory index to avoid duplicates?

Ju-usc: also another edge case I'm concerning is reflection lm's max context window due to mass react's trajectory tracing. 

i'm not fully sure what the right approach for this. do you have any thoughts? 

simple trim?
gather only failing tool call tracings?
skip it? just like how current flow is?

LakshyAAAgrawal:
@Ju-usc, were you able to address the discussion at https://discordapp.com/channels/1161519468141355160/1426031284794425374/1430169815707811841

Ju-usc:
> @Ju-usc, were you able to address the discussion at https://discordapp.com/channels/1161519468141355160/1426031284794425374/1430169815707811841

yes. also a lot of improvement coming soon!

Ju-usc:
still working on nested agent test, doc refine, and tutorial

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,409 @@ gepa = dspy.GEPA(
     auto="medium"
 )
 ```
+
+## ReAct Component Optimization
+
+### What is optimize_react_components?
```
LakshyAAAgrawal: Does dspy.Tool not work with any other components than dspy.React?

LakshyAAAgrawal: I suppose what you have built is truly general. We should be easily able to extend this to issues like #8962 (not suggesting we need to do it right here in this PR, but just trying to future proof!)

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,409 @@ gepa = dspy.GEPA(
...
+    )
+
+    improved_react_instruction = dspy.OutputField(
+        desc="Improved ReAct module instruction"
+    )
+    improved_extract_instruction = dspy.OutputField(
+        desc="Improved Extract module instruction",
+        default=""
+    )
+    # Note: Tool descriptions and arg descriptions are added dynamically via signature.append()
```
LakshyAAAgrawal: Has your experience been that generating improved version of all components (2 prompts + N tool descriptions) at once has been good?

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
Ju-usc: I do agree that the right direction is to make the system extendable to other component types beyond dspy.ReAct and to handle cases like #8962.

GEPA should automatically detect all modules in a program and optimize them based on how each module is structured for example:
- a ReAct module → predictor, extractor, and tools
- other modules → their own configurable predictors or tool sets.

Conceptually, we can reduce everything to two fundamental abstractions that cover most applications:

1. Instruction — directive text that tells the model what to do (system prompt) 
example: signature docstrings (signature.instruction)

2. Description — descriptive text that tells the model what something is (data model metadata)
example: signature field descriptions, tool descriptions, argument descriptions

Do instruction and description seem like the right foundational abstractions for generalizing optimization across modules (beyond dspy.ReAct)?

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
Ju-usc: Did some experiments- here's the gist with baseline/optimized prompts:
https://gist.github.com/Ju-usc/0298732be01123486c831d8f210abd36

Used nested ReAct (lead agent → subagent → web search) with 10 BrowseComp examples. GEPA updated both modules + tool descriptions together and the optimized prompts look pretty interesting although it didn't improve much due to simple setup.

Would love to hear your thoughts or feedback!

LakshyAAAgrawal: Hey, this is amazing! I am also wondering if we need to make it configurable whether each parameter is evolved or not. Hopefully the tool description can itself contain a summary of the parameters and each parameter does not need to be very detailed multi para words?

tests/teleprompt/test_gepa_react_optimization.py
```diff
@@ -0,0 +1,849 @@
...
+"""
+
+import json
+
+import dspy
+from dspy import Example
+from dspy.utils.dummies import DummyLM
+
+
+def setup_spy_for_base_program(monkeypatch):
```
LakshyAAAgrawal: Did something weird with replacing "d" happen here? Maybe spy is dspy?

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +317,236 @@ def __call__(
...
+                logger.error(f"Failed to deserialize config for {module_key}: {e}")
+                continue
+
+            # Reconstruct Tool objects from serialized schema
+            current_tools_dict = current_react_config.get("tools", {})
+            logger.info(f"Found {len(current_tools_dict)} tools: {list(current_tools_dict.keys())}")
+            tools_list = []
+            for tool_name, tool_info in current_tools_dict.items():
+                tool = dspy.Tool(
+                    func=lambda: None,
```
LakshyAAAgrawal: Can the func not be None?

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +317,236 @@ def __call__(
...
+                current_extract_instruction=current_react_config.get("extract", ""),
+                current_tools=tools_list,  # List of Tool objects for adapter formatting
+                examples_with_feedback=formatted_examples,
+            )
+
+            # Build improved config
+            logger.info("Building improved config from LM response...")
+            improved_react_config = {}
+
+            # Add react instruction (always improved)
```
LakshyAAAgrawal: Can you explain "always improved" and "only if improved" below?

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
LakshyAAAgrawal: Yes, they do. But I think let's move this discussion over to #8962, and merge this PR and create a separate PR for it.

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,409 @@ gepa = dspy.GEPA(
...
+    tool_calls = [
+        trajectory[key] 
+        for key in trajectory 
+        if key.startswith('tool_name_') and trajectory[key] != 'finish'
+    ]
+    
+    if tool_calls:
+        all_tool_names = ', '.join(tool_calls)
+        num_calls = len(tool_calls)
+        feedback = f"{'Correct Answer' if correct else 'Wrong Answer'}. Used {num_calls} tool calls: {all_tool_names}. Try to minimize tool calls."
```
LakshyAAAgrawal: Why are we highlighting minimizing tool calls as a callout objective in the documentation?

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,409 @@ gepa = dspy.GEPA(
...
+   - Return `dict[str, str]` with same keys
+
+**What this means:**
+- Your custom proposer receives ALL components: regular signatures AND ReAct modules
+- GEPA still does discovery and JSON serialization, but YOU handle the optimization logic
+- ReAct components are passed with keys like `"react_module"` or `"react_module:agent_name"`
+
+#### Implementing a Custom Proposer for ReAct
+
+If you need custom logic, you must handle ReAct components yourself. ReAct components are stored as JSON strings containing all 4 parts:
```
LakshyAAAgrawal: Instead of saying this, we can say that you can start with the existing implementation at X.

LakshyAAAgrawal:
Hey @Ju-usc, this looks great to me. Left a few comments, otherwise happy to merge!

Can you address the ruff issues as well?

dspy/teleprompt/gepa/instruction_proposal.py
Ju-usc: func cannot be None because Tool requires func: Callable (not Optional[Callable]).
We use lambda: None as a placeholder since:
1. Tool objects are reconstructed from deserialized JSON (which doesn't contain the actual function)
2. We only need the tool schema (name, desc, args) for the reflection LM - the function is never executed

Ju-usc:
@LakshyAAAgrawal Thanks for the thorough review! Addressed all 6 comments:

1. **"spy" naming** → Renamed to `setup_capture_for_base_program` (82dee257)
2. **Tool.func placeholder** → Added explanatory comments (ca84b9d8)
3. **Optional fields** → Changed to `str | None = None` for selective optimization (2eb89865)
4. **Prevent verbose outputs** → Refined reflection prompt with functional descriptions (bd4cdac4)
5. **Remove "minimize tool calls"** → Rewrote metric example to be general/extensible (0ad40776)
6. **Reference implementation** → Replaced inline example with link to `ReActModuleProposer` (ef5563e6, 1b10b65a)
     
Let me know if you have any other thoughts to move this PR forward!

Ju-usc:
@LakshyAAAgrawal I have also updated the PR description to reflect all the changes. Ready for rereview!

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
Ju-usc: I do agree the reflection LM can struggle to generate minimal yet effective descriptions in the right places. 

To address this, I refined the reflection prompt to guide appropriate granularity - field descriptions now provide functional context ("purpose of tool", "usage of parameter") and note progressive optimization. 

I think some applications might gain benefits from detailed descriptions (smaller models, complex tools), so this lets the LM decide based on what actually fixes failures rather than being prescriptive.

What's your thoughts?

Ju-usc: On configurability: The current design makes all components optional from the reflection LM's perspective (returns `None` to skip any component), so we rely on the reflection LM to determine what needs optimization based on trajectories and feedback. I wasn't sure how to cleanly expose per-parameter config flags as API parameters without making the interface messy. As a workaround, users could guide this through feedback strings (e.g., "don't optimize tool arg descriptions unless critical") though this is less reliable.

LakshyAAAgrawal:
When GEPA is invoked on a DSPy program which has ReAct module, but does not have "optimize_react=True" set, print a warning saying that "For ReAct programs, consider using `optimize_react=True` which can optimize ..."

Ju-usc:
@LakshyAAAgrawal accidentally hit "Close" while working on the warning feature. Reopened right away!

Implemented the warning. Ready for re-review when you have a chance!

LakshyAAAgrawal (approved):

tests/teleprompt/test_gepa_react_optimization.py
```diff
@@ -0,0 +1,849 @@
...
+    from dspy.teleprompt.gepa.gepa_utils import REACT_MODULE_PREFIX
+
+    captured_base_program = setup_capture_for_base_program(monkeypatch)
+    program = create_single_react_program()
+
+    optimizer, trainset = create_gepa_optimizer_for_detection()
+
+    try:
+        optimizer.compile(program, trainset=trainset, valset=trainset)
+    except Exception:
```
LakshyAAAgrawal: Do we expect any exception to actyally be raised here? If yes, we should catch exactly that exception.

tests/teleprompt/test_gepa_react_optimization.py
```diff
@@ -0,0 +1,849 @@
...
+    from dspy.teleprompt.gepa.gepa_utils import REACT_MODULE_PREFIX
+
+    captured_base_program = setup_capture_for_base_program(monkeypatch)
+    program = create_multi_react_workflow_program()
+
+    optimizer, trainset = create_gepa_optimizer_for_detection()
+
+    try:
+        optimizer.compile(program, trainset=trainset, valset=trainset)
+    except Exception:
```
LakshyAAAgrawal: Similarly here

tests/teleprompt/test_gepa_react_optimization.py
```diff
@@ -0,0 +1,849 @@
...
+
+    captured_base_program = setup_capture_for_base_program(monkeypatch)
+    program = create_orchestrator_with_workers_program()
+
+    optimizer, trainset = create_gepa_optimizer_for_detection()
+
+    try:
+        optimizer.compile(program, trainset=trainset, valset=trainset)
+    except Exception:
+        pass
```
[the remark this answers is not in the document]
LakshyAAAgrawal: This too?

LakshyAAAgrawal:
Hi @Ju-usc,

Echoing some discussion with @chenmoneygithub here.

One question we have is how tied is this to ReAct modules? Let's say I have a custom dspy Module, that is not a dspy.ReAct, but uses dspy.Tool, will these changes still be applicable? If yes, it makes sense to change `optimize_react_components=True` to `optimize_tool_descriptions=True`. If it does not, are there things specific to ReAct and these changes, that cannot be applied to other modules? Is there some way we could generalize there?

dspy/teleprompt/gepa/gepa.py
```diff
@@ -518,11 +533,65 @@ def feedback_fn(
             rng=rng,
             reflection_lm=self.reflection_lm,
             custom_instruction_proposer=self.custom_instruction_proposer,
-            warn_on_score_mismatch=self.warn_on_score_mismatch
+            warn_on_score_mismatch=self.warn_on_score_mismatch,
+            optimize_react_components=self.optimize_react_components,
```
chenmoneygithub: Instead of engineering this towards `dspy.ReAct`, I recommend covering tool calling in general: `optimize_react_components` => `optimize_tools`. ReAct is just one way for tool calling agent, and it's quite common for users to make customizations, and we may create other tool calling agent architectures in the near future.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -36,18 +44,18 @@ def __call__(
         - gold: The gold example.
         - pred: The predicted output.
         - trace: Optional. The trace of the program's execution.
-        - pred_name: Optional. The name of the target predictor currently being optimized by GEPA, for which 
+        - pred_name: Optional. The name of the target predictor currently being optimized by GEPA, for which
```
chenmoneygithub, line 48: Let's not mix style/lint fixes with actual code change in the same commit for clean commit history.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -518,11 +533,65 @@ def feedback_fn(
...
+
+            if self.optimize_react_components:
+                normalized_path = module_path.removeprefix("self.") if module_path != "self" else ""
+
+                # Get first predictor name as module identifier
+                for pred_name, _ in module.named_predictors():
+                    comp_name = pred_name if not normalized_path else f"{normalized_path}.{pred_name}"
+                    # Use full normalized path to avoid collapsing nested modules
+                    # e.g., "multi_agent.coordinator" not "multi_agent"
+                    module_key = f"{REACT_MODULE_PREFIX}:{normalized_path}" if normalized_path else REACT_MODULE_PREFIX
```
chenmoneygithub: I wonder if we need this custom name change - are you seeing duplicated keys without this code? could you share an example? If so, that's a bug we should fix on the module level.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -518,11 +533,65 @@ def feedback_fn(
...
+                    f"Detected ReAct module at '{module_path}'. Consider using "
+                    "`optimize_react_components=True` to jointly optimize react instructions, "
+                    "extract instructions, tool descriptions, and tool argument descriptions."
+                )
+
+        # Log base_program keys for debugging
+        logger.info(f"Initialized base_program with {len(base_program)} components:")
+        for key in sorted(base_program.keys()):
+            if key.startswith(REACT_MODULE_PREFIX):
+                logger.info(f"  {key}: <ReAct module JSON config>")
```
chenmoneygithub: I would get rid of these logs. These are internal implementation details, so users won't be able to follow

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -31,15 +39,17 @@ class ReflectiveExample(TypedDict):
 
     Each example contains the predictor inputs, generated outputs, and feedback from evaluation.
     """
-    Inputs: dict[str, Any]                              # Predictor inputs (may include str, dspy.Image, etc.)
-    Generated_Outputs: dict[str, Any] | str             # Success: dict with output fields, Failure: error message string
-    Feedback: str                                       # Always a string - from metric function or parsing error message
+
+    Inputs: dict[str, Any]  # Predictor inputs (may include str, dspy.Image, etc.)
```
chenmoneygithub, line 44: nit: separate style change to a separate PR.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -88,42 +100,187 @@ def __init__(
...
-
-        if self.custom_instruction_proposer is not None:
-            # We are only overriding the propose_new_texts method when a custom
-            # instruction proposer is provided. Otherwise, we use the GEPA
-            # default propose_new_texts.
-
-            def custom_propose_new_texts(
+        self.optimize_react_components = optimize_react_components
+
+        def build_propose_new_texts():
```
chenmoneygithub: This method is huge now, let's move the helper function out as private method.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -88,42 +100,187 @@ def __init__(
...
+                            },
+                        )["new_instruction"]
+                    return updated_components
+
+                instruction_proposer = default_instruction_proposer
+
+            # Init ReAct module proposer if tool optimization is enabled
+            react_module_proposer = None
+            if self.optimize_react_components:
+                from .instruction_proposal import ReActModuleProposer
```
chenmoneygithub: nit: always use absolute imports

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -88,42 +100,187 @@ def __init__(
...
+                instruction_components = [c for c in components_to_update if not c.startswith(REACT_MODULE_PREFIX)]
+
+                results: dict[str, str] = {}
+
+                # Handle regular instruction components
+                logger.debug(f"Routing {len(instruction_components)} instruction components to instruction_proposer")
                 if self.reflection_lm is not None:
                     with dspy.context(lm=self.reflection_lm):
-                        return self.custom_instruction_proposer(
+                        results.update(
```
chenmoneygithub: you can merge the if and else branch by something like:

```
with dspy.context(lm=self.reflection_lm or dspy.settings.lm):
```

Ju-usc:
@LakshyAAAgrawal @chenmoneygithub  Thank you so much for the awesome detailed feedback! I'll address all the comments soon. Love to polish this in the right direction!

Before to refactor architecture, i have a quick clarification on custom tool usage:
 
For custom modules using dspy.Tool (not dspy.ReAct), is it safe to assume they use dspy.ToolCalls as output from the predictor to handle tool calls? This would help determine a clean approach for generalizing the optimization, especially for constructing shared tracing for reflective dataset to jointly optimize predictor instruction that uses tools and tools descriptions.

chenmoneygithub:
@Ju-usc  Sorry about the late reply! 

> is it safe to assume they use dspy.ToolCalls as output from the predictor to handle tool calls? 

Unfortunately we cannot make this assumption. Help me understand - why do we need to care about the output type? I thought we only need to expose the Tool as GEPA candidate, and have GEPA propose new rules for Tool based on the feedback.

dspy/teleprompt/gepa/gepa.py
Ju-usc: Yes, we're seeing collisions without it. Here's an example:

```
class MultiAgent(dspy.Module):
def __init__(self):
self.coordinator = dspy.ReAct("task -> plan", tools=[...])
self.executor = dspy.ReAct("plan -> result", tools=[...])
```
The issue: Both ReAct modules have internal predictors with generic names (react, extract.predict). When we try to create module-level keys to group each ReAct's components together for joint optimization, we get:

Without normalized path:

```
     # First ReAct module (coordinator)
     base_program["react_module"] = {
         "react": "coordinate tasks...",
         "extract": "extract plan...",
         "tools": [...]
     }

     # Second ReAct module (executor) - OVERWRITES!
     base_program["react_module"] = {
         "react": "execute plans...",
         "extract": "extract results...",
         "tools": [...]
     }
     # Coordinator is lost!
```

With normalized path:

```
     # First ReAct module (coordinator)
     base_program["react_module:coordinator"] = {
         "react": "coordinate tasks...",
         "extract": "extract plan...",
         "tools": [...]
     }

     # Second ReAct module (executor) - separate key
     base_program["react_module:executor"] = {
         "react": "execute plans...",
         "extract": "extract results...",
         "tools": [...]
     }
     # Both preserved!
```
We need one key per ReAct module (not per predictor) because we jointly optimize react instructions + extract instructions + tool descriptions as a unit. For generic tool predictors, we don't need this - named_predictors() already gives unique names like agent1.predictor, agent2.predictor.

dspy/teleprompt/gepa/gepa.py
Ju-usc: I agree I'll address this!

dspy/teleprompt/gepa/gepa_utils.py
Ju-usc: Good point I'll address this!

Ju-usc:
> @Ju-usc Sorry about the late reply!
> 
> > is it safe to assume they use dspy.ToolCalls as output from the predictor to handle tool calls?
> 
> Unfortunately we cannot make this assumption. Help me understand - why do we need to care about the output type? I thought we only need to expose the Tool as GEPA candidate, and have GEPA propose new rules for Tool based on the feedback.

@chenmoneygithub Sorry about the confusion - I was overthinking and made a wrong assumption about needing ToolCalls output. Let me clarify the actual approach:

Detection (compile-time):
- Check if predictor signature has dspy.Tool input field (via type annotation)
- Could be single tool, list, or dict
- Initialize with empty tools list (lazy init) since we don't know which specific tools will be used at runtime

Tool Discovery (runtime via traces):
- Extract actual tools from predictor's trace inputs
- This handles edge cases:
  - Tools declared globally
  - Tools dynamically created
  - Tools conditionally passed based on runtime logic
- We only optimize tools that were actually used, not all declared tools

Why trace-based extraction matters:
- Predictor and its tools share the same trace (same execution context)
- We see exactly which tools were passed for each example
- Reflection LM gets real usage patterns, not just declarations

Update:
- Traverse new_prog to find Tool objects by name
- Update descriptions in-place
- Works regardless of where tools are stored (attribute name doesn't matter)

This is different from ReAct which knows its exact tool list at init time. For generic tool modules, we discover tools lazily from actual usage. Let me know if you have any thoughts/feedbacks about this!

Ju-usc:
@LakshyAAAgrawal @chenmoneygithub 
Question about ReAct handling:

Should we preserve the ReAct-specific path or adopt a fully generic approach? Both options use a single flag (optimize_tools).

Option 1 (preserve ReAct special handling):
- ReAct modules: jointly optimize react + extract + tools via existing ReActModuleProposer
- Generic tool modules: jointly optimize predictor + tools via new signature (similar pattern to ReActModuleProposer but with "predictor" instead of "react"/"extract")
- Benefit: Preserves existing ReAct joint optimization behavior that optimizes all components at once with the same tracing.
- Tradeoff: Two code paths under one flag, need new proposer signature for generic tools

Option 2 (fully generic, treat ReAct like any other module):
- ReAct's react predictor: optimized with tools (joint, detected as generic tool predictor)
- ReAct's extract predictor: optimized separately via default instruction proposer
- Generic tool modules: predictor + tools (joint)
- Benefit: Simpler, uniform architecture, one code path
- Tradeoff: Loses react+extract joint optimization

I'm leaning toward Option 1 to preserve the existing optimization quality for ReAct while extending support to generic tool modules for custom dspy.Tool programs. Love to hear feedbacks!

dspy/teleprompt/gepa/gepa_utils.py
Ju-usc: Thank you! I'll address this!

LakshyAAAgrawal:
@Ju-usc, if Option 2 will be detrimental to optimization quality, I think we should preserve option 1, i.e., allow for full generality, but leverage additional information from dspy.ReAct when available.

But the real question is that is there a way to achieve the same optimization quality even with option 2, the generalization.

dspy/teleprompt/gepa/gepa.py
Ju-usc: yes I'll refactor it without custom name change

chenmoneygithub:
@Ju-usc  @LakshyAAAgrawal  From my perspective the quality should be on par if we capture the tool trace correctly, because ReAct itself is just a way of interacting with tools. But definitely need to run experiments to find it out.

There is one prerequisite we need to build - capture tool usage in `dspy.settings.trace`. I see two viable options here:

1. Use the callback to inject the tool trace into the dspy.settings.trace, which is only enabled when running GPEA + `optimize_tool=True`.
2. Explicitly add tool calling inputs/outputs to `dspy.settings.trace`

I like 1 better because that doesn;t affect existing behavior. In general, I would like to avoid shipping an optimization algorithm that has many special logic for `dspy.ReAct`, which will create a maintenance blackhole for us in the long run. For example, it is subject to broken without awareness when ReAct code is changed. 

@Ju-usc This work is very challenging, but super cool. We really appreciate your contribution here.

Ju-usc:
@LakshyAAAgrawal @chenmoneygithub 

Thanks again for the thoughtful feedback — I've pushed toward as generic as safely possible while preserving ReAct behavior.

**Core idea:** Optimize tools jointly with the predictor that uses them.

For generic tool modules, I detect predictors with `dspy.Tool` in their signature at compile time, then discover actual tools from traces at runtime. For ReAct, we know tools statically from `module.tools`. Both share the same `ToolModuleProposer` and update path — the only difference is when tools are discovered.

**Here are my thoughts on a few design choices. Feel free to comment on these or anything else:**

1. **Prefix-based grouping** (`react_module:` / `tool_module:`) — I encode module type in string keys to preserve GEPA's `dict[str, str]` interface. A bit hacky, but avoids changing the core adapter API. 

2. **ReAct tracing uses `extract` only** — The `extract` predictor's trace already contains the full trajectory, so this avoids the duplicated-prefix issue from gepa-ai/gepa#97.

3. **No separate tool tracing yet** — Reflective dataset uses predictor-level traces only. Tool inputs are always captured (since `dspy.Tool` objects flow through predictor inputs), but tool outputs depend on how users design their module. If users need tool outputs in the reflective dataset, they can wire them back into a predictor. Is this enough for now, or should tool calls go into `dspy.settings.trace`?

4. **ReAct still has special handling** — Because it has 2 predictors (react + extract) that need joint optimization. Is this acceptable given the maintenance blackhole concern, or should I push for fully generic?

Ran an experiment with nested ReAct + custom tool module: https://gist.github.com/Ju-usc/80b9918fe07288204579df735e084cb4

Happy to iterate!

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +314,195 @@ def __call__(
...
+                if improved_tool_desc is not None:
+                    tool_info["desc"] = improved_tool_desc
+
+                # Update arg descriptions if LM proposed changes
+                for arg_name in tool_info["args"].keys():
+                    improved_tool_arg_desc = getattr(result, f"improved_tool_{tool_name}_arg_{arg_name}_desc", None)
+                    if improved_tool_arg_desc is not None:
+                        tool_info["args"][arg_name]["description"] = improved_tool_arg_desc
+
+                improved_module_config["tools"][tool_name] = tool_info
```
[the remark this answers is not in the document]
Ju-usc: I don't think this is an issue - `candidate[module_key]` is a json string, so `json.loads()` creates a new dict. Mutations don't affect the original.

dspy/teleprompt/gepa/instruction_proposal.py
```diff
@@ -310,3 +314,195 @@ def __call__(
...
+        updated_components = {}
+
+        for module_key in components_to_update:
+            if module_key not in candidate or module_key not in reflective_dataset:
+                logger.debug(f"Skipping {module_key}: not in candidate={module_key not in candidate}, not in reflective_dataset={module_key not in reflective_dataset}")
+                continue
+            current_module_config = json.loads(candidate[module_key])
+
+            # Predictor keys: 1 for tool modules, 2 for ReAct modules (extra extract predictor)
+            predictor_keys = [k for k, v in current_module_config.items() if isinstance(v, str)]
```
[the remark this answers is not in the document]
Ju-usc: I don't think this is needed - config is built internally by GEPA and always contains predictor keys. Edge case seems impossible.

docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
```diff
@@ -443,3 +443,306 @@ gepa = dspy.GEPA(
...
+#### ReAct Agent
+
+```python
+import dspy
+
+def search_web(query: str) -> str:
+    return f"Search results for: {query}"
+
+def calculate(expression: str) -> float:
+    return eval(expression)
```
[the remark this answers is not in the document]
Ju-usc: fixed in 09990a6f - replaced calculator/eval with get_weather tool.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -89,43 +98,173 @@ def __init__(
         self.reflection_lm = reflection_lm
         self.custom_instruction_proposer = custom_instruction_proposer
         self.warn_on_score_mismatch = warn_on_score_mismatch
+        self.enable_tool_optimization = enable_tool_optimization
+
+        self.propose_new_texts = self._build_propose_new_texts()
         self.reflection_minibatch_size = reflection_minibatch_size
 
+    def _build_propose_new_texts(self):
```
chenmoneygithub: This method is huge, and it is doing two things:

- Configure the proposer for predicts and tools
- Create a function that generates the proposal

We don't necessarily need to create `propose_component_texts` on the fly, so I recommend separating `propose_component_texts` out to be a private method.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -89,43 +98,173 @@ def __init__(
...
+                    )["new_instruction"]
+                return updated_components
+
+            instruction_proposer = default_instruction_proposer
+
+        # Init tool module proposer if tool optimization is enabled
+        tool_module_proposer = None
+        if self.enable_tool_optimization:
+            from dspy.teleprompt.gepa.instruction_proposal import ToolModuleProposer
+            tool_module_proposer = ToolModuleProposer()
```
chenmoneygithub: `module` literal here is confusing, maybe either `tool_proposer` or `tool_desc_proposer`

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -89,43 +98,173 @@ def __init__(
...
-        self.named_predictors = list(self.student.named_predictors())
+            return results
 
+        return propose_component_texts
 
     def build_program(self, candidate: dict[str, str]):
         new_prog = self.student.deepcopy()
+
+        # Start with plain string instructions from candidate
+        improved_predictors = {
```
chenmoneygithub: maybe call this `predict_candidates`? `improved` suggests that this has already been improved.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -89,43 +98,173 @@ def __init__(
...
+
+        # Update predictor instructions
         for name, pred in new_prog.named_predictors():
-            if name in candidate:
-                pred.signature = pred.signature.with_instructions(candidate[name])
+            if name in improved_predictors:
+                pred.signature = pred.signature.with_instructions(improved_predictors[name])
+
+        # Update tool descriptions
+        if improved_tools:
```
chenmoneygithub: nit: Make this a standalone method to reduce nesting.

dspy/teleprompt/gepa/gepa.py
```diff
@@ -546,11 +563,102 @@ def feedback_fn(
...
+                    }
+                }
+
+                base_program[module_key] = json.dumps(config, indent=2)
+        else:
+            # Warn if ReAct modules found but tool optimization disabled
+            for module_path, module in student.named_sub_modules():
+                if isinstance(module, ReAct):
+                    logger.warning(
+                        f"Detected ReAct module at '{module_path}'. Consider using "
```
chenmoneygithub: This should be a info instead warning IMO, unless we can articulate that involving tools in the optimization works strictly better than not involving tools.

dspy/teleprompt/gepa/gepa_utils.py
```diff
@@ -283,15 +441,97 @@ def make_reflective_dataset(self, candidate, eval_batch, components_to_update) -
...
                 continue
+
             ret_d[pred_name] = items
 
         if len(ret_d) == 0:
             raise Exception("No valid predictions found for any module.")
 
         return ret_d
 
+    def _update_candidate_tools(self, candidate, program, trajectories) -> None:
```
chenmoneygithub: could you help me understand the purpose of this method?

From my rough understanding, you are trying to link the tool to the predict's trace where the tool is getting used. Is that correct?

Ju-usc: Yes that is correct, the idea was to scan traces for predictors with tool inputs and extract them for optimization. 

Unlike ReAct where we can directly access module.tools, generic dspy.Tool doesn't easily expose a predictor→tool mapping. So the approach was to find Tool objects in the trace inputs, then update the candidate dict with tool configs - since that's the interface defining what gets optimized.

dspy/teleprompt/gepa/gepa_utils.py
Ju-usc: thank you!

dspy/teleprompt/gepa/gepa_utils.py
Ju-usc: yes just refactored!

dspy/teleprompt/gepa/gepa.py
Ju-usc: addressed!

Ju-usc:
> Thank you very much for the update!
> 
> The direction looks great to me, but the implementation, especially how we extract tool trace from the DSPy traces is a bit fragile. This is not directly caused by the PR, but in DSPy we are lacking this important lineage. As of now, we can probably do the following:
> 
> * Replace the generic tool capture + optimization code by a TODO.
> * Still protect ReAct-specific logic in an if branch.
> * Ship the ReAct code + tutorial.
> 
> This is still more generic than the last version, and leave a decent room for us to maintain and scale to generic tool optimization. Meanwhile, we will work on a robust way to connect tool calls to predict traces, after which we can revisit the generic tool optimization.

Thanks for all the suggestions! I've addressed everything and the PR is ready for review.

**Summary of changes:**
- `logger.warning` → `logger.info` for ReAct detection
- Extracted `_propose_component_texts` as private method  
- TODO'd out generic tool optimization (keeping ReAct-only for now)
- Removed generic tool module detection, kept ReAct only
- Naming: `improved_predictors` → `predictor_candidates`, `improved_tools` → `tool_candidates`
- Extracted `_update_tool_descriptions` and `_collect_tools` methods
- Renamed `ToolModuleProposer` → `ToolProposer`
- Unified prefix: `REACT_MODULE_PREFIX` → `TOOL_MODULE_PREFIX`
- Updated all docs and tests for ReAct-only support

Let me know if anything else needs to be addressed!

chenmoneygithub:
@Ju-usc  Thank you so much for the persistent work, and great PR! I will handle the rest of the work and play a bit more with it then merge.

Ju-usc:
@LakshyAAAgrawal @chenmoneygithub This wouldn't have been possible without all your guidance and feedback! Thank you so much. Really happy to make my first open source contribution and learned a lot!

LakshyAAAgrawal:
Dear @Ju-usc,

I would just like to say that this is a very very non-trivial contribution, and we all sincerely thank you for going through so many iterations to perfect the PR. Thanks a lot to @chenmoneygithub as well, for kindly offering his valuable advise to make this addition truly generalizable.

PR #8929, opened 2025-10-10
Added reference documentation

onel:
First batch of files with adding reference documentation to some files.
This is a test run to see if the style + content is helpful.

It follows the google docstring style but if the methods have types added it does not also add the `Args:` `Returns:` because that would be duplicate information. And also at risk of becoming out of sync.

Feedback more than welcomed.

#8930


onel:
Good idea. I reverted the other 2 files.

So we should do 1 file per PR? Or were the other 2 files not relevant?

PR #8955, opened 2025-10-20
refactor: rename output field from rationale to reasoning in MultiChainComparison

hellocybernetics:
# Summary
Renames the output field rationale → reasoning for consistency with Chain-of-Thought (CoT) terminology.
No behavioral change to parsing logic; we continue to accept both keys during extraction to remain backward-compatible.

# Changes
MultiChainComparison: output field renamed to reasoning.
Tests updated to use reasoning.
Extraction keeps accepting legacy rationale:
`rationale = c.get("rationale", c.get("reasoning")).strip().split("\n")[0].strip()`
(unchanged logic; still prefers rationale if present, otherwise falls back to reasoning)

# Motivation
Align naming with CoT and existing components that already use reasoning.
Reduce confusion between fields with the same semantic meaning.

#Backward Compatibility / Risk
Low. Callers reading only reasoning are supported; callers still sending rationale remain compatible via the fallback extraction.
External consumers relying strictly on the old key name may need to migrate, but the parser remains tolerant.

### DIFF
--- a/tests/predict/test_multi_chain_comparison.py
+++ b/tests/predict/test_multi_chain_comparison.py
@@ -32,7 +32,7 @@ def test_basic_example():
 
     # Call the MultiChainComparison on the completions
     question = "What is the color of the sky?"
-    lm = DummyLM([{"rationale": "my rationale", "answer": "blue"}])
+    lm = DummyLM([{"reasoning": "my rationale", "answer": "blue"}])
     dspy.settings.configure(lm=lm)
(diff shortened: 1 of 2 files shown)


chenmoneygithub:
@hellocybernetics  Thanks for the PR! 

> External consumers relying strictly on the old key name may need to migrate, but the parser remains tolerant.

This is not low risk, basically we should avoid changing any public facing arguments/field unless it's broken, or we see significant benefits from the breaking change.

PR #8964, opened 2025-10-24
[BAML Adapter] Skip fields that are explicitly excluded from the schema

yasyf:
Excluding a field from the JSON Schema manually normally implies you'd like the LLM to not be aware of it. This is useful in cases where a field is auto-populated and would be confusing to present to the LLM.


yasyf:
cc @okhat @prrao87 @TomeHirata since you have touched this file before :)

PR #8969, opened 2025-10-25
Update gepa[dspy] dependency version to 0.0.18

LakshyAAAgrawal:


LakshyAAAgrawal:
Dear @TomeHirata could you please take a look at the failing test which seems related to mlflow <> gepa integration.

TomeHirata:
Updated the test!

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
     "pillow>=10.1.0",
     "numpy>=1.26.0",
     "xxhash>=3.5.0",
-    "gepa[dspy]==0.0.17",
+    "gepa[dspy]==0.0.18",
```
TomeHirata, line 46: @LakshyAAAgrawal does this need to be [dspy]? It seems the dspy optional dependency group does not have any packages.

LakshyAAAgrawal: Yes, GEPA by default pulls in some dependencies, which may conflict with DSPy. Which is why I have created a dependency free version of GEPA for dspy under gepa[dspy]

TomeHirata: @LakshyAAAgrawal It seems gepa doesn't have any core dependencies though?
What package do you think is installed by default?
https://github.com/gepa-ai/gepa/blob/main/pyproject.toml#L20

LakshyAAAgrawal: That is true, but the idea is that even if gepa introduces any default dependencies in the future, it should have a separate dependency list so that it never introduces any conflict for dspy.

TomeHirata: Iirc, default depencies are installed even if you add [dspy]. So it won't serve for the purpose I think.

PR #8983, opened 2025-10-27
Replace SHA256 with xxhash for cache key generation

f0ster:
## Replace SHA256 with xxhash for cache key generation

Switched from SHA256 to xxhash64 for cache key hashing. SHA256 is cryptographically secure but overkill for cache keys where we just need fast, collision-resistant hashing.

### Performance
- **2.4x faster** cache key generation (0.50µs → 0.21µs)
- **59% time reduction** on every cache operation

### Changes
- `dspy/clients/cache.py`: Replace `sha256` with `xxhash.xxh64`
- `tests/clients/test_cache.py`: Update hash length test, add determinism check
- `tests/benchmarks/test_cache_key_performance.py`: Benchmark suite with before/after comparison

### Testing
- ✅ All 12 existing cache tests pass
- ✅ Cache keys remain deterministic
- ✅ 64-bit hash space sufficient for realistic workloads
- ✅ No behavioral changes

xxhash is already a dependency, so this is a pure performance win with zero additional overhead.

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -3,7 +3,7 @@ import inspect
 import logging
 import threading
 from functools import wraps
-from hashlib import sha256
+import xxhash
 from typing import Any
 
 import cloudpickle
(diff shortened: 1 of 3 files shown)


chenmoneygithub (commented):
Thanks for the PR! 

Changing hashing algorithm will invalidate users' local cache though, so we shouldn't touch it unless something is broken. Btw it's true that scale-wise the time cost reduction is significant, but the absolute time cost is very small for cache key computation. 

Thank again for the contribution, really appreciate it!

PR #8984, opened 2025-10-27
Add Disable Fallback Option in ChatAdapter

Ziems:
Useful for RL where you don't want any retries

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -1,4 +1,3 @@
-import logging
 import re
 import textwrap
 from typing import Any, NamedTuple
@@ -18,8 +17,6 @@ from dspy.clients.lm import LM
 from dspy.signatures.signature import Signature
 from dspy.utils.exceptions import AdapterParseError
 
-logger = logging.getLogger(__name__)
-
 field_header_pattern = re.compile(r"\[\[ ## (\w+) ## \]\]")
 
 
@@ -34,14 +31,14 @@ class ChatAdapter(Adapter):
         callbacks=None,
         use_native_function_calling: bool = False,
         native_response_types=None,
-        disable_adapter_fallback: bool = False,
+        use_json_adapter_fallback: bool = True,
     ):
         super().__init__(
             callbacks=callbacks,
             use_native_function_calling=use_native_function_calling,
             native_response_types=native_response_types,
         )
-        self.disable_adapter_fallback = disable_adapter_fallback
+        self.use_json_adapter_fallback = use_json_adapter_fallback
 
     def __call__(
         self,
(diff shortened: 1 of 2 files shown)


dspy/adapters/chat_adapter.py
```diff
@@ -40,10 +57,15 @@ def __call__(
             # fallback to JSONAdapter
             from dspy.adapters.json_adapter import JSONAdapter
 
-            if isinstance(e, ContextWindowExceededError) or isinstance(self, JSONAdapter):
+            if (
+                isinstance(e, ContextWindowExceededError)
+                or isinstance(self, JSONAdapter)
+                or self.disable_adapter_fallback
+            ):
+                logger.info(f"Skipping json fallback for adapter {self.__class__.__name__}")
```
chenmoneygithub: I think we can skip this. If we need some logging it should be the reverse case - we inform users that json is being retried because ChatAdapter fails, but in general I this can be hidden from the user.

dspy/adapters/chat_adapter.py
```diff
@@ -40,10 +57,15 @@ def __call__(
...
                 # On context window exceeded error or already using JSONAdapter, we don't want to retry with a different
```
chenmoneygithub: need to update the comment, also we may want to keep `raise e` to avoid breakage for users.

dspy/adapters/chat_adapter.py
```diff
@@ -60,10 +82,14 @@ async def acall(
...
-            if isinstance(e, ContextWindowExceededError) or isinstance(self, JSONAdapter):
+            if (
+                isinstance(e, ContextWindowExceededError)
+                or isinstance(self, JSONAdapter)
+                or self.disable_adapter_fallback
+            ):
                 # On context window exceeded error or already using JSONAdapter, we don't want to retry with a different
                 # adapter.
-                raise e
+                raise
```
chenmoneygithub: same here - update comment and keep the exception.

dspy/adapters/chat_adapter.py
```diff
@@ -26,6 +29,20 @@ class FieldInfoWithName(NamedTuple):
 
 
 class ChatAdapter(Adapter):
+    def __init__(
+        self,
+        callbacks=None,
+        use_native_function_calling: bool = False,
+        native_response_types=None,
+        disable_adapter_fallback: bool = False,
```
chenmoneygithub: we can explicitly list `json_adapter` here: `disable_json_adapter_fallback`. Or we can remove the negation form and call it `use_json_adapter_fallback`, and defaults to True.

Ziems:
Addressed the comments! Let me know what you think!

PR #8986, opened 2025-10-28
Introduce dspy.Reasoning to capture native reasoning from reasoning models

chenmoneygithub:
The reasoning models' response contains the reasoning content, and this PR introduce `dspy.Reasoning` so that the module  output captures the native reasoning, instead of letting LM regenerate one.

Usage:

```
import dspy

dspy.configure(lm=dspy.LM("anthropic/claude-3-7-sonnet-20250219", cache=False))

    reasoning: dspy.Reasoning = dspy.OutputField()


print(dspy.Predict(MySignature)(question="why did a chicken cross the kitchen?"))
```

For non-reasoning model, the above code will treat `dspy.Reasoning` field as a string field, which prompts the model to explicitly generate the reasoning.

There is a caveat of GPT-5 family models though - when using litellm chat completion + GPT-5 family models, the response doesn't contain the reasoning content. So temporarily we are treating GPT-5 family models as non-reasoning models.

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -192,7 +192,8 @@ def get_annotation_name(annotation):
     args = get_args(annotation)
     if origin is None:
         if annotation is Reasoning:
-            # Reasoning field type is treated as a string.
+            # Keep backward compatibility with the old behavior in `dspy.ChainOfThought`, where reasoning
+            # field type is treated as a string.
             return "str"
         if hasattr(annotation, "__name__"):
             return annotation.__name__


dspy/streaming/streaming_listener.py
```diff
@@ -147,6 +140,13 @@ def receive(self, chunk: ModelResponseStream):
                     is_last_chunk=self.stream_end,
                 )
 
+        try:
```
TomeHirata, line 148: nit: maybe add comment why this logic should come after native response handling?

dspy/adapters/utils.py
```diff
@@ -190,6 +191,9 @@ def get_annotation_name(annotation):
     origin = get_origin(annotation)
     args = get_args(annotation)
     if origin is None:
+        if annotation is Reasoning:
```
TomeHirata, line 194: Is there any way to implement the conversion more generically? Ideally this information should reside in Reasoning.

dspy/adapters/utils.py
```diff
@@ -84,7 +85,7 @@ def move_type_to_front(d):
 def translate_field_type(field_name, field_info):
     field_type = field_info.annotation
 
-    if get_dspy_field_type(field_info) == "input" or field_type is str:
+    if get_dspy_field_type(field_info) == "input" or field_type is str or field_type is Reasoning:
```
TomeHirata, line 88: ditto

dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,115 @@
...
+        elif "reasoning_effort" in lm.kwargs:
+            reasoning_effort = lm.kwargs["reasoning_effort"]
+
+        if reasoning_effort is None or not litellm.supports_reasoning(lm.model):
+            # If users explicitly set `reasoning_effort` to None or the LM doesn't support reasoning, we don't enable
+            # native reasoning.
+            return signature
+
+        if "gpt-5" in lm.model and lm.model_type == "chat":
+            # There is a caveat of Litellm as 1.79.0 that when using the chat completion API on GPT-5 family models,
```
TomeHirata, line 70: Is there any github issue for this on LiteLLM?

dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @classmethod
+    def adapt_to_native_lm_feature(
+        cls,
+        signature: type["Signature"],
+        field_name: str,
+        lm: "LM",
+        lm_kwargs: dict[str, Any],
+    ) -> type["Signature"]:
+        reasoning_effort = "unspecified"
```
TomeHirata: Is `unspecified` a valid reasoning effort?

dspy/adapters/types/reasoning.py
```diff
@@ -0,0 +1,115 @@
...
+        Parse a stream chunk into reasoning content if available.
+
+        Args:
+            chunk: A stream chunk from the LM.
+
+        Returns:
+            The reasoning content (str) if available, None otherwise.
+        """
+        try:
+            if hasattr(chunk, "choices") and chunk.choices:
```
TomeHirata: nit: 
```suggestion
            if choices := getattr(chunk, "choices"):
```

dspy/streaming/streaming_listener.py
chenmoneygithub: good call, done!

dspy/adapters/utils.py
chenmoneygithub: that's a good question. 

I did think about the same thing, but changing the `__name__` in `dspy.Reasoning` could lead to confusion, because essentially it's just a type, but from the perspective of DSPy Adapter, it is treated as string. So I kept the logic inside adapter utils.

dspy/adapters/utils.py
chenmoneygithub: same as above, let me know your thought!

dspy/adapters/types/reasoning.py
chenmoneygithub: good call! added

dspy/adapters/types/reasoning.py
chenmoneygithub: the code was a bit misleading, refactored

dspy/adapters/types/reasoning.py
chenmoneygithub: done!

dspy/adapters/utils.py
TomeHirata: Got it, we need this conversion so that LLM won't return `reasoning: {content: "xxx"}`?

chenmoneygithub: Yes exactly

TomeHirata: Understood, is `reasoning: {content: "xxx"}` that bad? Iirc, `ToolCalls` is handled in this way (something like `tool_calls: [{"name": "tool_a"}]`), so having a consistent behavior might not be a bad idea.

chenmoneygithub: This is for backward compatibility of ChainOfThought. Since ChainOfThought is the module that gets the most usage by DSPy users except from dspy.Predict, I would keep the behavior unchanged as much as possible. 

This PR only includes dspy.Reasoning, but we will have a followup to migrate ChainOfThought ot dspy.Reasnong. The purpose of dspy.Reasoning is providing a way to turn on the native reasoning, while not changing other behavior of ChainOfThought. Let me know if this makes sense to you!

TomeHirata: Just to confirm, you meant the consistency of the LLM response for `ChainOfThought` instead of the return signature of `ChainOfThought.__call__` by "backward compatibility"? For the latter, we can just return `dspy.Reasning.content` as reasoning field of `ChainOfThought` so that the return type is consistent to the existing behavior. But I agree if it's the former.

chenmoneygithub: Yes exactly, for example the result of `dspy.inspect_history()` will be changed without this conversion.

TomeHirata: Understood, can you add the comment? Also isn't the behavior changed anyway when native reasoning field is used?

TomeHirata (approved):
Approving to unblock

dspy/adapters/utils.py
chenmoneygithub: Added the comment! 

> Also isn't the behavior changed anyway when native reasoning field is used?

So for reasoning model, the behavior is changed, which is intentional. But for non-reasoning model, we want to keep it exactly the same behavior as before.

PR #8991, opened 2025-10-29
Fix PyPI download banner

TomeHirata:
Filing a fix for the broken badge in case this is not a temporal issue. 
image


dushmanta05:
Oops! I just send a PR #8995 and checked this is also there.

There was no issue, so I thought may be nobody has done it. :)

PR #9000, opened 2025-10-30
Fixed a serialization bug in `dspy.Example`

guydav:
`Example.toDict()` did not account for the case in which the value is a tuple. This happens in `Evaluate.__call__()`, where the `results` is a list of tuples, where each tuple includes `Prediction` objects internally. Right now, the failure to parse tuples means the internal `Prediction` objects do not get serialized to dicts, which means that the resultant dict is not, for example, json-serializable. 

I opted for a minimal fix, but there could be other approaches, e.g. adding a dependency to handle the serialization, etc.

### DIFF
--- a/dspy/primitives/example.py
+++ b/dspy/primitives/example.py
@@ -195,6 +195,8 @@ class Example:
                 return value.toDict()
             elif isinstance(value, list):
                 return [convert_to_serializable(item) for item in value]
+            elif isinstance(value, tuple):
+                return tuple(convert_to_serializable(item) for item in value)
             elif isinstance(value, dict):
                 return {k: convert_to_serializable(v) for k, v in value.items()}
             else:


guydav:
Hi @chenmoneygithub -- tl:dr; everything inherits from `dspy.Example`. `Evaluate.__call__` returns the following:
```python
EvaluationResult(
    score=round(100 * ncorrect / ntotal, 2),
    results=results,
)
```
`EvaluationResult` subclasses `Prediction`, which subclasses `Example`.
In the structure above, `results` has the [type signature](https://github.com/stanfordnlp/dspy/blob/b67732d510321048130a1f3ee726bf8f1cdcbe19/dspy/evaluate/evaluate.py#L57) `list[tuple["dspy.Example", "dspy.Example", Any]]`.
This is how we get an `Example` with, eventually, tuples that include `Example`s within them.  

If this doesn't make sense, I can generate a minimal example, but the thing that failed for me is cherry-picking the `save_as_json` logic implemented in `Example.__call__` in the current main ([here](https://github.com/stanfordnlp/dspy/blob/b67732d510321048130a1f3ee726bf8f1cdcbe19/dspy/evaluate/evaluate.py#L210))

PR #9006, opened 2025-10-31
Add a method to extract system message based on adapter and signature

chenmoneygithub:
In some scenarios users want to extract the actual prompt sent to LM programmatically. Right now we allow users to use `Adapter.format()` to get the multi-turn message, but sometimes users only want to extract the system message. This PR introduces a new method `Adapter.format_system_message()` which only extracts system message based on adapter and signature. 

Example usage:

```python
import dspy

signature = dspy.Signature("question -> answer")
system_message = dspy.ChatAdapter().format_system_message(signature)
print(system_message)
```

The output should resemble:

```
Your input fields are:
1. `question` (str):
Your output fields are:
1. `answer` (str):

[[ ## question ## ]]
{question}
[[ ## answer ## ]]
{answer}
[[ ## completed ## ]]
```


TomeHirata:
Can we tag related issues if any?

chenmoneygithub:
@TomeHirata  The related issue is our work on KD service/SDK - this PR just makes it easier to extract out the prompt.

PR #9013, opened 2025-11-03
Make `Teleprompt.compile` more typing friendly.

TyTodd:
I replaced `student: Module` in `Teleprompt.compile`'s signature with `student: M`, a TypeVar that tracks the type and returns that same type
```python
M = TypeVar("M", bound=Module)
class Teleprompter:
...
    def compile(
        self,
        student: M,
        ...
```
I also applied this signature to all other compilers in the SDK.

It doesn't change the functionality of anything at all. My main reason for adding this is I (and others) often add custom functionality to `dspy.Module`s and its just convenient to have the type checker still recognize those functions and attributes on the optimized program after running compile. 

For example, in this script the type checker will not highlight `weather_function` because the custom function is forgotten when the new program is returned from `compile`. With this new commit that is no longer the case. 

```python
class MyAgent(Module):
    def __init__(self):
        super().__init__()
        self.predict = dspy.ReAct("question -> answer", tools=[self.weather_function])

    def weather_function(self):
        return random.random()

optimizer = MIPROv2(metric=dspy.evaluate.answer_exact_match, auto="light")
compiled_agent = optimizer.compile(student=MyAgent(), trainset=trainset, teacher=teacher, valset=valset)
print(compiled_agent.weather_function())
```


TomeHirata:
Thanks @TyTodd! Is it hard to have `typing.cast()` in your codebase after `.compile()`?

TyTodd:
> Thanks @TyTodd! Is it hard to have `typing.cast()` in your codebase after `.compile()`?

Thanks for your response! It's not really _hard_. It's more of a taste thing I guess. 
Selfishly, I am building a library that extends DSPy and the type forwarding would just make things prettier on my end.

dspy/teleprompt/bettertogether.py
```diff
@@ -91,10 +95,9 @@ def _run_strategies(self, parsed_strategy, student, trainset, valset_ratio) -> M
         launched_flag = False
 
         for ind, step_code in enumerate(parsed_strategy):
-            current_strategy = self.STRAT_SEP.join(parsed_strategy[:ind + 1])
```
TomeHirata, line 94: Let's not include unrelated changes

TyTodd: Must be my formatter. Let me see if I can undo the other formatting changes.

TomeHirata:
I see, using base class in type hint is very common but I also understand where you come from. @okhat @chenmoneygithub what do you think about using generic in the type hint of .compile?

PR #9017, opened 2025-11-04
fix(audio): Normalize 'x-wav' audio format to 'wav'

### LINKED ISSUE #8996: [Bug] LiteLLM exception when using OpenAI Speech-To-Text models
### What happened?

I'm trying to use DSPy for an audio transcription task. From what I see in the documentation, this should be supported with a dedicated `dsp.yAudio` input class.
When I execute the attached code example I receive the following error:
> litellm.exceptions.BadRequestError: litellm.BadRequestError: OpenAIException - This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?

I do manage to use OpenAI's speech-to-text models using LiteLLM directly though:
```python

from litellm import transcription
from pathlib import Path

audio_file = Path("example_001.m4a").open("rb")
response = transcription(model="gpt-4o-transcribe", file=audio_file)
print(f"response: {response}")

```

Am I missing anything or is this a bug with the way LiteLLM is used behind the scenes?

### Steps to reproduce

```python
from pathlib import Path

import dspy


class TranscribeSignature(dspy.Signature):
    audio: dspy.Audio = dspy.InputField()
    transcription: str = dspy.OutputField()


if __name__ == "__main__":
    audio_path = Path("example_001.m4a")
    audio = dspy.Audio.from_file(audio_path)

    lm = dspy.LM(model="gpt-4o-mini-transcribe")
    dspy.configure(lm=lm)

    transcriber = dspy.Predict(TranscribeSignature)

    transcription = transcriber(audio=audio)  # <- raises litellm.exceptions.BadRequestError
    print(transcription)

```

### DSPy version

3.0.3

akshatvishu:
Closes #8996 

### Description:
The `dspy.Audio.from_file` (and `from_url`) method relies on Python's `mimetypes.guess_type()` to determine the audio format. On some operating systems, this function can return non-standard MIME types, such as `audio/x-wav` for `.wav` files.

These non-standard format strings, often prefixed with x- (like `x-wav` or `x-m4a`), are then passed to the LLM API (e.g., OpenAI). This can cause a `400 BadRequestError`, as the API typically only accepts compliant formats (e.g., wav, m4a).

This patch adds a check to `from_file`, `from_url`, and the data URI branch of `encode_audio` to normalize these formats by removing any x- prefix, ensuring an API-compliant format is always sent.

### DIFF
--- a/dspy/adapters/types/audio.py
+++ b/dspy/adapters/types/audio.py
@@ -67,7 +67,7 @@ class Audio(Type):
             raise ValueError(f"Unsupported MIME type for audio: {mime_type}")
         audio_format = mime_type.split("/")[1]
 
-        audio_format = _normalize_audio_format(audio_format)  # <-- Use the helper
+        audio_format = _normalize_audio_format(audio_format)
 
         encoded_data = base64.b64encode(response.content).decode("utf-8")
         return cls(data=encoded_data, audio_format=audio_format)
@@ -89,7 +89,7 @@ class Audio(Type):
 
         audio_format = mime_type.split("/")[1]
 
-        audio_format = _normalize_audio_format(audio_format)  # <-- Use the helper
+        audio_format = _normalize_audio_format(audio_format)
 
         encoded_data = base64.b64encode(file_data).decode("utf-8")
         return cls(data=encoded_data, audio_format=audio_format)


dspy/adapters/types/audio.py
```diff
@@ -61,6 +61,10 @@ def from_url(cls, url: str) -> "Audio":
         if not mime_type.startswith("audio/"):
             raise ValueError(f"Unsupported MIME type for audio: {mime_type}")
         audio_format = mime_type.split("/")[1]
+
+        if audio_format == "x-wav":
```
TomeHirata: can we use `audio_format = audio_format.replace("x-", "")`?

akshatvishu: sure!

dspy/adapters/types/audio.py
```diff
@@ -61,6 +61,10 @@ def from_url(cls, url: str) -> "Audio":
         if not mime_type.startswith("audio/"):
             raise ValueError(f"Unsupported MIME type for audio: {mime_type}")
         audio_format = mime_type.split("/")[1]
+
+        if "x-" in audio_format:
```
TomeHirata: This condition is probably unnecessary

TomeHirata:
Thanks @akshatvishu, can you add a unit test?

akshatvishu:
@TomeHirata  Added the unit-test and also I slightly changed the logic and used the  [removeprefix()](<https://docs.python.org/3/library/stdtypes.html#str.removeprefix>) instead of the [replace()](<https://docs.python.org/3/library/stdtypes.html#str.replace>) to safely remove only the prefix from audio format strings, preventing unintended replacements if `"x-"` appears elsewhere in the format.

TomeHirata (approved):
LGTM

PR #9018, opened 2025-11-04
fix(ChatAdapter): None/True/False parsing in Dict[str, Any] fields (#8820)

### LINKED ISSUE #8820: [Bug] Conflict between ChatAdapter.user_message_output_requirements and ChatAdapter.parse leads to high tool call failure rate
### What happened?

When describing dict fields with Any value type, LM is given output requirements that conflict with how the value is parsed which leads to high failure rate for tool call args.

Minimal example:
```
class ActionToolCallSignature(dspy.Signature):

    function_name: str = dspy.OutputField(
        desc="Name of the function to be called."
    )
    arguments: Dict[str, Any] = dspy.OutputField(
        desc="Arguments for the function to be called."
    )
```

Chat adapter will generate the following prompt component:
`Respond with the corresponding output fields, starting with the field [[ ## reasoning ## ]], then [[ ## function_name ## ]], then [[ ## arguments ## ]] (must be formatted as a valid Python dict[str, Any]), and then ending with the marker for [[ ## completed ## ]].`

Note that the LM is instructed to format the arguments as a Python dict.

To which the LM will generate the following correct output:
`[[ ## arguments ## ]] {"title": "Wikipedia:Featured and good topic candidates/Featured log/November 2016", "revision_id": None}`

which will then be parsed as:
```
{'title': 'Wikipedia:Featured and good topic candidates/Featured log/November 2016',
 'revision_id': 'None'}
```
Note that the `None` value is converted into a `str`.

I believe this is from
`candidate = json_repair.loads(value)  # json_repair.loads returns "" on failure.`
ln 165 in dspy.adapters.utils.parse_value

I would say perhaps the output requirements should tell the LM to format dicts as json, not as python dicts, but I am not sure about all cases that need to be handled.

plz lmk what you think and if a PR would be helpful

I put this in the discord forum-discussion as well. wasn't sure where was better.

### Steps to reproduce

1. create a signature as described above
2. see chat adapter generated prompt
3. use chat adapter parse to parse the example LM output above 
4. note the incorrect conversion from `None` to `"None"`

### DSPy version

3.0.2

isaacbmiller:
When LLMs output Python-style dicts with None values, json_repair.loads was converting None to the string "None" instead of null. This caused high tool call failure rates for Dict[str, Any] fields.

Changes:
- Update ChatAdapter to instruct "valid JSON" instead of "valid Python dict"
- Reorder parse_value to try ast.literal_eval before json_repair.loads
- Preserve ast.literal_eval results as fallback when json_repair fails
- Add comprehensive tests for None, True, False, and null handling

This maintains backward compatibility by accepting both Python and JSON syntax while correctly parsing None/True/False/null values.

Fixes #8820

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -167,12 +167,20 @@ def parse_value(value, annotation):
         # Handle union annotations, e.g., `str | None`, `Optional[str]`, `Union[str, int, None]`, etc.
         return TypeAdapter(annotation).validate_python(value)
 
-    candidate = json_repair.loads(value)  # json_repair.loads returns "" on failure.
+    # Try Python literal parsing first (handles None, True, False, single quotes, tuples)
+    ast_result = None
+    try:
+        ast_result = ast.literal_eval(value)
+        return TypeAdapter(annotation).validate_python(ast_result)
+    except (ValueError, SyntaxError):
+        pass  # Parsing failed
+    except pydantic.ValidationError:
+        pass  # Parsing succeeded but validation failed, will try json_repair
+
+    # Try JSON repair as fallback
+    candidate = json_repair.loads(value)
     if candidate == "" and value != "":
-        try:
-            candidate = ast.literal_eval(value)
-        except (ValueError, SyntaxError):
-            candidate = value
+        candidate = ast_result if ast_result is not None else value
 
     try:
         return TypeAdapter(annotation).validate_python(candidate)
--- a/tests/adapters/test_chat_adapter.py
+++ b/tests/adapters/test_chat_adapter.py
@@ -1,4 +1,4 @@
-from typing import Literal
+from typing import Any, Literal
 from unittest import mock
 
 import pydantic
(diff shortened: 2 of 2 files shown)


dspy/adapters/utils.py
```diff
@@ -167,12 +167,20 @@ def parse_value(value, annotation):
...
-    candidate = json_repair.loads(value)  # json_repair.loads returns "" on failure.
+    # Try Python literal parsing first (handles None, True, False, single quotes, tuples)
+    ast_result = None
+    try:
+        ast_result = ast.literal_eval(value)
+        return TypeAdapter(annotation).validate_python(ast_result)
+    except (ValueError, SyntaxError):
+        pass  # Parsing failed
+    except pydantic.ValidationError:
+        pass  # Parsing succeeded but validation failed, will try json_repair
```
TomeHirata, line 178:
```suggestion
    except (ValueError, SyntaxError, pydantic.ValidationError):
        pass
```

TomeHirata:
>Update ChatAdapter to instruct "valid JSON" instead of "valid Python dict"

Is this included in this PR? Also it might make sense to continue to let LLM return python literals for ChatAdapter and have separate utilities between ChatAdapter and JSONAdapter.

tests/adapters/test_chat_adapter.py
```diff
@@ -641,3 +640,67 @@ class MySignature(dspy.Signature):
 In adhering to this structure, your objective is: 
         Answer the question with multiple answers and scores"""
     assert system_message == expected_system_message
+
```
TomeHirata, line 643: Can we add some test cases in tests/adapters/test_adapter_utils.py?

PR #9019, opened 2025-11-04
Move numpy dependency into an optional dep group

isaacbmiller:
Numpy is one of the heaviest dependencies in DSPy (20MB), and it not necessary for most usecases.

This adds an import guard to the uses so that users can opt in to the dependency if needed.

Refactors some small usecase logic in cobro, simba, and miprov2 to not use numpy at all.

Based on how sci-kit learn does it: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_optional_dependencies.py

cc @chenmoneygithub @TomeHirata


isaacbmiller:
Closing after discussing offline.

PR #9025, opened 2025-11-06
Refactor: update type hints for adapter and LM methods

TomeHirata:
- Changed output type in Adapter class to support both dict and str.
- Updated __call__, acall, forward, and aforward methods in BaseLM and LM classes to use more explicit type hints for prompt and messages parameters.
- Enhanced documentation to clarify expected response formats for forward and aforward methods.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -114,7 +114,7 @@ class Adapter:
         self,
         processed_signature: type[Signature],
         original_signature: type[Signature],
-        outputs: list[dict[str, Any]],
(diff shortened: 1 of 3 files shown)


dspy/adapters/base.py
```diff
@@ -114,7 +114,7 @@ def _call_postprocess(
         self,
         processed_signature: type[Signature],
         original_signature: type[Signature],
-        outputs: list[dict[str, Any]],
+        outputs: list[dict[str, Any] | str],
```
TomeHirata, line 117: This was the main motivation, we support a list of string actually

PR #9028, opened 2025-11-06
docs: add note on Python version for pre-commit

akshatvishu:
This PR adds a warning note to the `CONTRIBUTING.md` file.

**The Problem:**
The "Environment Setup" guide recommends using `Python 3.11`, but the `.pre-commit-config.yaml` file requires `Python 3.10` . This causes `pre-commit install` to fail for new contributors who follow the instructions and only have `Python 3.11` .

**The Fix:**
This note clarifies the discrepancy *before* the setup instructions. It advises contributors to either ensure `python3.10` is also available on their PATH or simply use `python=3.10` for their virtual environment to satisfy both requirements.

Hi @TomeHirata , I've pushed the requested changes. Thanks!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,10 +91,6 @@ Once the pull request is approved, a team member will take care of merging.
 
 Python 3.10 or later is required.
 
-> **Note on Python Versions:** While the setup instructions below recommend Python 3.11 for development, our `pre-commit` hooks (in `.pre-commit-config.yaml`) are configured to use **Python 3.10**.
->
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -91,6 +91,10 @@ Once the pull request is approved, a team member will take care of merging.
 
 Python 3.10 or later is required.
 
+> **Note on Python Versions:** While the setup instructions below recommend Python 3.11 for development, our `pre-commit` hooks (in `.pre-commit-config.yaml`) are configured to use **Python 3.10**.
```
TomeHirata: Can we change the instruction below to 3.10 instead? We should make sure our package works with the oldest supported version.

akshatvishu: Sure, I updated the `venv` setup to now install with Python 3.10 by default!

Issue #9029, opened 2025-11-06
[Feature] Pass a Sampler to MIPROv2 to parallelize at the prompt level


sign-of-fourier:
### What feature would you like to see?

I want to be able to pass a sampler to the MIPROv2 optimizer. Right now, MIPROv2 uses optuna and can parallelize at the training set evaluation level, meaning, the evaluation of each observation in the training set can be evaluated in parallel.

However, it can be even faster to also be able to evaluate at the prompt level. In other words, evaluate different instruction, few-shot combinations at the same time.

This type of parallelization is already native to optuna. You simply need to pass a value for n_jobs. However, it requires a different kind of sampler. Right now, MIPROv2 uses the TPESampler by default with a Categorical Distribution. 

The sampler to be passed should be allowed to be anything but I want to be able to pass a qEI, batch Expected Improvement Sampler from mpromptune.

The MIPROv2 teleprompter does not need to import the sampler or be dependent on it in any way. The only change is that sampler could be passed. For reverse compatibility, the parameter should optional and the default behavior will be the same as it is today: use the TPESampler.

The second advantage to the qEI sampler is that it uses a Gaussian Process based on the embeddings of the prompts. The TPESampler uses a categorical distribution. The categorical distribution treats the instructions and the few independently and does not search the space at the semantic level. The Gaussian Process is a better way to perform Bayesian Optimization and it at the semantic level rather than simple categorical.

This requires a second, one-line change that the sampler is initialized after the instructions and few-shot demonstrations are created.

1. Ability to pass sampler, otherwise same as today: TPESampler
2. Ability to pass n_jobs allowing optuna to parallelize at the instruction, few-shot combination level
3. MIPROv2 calls initilization function after instruction, few-shot demonstrations are created. 

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Additional Context

There are many references about batch Expected Improvement. Here are two:
- https://hal.science/hal-00732512v2/document
- https://botorch.org/docs/tutorials/closed_loop_botorch_only/

Python
(qEI) Batch Expected Improvement Sampler
 https://pypi.org/project/m-promptune/

These three changes I've described are implemented in a forked repository here:
https://github.com/sign-of-fourier/dspy/blob/main/dspy/teleprompt/mipro_optimizer_v2.py

TomeHirata:
cc: @klopsahlong can you take a look at this proposal?

PR #9030, opened 2025-11-06
Miprov2 option to pass sampler as an argument

sign-of-fourier:
Feature request 9029

https://github.com/stanfordnlp/dspy/issues/9029


sign-of-fourier:
Thanks for looking at this!

Output is an a gist [here](https://gist.github.com/sign-of-fourier/ce6dc4cd3a1a628baed59c30149e10e6).

The notebook from where this output came from is here:
https://github.com/sign-of-fourier/mpromptune/blob/main/notebooks/tutorial.ipynb

Let me know how else I can help.

PR #9035, opened 2025-11-09
Add optional timeout to PythonInterpreter to fix infinite/long hangs

adaminsky:
## Summary:
Adds a timeout parameter to `dspy.primitives.python_interpreter.PythonInterpreter` to cap how long sandboxed code is allowed to run. When a timeout is provided, stdout is read via a background thread; if no output arrives before the deadline, the Deno subprocess is killed and an InterpreterError is raised. **This guards against code hanging indefinitely.**

## Changes:
- Added `timeout: float | None` in the constructor and store it on the instance.
- Wrap the stdout read in `_read_output_line`, which either blocks normally or enforces the timeout by joining a reader thread and, on timing out, terminating the sandbox process.
- Ensure timed-out executions clean up the underlying Deno process so future runs can restart cleanly.

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -1,7 +1,9 @@
 import json
 import os
 import subprocess
+import threading
 from os import PathLike
+from queue import Empty, Queue
 from types import TracebackType
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


chenmoneygithub:
@isaacbmiller Mind taking a look?

PR #9036, opened 2025-11-09
Add Per-Instance Program Synthesis module and tests

adaminsky:
This adds Per-Instance Program Synthesis (PIPS) from the NeurIPS '25 paper *Once Upon an Input: Reasoning via Per-Instance Program Synthesis* (https://openreview.net/forum?id=XsQNqRdcdh).

PIPS is an agent which solves problems by either using code generation and execution or Chain of Thought, choosing the reasoning mode by having the model select on a per-instance basis. When code execution is chosen, the final answer is computed entirely via code from an abstracted version of the input. PIPS also refines the generated code based on feedback from another LLM process which critiques the code. This refinement process is vital since frontier models tend to hardcode (wrong) answers within code rather than use the code to actually compute anything. On many reasoning benchmarks, PIPS significantly outperforms Chain of Thought, Program of Thought, and CodeAct.

### DIFF
--- a/dspy/predict/per_instance_synthesis.py
+++ b/dspy/predict/per_instance_synthesis.py
@@ -1,0 +1,442 @@
+import json
+import re
+import textwrap
+from dataclasses import dataclass
+from typing import Any
+
+import dspy
+from dspy.primitives.python_interpreter import InterpreterError, PythonInterpreter
+from dspy.signatures.signature import Signature, ensure_signature
+
+
+@dataclass
+class PIPSResult:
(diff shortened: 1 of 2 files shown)


chenmoneygithub:
@adaminsky Thanks for the PR! 

This paper is very new, and doesn't have reference yet, so it's may not be a good time to ship to DSPy directly. However, if you can share a github gist/colab notebook to demonstrate it improves the performance significantly compared to existing modules, we can check in the code. Thank you!

PR #9041, opened 2025-11-10
Support Python 3.14

TomeHirata:
This PR allows to use DSPy with Python 3.14


dspy/utils/magicattr.py
```diff
@@ -0,0 +1,127 @@
+"""
+Compatibility layer for magicattr that works with Python 3.14+
+
+This module provides a patched version of magicattr's functionality
+that is compatible with Python 3.14's removal of ast.Num and ast.Str.
```
TomeHirata, line 5: magicattr is not compatible with 3.14 and is not actively maintained

PR #9048, opened 2025-11-12
Add guards against loading pkl files

isaacbmiller:
Adds a guard against loading pkl files in individual programs without an argument or environment variable set.

### DIFF
--- a/tests/primitives/test_base_module.py
+++ b/tests/primitives/test_base_module.py
@@ -229,7 +229,7 @@ def test_load_with_version_mismatch(tmp_path):
         assert len(handler.messages) == 4
 
         # First message is about pickle loading
-        assert "Loading .pkl files can run arbitrary code" in handler.messages[0]
+        assert ".pkl" in handler.messages[0]
 
         # Rest are version mismatch warnings
         for msg in handler.messages[1:]:


isaacbmiller:
@codex review

dspy/primitives/base_module.py
```diff
@@ -233,26 +234,35 @@ def save(self, path, save_program=False, modules_to_serialize=None):
                     "with `.pkl`, or saving the whole program by setting `save_program=True`."
                 )
         elif path.suffix == ".pkl":
+            logger.warning("Loading .pkl files can run arbitrary code, which may be dangerous. Prefer "
```
chenmoneygithub: For the right action, let's include the code details for clear action, e.g., "Prefer saving with json format, e.g., module.save("module.json")"

dspy/primitives/base_module.py
```diff
@@ -233,26 +234,35 @@ def save(self, path, save_program=False, modules_to_serialize=None):
...
+                          "saving with .json files if possible.")
             state = self.dump_state(json_mode=False)
             state["metadata"] = metadata
             with open(path, "wb") as f:
                 cloudpickle.dump(state, f)
         else:
             raise ValueError(f"`path` must end with `.json` or `.pkl` when `save_program=False`, but received: {path}")
 
-    def load(self, path):
+    def load(self, path, dangerously_allow_pickle=False):
```
chenmoneygithub: I feel a bit strange about the name, I recommend going with `allow_pickle` to keep consistency with numpy

dspy/utils/saving.py
```diff
@@ -24,17 +25,21 @@ def get_dependency_versions():
     }
 
 
-def load(path: str) -> "Module":
+def load(path: str, dangerously_allow_pickle: bool = False) -> "Module":
```
chenmoneygithub: same here, probably just `allow_pickle`

dspy/primitives/base_module.py
```diff
@@ -233,26 +234,35 @@ def save(self, path, save_program=False, modules_to_serialize=None):
...
+            dangerously_allow_pickle (bool): If True, allow loading .pkl files, which can run arbitrary code.
+                This is dangerous and should only be used if you are sure about the source of the file and in a trusted environment.
         """
         path = Path(path)
 
         if path.suffix == ".json":
             with open(path, "rb") as f:
                 state = orjson.loads(f.read())
         elif path.suffix == ".pkl":
+            if not dangerously_allow_pickle and not os.getenv("DSPY_ALLOW_PICKLE") == "1":
```
chenmoneygithub: Here we are basically implicitly declaring an env var `DSPY_ALLOW_PICKLE` that controls actual logic, which could be confusing to our users. 

If we want to adopt the env var-based control, we can follow [how MLflow handles it](https://github.com/mlflow/mlflow/blob/1d36f00c7e349844a0abfbd620536ad27e322b06/mlflow/environment_variables.py). But IMO we don't need this second lock, `allow_pickle` by itself should be a decent guard for now, unless the reporter disagrees with our solution.

isaacbmiller:
Removed environment variable, changed from `dangerously_allow_pickle` to just `allow_pickle`, and suggested `module.save("module.json")`.

chenmoneygithub (approved):
LGTM!

PR #9049, opened 2025-11-12
adding a logical branch to model_dump pydantic objects in Example.to_…

crawftv:
adding a logical branch to model_dump pydantic objects in Example.to_dict()


crawftv:
I’ll close it. Someone else did a better implementation of my bug report with copilot

PR #9050, opened 2025-11-12
Enable callback logging only for full eval on GEPA

TomeHirata:
Instead of relying on "capture_trace" option, enable/disable logging based on the sample size since sample size is a more accurate indicator.


LakshyAAAgrawal:
Hi @TomeHirata, one thing to be aware of is that GEPA will soon have support for not running eval on len(valset) but instead a subset of it, similar to what MIPRO does, hence this scheme may stop working then. I am trying to think what might be a good solution, maybe tracking reflection_minibatch_size may be better here?

TomeHirata:
@LakshyAAAgrawal thanks for sharing the plan, so should we disable logging when `len(valset)== reflection_minibatch_size`?

LakshyAAAgrawal:
I think `len(valset) <= reflection_minibatch_size` should be good.

TomeHirata:
@LakshyAAAgrawal sounds good, updated

PR #9051, opened 2025-11-12
Add FieldwiseChainOfThought predictor class

youngbinkim0:
Introduces the FieldwiseChainOfThought class, which extends Predict to prepend reasoning fields before each output field in a signature. Includes logic for signature extension, state dumping/loading, and activation control to support step-by-step reasoning for model outputs.


chenmoneygithub:
@youngbinkim0 Thanks for the PR! 

I am not sure if we want this, are you seeing performance gain by using this instead of `ChainOfThought`? If so, could you share a github gist/colab notebook for demonstration?

PR #9055, opened 2025-11-13
Add param to load_memory_cache to stop pkl files without explicit loading

isaacbmiller:
Add param to load_memory_cache to stop loading pkl files without explicit parameter

The downside is that this doesn't provide an alternative to using pkl for cache storage

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -173,7 +173,7 @@ class Cache:
     def load_memory_cache(self, filepath: str, allow_pickle: bool = False) -> None:
         if not allow_pickle:
             raise ValueError("Loading untrusted .pkl files can run arbitrary code, which may be dangerous. \
-            Set allow_pickle=True to load if you are running in a trusted environment and the file is from a trusted source.")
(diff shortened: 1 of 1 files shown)


dspy/clients/cache.py
```diff
@@ -170,7 +170,11 @@ def save_memory_cache(self, filepath: str) -> None:
             with open(filepath, "wb") as f:
                 cloudpickle.dump(self.memory_cache, f)
 
-    def load_memory_cache(self, filepath: str) -> None:
+    def load_memory_cache(self, filepath: str, allow_pickle: bool = False) -> None:
+        if not allow_pickle:
+            raise ValueError("Loading untrusted .pkl files can run arbitrary code, which may be dangerous. \
+            Set allow_pickle=True to load if you are running in a trusted environment and the file is from a trusted source.")
```
chenmoneygithub: nit: backtick around code.
```suggestion
            Set `allow_pickle=True` to load if you are running in a trusted environment and the file is from a trusted source.")
```

chenmoneygithub (approved):
LGTM with one minor comment, feel free to merge after addressing!

PR #9056, opened 2025-11-13
feat: integrate nest-asyncio for async tool calls

TomeHirata:
- Added support for using async tools within a running event loop by integrating `nest-asyncio`.
- Updated `pyproject.toml` to include `nest-asyncio` as a test dependency.
- Enhanced error handling to prompt installation of `nest-asyncio` if not found.
- Added tests to verify async tool calls from a running event loop.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "cloudpickle>=3.0.0",
     "numpy>=1.26.0",
     "xxhash>=3.5.0",
-    "gepa[dspy]==0.0.18",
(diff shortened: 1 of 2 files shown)


chenmoneygithub (commented):
I need to do some testing on this, as I remember the nest-asyncio was causing unexpected errors on async calling. 

Btw, the uv.lock update seems wrong to me, it shouldn't change every line.

TomeHirata:
@chenmoneygithub Yes please! Also updated uv.lock.

PR #9061, opened 2025-11-13
Bump DSPy version

chenmoneygithub:


TomeHirata:
This was not updated because we used b2 tag for 3.0.4 release?

chenmoneygithub:
@TomeHirata The reason was a bit strange. In this 3.0.4 release, I was not using main branch as the target, but 3.0.4b2 branch instead, so there was a failure at the last step: https://github.com/stanfordnlp/dspy/actions/runs/19240679329/job/55002405672

Issue #9063, opened 2025-11-15
[docs] Add Google-style docstrings for dspy/adapters/chat_adapter.py ChatAdapter class


azai91:
### What feature would you like to see?

I would like to improve docstring coverage for some of the public adapter APIs - starting with the ChatAdapter init method

Ref: https://github.com/stanfordnlp/dspy/issues/8926

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

chenmoneygithub:
@azai91  Thank you for taking on this! Looking forward to your PR

PR #9072, opened 2025-11-18
[docs] Add Google-style docstrings for dspy/adapters/chat_adapter.py ChatAdapter class #9063

azai91:
This PR adds a Google-style docstring for ChatAdapter class and constructor
in dspy/adapters/chat_adapter.py.

Behavior verified in IPython (returns a NEW Signature class with same fields and new instructions).
No logic changes.

```
Help on function __init__ in module dspy.adapters.chat_adapter:

__init__(self, callbacks=None, use_native_function_calling: bool = False, native_response_types=None, use_json_adapter_fallback: bool = True)
    Args:
        callbacks: List of callback functions to execute during `format()` and `parse()` methods. Callbacks can be
            used for logging, monitoring, or custom processing. Defaults to None (empty list).
        use_native_function_calling: Whether to enable native function calling capabilities when the LM supports it.
            If True, the adapter will automatically configure function calling when input fields contain `dspy.Tool`
            or `list[dspy.Tool]` types. Defaults to False.
        native_response_types: List of output field types that should be handled by native LM features rather than
            adapter parsing. For example, `dspy.Citations` can be populated directly by citation APIs
            (e.g., Anthropic's citation feature). Defaults to `[Citations]`.
        use_json_adapter_fallback: Whether to automatically fallback to JSONAdapter if the ChatAdapter fails.
            If True, when an error occurs (except ContextWindowExceededError), the adapter will retry using
            JSONAdapter. Defaults to True.
```

Refs: https://github.com/stanfordnlp/dspy/issues/9063

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -15,6 +15,7 @@ from dspy.adapters.utils import (
 )
 from dspy.clients.lm import LM
 from dspy.signatures.signature import Signature
+from dspy.utils.callback import BaseCallback
 from dspy.utils.exceptions import AdapterParseError
 
 field_header_pattern = re.compile(r"\[\[ ## (\w+) ## \]\]")
@@ -39,12 +40,20 @@ class ChatAdapter(Adapter):
 
     def __init__(
         self,
-        callbacks=None,
+        callbacks: list[BaseCallback] | None = None,
         use_native_function_calling: bool = False,
-        native_response_types=None,
+        native_response_types: list[type[type]] | None = None,
         use_json_adapter_fallback: bool = True,
     ):
-
+        """
+        Args:
+            callbacks: List of callback functions to execute during adapter methods.
+            use_native_function_calling: Whether to enable native function calling capabilities.
+            native_response_types: List of output field types handled by native LM features.
+            use_json_adapter_fallback: Whether to automatically fallback to JSONAdapter if the ChatAdapter fails.
+                If True, when an error occurs (except ContextWindowExceededError), the adapter will retry using
+                JSONAdapter. Defaults to True.
+        """
         super().__init__(
             callbacks=callbacks,
             use_native_function_calling=use_native_function_calling,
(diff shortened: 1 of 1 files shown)


azai91:
@chenmoneygithub can you review this PR when you get the chance?

dspy/adapters/chat_adapter.py
```diff
@@ -26,13 +26,44 @@ class FieldInfoWithName(NamedTuple):
 
 
 class ChatAdapter(Adapter):
+    """Chat-based Adapter for LM interactions.
+
+    The ChatAdapter formats DSPy signatures into a conversational chat format with structured field markers.
```
chenmoneygithub: This is not true. Although we call this class `ChatAdapter`, the format has nothing related to chat/conversation. It's just an arbitrary structure that works robustly for language models.

dspy/adapters/chat_adapter.py
```diff
@@ -26,13 +26,44 @@ class FieldInfoWithName(NamedTuple):
...
     def __init__(
         self,
         callbacks=None,
         use_native_function_calling: bool = False,
         native_response_types=None,
         use_json_adapter_fallback: bool = True,
     ):
+        """
+        Args:
+            callbacks: List of callback functions to execute during `format()` and `parse()` methods. Callbacks can be
```
chenmoneygithub: Here we are replicating the docstring of the parent class, can we use a placeholder and replace the docstring programmatically?

chenmoneygithub (commented):
Thanks for the PR!

dspy/adapters/chat_adapter.py
azai91: couldn't find any example in the codebase so I added `__init__.__doc__ = Adapter.__init__.__doc__` which is a pattern [lightgbm](https://lightgbm.readthedocs.io/en/latest/_modules/lightgbm/sklearn.html#LGBMRegressor) uses. lmk what you think.

dspy/adapters/chat_adapter.py
azai91: update the class docs to be less tried to chat / more agnostic to language models.

dspy/adapters/chat_adapter.py
```diff
@@ -26,20 +26,34 @@ class FieldInfoWithName(NamedTuple):
...
     ):
+
         super().__init__(
             callbacks=callbacks,
             use_native_function_calling=use_native_function_calling,
             native_response_types=native_response_types,
         )
         self.use_json_adapter_fallback = use_json_adapter_fallback
 
+    __init__.__doc__ = Adapter.__init__.__doc__
```
chenmoneygithub: Let's move this to the bottom of the file. Btw looks like we are missing the additional args section?

azai91: moved to bottom and added use_json_adapter_fallback arg. updated description with the preview.

onel:
I was also thinking of helping with chat adapter and happy to see work has started. 

@chenmoneygithub I think this is a super important module and the docs should include an example, what do you think?
I also think the other methods on the class should be documented as well.
I can help with that if needed

PR #9073, opened 2025-11-18
FIX: streamify was appending StatusStreamingCallback directly to the shared settings.callbacks list

glesperance:
### Fix streaming status message callback mutation

**Bug**  
`streamify` was appending `StatusStreamingCallback` directly to the shared `settings.callbacks` list. 

That list was mutated in place and reused across calls, callbacks (and their status message providers) leaked between different `dspy.streamify(...)` invocations. 

**Fix**  
Each call to `streamify` now works on an isolated callback list by copying the global callbacks before modification. 

This prevents callback leakage between programs, ensures that custom status message providers only affect the specific program they are passed to, and restores the correct, deterministic status message behavior required by the tests.

### DIFF
--- a/tests/streaming/test_streaming.py
+++ b/tests/streaming/test_streaming.py
@@ -135,7 +135,7 @@ async def test_custom_status_streaming():
 
 
 @pytest.mark.anyio
-async def test_default_then_custom_status_message_provider():
+async def test_concurrent_status_message_providers():
     class MyProgram(dspy.Module):
         def __init__(self):
             self.generate_question = dspy.Tool(lambda x: f"What color is the {x}?", name="generate_question")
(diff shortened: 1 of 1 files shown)


dspy/streaming/streamify.py
```diff
@@ -161,7 +162,7 @@ async def use_streaming():
     elif not iscoroutinefunction(program):
         program = asyncify(program)
 
-    callbacks = settings.callbacks
+    callbacks = deepcopy(settings.callbacks)
```
chenmoneygithub: I am concerned about how robust this line is, since there are many cases where `deepcopy` could fail. Maybe we should just do a shallow copy and use `dspy.context(callbacks=callbacks)` instead?

chenmoneygithub (commented):
Thanks for spotting the bug and putting up the PR!

dspy/streaming/streamify.py
glesperance: agreed. added a commit amend the changes

glesperance:
@chenmoneygithub just bumping this in case you missed the notif about the new commit fixing the issue you brought up

tests/streaming/test_streaming.py
```diff
@@ -134,6 +134,100 @@ def module_start_status_message(self, instance, inputs):
         assert status_messages[2].message == "Predict starting!"
 
 
+@pytest.mark.anyio
+async def test_default_then_custom_status_message_provider():
```
chenmoneygithub: We may want to refactor the testing a bit - instead of default => custom, we can put up two threads that apply different `MyStatusMessageProvider`, and verify that both threads work as expected. Then at the end we verify that the callback is not being modified.

glesperance:
@chenmoneygithub The latest commit incorporates your feedback on the tests. Please let me know if you see anything else that needs adjustment.

Just flagging this touches a production bug -- ie we're running on our fork but looking forward to moving back to main whenever you get a chance to review.

chenmoneygithub (approved):
Awesome, LGTM!

PR #9089, opened 2025-11-26
Properly yield the last chunk in streaming

chenmoneygithub:
Fix #9085 

Earlier we introduced the logic to immediately yield when the chunk cannot form the last chunk, which is leading to the case where the last chunk only contains the end identifier, whereas we still want to yield it.

PR #9094, opened 2025-11-26
Fix/is last chunk 9085

### LINKED ISSUE #9085: [Bug] `is_last_chunk` never set
### What happened?

I'm not sure if it's because of #8890, but with the built-in `ReAct` module and `streamify`, I'm never seeing `is_last_chunk` on `StreamResponse`'s.

FYSA, I'm running against vLLM using GPT OSS 120b.

### Steps to reproduce

Notebook content:

```python
import logging
import dspy  # pyright: ignore[reportMissingTypeStubs]
from dspy.streaming import streamify  # pyright: ignore[reportMissingTypeStubs]
from dspy.streaming.messages import StreamResponse  # pyright: ignore[reportMissingTypeStubs]
from dspy.streaming.streamify import StreamListener  # pyright: ignore[reportMissingTypeStubs]
# Set up logging
logging.basicConfig(level=logging.INFO, format="%(message)s")

# Configure LLM (using your existing client)
lm = dspy.LM(
    model="hosted_vllm/gpt-oss-120b",
    api_key="None",
    api_base="https://my-vllm-server/v1",
)
dspy.configure(lm=lm)  # pyright: ignore[reportUnknownMemberType]
dspy.configure_cache(
    enable_disk_cache=False,
    enable_memory_cache=False,
)

# Use built-in ReAct
react = dspy.ReAct("question -> answer", max_iters=3, tools=[])  # pyright: ignore[reportArgumentType]

# Create stream listener
stream_listeners = [
    StreamListener("next_thought", allow_reuse=True),
    StreamListener("reasoning", allow_reuse=True),
    StreamListener("answer", allow_reuse=True),
]

# Streamify the ReAct module
stream_react = streamify(react, stream_listeners=stream_listeners)

# Stream and log every chunk
chunk_count = 0
last_chunk_count = 0

async for chunk in stream_react(question="What is 2+2?"):  # pyright: ignore[reportCallIssue, reportUnknownVariableType]
    if isinstance(chunk, StreamResponse):
        chunk_count += 1
        if chunk.is_last_chunk:
            last_chunk_count += 1

print(f"Total chunks: {chunk_count}")
print(f"Chunks with is_last_chunk=True: {last_chunk_count}")

if last_chunk_count == 0:
    print("No last chunks found")
else:
    print("Some last chunks found")

```

Output in 3.0.4:
```
Total chunks: 51
Chunks with is_last_chunk=True: 0
No last chunks found
```

Output in 3.0.3:
```
Total chunks: 45
Chunks with is_last_chunk=True: 3
Some last chunks found
```

### DSPy version

3.0.4

baughmann:
## Summary

Fixes #9085 - `is_last_chunk` was never being set to `True` in `StreamResponse` objects in DSPy 3.0.4, breaking stream completion detection. I noticed this behaviour both with Qwen 3 Next 80b and with GPT OSS 120b with vLLM. Given that @chenmoneygithub fixed #8890 it would be really nice to get their two cents on this PR if possible. Not sure if the model they used behaved differently to these ones.

## Background

I discovered that `is_last_chunk` was always `False` when using `StreamListener` with the built-in `ReAct` module. This made it impossible to reliably detect when streaming had completed for a given field. My original goal was straightforward: make sure `is_last_chunk` gets sent.

## Why This Fix Couldn't Be More Surgical

The root cause is in `_default_handle_stream_chunk()`:

## Original code (line 295)
```python
if token:
    return StreamResponse(...)
```
The method only returned a `StreamResponse` when `token` was non-empty. Since the final chunk is detected by seeing the next field's marker (or end of stream), the token buffer is often empty at that point - so no final chunk was ever emitted.

I considered several approaches:

1. **Minimal fix:** Only emit when `self.stream_end=True` - This would still end up not emitting the last chunk most of the time
2. **Separate final chunk:** Basically keep the existing `if token:` condition as-is, but add logic to track whether we've already emitted a final chunk and emit it if we haven't. This would've added more state to keep track of in the stream listener.
3. **Current approach:** Change condition to `if token or self.stream_end:` - Simpler but changes behavior

I went with option 3 because it ensures `is_last_chunk=True` is always emitted, makes the behavior predictable, and aligns with the existing `finalize()` method behavior added in #8890.

## Behavioral Changes

**This changes user-facing behavior in a significant way.**

### Before (v3.0.4):
```python
chunks = [
    StreamResponse(chunk='To', is_last_chunk=False),
    StreamResponse(chunk=' get to', is_last_chunk=False),
    StreamResponse(chunk=' the other side', is_last_chunk=False),
]
# No chunk with is_last_chunk=True was emitted
```
### After (this PR):
```python
chunks = [
    StreamResponse(chunk='To', is_last_chunk=False),
    StreamResponse(chunk=' get to', is_last_chunk=False),
    StreamResponse(chunk=' the other side', is_last_chunk=False),
    StreamResponse(chunk='', is_last_chunk=True),  # New: empty final chunk
]
```
**What's different:**
1. Every field now gets a final chunk with `chunk=""` and `is_last_chunk=True`
2. Users will receive one additional `StreamResponse` per field
3. The empty string in `chunk` is intentional - it's an end-of-stream marker

### Why I Think This Is Correct

Looking at the history:

1. Commit `ab7ac0b9` (July 2025) initially added `is_last_chunk` with the same `if token:` condition
2. Commit `a0b2155a` (October 2025) added the `finalize()` method specifically to handle missing completion markers by emitting empty final chunks
3. Several tests were updated to expect `all_chunks[-2]` instead of `all_chunks[-1]` because of these empty chunks

This suggests empty final chunks were already the intended behavior (maybe?), but the implementation was incomplete. This PR makes it consistent across all streaming scenarios, not just when completion markers are missing.

### Meat-and-potatoes
- Modified `dspy/streaming/streaming_listener.py::_default_handle_stream_chunk()` (1 line change)
- Changed condition from `if token:` to `if token or self.stream_end:`

### Tests
- Fixed `test_stream_listener_returns_correct_chunk_chat_adapter_untokenized_stream` to expect empty final chunks
- Added 8 tests in `tests/streaming/test_streaming_is_last_chunk.py` (willing to append these to `test_streaming.py` if desired:
  - Coverage for ChatAdapter, JSONAdapter, XMLAdapter
  - Multiple listeners scenario
  - `allow_reuse=True` and `allow_reuse=False` behavior
  - Edge cases (few tokens, varied chunk sizes, order invariance)

### Documentation
- Enhanced `StreamResponse` docstring to document all 4 fields including `is_last_chunk`
- Updated `StreamListener` docstring with usage examples
- Added comprehensive section in streaming tutorial explaining when `is_last_chunk=True` occurs, why the final chunk is empty, and how to use it

## Testing

All 41 streaming tests pass (36 passed, 3 skipped). The new tests ensure exactly one `is_last_chunk=True` per field per streaming session and consistent behavior across all adapters. I think I am unable to run some of the other tests locally, so I am hoping that CI picks up on this and gives me some more answers.

## Questions for Reviewers

I'm specifically looking for feedback on:

1. **Is this the right approach?** The behavioral change affects all users. Are there backward compatibility concerns I should address?

2. **Alternative solutions?** Are there better ways to fix `is_last_chunk` that would be less invasive? For example:
   - Emitting final chunk only from `finalize()` instead of in the main flow?
   - Making empty final chunks opt-in via a flag?
   - Different signaling mechanism entirely?

3. **Documentation:** Is the explanation of empty final chunks clear enough? Should we add migration guidance?

4. **Edge cases:** Are there scenarios where this change could break existing usage patterns that I haven't considered?

## Compatibility Notes

I think the impact is minimal because:

1. The bug made the feature unusable - `is_last_chunk` was always `False`, so users couldn't rely on it anyway
2. Empty chunks are filterable - Users can `if chunk.chunk:` to skip them if needed
3. Predictable behavior - The guarantee of exactly one `is_last_chunk=True` per field makes completion detection reliable

That said, users who iterate over all chunks will see additional entries. If this is a concern, I'm open to exploring other approaches.

Thanks in advance, and please be gentle I'm new here!

### DIFF
--- a/docs/docs/tutorials/index.md
+++ b/docs/docs/tutorials/index.md
@@ -42,6 +42,12 @@ Welcome to DSPy tutorials! We've organized our tutorials into three main categor
     - [GEPA for Code Backdoor Classification (AI control)](gepa_trusted_monitor/index.ipynb)
 
 
+- Experimental RL Optimization:
+    - [Overview](rl_ai_program/index.md)
+    - [RL for Privacy-Conscious Delegation](rl_papillon/index.ipynb)
+    - [RL for Multi-Hop Research](rl_multihop/index.ipynb)
+
+
 - Tools, Development, and Deployment
     - [Use MCP in DSPy](mcp/index.md)
     - [Output Refinement](output_refinement/best-of-n-and-refine.md)
@@ -52,5 +58,15 @@ Welcome to DSPy tutorials! We've organized our tutorials into three main categor
     - [Tracking DSPy Optimizers](optimizer_tracking/index.md)
     - [Streaming](streaming/index.md)
     - [Async](async/index.md)
+
+
+- Real-World Examples:
+    - [Overview](real_world_examples/index.md)
+    - [Generating llms.txt](llms_txt_generation/index.md)
+    - [Email Information Extraction](email_extraction/index.md)
+    - [Memory-Enabled ReAct Agents with Mem0](mem0_react_agent/index.md)
+    - [Financial Analysis with Yahoo Finance](yahoo_finance_react/index.md)
+    - [Automated Code Generation from Documentation](sample_code_generation/index.md)
+    - [Building a Creative Text-Based AI Game](ai_text_game/index.md)
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "pydantic>=2.0",
     "litellm>=1.64.0",
     "diskcache>=5.6.0",
-    "json-repair>=0.30.0",
+    "json-repair>=0.54.2",
     "tenacity>=8.2.3",
     "anyio",
     "asyncer==0.0.8",
--- a/tests/adapters/test_chat_adapter.py
+++ b/tests/adapters/test_chat_adapter.py
@@ -649,6 +649,37 @@ def test_chat_adapter_native_reasoning():
         assert result[0]["reasoning"] == dspy.Reasoning(content="Step-by-step thinking about the capital of France")
 
 
+def test_chat_adapter_parses_float_with_underscores():
+    """
+    This test verifies that ChatAdapter can parse float numbers with underscores.
+    After json-repair version 0.54.1, floats like "123_456.789" are treated as normal float numbers.
+    """
+
+    class Score(pydantic.BaseModel):
+        score: float
+
+    class MySignature(dspy.Signature):
+        question: str = dspy.InputField()
+        score: Score = dspy.OutputField()
+
+    adapter = dspy.ChatAdapter()
+
+    # Simulate a response with a float number containing underscores
+    with mock.patch("litellm.completion") as mock_completion:
+        mock_completion.return_value = ModelResponse(
+            choices=[
+                Choices(message=Message(content="[[ ## score ## ]]\n{'score': 123_456.789}\n[[ ## completed ## ]]"))
+            ],
+            model="openai/gpt-4o-mini",
+        )
+
+        lm = dspy.LM("openai/gpt-4o-mini", cache=False)
+        result = adapter(lm, {}, MySignature, [], {"question": "What is the score?"})
+
+        # The underscore-separated float should be parsed as a normal float
+        assert result[0]["score"].score == 123456.789
+
+
 def test_format_system_message():
     class MySignature(dspy.Signature):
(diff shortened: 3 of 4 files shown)


chenmoneygithub:
duplicates #9089 

Anyway, thanks for the PR! The fix is mostly correct, but it's incorrect to claim that the last chunk is always empty.

baughmann:
@chenmoneygithub oops yep good catch thank you!

PR #9104, opened 2025-12-01
Fix ParallelExecutor crash when Parallel called from within module forward()

LukeDevs:
**## Summary**

Fixes a bug where `dspy.Parallel` silently returns `None` for all tasks when called from within a module's `forward()` method while MLflow autologging is enabled.

**## Problem

When using `mlflow.dspy.autolog()`, calling `dspy.Parallel` from inside a module's `forward()` method causes all results to silently return `None`. 

The root cause is twofold:

**1. Incorrect ContextVar access in `parallelizer.py` (line 95)**

`thread_local_overrides` is a `ContextVar`, which requires `.get()` and `.set()` methods for access. The existing code incorrectly attempts to access an `.overrides` attribute directly:

```python
thread_local_overrides.overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
```

This raises `AttributeError: '_contextvars.ContextVar' object has no attribute 'overrides'`.

The bug only manifests when `usage_tracker` is present in the parent context, which occurs when MLflow autologging is enabled.

**2. Silent exception handling (line ~155)**

Worker exceptions are caught and silently discarded:

```python
try:
    index, outcome = f.result()
except Exception:
    pass
```

This masked the underlying `AttributeError`.

**## Solution

**1. Restructured ContextVar access to build the complete overrides dictionary before setting:**

```python
original = thread_local_overrides.get()
new_overrides = {**original, **parent_overrides.copy()}
if parent_overrides.get("usage_tracker"):
    new_overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
token = thread_local_overrides.set(new_overrides)
```

**2. Added error logging for worker failures:**

```python
except Exception as e:
    logger.error(f"Worker failed: {e}")
    if self.provide_traceback:
        traceback.print_exc()
```

**## Reproduction

```python
import dspy
import mlflow

mlflow.dspy.autolog()  # Enables usage_tracker, triggering the bug

class ProcessItem(dspy.Module):
    def __init__(self):
        self.predictor = dspy.ChainOfThought("item -> result")

    def forward(self, item):
        return self.predictor(item=item)

class GetItems(dspy.Module):
    def __init__(self):
        self.predictor = dspy.ChainOfThought("objective -> task_items: list[str]")

    def forward(self, objective):
        return self.predictor(objective=objective)

class Pipeline(dspy.Module):
    def __init__(self):
        self.get_items = GetItems()
        self.process_item = ProcessItem()

    def forward(self, objective):
        result = self.get_items(objective=objective)

        parallel = dspy.Parallel(num_threads=4, provide_traceback=True)
        results = parallel([
            (self.process_item, {"item": item})
            for item in result.task_items
        ])
        return dspy.Prediction(results=results)  # Returns [None, None, None, ...] before fix

lm = dspy.LM("azure/gpt-4")
dspy.configure(lm=lm)

pipeline = Pipeline()
results = pipeline("do something")
print(results)
```

### DIFF
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -90,9 +90,9 @@ class ParallelExecutor:
 
             original = thread_local_overrides.get()
             new_overrides = {**original, **parent_overrides.copy()}
-            if parent_overrides.get("usage_tracker"):
+            if new_overrides.get("usage_tracker"):
                 # Usage tracker needs to be deep copied across threads so that each thread tracks its own usage
-                new_overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
+                new_overrides["usage_tracker"] = copy.deepcopy(new_overrides["usage_tracker"])
             token = thread_local_overrides.set(new_overrides)
 
             try:
--- a/docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
+++ b/docs/docs/api/optimizers/GEPA/GEPA_Advanced.md
@@ -443,3 +443,146 @@ gepa = dspy.GEPA(
     auto="medium"
 )
 ```
+
+## Tool Optimization
+
+### What is enable_tool_optimization?
+
+When `enable_tool_optimization=True`, GEPA jointly optimizes `dspy.ReAct` modules: predictor instructions and tool descriptions and argument descriptions are updated together, instead of being tuned in isolation. This lets the model learn better patterns for when to call a tool and how to use it from the same execution traces and feedback that drive core GEPA.
+
+### Usage and constraints
+
+- **Expose tools as `dspy.Tool` in signatures and examples.** GEPA only optimizes tools that are represented as `dspy.Tool` and actually passed as `dspy.Tool` objects into your modules.
+- **Treat `Tool.name` as a stable identifier.** `Tool.name` is the tool's name, and GEPA uses it to attach improved descriptions and argument descriptions. If you reuse the same `Tool.name` for different tools, they will share the same text updates.
+- **Avoid custom tools named `"finish"`.** The built-in ReAct `"finish"` tool is reserved and excluded from optimization. Custom tools with the name `"finish"` are also not optimized.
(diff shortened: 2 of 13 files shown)


dspy/utils/parallelizer.py
```diff
@@ -154,7 +154,9 @@ def all_done():
                         try:
                             index, outcome = f.result()
                         except Exception:
-                            pass
+                            logger.error(f"Worker failed: {e}")
```
TomeHirata: This is not really helpful since we catch exceptions by wrapping the original function in `_wrap_function`

dspy/utils/parallelizer.py
```diff
@@ -89,10 +89,10 @@ def worker(parent_overrides, submission_id, index, item):
             from dspy.dsp.utils.settings import thread_local_overrides
 
             original = thread_local_overrides.get()
-            token = thread_local_overrides.set({**original, **parent_overrides.copy()})
+            new_overrides = {**original, **parent_overrides.copy()}
             if parent_overrides.get("usage_tracker"):
-                # Usage tracker needs to be deep copied across threads so that each thread tracks its own usage
-                thread_local_overrides.overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
+                new_overrides["usage_tracker"] = copy.deepcopy(parent_overrides["usage_tracker"])
```
TomeHirata: Looks reasonable, can we keep the comment? And also can you add a test?

dspy/utils/parallelizer.py
LukeDevs: @TomeHirata Understood. I've reverted back to pass.

dspy/utils/parallelizer.py
LukeDevs: @TomeHirata Added the comment back in directly above the deep copy.

LukeDevs:
@TomeHirata. Implemented a unit test as requested. Tried to use the existing tests as reference.

PR #9109, opened 2025-12-04
Fix two bugs in bootstrap_trace.py

CDBiddulph:
Previously, bootstrap_trace_data would capture execution traces in the `prediction` parameter of wrapped_metric, but then discarded them. Metrics would always receive trace=None instead of the actual trace. This would break metrics that use the trace.

Also fixed attempted division of two lists, dividing their lengths instead.


CDBiddulph:
Ping - could somebody please take a look?

PR #9112, opened 2025-12-04
Enhance StreamListener to support generic type annotations for output

### LINKED ISSUE #9099: [Bug] StreamListener crashes when `_output_type` is a Generic
### What happened?

StreamListener crashes when `_output_type` is a Generic with the following error message:

```txt
TypeError: issubclass() arg 1 must be a class
```

### Steps to reproduce

MNE
```python
import asyncio
import dspy

dspy.configure(lm=dspy.LM("openai/gpt-4o-mini", cache=False))

# Create the program and wrap it with streaming functionality
class Hello(dspy.Signature):
    world: List[str] = dspy.OutputField()

predict = dspy.Predict(Hello)
stream_listeners = [
    dspy.streaming.StreamListener(signature_field_name="world"),
]
stream_predict = dspy.streamify(predict, stream_listeners=stream_listeners)

async def use_streaming():
    output = stream_predict(
        question="why did a chicken cross the kitchen?",
    )
    return_value = None
    async for value in output:
        if isinstance(value, dspy.Prediction):
            return_value = value
        else:
            print(value)
    return return_value

output = asyncio.run(use_streaming())
print(output)
```

this happens because `issubclass` expects a non-generic class.

https://github.com/stanfordnlp/dspy/blob/d42f29ef4489a30a22ba783fdfb84ae1e598142b/dspy/streaming/streaming_listener.py#L138

the following snippet shows the core components at work:

```python
import dspy
from typing import List
    
class Hello(dspy.Signature):
    world: List[str] = dspy.OutputField()

t=Hello.output_fields["world"].annotation
print(f"{t=}")

# Only call issubclass if t is actually a class
import inspect
if inspect.isclass(t):
    issubclass(t, str)
else: print("not a class")
```

it seem that an easy way to fix the issue would be to check if `self._output_type` is an instantiable class before running `issubclass`.

i'm bypassing that check entirely by overloading the `_output_type` property with `type(str)`

### DSPy version

3.1.0b1

TomeHirata:
Closes #9099

### DIFF
--- a/tests/streaming/test_streaming.py
+++ b/tests/streaming/test_streaming.py
@@ -1206,6 +1206,52 @@ async def test_chat_adapter_simple_pydantic_streaming():
 
 
 @pytest.mark.anyio
+async def test_chat_adapter_with_generic_type_annotation():
+    class TestSignature(dspy.Signature):
+        question: str = dspy.InputField()
+        response: list[str] | int = dspy.OutputField()
+
+    class MyProgram(dspy.Module):
+        def __init__(self):
+            self.predict = dspy.Predict(TestSignature)
+
+        def forward(self, question, **kwargs):
+            return self.predict(question=question, **kwargs)
+
+    async def chat_stream(*args, **kwargs):
+        yield ModelResponseStream(model="gpt-4o-mini", choices=[StreamingChoices(delta=Delta(content="[[ ##"))])
+        yield ModelResponseStream(model="gpt-4o-mini", choices=[StreamingChoices(delta=Delta(content=" response"))])
+        yield ModelResponseStream(model="gpt-4o-mini", choices=[StreamingChoices(delta=Delta(content=" ## ]]\n\n"))])
+        yield ModelResponseStream(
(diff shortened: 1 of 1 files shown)


tests/streaming/test_streaming.py
```diff
@@ -1932,3 +1932,24 @@ async def non_reasoning_stream(*args, **kwargs):
...
+
+    assert listener._output_type == list[str] | int
+
+    mock_chunk = ModelResponseStream(
+        model="gpt-4o-mini",
+        choices=[StreamingChoices(delta=Delta(content="test"))],
+    )
+
+    with dspy.context(adapter=dspy.JSONAdapter()):
+        result = listener.receive(mock_chunk)
```
chenmoneygithub: I do get what the test case is testing, but our future maintainers could be confused. Maybe add a few more chunks + flush to make the test easier to follow?

TomeHirata: Good point, updated the test case into a full example of streaming

PR #9014, opened 2025-11-04
feat(File): add File type for handling file data

TomeHirata:
Support File input by introducing dspy.File. This is the last content type supported by OpenAI chat completion that is not supported by DSPy natively.

Closes #8974 #8916

```python
lm = dspy.LM(
    model="openai/gpt-5-nano",
    temperature=1.0,
)

dspy.configure(lm=lm)

class QA(dspy.Signature):
    file: dspy.File = dspy.InputField()
    summary = dspy.OutputField()
program = dspy.Predict(QA)
program(file=dspy.File.from_path("./2507.19457v1.pdf"))
```
Inspect history:
```
[2025-11-13T15:33:25.005187]

System message:

Your input fields are:
1. `file` (File):
Your output fields are:
1. `summary` (str):
All interactions will be structured in the following way, with the appropriate values filled in.

[[ ## file ## ]]
{file}

[[ ## summary ## ]]
{summary}

[[ ## completed ## ]]
In adhering to this structure, your objective is: 
        Given the fields `file`, produce the fields `summary`.


User message:

[[ ## file ## ]]
<file: name:2507.19457v1.pdf, id:, data_length:3825976>

Respond with the corresponding output fields, starting with the field `[[ ## summary ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.


Response:

[[ ## summary ## ]]
[...]

PR #9114, opened 2025-12-04
support optional `mime_type` in `dspy.File.from_file_id`

dimitry12:
Support optional `mime_type` in `dspy.File.from_file_id`.

The main effect of this is: LiteLLM will not attempt to download file from `file_id` to determine its `format`.

This is critical when using Google Gemini and its Files API, because prior to this change the flow for multi-modal context was:
- use Files API to upload files;
- use each file's `.uri` as `file_id` and instantiate `dspy.File.from_file_id` which is then converted/"adaptered" into LiteLLM-schema dict,
- but LiteLLM attempts to [determine `format`](https://docs.litellm.ai/docs/providers/gemini#https-file) by downloading the file from URL contained in `file_id`. This is not possible because Google Gemini Files API is write-only, and the whole generation fails.

With this change we can pass Files API URIs as `file_id` and we can now provide optional `mime_type` explicitly:
```
dspy.File.from_file_id(f.uri, mime_type=f.mime_type)
```
preventing LiteLLM from attempting to download the file.

This change addresses:
- https://github.com/stanfordnlp/dspy/pull/9014#discussion_r2584738117
- https://github.com/stanfordnlp/dspy/pull/9014#discussion_r2508295005

### DIFF
--- a/dspy/adapters/types/file.py
+++ b/dspy/adapters/types/file.py
@@ -31,6 +31,7 @@ class File(Type):
     file_data: str | None = None
     file_id: str | None = None
     filename: str | None = None
+    mime_type: str | None = None
 
     model_config = pydantic.ConfigDict(
         frozen=True,
@@ -47,6 +48,7 @@ class File(Type):
                 "file_data": values.file_data,
                 "file_id": values.file_id,
                 "filename": values.filename,
+                "mime_type": values.mime_type,
             }
 
         if isinstance(values, dict):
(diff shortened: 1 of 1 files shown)


tyrael147:
Hi,

I am dealing with the same problem here. I have PDF files that were uploaded through the genai api and the current version of LiteLLM fails because the exact issue you mentioned.

I tested your fork and works just fine, thank you for that. However, in the latest version of dspy (3.3.1), some logic inside modules like the adapter have changed and an additional fix is needed so the ``format`` field is passed to the LiteLLM engine (1.96.0). 

For the moment I added some changes to your fix so I can still use the latest dspy version: https://github.com/tyrael147/dspy/commit/939f387d5da0824c3c9dedfb8e797b1d47966ba0 

Is something related to this planned?

PR #9115, opened 2025-12-05
[WIP] feat(history): Allow history to be more flexible to new kinds of user inputs

isaacbmiller:
## 📝 Changes Description

Refactors both dspy.History.

Goal
Multi turn chat has been a weaker part of the DSPy dx. It is not clear how you should maintain history, nor particularly easy to do so. These PRs should allow you to build multi turn, (optionally tool use) chatbots more easily.

This PR is the start of efforts to alleviate the concerns.

Overview
History

DSPy.History used to be set to only a single input type.

All inputs and outputs needed to match the current signature that is being used otherwise each step would be filtered to match those fields. This was done to guarantee that we could separate into proper user/assistant pairs. We relax that constraint and give the user more flexibility on how to collect and display messages.

History now has 4 possible modes that it can handle:

signature (the old default) - will filter strictly to the current signature
demos - has input/output fields labeled via a nested dict with: `{"input_fields": {"k":"v"}, "output_fields":{"k":"v"}}, but does not require them to match a signature. This still allows you to get user/assistant splitting.
flat - a dict of kv pairs - all messages go into the user side
raw - uses user/assistant messages

Next PRs:
1. Change ReAct to use History as well
2. Update documentation to reflect new usage for both History and ReAct
3. Create a tutorial showing how to make a multi turn chatbot with history

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import TYPE_CHECKING, Any, get_origin
 
@@ -476,11 +477,15 @@ class Adapter:
                 return name
         return None
 
-    def _serialize_kv_value(self, v: Any) -> Any:
-        """Safely serialize values for kv-mode formatting."""
-        if isinstance(v, (str, int, float, bool)) or v is None:
-            return v
-        return serialize_for_json(v)
+    def _serialize_kv_value(self, v: Any) -> str:
+        """Serialize a value to string for flat-mode history formatting.
+
+        Uses the same pattern as format_field_value in adapters/utils.py.
+        """
+        jsonable = serialize_for_json(v)
+        if isinstance(jsonable, (dict, list)):
+            return json.dumps(jsonable, ensure_ascii=False)
+        return str(jsonable)
 
     def _make_dynamic_signature_for_inputs(self, keys: list[str]) -> type[Signature]:
         """Create a dynamic signature with input fields only (no instructions)."""
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -8,6 +8,7 @@ from dspy.adapters.types import History, Type
 from dspy.adapters.types.base_type import split_message_content_for_custom_types
 from dspy.adapters.types.reasoning import Reasoning
 from dspy.adapters.types.tool import Tool, ToolCalls
+from dspy.adapters.utils import serialize_for_json
 from dspy.experimental import Citations
 from dspy.signatures.field import InputField, OutputField
 from dspy.signatures.signature import Signature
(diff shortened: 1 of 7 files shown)


ryanh-ai:
@isaacbmiller  Any reason not to pull this into local branch and use it at this point?

isaacbmiller:
Yes this will likely change. Going over different design iterations currently.

ryanh-ai:
Hi @isaacbmiller checking back here to see if you had any thoughts.  I'm interested in getting native tool calling in an optimizable agent loop working, was going to start from here.

immu4989:
Just probed #9755 (the canonical `ToolCalls` refactor) — comment here:
https://github.com/stanfordnlp/dspy/pull/9755#issuecomment-4527518550. Noticed a small integration gap with this
PR worth flagging while the design is still in flux.

After #9755 lands, the natural way to hold a ReAct turn's tool calls is as
`ToolCalls.ToolCall(name, args, id)` objects. But History here expects raw-
mode messages as wire format dicts, so round tripping a tool using turn
requires:

```python
result1 = agent(request="...")
# result1.trajectory[...].tool_calls -> list[ToolCalls.ToolCall]

history = dspy.History.from_raw([
    {"role": "assistant", "content": None,
     "tool_calls": [tc.to_lm_tool_call("chat") for tc in tool_calls]},
    {"role": "tool", "content": str(tool_result), "tool_call_id": tc.id},
    # ... one tool message per call
])
```
That's a fair amount of glue between two adjacent PRs in the same stack.
Cleanest fix probably looks like a History.from_trajectory(trajectory)
factory that handles the conversion (and maybe a symmetric
history.to_trajectory() for the inverse).

Two specific questions to help me plan:

1. Per the original PR description ("Next PRs: change ReAct to use History"), is that ReAct integration still on the roadmap, or has the direction moved since December? Asking because the design iterations you mentioned in December never surfaced and the thread's been quiet.
2. If the ReAct integration is still planned but waiting on capacity: would you take a small follow-on PR adding History.from_trajectory(...) so the #9755 ↔ History interop lands cleanly without waiting for the larger ReAct refactor?

Seconding @ryanh-ai's interest — native-tool-calling-in-an-optimizable-
agent-loop is the story I'm trying to land in my own pipelines too.

ryanh-ai:
I'm happy to work on this collaboratively once the semantics from the Active PR lock down.

I haven't reviewed the other PR in detail yet, does it have all the necessary pieces to build tool inclusive multi-turn trajectories in your view?

isaacbmiller:
This is complete now with ReActV2 merged - please test and let me know if it works for your usecases. Closing this PR now

immu4989:
Great timing, thanks. Pulling #9823 / #9824 / ReActV2 (#9825) and will
report back with a multi turn tool probe, same shape as the #9755
round trip test, but exercising the new replay + tool-result path
end to end. Will tag findings on the ReActV2 PR.

@ryanh-ai — happy to share what I find; ping me if you want to coordinate
test cases.

PR #9118, opened 2025-12-09
Add compatibility of type aliases (Python 3.12+) to `make_signature`

stevapple:
This will allow `Signature`s involving type aliases to be used with `ReAct`.


stevapple:
May @chenmoneygithub review this patch or hand it to another reviewer? Thanks in advance

PR #9121, opened 2025-12-10
fix: skip merging when both usage entry values are None

### LINKED ISSUE #9120: [Bug] TypeError in usage_tracker.py: object of type 'int' has no len() when merging usage entries
### What happened?

Image

When using `dspy.streaming.streamify()` with MLflow's `mlflow.dspy.autolog()`, a `TypeError` occurs in the usage tracker:

```
TypeError: object of type 'int' has no len()
```

The error originates from `dspy/utils/usage_tracker.py` at line 35:
```python
if usage_entry2 is None or len(usage_entry2) == 0:
```

The issue appears to be that `usage_entry2` is expected to be a list or dict, but receives an `int` instead, causing the `len()` call to fail.

### Steps to reproduce

This a mini reproduction code.
```python
import dspy
import dspy.streaming 
import asyncio
import mlflow
import mlflow.dspy
import random
mlflow.set_tracking_uri("http://localhost:5000")  # Use local MLflow server
mlflow.set_experiment("dspy-mini-reproduction")
mlflow.dspy.autolog()

lm = dspy.LM(model="gpt-4o-mini")
dspy.configure(lm=lm)

def search_weather(city: str):
    return {
        "city": city,
        "weather": random.choice(["sunny", "cloudy", "rainy", "snowy"]),
        "temperature": random.randint(0, 40),
        "humidity": random.randint(0, 100),
        "pressure": random.randint(900, 1100),
        "wind_speed": random.randint(0, 100),
        "wind_direction": random.choice(["N", "S", "E", "W"]),
        "wind_gust": random.randint(0, 100),
        "wind_gust_direction": random.choice(["N", "S", "E", "W"]),
        "wind_gust_speed": random.randint(0, 100),
    }

react = dspy.ReAct("question -> answer", tools=[search_weather])


stream_react = dspy.streaming.streamify(react,stream_listeners=[
    dspy.streaming.StreamListener("answer")
])


async def main():
    city = random.choice(["Tokyo", "London", "Paris", "Berlin", "Rome", "Madrid", "Berlin", "Rome", "Madrid", "Berlin", "Rome", "Madrid"])
    pred = stream_react(question=f"What is the weather in {city}?")
    async for chunk in pred:
        print(chunk)

asyncio.run(main())
```

### DSPy version

3.0.4

chizukicn:
close #9120

### DIFF
--- a/tests/utils/test_usage_tracker.py
+++ b/tests/utils/test_usage_tracker.py
@@ -190,6 +190,16 @@ def test_merge_usage_entries_with_none_values():
                 "prompt_tokens": 800,
                 "completion_tokens": 100,
                 "total_tokens": 900,
+                "prompt_tokens_details": None,
+                "completion_tokens_details": None,
+            },
+        },
+        {
+            "model": "gpt-4o-mini",
+            "usage": {
+                "prompt_tokens": 100,
+                "completion_tokens": 100,
+                "total_tokens": 200,
                 "prompt_tokens_details": {"cached_tokens": 50, "audio_tokens": 50},
                 "completion_tokens_details": None,
             },
@@ -216,9 +226,9 @@ def test_merge_usage_entries_with_none_values():
 
     total_usage = tracker.get_total_tokens()
 
-    assert total_usage["gpt-4o-mini"]["prompt_tokens"] == 2717
-    assert total_usage["gpt-4o-mini"]["completion_tokens"] == 246
-    assert total_usage["gpt-4o-mini"]["total_tokens"] == 2963
+    assert total_usage["gpt-4o-mini"]["prompt_tokens"] == 2817
+    assert total_usage["gpt-4o-mini"]["completion_tokens"] == 346
+    assert total_usage["gpt-4o-mini"]["total_tokens"] == 3163
     assert total_usage["gpt-4o-mini"]["prompt_tokens_details"]["cached_tokens"] == 50
     assert total_usage["gpt-4o-mini"]["prompt_tokens_details"]["audio_tokens"] == 50
     assert total_usage["gpt-4o-mini"]["completion_tokens_details"]["reasoning_tokens"] == 1
@@ -382,25 +392,3 @@ def test_parallel_executor_with_usage_tracker():
 
     # Parent tracker should remain unchanged (workers have independent copies)
     assert len(parent_tracker.usage_data) == 0
-
-
-def test_merge_usage_entries_with_mixed_types():
-    """Test merging usage entries with mixed types (dict, int, None).
-
-    This test verifies the fix for the bug where _merge_usage_entries would fail
-    with TypeError when trying to call len() on a non-dict value during recursive merging.
-    """
-    tracker = UsageTracker()
-
-    # Test None values in details
-    tracker.add_usage("gpt-4o-mini", {"completion_tokens": 129, "prompt_tokens": 975, "completion_tokens_details": None, "prompt_tokens_details": None})
-    tracker.add_usage("gpt-4o-mini", {"completion_tokens": 144, "prompt_tokens": 1418, "completion_tokens_details": None, "prompt_tokens_details": {"cached_tokens": 0}})
-    tracker.add_usage("gpt-4o-mini", {"completion_tokens": 121, "prompt_tokens": 1167, "completion_tokens_details": {"reasoning_tokens": 0}, "prompt_tokens_details": None})
-
-    total_usage = tracker.get_total_tokens()
-    assert total_usage["gpt-4o-mini"]["completion_tokens"] == 394
-    assert total_usage["gpt-4o-mini"]["prompt_tokens"] == 3560
-    assert isinstance(total_usage["gpt-4o-mini"]["prompt_tokens_details"], dict)
-    assert isinstance(total_usage["gpt-4o-mini"]["completion_tokens_details"], dict)
-    assert total_usage["gpt-4o-mini"]["prompt_tokens_details"] == {"cached_tokens": 0}
-    assert total_usage["gpt-4o-mini"]["completion_tokens_details"] == {"reasoning_tokens": 0}


dspy/utils/usage_tracker.py
```diff
@@ -35,18 +35,37 @@ def _flatten_usage_entry(self, usage_entry: dict[str, Any]) -> dict[str, Any]:
...
+            old_is_dict = isinstance(old_value, dict)
+            new_is_dict = isinstance(new_value, dict)
+
+            # Case 1: Both are dicts - recursively merge them
+            if old_is_dict and new_is_dict:
+                result[key] = self._merge_usage_entries(old_value, new_value)
+
+            # Case 2: One is dict, the other is not (int/None)
+            # Use new value if it's not None (right value takes precedence), otherwise keep old value
+            elif old_is_dict != new_is_dict:
```
chenmoneygithub: would this even happen? it's possible that one is dict and the other is None, but I doubt if they can be of different types. If that's the case, that implies an error on the litellm side, which should provide a unified format for the same model.

chenmoneygithub (commented):
Thanks for the PR! But same as @TomeHirata commented in the issue, I could not reproduce the issue.

Dropped a comment on the code change, but basically litellm shouldn't produce different format for the same key in the usage dict for the same model, which is why we made the strong assumption of format match in the code.

dspy/utils/usage_tracker.py
chizukicn: @chenmoneygithub Regarding this issue, I was also able to reproduce it in a GitHub Actions environment.
Just as a note, I’m using a DeepSeek model served through a third-party OpenAI-compatible API:
https://github.com/chizukicn/dspy-mini-reproduction/actions/runs/20153004759/job/57849614753?pr=3

chizukicn:
@chenmoneygithub 
In my previous reproduction tests, when the program runs normally, the value of self.usage_data looks like this:
```python
    {
        'openai/deepseek/deepseek-v3.2': [
            {
                'completion_tokens': 108,
                'prompt_tokens': 741,
                'total_tokens': 849,
                'completion_tokens_details': None,
                'prompt_tokens_details': None
            },
            {
                'completion_tokens': 158,
                'prompt_tokens': 454,
                'total_tokens': 612,
                'completion_tokens_details': {
                    'accepted_prediction_tokens': None,
                    'audio_tokens': None,
                    'reasoning_tokens': 0,
                    'rejected_prediction_tokens': None,
                    'text_tokens': None,
                    'image_tokens': None
                },
                'prompt_tokens_details': {
                    'audio_tokens': 0,
                    'cached_tokens': 128,
                    'text_tokens': None,
                    'image_tokens': None,
                    'cache_creation_input_tokens': 0,
                    'cache_read_input_tokens': 0
                }
            }
        ]
    }
```
After adding the following logic to simulate a first-time tool call failure:
```python
first_time = True

def search_weather(city: str):
    global first_time
    if first_time:
        raise RuntimeError("First time")
    first_time = False
    
    return {
        "city": city,
        "weather": random.choice(["sunny", "cloudy", "rainy", "snowy"]),
        "temperature": random.randint(0, 40),
        "humidity": random.randint(0, 100),
        "pressure": random.randint(900, 1100),
        "wind_speed": random.randint(0, 100),
        "wind_direction": random.choice(["N", "S", "E", "W"]),
        "wind_gust": random.randint(0, 100),
        "wind_gust_direction": random.choice(["N", "S", "E", "W"]),
        "wind_gust_speed": random.randint(0, 100),
    }
```
The output of self.usage_data becomes:
```json
{
  "openai/deepseek/deepseek-v3.2": [
    {
      "completion_tokens": 129,
      "prompt_tokens": 975,
      "total_tokens": 1104,
      "completion_tokens_details": null,
      "prompt_tokens_details": null
    },
    {
      "completion_tokens": 144,
      "prompt_tokens": 1418,
      "total_tokens": 1562,
      "completion_tokens_details": null,
      "prompt_tokens_details": {
        "audio_tokens": 0,
        "cached_tokens": 0,
        "text_tokens": null,
        "image_tokens": null,
        "cache_creation_input_tokens": 0,
        "cache_read_input_tokens": 0
      }
    },
    {
      "completion_tokens": 121,
      "prompt_tokens": 1167,
      "total_tokens": 1288,
      "completion_tokens_details": {
        "accepted_prediction_tokens": null,
        "audio_tokens": null,
        "reasoning_tokens": 0,
        "rejected_prediction_tokens": null,
        "text_tokens": null,
        "image_tokens": null
      },
      "prompt_tokens_details": null
    }
  ]
}

```
Notice that the array now contains three entries, and the first two entries have completion_tokens_details set to None.

Looking at the _merge_usage_entries method:
```python
def _merge_usage_entries(self, usage_entry1: dict[str, Any] | None, usage_entry2: dict[str, Any] | None) -> dict[str, Any]:
    if usage_entry1 is None or len(usage_entry1) == 0:
        return dict(usage_entry2)
    if usage_entry2 is None or len(usage_entry2) == 0:
        return dict(usage_entry1)

    result = dict(usage_entry2)
    for k, v in usage_entry1.items():
        current_v = result.get(k)
        if isinstance(v, dict) or isinstance(current_v, dict):
        else:
            result[k] = (current_v or 0) + (v or 0)
    return result
```
According to this logic, the first two `completion_tokens_details` entries, being None, are treated as 0 during the merge. Then the third entry’s `completion_tokens_details` gets recursively merged with 0 (corresponding to `usage_entry2`). This explains why the merged result is not as expected, and why None values are causing unintended behavior during recursive merging.

The new reproduction code can be found here: https://github.com/chizukicn/dspy-mini-reproduction/blob/main/main.py

chenmoneygithub:
@chizukicn  Thanks for the detailed explanation, which does make it very clear!

Regarding the implementation, I think we can change the line:

```
result[k] = (current_v or 0) + (v or 0)
```

to check `None` explicilty? 

```
if not current_v and not v:
   result[k] = None
else:
   result[k] = (current_v or 0) + (v or 0)
```

chizukicn:
> @chizukicn Thanks for the detailed explanation, which does make it very clear!
> 
> Regarding the implementation, I think we can change the line:
> 
> ```
> result[k] = (current_v or 0) + (v or 0)
> ```
> 
> to check explicilty?`None`
> 
> ```
> if not current_v and not v:
>    result[k] = None
> else:
>    result[k] = (current_v or 0) + (v or 0)
> ```

@chenmoneygithub But I’m a bit concerned that this might change the original logic. If both values are 0, `not current_v and not v` would treat them as None, whereas they were originally valid zeros. Maybe we could change the condition to `current_v is None and v is None` instead.

chenmoneygithub:
@chizukicn Yes that sounds good!

chizukicn:
@chenmoneygithub  I made a minimal change so that when both `current_v` and `v` are `None`, they are skipped.

PR #9123, opened 2025-12-11
Fix responses structured outputs + cache retrieval error

### LINKED ISSUE #9122: [Bug] Structured outputs always fail when using responses model
## Summary

When using a responses model, attempting to use structured outputs always fails. The request then falls back to normal `JSONAdapter` outputs, which are considerably more brittle.

## Details

Attached is a minimal example, which runs a simple `dspy.Predict` module with a chat completions model first, followed by an attempt with a responses model (disable cache by uncommenting the relevant lines if needed).
Using the `JSONAdapter` to enable structured outputs and a responses model `dspy.LM(model="responses", ...)` always results in:

```
WARNING: Failed to use structured output format, falling back to JSON mode.
```

This is because the way the request parameters pertaining to structured output are constructed and passed on to LiteLLM is incompatible with the Responses API. Notably, passing a Pydantic model to `text.format` is *not* valid from the point of view of the API, despite documentation that might imply otherwise.

### Steps to reproduce

```python
import os
import pydantic
import dspy

OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")

# dspy.configure_cache(
#    enable_disk_cache=False,
#    enable_memory_cache=False,
# )


class Answer(pydantic.BaseModel):
    answer: str
    number: int


class TestSignature(dspy.Signature):
    answer: Answer = dspy.OutputField()


completions_lm = dspy.LM(
    "openai/gpt-5-mini",
    api_key=OPENAI_API_KEY,
    model_type="chat",
    reasoning_effort="low",
    allowed_openai_params=["reasoning_effort"],
)
responses_lm = dspy.LM(
    "openai/gpt-5-mini",
    api_key=OPENAI_API_KEY,
    model_type="responses",
    reasoning={"effort": "low"},
    allowed_openai_params=["reasoning"],
)

predict = dspy.Predict(TestSignature)
adapter = dspy.JSONAdapter()


print("Trying with completions API...")
with dspy.context(lm=completions_lm, adapter=adapter):
    prediction = predict(question="What is a good test answer and number?")
    print(f"Answer: {prediction.answer}")

print("Trying with responses API...")
with dspy.context(lm=responses_lm, adapter=adapter):
    prediction = predict(question="What is a good test answer and number?")
    print(f"Answer: {prediction.answer}")

```

### Proposed fix

See PR #9123.

### DSPy version

main@b95e101

Olocool17:
Closes #9122. See the linked issue  for an extended problem description + minimal reproducible example.

### Structured outputs fix

My proposed fix mutates the `response_format` request argument into the proper form if it is passed as a Pydantic model.
Documentation around this is somewhat unreliable: the litellm docs related to this are outdated/incorrect and the OpenAI docs seem to imply that passing a Pydantic model into `text.format` is valid, but this causes JSON serialization errors down the line. 

So far, the specific formulation as presented in this PR has rendered no issues with our testing.

### Cache retrieval fix

While implementing the fix above, I additionally stumbled on an error pertaining to responses models when using caching.
When an item is successfully retrieved from the cache, an ad-hoc attribute `.cache_hit` is created and set to `True` on the `response` object.

 Unfortunately, this is only possible for `litellm.ModelResponse` (the response for chat models) and not for `litellm.ResponsesAPIResponse` (the response for response models), because the latter is a Pydantic model without `extra="allow"` set in its config.

My proposed fix for this is to simply remove this ad-hoc attribute alltogether, since it is actually superflueous: the `response.usage` attribute is cleared on cache hit anyways, which makes `settings.usage_tracker.add_usage(self.model, dict(results.usage)` a null-op.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -164,7 +164,7 @@ class LM(BaseLM):
 
         self._check_truncation(results)
 
-        if not getattr(results, "cache_hit", False) and dspy.settings.usage_tracker and hasattr(results, "usage"):
+        if dspy.settings.usage_tracker and hasattr(results, "usage"):
             settings.usage_tracker.add_usage(self.model, dict(results.usage))
         return results
 
@@ -202,7 +202,7 @@ class LM(BaseLM):
 
         self._check_truncation(results)
 
-        if not getattr(results, "cache_hit", False) and dspy.settings.usage_tracker and hasattr(results, "usage"):
+        if dspy.settings.usage_tracker and hasattr(results, "usage"):
             settings.usage_tracker.add_usage(self.model, dict(results.usage))
         return results
 
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -122,7 +122,6 @@ class Cache:
         if hasattr(response, "usage"):
             # Clear the usage data when cache is hit, because no LM call is made
             response.usage = {}
-            response.cache_hit = True
         return response
 
     def put(


dspy/clients/lm.py
```diff
@@ -501,6 +502,12 @@ def _convert_chat_request_to_responses_request(request: dict[str, Any]):
     # Convert `response_format` to `text.format` for Responses API
     if "response_format" in request:
         response_format = request.pop("response_format")
+        if isinstance(response_format, type) and issubclass(response_format, pydantic.BaseModel):
```
chenmoneygithub, line 505: Instead of one more serialization, can we use the `text_format` arg? 

```
response_format = request.pop("response_format")
request["text_format"] = response_format
```

chenmoneygithub (commented):
Thanks for the PR! both fixes look good to me, but could you please separate them into 2 PRs? Ideally we want to keep a clean commit history, and each PR only does one thing. 

Please also add unit test for the responses API conversion, thank you again!

dspy/clients/lm.py
Olocool17: Unfortunately not, I tried just the `text_format` arg at first, but I think `litellm` refused to play nice with it.
Serializing it ourselves might a bit more verbose but being explicit is warranted here, since we modify the Pydantic model's `model_json_schema` in `JSONAdapter` manually anyways.

Olocool17:
Thanks! I'll close this PR and spin up two new ones (see https://github.com/stanfordnlp/dspy/pull/9130 and https://github.com/stanfordnlp/dspy/pull/9131) .

PR #9128, opened 2025-12-14
Fix: enforce positive `memory_max_entries` for in-memory cache

mshr-h:
Passing invalid `memory_max_entries` to `Cache` can crash cachetool with unclear errors. This PR validates it upfront and fails with clearer messages.

Repros:

```python
from dataclasses import dataclass
from dspy.clients.cache import Cache

@dataclass
class DummyResponse:
    message: str
    usage: dict

if __name__ == "__main__":
    memory_cache = Cache(
        enable_disk_cache=False,
        enable_memory_cache=True,
        disk_cache_dir="",
        disk_size_limit_bytes=0,
        memory_max_entries=1, # negative entry size
    )

    request = {"prompt": "Hello", "model": "openai/gpt-4o-mini", "temperature": 0.7}
    value = DummyResponse(message="This is a test response", usage={"prompt_tokens": 10, "completion_tokens": 20})
    memory_cache.put(request, value)  # -> ValueError: value too large
```

```python
from dataclasses import dataclass
from dspy.clients.cache import Cache

@dataclass
class DummyResponse:
    message: str
    usage: dict

if __name__ == "__main__":
    memory_cache = Cache(
        enable_disk_cache=False,
        enable_memory_cache=True,
        disk_cache_dir="",
        disk_size_limit_bytes=0,
        memory_max_entries=None, # None
    )

    request = {"prompt": "Hello", "model": "openai/gpt-4o-mini", "temperature": 0.7}
    value = DummyResponse(message="This is a test response", usage={"prompt_tokens": 10, "completion_tokens": 20})
    memory_cache.put(request, value)  # -> TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

### DIFF
--- a/dspy/clients/__init__.py
+++ b/dspy/clients/__init__.py
@@ -14,6 +14,8 @@ logger = logging.getLogger(__name__)
 
 DISK_CACHE_DIR = os.environ.get("DSPY_CACHEDIR") or os.path.join(Path.home(), ".dspy_cache")
 DISK_CACHE_LIMIT = int(os.environ.get("DSPY_CACHE_LIMIT", 3e10))  # 30 GB default
+
+
 def configure_cache(
     enable_disk_cache: bool | None = True,
     enable_memory_cache: bool | None = True,
@@ -41,12 +43,14 @@ def configure_cache(
     )
 
     import dspy
+
     # Update the reference to point to the new cache
     dspy.cache = DSPY_CACHE
 
 
 litellm.telemetry = False
 litellm.cache = None  # By default we disable LiteLLM cache and use DSPy on-disk cache.
+
 
 def _get_dspy_cache():
     disk_cache_dir = os.environ.get("DSPY_CACHEDIR") or os.path.join(Path.home(), ".dspy_cache")
@@ -71,6 +75,7 @@ def _get_dspy_cache():
             memory_max_entries=1000000,
         )
     return _dspy_cache
+
 
 DSPY_CACHE = _get_dspy_cache()
 
--- a/tests/clients/test_cache.py
+++ b/tests/clients/test_cache.py
@@ -48,39 +48,6 @@ def test_initialization(tmp_path):
     assert memory_cache.memory_cache.maxsize == 50
     assert memory_cache.disk_cache == {}
 
-    # Test memory-only cache with invalid memory_max_entries
-    with pytest.raises(ValueError, match=r"`memory_max_entries` must be a positive number, but received -1"):
-        memory_cache = Cache(
-            enable_disk_cache=False,
-            enable_memory_cache=True,
-            disk_cache_dir="",
-            disk_size_limit_bytes=0,
-            memory_max_entries=-1,
-        )
-
-    # Test memory-only cache with invalid memory_max_entries
-    with pytest.raises(ValueError, match=r"`memory_max_entries` cannot be None. Use `math.inf` if you need an unbounded cache."):
-        memory_cache = Cache(
-            enable_disk_cache=False,
-            enable_memory_cache=True,
-            disk_cache_dir="",
-            disk_size_limit_bytes=0,
-            memory_max_entries=None,
-        )
-
-    # Test memory-only cache with unbounded size
-    import math
-    memory_cache = Cache(
-        enable_disk_cache=False,
-        enable_memory_cache=True,
-        disk_cache_dir="",
-        disk_size_limit_bytes=0,
-        memory_max_entries=math.inf,
-    )
-    assert isinstance(memory_cache.memory_cache, LRUCache)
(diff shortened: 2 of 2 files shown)


dspy/clients/__init__.py
```diff
@@ -19,7 +19,7 @@ def configure_cache(
     enable_memory_cache: bool | None = True,
     disk_cache_dir: str | None = DISK_CACHE_DIR,
     disk_size_limit_bytes: int | None = DISK_CACHE_LIMIT,
-    memory_max_entries: int | None = 1000000,
+    memory_max_entries: int = 1000000,
```
chenmoneygithub, line 24: Why changing this? `cachetools.LRUCache(maxsize=None)` is a legit usage to remove the size limit.

dspy/clients/cache.py
```diff
@@ -43,6 +43,8 @@ def __init__(
         self.enable_disk_cache = enable_disk_cache
         self.enable_memory_cache = enable_memory_cache
         if self.enable_memory_cache:
+            if memory_max_entries <= 0:
+                raise ValueError("memory_max_entries must be a positive integer")
```
chenmoneygithub: nit: backtick around variables, and let's print out the received value. 
```suggestion
                raise ValueError(f"`memory_max_entries` must be a positive integer, but received {memory_max_entries}")
```

dspy/clients/__init__.py
mshr-h: Thank you for your comment. According to the [official docs](https://cachetools.readthedocs.io/en/latest/), we need to pass `math.inf` or something similar to remove the size limit. Added that to the docstring.  @chenmoneygithub 
> Note Please be aware that *maxsize* must be a positive number. If you really want your cache to grow without bounds, use math.inf or something similar.

chenmoneygithub: Interesting, looks like the recommended approach is changed, but it's still a valid syntax to set maxsize=None

mshr-h: When the maxsize=None, the following line throws a `TypeError: '>' not supported between instances of 'int' and 'NoneType'`. Please check the 2nd repro int the PR comment. I'm using `cachetools==6.1.0`. @chenmoneygithub 
https://github.com/tkem/cachetools/blob/v6.1.0/src/cachetools/__init__.py#L73

mshr-h: Another option may be to replace `None` with `math.inf` on our side when `None` is passed to `configure_cache()`.

chenmoneygithub: Disallowing None sounds good to me, I did a second look, the one allowing None value is `@functools.lru_cache` not `cachetools.LRUCache`.

tests/clients/test_cache.py
```diff
@@ -48,6 +48,16 @@ def test_initialization(tmp_path):
     assert memory_cache.memory_cache.maxsize == 50
     assert memory_cache.disk_cache == {}
 
+    # Test memory-only cache with invalid memory_max_entries
+    with pytest.raises(ValueError):
```
chenmoneygithub: Let's also verify the error string is as expected.

dspy/clients/cache.py
```diff
@@ -43,6 +43,8 @@ def __init__(
         self.enable_disk_cache = enable_disk_cache
         self.enable_memory_cache = enable_memory_cache
         if self.enable_memory_cache:
+            if memory_max_entries <= 0:
```
chenmoneygithub: we can also add the None validation here according.

dspy/clients/__init__.py
```diff
@@ -19,7 +19,7 @@ def configure_cache(
     enable_memory_cache: bool | None = True,
     disk_cache_dir: str | None = DISK_CACHE_DIR,
     disk_size_limit_bytes: int | None = DISK_CACHE_LIMIT,
-    memory_max_entries: int | None = 1000000,
+    memory_max_entries: int | float = 1000000,
```
mshr-h: fyi: `math.inf` is float type so it should also accept it.

mshr-h:
@chenmoneygithub 
Thank you for your review! I updated the code accordingly.

dspy/clients/__init__.py
chenmoneygithub: yes this definitely reads weird in type hint, but that's sticking with cachetools' best practice, so acceptable.

mshr-h: Ok. reverted. @chenmoneygithub

mshr-h: nit: Maybe just `float` type is better than `int | float` or `int`?
https://github.com/python/typeshed/blob/main/stubs/cachetools/cachetools/__init__.pyi#L20
```python
    def __init__(self, maxsize: float, getsizeof: None = None) -> None: ...
```

PR #9130, opened 2025-12-15
Fix Responses API structured outputs

Olocool17:
Previous PR: https://github.com/stanfordnlp/dspy/pull/9123

Closes https://github.com/stanfordnlp/dspy/issues/9122. See the linked issue for an extended problem description + minimal reproducible example.

### Structured outputs fix

My proposed fix mutates the response_format request argument into the proper form if it is passed as a Pydantic model.
Documentation around this is somewhat unreliable: the litellm docs related to this are outdated/incorrect and the OpenAI docs seem to imply that passing a Pydantic model into text.format is valid, but this causes JSON serialization errors down the line.

So far, the specific formulation as presented in this PR has rendered no issues with our testing. I've also added a small unit test to verify that the chat -> responses api request conversion is valid for this case.

### DIFF
--- a/tests/adapters/test_json_adapter.py
+++ b/tests/adapters/test_json_adapter.py
@@ -978,8 +978,11 @@ def test_json_adapter_with_responses_api():
     call_kwargs = mock_responses.call_args.kwargs
     assert "response_format" not in call_kwargs
     assert "text" in call_kwargs
-    assert isinstance(call_kwargs["text"]["format"], type)
-    assert issubclass(call_kwargs["text"]["format"], pydantic.BaseModel)
+
+    assert isinstance(call_kwargs["text"]["format"], dict)
+    assert isinstance(call_kwargs["text"]["format"]["name"], str)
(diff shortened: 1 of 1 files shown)


tests/clients/test_lm.py
```diff
@@ -931,3 +931,56 @@ def test_responses_api_with_image_input():
...
+
+        # Verify the request was converted correctly
+        assert "text" in call_args
+        response_format = call_args["text"]["format"]
+
+        assert response_format == {
+            "name": TestModel.__name__,
+            "type": "json_schema",
+            "schema": TestModel.model_json_schema(),
+        }
```
TomeHirata: Can we move lines other than `lm_result = lm("What is a good test answer?", response_format=TestModel)` outside the patch block?

Olocool17: Sure thing! That makes sense.

Olocool17:
The `test_json_adapter_with_responses_api` test slipped through the cracks. Fixed it just now.

PR #9131, opened 2025-12-15
 Fix Responses API cache retrieval error

Olocool17:
Previous PR: https://github.com/stanfordnlp/dspy/pull/9123

### Cache retrieval fix

While implementing the fix outlined in https://github.com/stanfordnlp/dspy/pull/9130 I additionally stumbled on an error pertaining to responses models when using caching.
When an item is successfully retrieved from the cache, an ad-hoc attribute `.cache_hit` is created and set to `True` on the `response` object.

 Unfortunately, this is only possible for `litellm.ModelResponse` (the response for chat models) and not for `litellm.ResponsesAPIResponse` (the response for response models), because the latter is a Pydantic model without `extra="allow"` set in its config.

My proposed fix for this is to simply remove this ad-hoc attribute alltogether, since it is actually superflueous: the `response.usage` attribute is cleared on cache hit anyways, which makes `settings.usage_tracker.add_usage(self.model, dict(results.usage)` a null-op.

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -122,7 +122,6 @@ class Cache:
         if hasattr(response, "usage"):
             # Clear the usage data when cache is hit, because no LM call is made
             response.usage = {}
-            response.cache_hit = True
         return response
 
     def put(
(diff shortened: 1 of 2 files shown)


okhat:
Ah. Interesting. Maybe we should instead make the object more flexible? It's possible that some folks have been relying on the attribute this PR is removing?

PR #9141, opened 2025-12-25
Improve `dspy.primitives.module` Typing

max-muoto:
This PR improves the typing of `dspy.primitives.module` by adding complete types for the `Module` class as we're currently using Pyright + DSPy and hitting a number of false positives due to missing types .

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -1,7 +1,9 @@
 import inspect
 import logging
-from typing import Any
+from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


okhat:
Seems like a good idea but the imports are failing?

max-muoto:
> Seems like a good idea but the imports are failing?

Apologies, it doesn't look we're deferring annotations with the imports under `if TYPE_CHECKING` here will push up a quick fix.

PR #9143, opened 2025-12-25
Add typing to `dspy.datasets.dataset`

max-muoto:
This PR adds a basic level typing to the `dspy.datasets.dataset` module, primarily for external users who currently hit false positives when using the `Dataset` class. Python type-checkers will generally assume types based on defaults when not provided, meaning users will hit FPs with their type-checkers when passing in arguments like `train_seed` or `test_size`.


isaacbmiller:
LGTM. Thanks for adding!

Issue #9142, opened 2025-12-25
[Bug] `apply_sync_streaming` swallows exceptions


max-muoto:
### What happened?

When using `streamify` with `async_streaming=False`, exceptions raised in the async generator end up silently swallowed due to how `apply_sync_streaming` is implemented:

```python
def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
    """Convert the async streaming generator to a sync generator."""
    queue = Queue()  # Queue to hold items from the async generator
    stop_sentinel = object()  # Sentinel to signal the generator is complete

    # To propagate prediction request ID context to the child thread
    context = contextvars.copy_context()

    def producer():
        """Runs in a background thread to fetch items asynchronously."""

        async def runner():
            try:
                async for item in async_generator:
                    queue.put(item)
            finally:
                # Signal completion
                queue.put(stop_sentinel)

        context.run(asyncio.run, runner())

    # Start the producer in a background thread
    thread = threading.Thread(target=producer, daemon=True)
    thread.start()

    # Consume items from the queue
    while True:
        item = queue.get()  # Block until an item is available
        if item is stop_sentinel:
            break
        yield item
```

Instead of explicitly catching exceptions and putting them in the queue to propagate to the caller, exceptions simply lead to the stop sentinel being sent..

### DSPy version

3.0.4

SamSi0322:
I'll take this. The fix is to catch exceptions in the producer, put them in the queue, and re-raise on the consumer side. PR incoming.

PR #9145, opened 2025-12-29
fix(streaming): prevent silent exception swallowing in apply_sync_streaming

akshatvishu:
This PR fixes a bug where `apply_sync_streaming` silently swallowed exceptions from the background `async` generator. Previously, the `finally` block always sent a `stop_sentinel`; allowing the synchronous caller to proceed as though no error had occurred!
 
**Changes:**
- Catch `BaseException` in the producer thread and send it through the queue.    
- Use a `sentinel` tuple to distinguish exceptions from regular data (in case someone's yielding Exception objects as actual content).     
- Unwrap `anyio ExceptionGroups` when there's only a single exception—this way callers get the actual error instead of a wrapper.     
- Added test that verifies exceptions propagate correctly while status messages before the crash are still received. 

Fixes #9142

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -224,38 +224,52 @@ def streamify(
         return sync_streamer
 
 
-def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
-    """Convert the async streaming generator to a sync generator."""
-    queue = Queue()  # Queue to hold items from the async generator
-    stop_sentinel = object()  # Sentinel to signal the generator is complete
 
-    # To propagate prediction request ID context to the child thread
+def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
+    """
+    Convert an async streaming generator to a sync generator.
+    """
+    queue = Queue()
+    stop_sentinel = object()
+    exception_sentinel = object()
+
     context = contextvars.copy_context()
 
     def producer():
         """Runs in a background thread to fetch items asynchronously."""
-
         async def runner():
             try:
                 async for item in async_generator:
                     queue.put(item)
-            finally:
-                # Signal completion
                 queue.put(stop_sentinel)
+            except BaseException as e:
+                # Catch BaseException (KeyboardInterrupt, SystemExit, etc.)
+                # and transport it to the main thread via the queue.
+                queue.put((exception_sentinel, e))
 
         context.run(asyncio.run, runner())
 
-    # Start the producer in a background thread
     thread = threading.Thread(target=producer, daemon=True)
     thread.start()
 
-    # Consume items from the queue
-    while True:
-        item = queue.get()  # Block until an item is available
-        if item is stop_sentinel:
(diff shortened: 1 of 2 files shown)


jayhemnani9910:
## Tested and Verified ✅

  I tested this PR locally and verified it works correctly.

  ### Testing Results
  - `test_sync_streaming_keyboard_interrupt_propagation` ✅
  - `test_sync_status_streaming_with_error_propagation` ✅
  - Full streaming test suite (36 tests) ✅
  - Manual exception propagation test ✅

  ### Code Review

  The implementation is solid:
  1. **Exception sentinel pattern** - Using `(exception_sentinel, e)` tuple correctly distinguishes exceptions from regular yielded data
  2. **BaseException catch** - Correctly catches all exceptions including `KeyboardInterrupt`
  3. **ExceptionGroup unwrapping** - Nice touch for cleaner error messages
  4. **Thread cleanup** - Adding `thread.join(timeout=1.0)` is good practice

  Great fix! 🎉

TomeHirata:
@chenmoneygithub can you help take a look?

dspy/streaming/streamify.py
```diff
@@ -224,38 +224,52 @@ def sync_streamer(*args, **kwargs):
...
         """Runs in a background thread to fetch items asynchronously."""
-
         async def runner():
             try:
                 async for item in async_generator:
                     queue.put(item)
-            finally:
-                # Signal completion
                 queue.put(stop_sentinel)
+            except BaseException as e:
```
chenmoneygithub: Why do we only capture BaseException? I might be missing something, but I think the issue is all exceptions are swallowed by the try-finally block?

chenmoneygithub (commented):
Thanks for the PR! Left a comment about the exception handling

dspy/streaming/streamify.py
jayhemnani9910: I might be wrong here, but my understanding is that BaseException is actually the root of Python's exception hierarchy—so it should   
  catch everything, including regular Exception subclasses.                                                                             
                                                                                                                                        
  The reason I used BaseException instead of Exception was to also catch things like KeyboardInterrupt and SystemExit, which don't      
  inherit from Exception. Felt like those should propagate properly too rather than being swallowed.                                    
                                                                                                                                        
  As for the original bug—I think the issue was that the finally block always ran queue.put(stop_sentinel) regardless of whether an     
  exception occurred. The exception would still propagate in the background thread, but the main thread just saw stop_sentinel and      
  assumed everything was fine. So the exception got lost between threads.                                                               
                                                                                                                                        
  Let me know if I'm misunderstanding something though!

akshatvishu: Currently, the `finally` block hides errors by sending a "success" sentinel even when the background thread crashesWe catch `BaseException` so system signals like `KeyboardInterrupt` (Ctrl+C) actually reach the main thread instead of being mistaken for clean exits.

PR #8655, opened 2025-08-14
Replace ujson by orjson

chenmoneygithub:
This is continued work of #8584 due to the inactivity of the original contributor.

Verified that json saved with the old `ujson` path is still loadable by the `orjson` code. Specifically I optimized a dspy.ReAct and saved the state as:

```
{
  "react": {
    "traces": [],
    "train": [],
    "demos": [
      {
        "augmented": true,
        "question": "That Darn Cat! and Never a Dull Moment were both produced by what studio?",
        "trajectory": "[[ ## thought_0 ## ]]\nI need to find out which studio produced both \"That Darn Cat!\" and \"Never a Dull Moment.\" This information is likely available on Wikipedia, so I will search for it there.\n\n[[ ## tool_name_0 ## ]]\nsearch_wikipedia\n\n[[ ## tool_args_0 ## ]]\n{\"query\": \"That Darn Cat! and Never a Dull Moment studio production\"}\n\n[[ ## observation_0 ## ]]\n[1] «That Darn Cat! | That Darn Cat! is a 1965 American Walt Disney Productions thriller comedy film starring Hayley Mills (in her last of the six films she made for the Walt Disney Studios) and Dean Jones (starring in his first film for Disney) in a story about bank robbers, a kidnapping and a mischievous cat. The film was based on the 1963 novel \"Undercover Cat\" by Gordon and Mildred Gordon and was directed by Robert Stevenson. The title song was written by the Sherman Brothers and sung by Bobby Darin. The 1997 remake includes a cameo appearance by Dean Jones.»\n[2] «Never a Dull Moment (1968 film) | Never a Dull Moment is a 1968 American comedy crime film from Walt Disney Productions starring Dick Van Dyke and Edward G. Robinson and directed by Jerry Paris. The script by AJ Carothers was based on a novel by John Godey. The supporting cast features Dorothy Provine, Henry Silva, Slim Pickens and Jack Elam.»\n[3] «That Darn Cat (1997 film) | That Darn Cat is a 1997 American mystery comedy film starring Christina Ricci and Doug E. Doug. It is a remake of the 1965 film \"That Darn Cat! \", which in turn was based on the book \"Undercover Cat\" by Gordon and Mildred Gordon. It is directed by British TV veteran Bob Spiers (most famous for \"Fawlty Towers\", as well as \"Spice World\") and written by Scott Alexander and Larry Karaszewski, best known for \"Ed Wood\" and the first two \"Problem Child\" films.»",
[...]

PR #9148, opened 2026-01-02
Replace ujson with orjson in docs

togimoto:
Replaces any remaining mention of `ujson` in the codebase to `orjson`.

As per #8655, DSPy switched from `ujson` to `orjson`.

But `ujson` instances still remain in the documentation (and in a piece of commented-out code in tests), causing the error `ModuleNotFoundError: No module named 'ujson’` when any example in the documentation using `ujson` is run.


chenmoneygithub (approved):
Thanks for the PR, LGTM!

PR #9169, opened 2026-01-07
Patch `dspy.gepa` to handle `list[dict]` output from a `dspy.LM`

### LINKED ISSUE #9168: [Bug] `dspy==3.1.0` no longer guarantees a `str` output from `dspy.LM.__call__`, but GEPA assumes it and crashes
### What happened?

The recent release of `dspy==3.1.0` significantly changes the output of a `dspy.LM` from a `str` to possibly a `dict[str, str]` for reasoning models, that includes two keys: "text" and "reasoning_output".

Specifically, `gepa` expects a `str` here:
https://github.com/gepa-ai/gepa/blob/1b5eff5133be1015210e0512953c25a4b85ad454/src/gepa/proposer/reflective_mutation/base.py#L48

This happens for the OpenAI family of models, and many others.

### Steps to reproduce

Pre-requisites: `uv` and `uv python install 3.13`

The following minimal example runs fine using `uv run --with gepa==0.0.24 example.py`:

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "dspy==3.0.4",
# ]
# ///

import os
from typing import Optional

import dspy
from dspy.teleprompt.gepa.gepa_utils import (
    DSPyTrace,
    ScoreWithFeedback,
)

lm = dspy.LM(
    model="openrouter/nvidia/nemotron-3-nano-30b-a3b:free",
    api_key=os.environ["OPENROUTER_API_KEY"],
    temperature=1.0,
    cache=False,
)
dspy.configure(lm=lm)

reflection_lm = dspy.LM(
    model="openrouter/nvidia/nemotron-3-nano-30b-a3b:free",
    api_key=os.environ["OPENROUTER_API_KEY"],
    temperature=1.0,
    cache=False,
)

    input: str = dspy.InputField()
    output: str = dspy.OutputField()


def metric(
    gold: dspy.Example,
    pred: dspy.Prediction,
    trace: Optional[DSPyTrace] = None,
    pred_name: Optional[str] = None,
    pred_trace: Optional[DSPyTrace] = None,
) -> float | ScoreWithFeedback:
    return 0.0

student = dspy.Predict(MySignature)

gepa = dspy.GEPA(
    metric=metric,
    track_stats=True,
    max_metric_calls=2,
    reflection_lm=reflection_lm
)
trainset = [dspy.Example(input="input", output="output").with_inputs("input")]
new_prog = gepa.compile(student, trainset=trainset, valset=trainset)
```

However, upgrading to `dspy==3.1.0` by editing the script metadata:
```python
...
# dependencies = [
#     "dspy==3.1.0",
# ]
# ///

...
```

Loudly crashes with:
```python
AttributeError: 'dict' object has no attribute 'strip'
```

### DSPy version

3.1.0

mariusarvinte:
Fixes #9168 

The current reflection model in GEPA always expects a `str` to come from the `dspy.LM`:
https://github.com/gepa-ai/gepa/blob/1b5eff5133be1015210e0512953c25a4b85ad454/src/gepa/proposer/reflective_mutation/base.py#L27-L28

This PR modifies `dspy.gepa` to extract the `"text"` values - assuming they exist - in case a model returns a `list[dict]`.

### DIFF
--- a/tests/teleprompt/test_gepa_instruction_proposer.py
+++ b/tests/teleprompt/test_gepa_instruction_proposer.py
@@ -347,7 +347,8 @@ def test_default_proposer(reasoning: bool, caplog):
     )
 
     with caplog.at_level(logging.INFO, logger="dspy.teleprompt.gepa.gepa"):
-        # Let logs propagate up to root
+        # Let logs propagate up to root because gepa uses try-catch and logs the error
+        # https://github.com/gepa-ai/gepa/blob/1b5eff5133be1015210e0512953c25a4b85ad454/src/gepa/proposer/reflective_mutation/reflective_mutation.py#L128
         dspy_logger = logging.getLogger("dspy")
         original_propagate = dspy_logger.propagate
         dspy_logger.propagate = True


dspy/clients/base_lm.py
```diff
@@ -220,8 +220,8 @@ def _process_completion(self, response, merged_kwargs):
 
             outputs.append(output)
 
-        if all(len(output) == 1 for output in outputs):
-            # Return a list if every output only has "text" key
+        if all(len(output) == 1 for output in outputs) or merged_kwargs.get("text_output", False):
```
TomeHirata: Can you not modify BaseLM and handle the list[dict] in dspy.GEPA?

mariusarvinte: Reverted to a self-contained `dspy.gepa` patch

Fwiw the `BaseLM` change in 3.1.0 is breaking for any applications that use direct calls of `dspy.LM` instances, would be nice to give users a chance to specify string-only outputs for backward compatibility.

mariusarvinte: @TomeHirata does this look good? GEPA is still non-functional with reasoning models

TomeHirata: `dspy.LM` has been returning list[dict] for some cases before 3.1.0. It just started to parse reasoning response in that version.

>GEPA is still non-functional with reasoning models

Why is this the case?

mariusarvinte: > `dspy.LM` has been returning list[dict] for some cases before 3.1.0. It just started to parse reasoning response in that version.
> 
> > GEPA is still non-functional with reasoning models
> 
> Why is this the case?

Do you have a working example of GEPA and said output? I doubt GEPA was working before, with any kind of `dict`

TomeHirata: I think it's a bug in GEPA<>DSPy integration. If it doesn't handle list[dict], it has been nonfunctional for some cases. Current fix looks good to me. Did you test this branch with GEPA and reasoning models?

mariusarvinte: I ran the example in https://github.com/stanfordnlp/dspy/issues/9168 with 
```
OPENROUTER_API_KEY="some-key" uv run --with /path/to/local/dspy example.py
```

eventually producing
```
2026/01/14 18:37:13 INFO dspy.teleprompt.gepa.gepa: Iteration 1: Proposed new text for self: You are given a JSON object that contains exactly one field named **`input`**.
Your task is to produce a new JSON object that contains exactly one field named **`output`**, where the value of **`output`** must be identical to the value of the **`input`** field you received.

**Requirements:**
1. The output must be a valid JSON object.
2. It must contain only the key **`output`**; no additional keys or properties are allowed.
3. The value associated with **`output`** must be a direct copy of the value that was associated with **`input`** in the input object.
4. If the input object does not contain the **`input`** key, output an empty JSON object (`{}`).
5. Do not include any explanatory text, comments, or extra whitespace beyond what is necessary for valid JSON.

**Example:**
- Input: `{"input": "hello"}`
- Output: `{"output": "hello"}`
```

Nemotron-30B-A3B is a reasoning model.

TomeHirata: Great, can you add tests for this change?

mariusarvinte: @TomeHirata Added a test that now passes, but fails on `main`.

To reproduce the failing test on `main`, run:
```
git checkout main
git checkout fix_gepa_adapter_reasoning -- tests\teleprompt\test_gepa_instruction_proposer.py dspy\utils\dummies.py

uv run pytest -k test_gepa_instruction_proposer
```

This will fail on `main` on the newly added test variant with `reasoning=True`.
```
FAILED tests/teleprompt/test_gepa_instruction_proposer.py::test_default_proposer[True] - assert 'Exception d...ion/proposal' not in 'INFO     ds... candidate\n'
1 failed, 4 passed, 1 skipped, 926 deselected, 1 warning in 1.86s 
```

But will pass on this branch
```
5 passed, 1 skipped, 926 deselected, 1 warning in 1.66s
```

mariusarvinte: I think the failing test needs a re-trigger, https://github.com/stanfordnlp/dspy/pull/9169/commits/f0a97ef4fc2e15f36464f15b5a2d77954180a9c9 seems to pass it and the change in the latest https://github.com/stanfordnlp/dspy/pull/9169/commits/49bbc9443410d40aa0e656145ffb9a0e8a65579a is immaterial for it

okhat:
Hey @mariusarvinte , thanks so much for catching this and preparing a great PR for it, as well as tests. Looks great.

PR #9193, opened 2026-01-12
feat(rlm): Add RLM Module and improve PythonInterpreter

isaacbmiller:
# RLMs

RLMs with only a single layer of recursion through an LLM prompting tool, as done in the RLM paper, are effectively equivalent to a ReAct agent with a REPL tool. The key difference is that inputs are injected as variables so the LLM does not need to rewrite all the contents if it wants to use these inputs inside a REPL. For smaller prompts/contexts, this difference is negligible. For larger contexts is where we start to see real benefits from this approach.

We see this especially inside a benchmark like Oolong, where the LLM is given up to 4MM tokens of context, and needs to process them programmatically. It can split up the context and run operations on smaller chunks of it.

The approach for implementing it inside of dspy follows a similar pattern to ReAct. We create a tool calling agent whose only tool is to write code. Inside of the code environment that the agent is dropped into has all of the inputs in it as parts of a context variable by default, along with a description of what those inputs are, where the description is given by the signature.

## Approach

We use Deno with a WASM runtime inside to run our PythonInterpreter.

We pass messages to and from the runtime via stdin and stdout.

You can pass any tool to the agent, but it must be a python callable, which will be used by the agent as a callable function in the REPl, rather than a native model tool call.

Importantly, the agent needs a way to escape the REPL loop. As is done in the [RLM repo](https://github.com/alexzhang13/rlm/tree/main), we implement FINAL and FINAL_VAR functions. FINAL is used to submit an answer directly, and FINAL_VAR is used to submit answers inside a variable. If this result is not parseable to the output types, we raise an error and ask the LM to retry.

If the RLM reaches `max_iters` without calling FINAL, we pass the entire trajectory to an `extract` module, similar to how `dspy.ReAct` has an extract module, and ask it to parse the output fields.

### Tool Design

The default tools that the model is provided are an `llm_query`, and an `llm_query_batched` tool. In order to avoid excessive llm calls, we provide a parameter called `MAX_LLM_CALLS`.

These tools are hard coded into the system prompt.

If max_iters(default 20) is reached, we use an 'extract' tool, where we create a signature with the REPL history as the inputs, and the original signature outputs as the outputs. For the extract step, there is no access to submit a variable.

#### Interpreter

The bulk of the work in this PR is related to the `Interpreter`. This is a Protocol for interacting with code environments. We primarily need an `execute()` function. I update the DSPy `PythonInterpreter` to handle tool calls.

The default interpreter is `PythonInterpreter`, This uses deno in order to create a local WASM runtime, which provides a secure way to run python code locally. There is no file read or write access unless the path is specified in your `PythonInterpreter` initialization. All network calls are also sandboxed.

The most complicated parts of a `Interpreter` are (1) how tool calls work, and (2) how the final variables are written and read. 
(1) Tool Calls:
For the `PythonInterpreter`, we can pass any python callable as a tool. This does NOT allow the model to run this callable directly. When this function is called, the WASM runtime outputs a payload to stdout via console.log. The host machine's python process is watching stdout, and will run this function with the given args. When this function returns, it will write to stdin, which the pyodide process is watching for, and will then return the value into the REPL as the result of the function (notably not just plaintext in the prompt!).

Importantly, you can use any python callable for this. By default, we only provide `llm_query` and `llm_query_batched`.

(2) Final Variables:
For `FINAL` and `FINAL_VAR`, we raise an exception which Deno catches and outputs our results in for `PythonInterpreter`. Our setup has explicit passthroughs for Exceptions, so when a FinalAnswer error is raised, Deno outputs it to stdout. The host process will parse this output as json, and extract the value, ensuring that the final answer matches the output fields of the signature.

TODO: 
- [ ] Write docs
- [ ] Test Optimization
- [ ] Create example notebook showing usage
- [ ] Test Multi turn chat

Limitations:
- Pydantic objects not yet supported as output types.
- non-python based tools not yet supported
- Images are not supported.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -49,6 +49,7 @@ Available:
 - `llm_query_batched(prompts)` - query multiple prompts concurrently (much faster for multiple queries)
 - `print()` - ALWAYS print to see results
 - `FINAL({final_output_names})` - submit final answer when done
+- `FINAL_VAR({final_var_output_names})` - submit final answer using variable names
 - Standard libraries: re, json, collections, math, etc.
 
 IMPORTANT: This is ITERATIVE. Each code block you write will execute, you'll see the output, then you decide what to do next. Do NOT try to solve everything in one step.
@@ -58,7 +59,7 @@ IMPORTANT: This is ITERATIVE. Each code block you write will execute, you'll see
 3. VERIFY BEFORE SUBMITTING - If results seem wrong (zeros, empty, unexpected), reconsider your approach.
 4. USE llm_query FOR SEMANTICS - String matching finds WHERE things are; llm_query understands WHAT things mean.
 
-You have max {max_llm_calls} sub-LLM calls. When done, call FINAL() with your answer."""
+You have max {max_llm_calls} sub-LLM calls. When done, call FINAL() or FINAL_VAR() with your answer."""
 
 # Pattern to match markdown code fences: ```python\n...\n``` or ```\n...\n```
 _CODE_FENCE_PATTERN = re.compile(r"^```(?:python|py)?\s*\n(.*?)\n```\s*$", re.DOTALL)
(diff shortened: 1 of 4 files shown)


isaacbmiller:
@codex

noahvandal:
I was going to see if there was an RLM module for DSPy in the works, otherwise I was going to make one myself! However this seems to be a nice implementation of it.

isaacbmiller:
@codex

okhat:
This is really great @isaacbmiller . LGTM except for taking up `dspy.Interpreter` at the top-level.

PR #9195, opened 2026-01-13
Fix(gepa): handle reasoning LM outputs in GEPA optimizer

mariam851:
Description
This PR fixes the AttributeError: 'dict' object has no attribute 'strip' that occurs when using reasoning models (like OpenAI o1) as a reflection_lm in GEPA. These models return dict or Prediction objects which the core GEPA engine doesn't expect.

Key Changes
Normalization: Added a wrapper in GEPA.compile to ensure reflection_lm output is always a str. It handles list, dict, and dspy.Prediction types.

Testing: Added tests/teleprompt/test_gepa.py to verify the fix across different output formats.

### DIFF
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -647,13 +647,21 @@ class GEPA(Teleprompter):
         # Build the seed candidate configuration
         seed_candidate = self._build_seed_candidate(student)
 
+        def reflection_wrapper(prompt):
+            res = self.reflection_lm(prompt)
+            if isinstance(res, list) and len(res) > 0:
+                res = res[0]
+            if hasattr(res, 'answer'):
+                return str(res.answer)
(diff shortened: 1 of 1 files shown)


okhat:
Thanks so much @mariam851 !  This may actually be a cleaner fix than the PR we merged yesterday. Sadly we saw the other one first..

mariusarvinte:
> Thanks so much @mariam851 ! This may actually be a cleaner fix than the PR we merged yesterday. Sadly we saw the other one first..

The "cleanest" fix is https://github.com/stanfordnlp/dspy/pull/9169/changes/35919ded2a4983f8f40f02d92bd8992945358402 or something in the spirit of, that deals with `dspy.BaseLM` - but this was explicitly requested to not be changed

And that fix would also render https://github.com/stanfordnlp/dspy/pull/9219 un-necessary

PR #9199, opened 2026-01-14
Expose timeout and straggler_limit params in Parallel

halfprice06:
ParallelExecutor has `timeout` (default 120s) and `straggler_limit` (default 3) params but Parallel doesn't expose them.

If you run ≤3 tasks that each take >2 min, they get resubmitted as "stragglers" after the timeout, causing duplicate work.

This adds both params to `Parallel.__init__()` and `Module.batch()`. Setting `timeout=0` disables resubmission.


okhat:
Thank you. LGTM

Issue #9178, opened 2026-01-08
[Bug] KeyError: 'topk' running `dspy.ColBERTv2`


brenorb:
### What happened?

I just tried to run this code snippet from [docs](https://dspy.ai/#__tabbed_2_2) and got the following key error. 
Is there some issue with colbert? 

Error:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[20], line 8
      5 rag = dspy.ChainOfThought("context, question -> response")
      7 question = "What's the name of the castle that David Gregory inherited?"
----> 8 rag(context=search_wikipedia(question), question=question)

Cell In[20], line 2, in search_wikipedia(query)
      1 def search_wikipedia(query: str) -> list[str]:
----> 2     results = dspy.ColBERTv2(url="http://20.102.90.50:2017/wiki17_abstracts")(query, k=3)
      3     return [x["text"] for x in results]

File ~/Documents/code/.venv/lib/python3.12/site-packages/dspy/dsp/colbertv2.py:32, in ColBERTv2.__call__(self, query, k, simplify)
     30     topk: list[dict[str, Any]] = colbertv2_post_request(self.url, query, k)
     31 else:
---> 32     topk: list[dict[str, Any]] = colbertv2_get_request(self.url, query, k)
     34 if simplify:
     35     return [psg["long_text"] for psg in topk]

File ~/Documents/code/.venv/lib/python3.12/site-packages/dspy/clients/cache.py:249, in request_cache.<locals>.decorator.<locals>.sync_wrapper(*args, **kwargs)
    246 # Otherwise, compute and store the result
    247 # Make a copy of the original request in case it's modified in place, e.g., deleting some fields
    248 original_request = copy.deepcopy(modified_request)
--> 249 result = fn(*args, **kwargs)
    250 # `enable_memory_cache` can be provided at call time to avoid indefinite growth.
    251 cache.put(original_request, result, ignored_args_for_cache_key, enable_memory_cache)

File ~/Documents/code/.venv/lib/python3.12/site-packages/dspy/dsp/colbertv2.py:54, in colbertv2_get_request_v2_wrapped(*args, **kwargs)
     52 @request_cache()
     53 def colbertv2_get_request_v2_wrapped(*args, **kwargs):
---> 54     return colbertv2_get_request_v2(*args, **kwargs)

[...]

PR #9223, opened 2026-01-21
fix(gepa): Remove enable_tool_optimization feature

### LINKED ISSUE #9202: [Bug] dspy.ReAct auto-generated tool schema lost when instruction is modified
### What happened?

`dspy.ReAct` auto-populates tool schema (name, args, types) into the instruction at `__init__`:
```
(1) search, whose description is <desc>...</desc>. It takes arguments {'queries': {'type': 'array', ...}}.
```

However, this instruction is baked once and never regenerated. When:
1. `Tool.desc` or `Tool.args` are updated - instruction remains stale
2. `with_instructions()` is called - auto-generated tool schema is completely lost

This breaks GEPA tool optimization - when reflection LM proposes new instructions and GEPA applies them, the optimized program loses tool schema, causing tool call failures.

### Reproduction

```python
import dspy

def search(queries: list[str]) -> str:
    '''Initial description.'''
    return f'results for {queries}'

react = dspy.ReAct('task -> answer', tools=[search])

# Initial - tool schema auto-populated in instruction
assert 'queries' in react.react.signature.instructions  # ✓ schema baked in
assert 'Initial description' in react.react.signature.instructions  # ✓ desc baked in

# Update tool description
react.tools['search'].desc = 'Updated description.'
assert 'Initial description' in react.react.signature.instructions  # stale
assert 'Updated description' not in react.react.signature.instructions  # not updated

# with_instructions() loses schema entirely
react.react.signature = react.react.signature.with_instructions('New instruction.')
assert 'queries' not in react.react.signature.instructions  # schema lost
```

### Expected
- Tool updates should reflect in instruction, OR
- `with_instructions()` should preserve/regenerate tool schema

### Actual
Instruction is baked at `__init__`, never regenerated.

### DSPy version

3.1.0b1

Ju-usc:
Closes #9202

Removes the `enable_tool_optimization` feature (PR #8928). Mutating tool objects directly is not the right approach — GEPA's default instruction optimization is already general enough to optimize tool-using layers (e.g., ReAct) via their predictor instructions without special-casing tool modules.


Ju-usc:
@chenmoneygithub Working on a fix for the duplication issue that codex flagged, but running into a tricky part—would really appreciate your input.

To avoid duplicating the tool schema, we need to strip the existing block before re-appending a fresh one. The challenge is reliably detecting where the auto-generated schema starts and ends. 

One approach is finding the "When providing next_tool_args..." line and walking backwards over tool lines. But if GEPA proposes instructions that happen to look like tool lines (e.g., "(1) First, analyze the query..."), we'd strip them incorrectly.

A few options:

**Option A:** More reliable tool line detection (e.g., check for `whose description is` or `It takes arguments`). But I'm not fully sure this is enough since GEPA can propose instructions that may incorrectly trigger tool line detection. 

**Option B:** Restructure so tool schema comes after the anchor line:
```
{base instruction}
When providing `next_tool_args`, the value inside the field must be in JSON format:
(1) search, whose description is ...
(2) calculate, whose description is ...
```
Then we can safely strip everything after the anchor. But we need to touch dspy.ReAct.

**Option C:** Use XML tags to delimit:
```
{base instruction}
<tool_schema>
(1) search, whose description is ...
When providing `next_tool_args`, the value inside the field must be in JSON format
</tool_schema>
```

Any suggestions? Thank you for review!

chenmoneygithub:
@Ju-usc  I think you can enforce the same data format in the [signature](https://github.com/stanfordnlp/dspy/blob/1b7b875cb74d9b82916d49f4e454065b78f974cb/dspy/teleprompt/gepa/instruction_proposal.py#L319) used for tool proposing? You can try letting the LM follow the format, or hardcode the string to separate the instruction and tool descriptions. 

This does trigger my another question - for ReAct optimization, since the `ReAct.react` already incorporates the tool info in its instruction, would it be possible to just write your proposer to do inline adjustments? Would it work worse than the current approach?

Ju-usc:
> @Ju-usc I think you can enforce the same data format in the [signature](https://github.com/stanfordnlp/dspy/blob/1b7b875cb74d9b82916d49f4e454065b78f974cb/dspy/teleprompt/gepa/instruction_proposal.py#L319) used for tool proposing? You can try letting the LM follow the format, or hardcode the string to separate the instruction and tool descriptions.
> 
> This does trigger my another question - for ReAct optimization, since the `ReAct.react` already incorporates the tool info in its instruction, would it be possible to just write your proposer to do inline adjustments? Would it work worse than the current approach?

Thanks for the suggestion! Quick clarification - by "inline adjustments," do you mean having the proposer optimize `react.react`/`react.extract` instructions directly (which already contain embedded tool schema), rather than separately managing `dspy.Tool` objects?

If so, I like this direction for the future. I think it would simplify a lot of internal logic of gepa tool optimization and make a general tool proposer easier to implement - I previously attempted to extract `dspy.Tool` instances from tracing for custom tool-call agents and it got complex. With inline adjustments, we could just rely on existing predictor signatures regardless of agent type.

My main concern: if the reflection LM omits or mangles the tool schema in its proposal, that candidate fails all tool calls → 0.0 score → wasted iteration. With many tools, this risk increases. Though maybe the genetic algorithm + strong reflection LM would self-correct over iterations?

This brings me to a real question: do you know of a cheap/standard test environment for tool-calling agents? I have been testing on BrowseComp+ but it is costly. Would help to iterate faster on proposer designs.

For now, I will proceed with the structured format approach (hardcode separator between instruction and tool descriptions) to keep things deterministic. We can revisit the inline approach once we have a good test environment.

Ju-usc:
Ran a quick experiment with this fix - modified the [DSPy agents tutorial](https://dspy.ai/tutorials/agents/) to use HoVer dataset + BM25S local retrieval (ColBERTv2 server is down, see #9178).

Baseline: 27.7% → After GEPA with `enable_tool_optimization=True`: **35.3%**

The optimized tool descriptions work well, though a bit verbose since we append full tool schema after `with_instructions()`. I'm optimistic about trying the inline approach you suggested to simplify internals.

Notebook: https://gist.github.com/Ju-usc/3ad7783179adf45c6a02b85ba90a9881

Ju-usc:
@chenmoneygithub I'm wondering about the real benefits of mutating Tool objects here - I think I overcomplicated the implementation. Seems like if we carefully design the ToolProposer signature to capture tool specifics well, we could keep the internals much simpler - and it would generalize to non-ReAct agents that use tools too.

Thinking the next step is:
- Find a good experiment environment to stress test for Non-ReAct, ReAct agent (maybe this above notebook setup is good starting point? LMK if you are aware of better setup)
- Compare vanilla GEPA vs GEPA with improved ToolProposer to isolate the actual usefulness of tool-specific optimization (since the only difference would then be the ReAct-specific handling + ToolProposer prompt)
- Measure cost, runtime, and accuracy (anything else?)

Do you have any thoughts on this direction or suggestions for next steps?

Thank you for review!

Ju-usc:
@codex review

chenmoneygithub:
@Ju-usc  Thanks for the update!

>  by "inline adjustments," do you mean having the proposer optimize react.react/react.extract instructions directly

Yes exactly, my gut feeling is the powerful models like 4.5-opus should be able to understand our goal pretty well, so worth giving it a try! 

> Find a good experiment environment to stress test for Non-ReAct, ReAct agent (maybe this above notebook setup is good starting point? LMK if you are aware of better setup)

Your notebook setup looks good to me, we can see how far we can push it. 

> My main concern: if the reflection LM omits or mangles the tool schema in its proposal, that candidate fails all tool calls → 0.0 score → wasted iteration. With many tools, this risk increases. Though maybe the genetic algorithm + strong reflection LM would self-correct over iterations?

That's a valid concern, I would suggest just giving it a try. Reasoning models are doing extremely well in breaking down the prompts, so my feeling is as long as we write clearly in the proposer "do not change the schema", then the problem should rise rarely.

So my overall suggestion is giving the inline optimization a try and see how far we can push it based on your notebook. The main difference from generic GEPA optimizer is here we are plugging in a ToolProposer for specific purposes. Looking forward to your findings!

okhat:
I don't agree with this PR. We should not be modifying things in-place silently.......... this can be a mess for parallelism.

Things *are* baked in at `init` for a reason. It's not an oversight.

Ju-usc:
I wanted to share an update on the tool optimization pr.

Ran some experiments comparing vanilla GEPA vs the tool proposer inline approach. tried several prompt versions for the tool proposer but didn't find meaningful gain against vanilla GEPA.

**Experiment Setup**
- Task: ReAct agent with `search` tool, HotPotQA-style questions
- Student LM: `llama-3.2-3b-instruct` 
- Reflection LM: `grok-4.1-fast` 
- GEPA config: 3 candidates, 3 iterations, 10 train / 300 eval examples / component_selector = all
- Tried multiple prompt variations for the tool proposer

| | Baseline | Vanilla GEPA | Tool Optimization |
|---|---|---|---|
| Accuracy range | 23–28% | **35–39%** | 21–32% |

Inline tool optimization approach did not outperformed vanilla GEPA across any prompt variation I tried.

The only structural difference in my test setup was: GEPA's default proposer takes `extract` and `react` predictors separately and optimizes them independently. on the other hand, the tool optimization proposer takes the extract trajectory and co-optimizes both `extract` and `react` together. both approaches update instructions only — no tool mutations.

i'm honestly surprised that default GEPA works better empirically in this setup. I thought in theory co-optimizing for ReAct makes sense, but i think this turned out to be a premature feature — i couldn't find a meaningful way to outperform the default proposer across multiple iterations.

i'm leaning toward removing this feature and cleaning up the codebase for a few reasons:
- I saw on discord that @LakshyAAAgrawal plans a major GEPA overhaul soon, so i think it's better to remove this and keep the codebase clean for future features rather than ship something that doesn't show clear gains
- I believe future directions like supporting RLM module for reflection would be a better investment than this approach (I think this is very exciting idea)
- I also didn't fully internalize how `dspy.ReAct` works — tool descriptions are actually baked inside the signature's instructions layer, and that's where the entire prompt for react lives. so it doesn't really make sense to touch tool objects directly, both for parallelism reasons in GEPA and just simplicity. 
- Vanilla GEPA is really general and no need to support for separate tool optimization feature — it can optimize tool description already.

I cleaned up the mess I created in this PR. thank you @okhat and @chenmoneygithub for all the guidance throughout — hopefully i can make a better useful contribution next time!

PR #9231, opened 2026-01-22
feat(RLMs): Fix code fence parsing

isaacbmiller:
The problem: The history was storing the raw LLM output, which often included markdown code fences like ```python ... ```. This added noise to the context the LLM sees on every subsequent iteration.

The fix: Store only the stripped code (without markdown fences) in history. The code was already being stripped before execution via _strip_code_fences() rlm.py:69-74, but the raw version was being saved to history.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -448,22 +448,22 @@ class RLM(Module):
         result: Any,
         history: REPLHistory,
         output_field_names: list[str],
-        code: str,
     ) -> Prediction | REPLHistory:
         """Process interpreter result, returning Prediction if final, else updated history.
 
         This shared helper reduces duplication between sync and async execution paths.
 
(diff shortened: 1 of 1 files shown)


isaacbmiller:
@codex

dspy/predict/rlm.py
```diff
@@ -448,34 +448,36 @@ def _process_execution_result(
...
         output_field_names: list[str],
+        code: str,
     ) -> Prediction | REPLHistory:
         """Process interpreter result, returning Prediction if final, else updated history.
 
         This shared helper reduces duplication between sync and async execution paths.
 
         Args:
-            pred: The prediction containing reasoning and code attributes
+            pred: The prediction containing reasoning attribute
```
okhat: If this is intentional, then all LGTM

PR #9234, opened 2026-01-23
fix(RLM): no longer get stuck on imports

isaacbmiller:
The DSPy code interpreter communicates with a sandboxed Deno/Pyodide environment via JSON-RPC messages over stdout. However, Pyodide can emit non-JSON messages (e.g., package loading status) that previously caused incorrect behavior.

Before: Non-JSON lines were treated as valid output and returned to the caller.

After: Non-JSON lines are logged and skipped; the interpreter continues waiting for a proper JSON-RPC response.

```
# Before (in execute loop):
try:
    msg = json.loads(output_line)
except json.JSONDecodeError:
    return output_line  # BUG: returns noise as result
# After:
if not output_line.startswith("{"):
    continue  # skip non-JSON lines
try:
    msg = json.loads(output_line)
except json.JSONDecodeError:
    continue  # skip malformed JSON too
```

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -479,7 +479,7 @@ class PythonInterpreter:
 
             # Skip non-JSON lines (e.g., Pyodide package loading messages)
             if not output_line.startswith("{"):
-                logger.info(f"Skipping non-JSON output: {output_line}")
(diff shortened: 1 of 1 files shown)


okhat:
LGTM

Issue #7902, opened 2025-03-04
[Feature] Thread for Async - Description of our production workflows


williambrach:
### What feature would you like to see?

Hi @chenmoneygithub,

I am creating a feature request as you requested in Discord - in the DSPy async thread.

I haven't deeply explored the current dspy.asyncify implementation, so I can't request specific features, but I think I can provide you with our production workflows and how we use DSPy. All of these should be as performant as possible and with async support and stable. 

## Inference workflow

This is the simple, default use case. This represents about 70% of our production use cases, where we just "make an API call to our litellm server".

```python
class Extractor(dspy.Module):
    def __init__(self) -> None:
        self.extract = dspy.Predict(ExtractorSignature)
        
    def forward(self, passage: str) -> str:
        response = self.extract(passage=passage)
        return response


dspy.settings.configure(lm=lm, async_max_workers=8)
extractor = Extractor()
extractor = dspy.asyncify(extractor)

r = await extractor(passage="...")
```

## "More advanced" inference workflow

This is a use case where we combine multiple Signatures in one module, which is also common in our production settings. I'm intentionally not using a multi-module approach (e.g., loading a trained module within a module). I'm trying to combine module calls in separate async function and I am asyncifying all modules. If it's possible to have one main module where I could load multiple modules from `.json` and use their `forward` functions, that would be awesome.

```python
class Extractor(dspy.Module):
    def __init__(self) -> None:
        self.extract = dspy.Predict(ExtractorSignature)
        self.find_items = dspy.ChainOfThought(FindSignature)
        
    def forward(self, passage: str, find : bool = True) -> str:
        response = self.extract(passage=passage)
        if find:
             response = self.find_items(response=response)
        return response

dspy.settings.configure(lm=lm, async_max_workers=8)
extractor = Extractor()
extractor = dspy.asyncify(extractor)

r = await extractor(passage="...")
```

## Batch workflow

The last use case we deal with is batch inference via `Evaluate`. Performance is okay, but a dedicated interface would be nice. We typically process about 5k - 10k dspy.Examples per batch. We mainly use programs from the "inference workflows" above, and we chain them to maintain at least some checkpoints. This is a approach I use also a lot in running experiments for my research mostly. 

```python
evaluate = dspy.Evaluate(
    devset=batch, # list of dspy.Examples
    metric=lambda x, y: True,
    num_threads=20,
    display_progress=False,
    return_outputs=True,
)
_, outputs = evaluate(program)
[...]

Issue #1786, opened 2024-11-11
support for async dspy.LM


JacksonCakes:
Hi, does Dspy currently support the asynchronous version of dspy.LM?
which might be equivalent to the following:
```
async def call_llm(prompt: str,file:str, cur_type: str):
    headers = {"Content-Type": "application/json"}
    res_args = {
        "model": LLM_NAME,
        "messages": [
            {
                "role": "user",
                "content": prompt,
            },
        ],
        "temperature": 0,
    }
    async with httpx.AsyncClient(timeout=httpx.Timeout(LLM_TIMEOUT)) as client:
        response = await client.post(LLM_URL, json=res_args, headers=headers)
    response = response.json()["choices"][0]["message"]["content"].strip()
    
task = [call_llm(x) for x in item] 
result = asyncio.gather(*task)
```

okhat:
Hey @JacksonCakes ! See #1734 and let us know what you think!

JacksonCakes:
Thanks!! That's exact what I am looking for!

okhat:
Wonderful. May I ask what you need async for? We're considering the right form of async to merge.

JacksonCakes:
Any LLM call that can be run in parallel would be super useful here! A great use case from my side is finding the right SQL table based on user queries from a huge database. With so many tables, compress every table metadata into single prompt or doing this sequentially just isn't practical

okhat:
Would you find this #1690 useful? We're probably going to go with that.

JacksonCakes:
Yup, I think that would work too! But I am just curious why multi-threaded is preferred over #1734 ?

PR #9235, opened 2026-01-24
feat(async): Add native aforward() to remaining modules

### LINKED ISSUE #1975: Async support in dspy.ReAct

trmanish:
## Summary

Adds `aforward()` implementations to DSPy modules that were missing async support, completing the native async API surface.

## Motivation

DSPy already has async infrastructure (`LM.aforward()`, `Module.acall()`, `Predict.aforward()`, `ReAct.aforward()`), but several modules were missing `aforward()`, forcing users to use the less idiomatic `dspy.asyncify()` wrapper.

**Related Issues:** #1975, #7902, #1786

## Changes

Added `aforward()` to:
- `MultiChainComparison`
- `BestOfN`
- `ProgramOfThought`
- `CodeAct`
- `Refine`

## Usage

**Before:**
```python
async_module = dspy.asyncify(my_module)
result = await async_module(question="...")
```

**After:**
```python
result = await my_module.acall(question="...")
```

## Testing

Added `tests/predict/test_async_modules.py` with unit tests for new async methods.

Closes #1975

### DIFF
--- a/dspy/predict/best_of_n.py
+++ b/dspy/predict/best_of_n.py
@@ -86,3 +86,38 @@ class BestOfN(Module):
         if best_trace:
             dspy.settings.trace.extend(best_trace)
         return best_pred
+
+    async def aforward(self, **kwargs):
+        lm = self.module.get_lm() or dspy.settings.lm
+        start = lm.kwargs.get("rollout_id", 0)
+        rollout_ids = [start + i for i in range(self.N)]
+        best_pred, best_trace, best_reward = None, None, -float("inf")
+
+        for idx, rid in enumerate(rollout_ids):
+            lm_ = lm.copy(rollout_id=rid, temperature=1.0)
+            mod = self.module.deepcopy()
+            mod.set_lm(lm_)
+
+            try:
+                with dspy.context(trace=[]):
+                    pred = await mod.acall(**kwargs)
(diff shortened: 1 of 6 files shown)


okhat:
Thank you @trmanish . IMO there surely has to be a better way than just adding so much code to every single module :(

Is Python async really that messed up that we can't do this without repeating ourselves 17 times?

aazizisoufiane:
I agree the duplication isn't ideal. That said, this approach is consistent with how ReAct.aforward() and Predict.aforward() are already implemented in DSPy.
Would it make sense to merge this for now and open a separate issue to explore reducing duplication across all modules? That refactor would be a bigger effort best handled separately.

PR #9238, opened 2026-01-24
Update gepa[dspy] dependency to version 0.0.26 adding support for cached evals in GEPA expected to reduce metric calls, fix MLFlow logging, and bug fixes

LakshyAAAgrawal:


okhat:
LGTM other than @chenmoneygithub 's request

PR #9247, opened 2026-01-29
fix(RLM): Refactor tools to take list instead of dict and properly serialize None/null values

isaacbmiller:
This PR makes two changes:

1. Tools API simplified from dict to list

Tool names are now inferred from function names instead of requiring explicit dict keys:

```python
# Before
rlm = RLM("ctx -> answer", tools={"my_tool": fn})
# After
rlm = RLM("ctx -> answer", tools=[fn])
For custom names, wrap with dspy.Tool(fn, name="custom").
```

Internally, tools are now stored as Tool objects rlm.py:135 which provide structured metadata (tool.name, tool.args, tool.desc) instead of requiring inspect.signature() introspection.

**BREAKING: This change removes the ability for you to pass tools via a dict, instead requiring that you pass tools as a list, as is done inside ReAct**

2. Fixed None handling in variable serialization

The PythonInterpreter now correctly serializes nested None values when injecting variables into the sandbox python_interpreter.py:433-450.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -10,7 +10,6 @@ Reference: "Recursive Language Models" (Zhang, Kraska, Khattab, 2025)
 
 from __future__ import annotations
 
-import inspect
 import logging
 import re
 import threading


isaacbmiller:
reviewed offline

PR #9248, opened 2026-01-30
Fix the request header in streaming mode

chenmoneygithub:
The request header was unintentionally dropped in the streaming mode.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -321,8 +321,6 @@ def _get_stream_completion_fn(
     if dspy.settings.track_usage:
         request["stream_options"] = {"include_usage": True}
 
-    headers = _get_headers(headers)
-
     async def stream_completion(request: dict[str, Any], cache_kwargs: dict[str, Any]):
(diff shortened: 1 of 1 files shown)


dspy/clients/lm.py
```diff
@@ -320,12 +321,13 @@ def _get_stream_completion_fn(
     if dspy.settings.track_usage:
         request["stream_options"] = {"include_usage": True}
 
+    headers = _get_headers(headers)
```
isaacbmiller: can we rename `_get_headers` to something more informative like `add_dspy_header` and also move it to line 359 so all of our header manipulation logic is in one place?

isaacbmiller (changes requested):

dspy/clients/lm.py
chenmoneygithub: Makes sense. `dspy_header` is still a bit strange, renamed to `_add_dspy_identifier_to_headers`.

isaacbmiller (approved):
lgtm

PR #9252, opened 2026-02-02
feat(refine): Support Prediction and tuple return types in reward_fn

aazizisoufiane:
This MR adds support for `dspy.Prediction` as a return type in `reward_fn` for the `Refine` module.

Right now, `reward_fn` can only return a float. The problem is that `OfferFeedback` has no idea *why* the reward was low — it can only see the score. By returning a `Prediction(score=0.3, feedback="Missing header row")`, the reward function can tell `OfferFeedback` exactly what went wrong, leading to better retry advice.

Two return types supported:
- `float` (existing, unchanged)
- `dspy.Prediction` with `score` and optional `feedback`

Tuple return was considered but dropped — `Prediction` is cleaner and more consistent with DSPy patterns.

### DIFF
--- a/dspy/predict/refine.py
+++ b/dspy/predict/refine.py
@@ -46,24 +46,33 @@ class Refine(Module):
     def __init__(
             self,
             module: Module,
-            N: int,
+            N: int,  # noqa: N803
             reward_fn: Callable[[dict, Prediction], float] | Callable[[dict, Prediction], Prediction],
             threshold: float,
             fail_count: int | None = None,
     ):
         """
+        Refines a module by running it up to N times with different rollout IDs at `temperature=1.0`
+        and returns the best prediction.
+
+        This module runs the provided module multiple times with varying rollout identifiers and selects
+        either the first prediction that exceeds the specified threshold or the one with the highest reward.
+        If no prediction meets the threshold, it automatically generates feedback to improve future predictions.
+
         Args:
-            ...
+            module (Module): The module to refine.
+            N (int): The number of times to run the module.
             reward_fn (Callable): The reward function. Can return:
                 - A simple float score
                 - A Prediction with `score` and optional `feedback` attributes
-            ...
+            threshold (float): The threshold for the reward function.
+            fail_count (Optional[int], optional): The number of times the module can fail before raising an error
 
         Example:
-        ```python
+```python
             import dspy
 
-            dspy.settings.configure(lm=dspy.LM("openai/gpt-4o-mini"))
+            dspy.configure(lm=dspy.LM("openai/gpt-4o-mini"))
 
             # Define a QA module with chain of thought
             qa = dspy.ChainOfThought("question -> answer")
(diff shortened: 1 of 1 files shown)


okhat:
Ah this is probably nice. Do we need a tuple return value? Tuples cause a mess IMO, but dspy.Prediction makes a lot of sense here.

aazizisoufiane:
Agreed, the tuple is unnecessary here. A simple `float` and `dspy.Prediction` are fair enough . I'll update the MR.

dspy/predict/refine.py
```diff
@@ -29,60 +31,64 @@ class OfferFeedback(Signature):
...
-        If no prediction meets the threshold, it automatically generates feedback to improve future predictions.
-
-
         Args:
-            module (Module): The module to refine.
-            N (int): The number of times to run the module. must
-            reward_fn (Callable): The reward function.
-            threshold (float): The threshold for the reward function.
-            fail_count (Optional[int], optional): The number of times the module can fail before raising an error
+            ...
```
okhat: wait why did the docstring get dropped for `module`, `N`, etc?

dspy/predict/refine.py
```diff
@@ -169,9 +201,10 @@ def __call__(self, lm, lm_kwargs, signature, demos, inputs):
 
             except Exception as e:
                 print(f"Refine: Attempt failed with rollout id {rid}: {e}")
-                if idx > self.fail_count:
+                if idx >= fail_count:
```
okhat: oh wow good catch, it's very weird that this was decrementing an attribute on the object itself!?! Not thread safe I imagine?

aazizisoufiane: Yes exactly! Mutating `self.fail_count` directly meant any concurrent call to the same Refine instance would share (and potentially corrupt) that state. Using a local variable keeps each invocation isolated and thread safe 🙂

dspy/predict/refine.py
okhat: happy to merge once this is resolved and the >= vs. > is clarified (ideally we stick to the old behavior of > unless it's simply wrong)

dspy/predict/refine.py
aazizisoufiane: I kept the original `>` behavior and the decrement logic — the only change is using a local `fail_count` variable instead of mutating `self.fail_count`. So the semantics are identical, just thread safe now

dspy/predict/refine.py
aazizisoufiane: You're right, I accidentally dropped the top-level description and most of the Args when updating the docstring. I restored the full original docstring and only extend the `reward_fn` entry to document the new return types.

dspy/predict/refine.py
aazizisoufiane: Gentle ping 🙂 @okhat 
Let me know if there’s anything you'd like me to refine here.

CharlesGaydon-Servier:
Hi, is this still in the work? :)

PR #9253, opened 2026-02-02
Refactor Docker command execution to use list-based subprocess for improved robustness

RinZ27:
I noticed that `HFServerTGI` builds the `docker run` command using string concatenation and executes it via `shell=True`. This approach can lead to unexpected behavior or potential command injection if parameters like `model_name` contain special characters.

I've refactored the `run_server` method to use a list-based argument structure with `shell=False`. This ensures that all arguments are properly handled as literals by the operating system, making the Docker command execution much more robust and secure against malformed inputs.

Testing:
- Verified the generated command structure matches the original intent.
- Confirmed that arguments with spaces or special characters no longer risk breaking out of the intended command context.

### DIFF
--- a/dsp/modules/hf_client.py
+++ b/dsp/modules/hf_client.py
@@ -1,0 +1,444 @@
+import os
+import random
+import re
+import shutil
+import subprocess
+
+# from dsp.modules.adapter import TurboAdapter, DavinciAdapter, LlamaAdapter
+import backoff
+import requests
+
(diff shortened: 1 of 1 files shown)


RinZ27:
Closing this PR as I realized the `dsp/modules/hf_client.py` module has been refactored or removed in the current `main` branch. My local environment was out of sync with upstream. Apologies for the noise!

Issue #9251, opened 2026-02-01
[Bug] prompt constructed with dspy.Code has unintended text


arjunguha:
### What happened?

I've observed this on the latest release and on HEAD. I think there is a bug in the implementation of dspy.Code. The prompt has contradictory directions,  asking the model to response both in Markdown and in JSON. I am pretty sure the JSON instructions are not intended, and were instead picked up from the docstring.

Here is an example of a system prompt with ChatAdapter:

``````
Your input fields are:
1. `question` (str):
Your output fields are:
1. `code` (Code): 
    Type description of Code: Code represented in a string, specified in the `code` field. If this is an output field, the code field should follow the markdown code block format, e.g. 
```python
{code}
```
Programming language: python
All interactions will be structured in the following way, with the appropriate values filled in.

[[ ## question ## ]]
{question}

[[ ## code ## ]]
[...]

PR #9257, opened 2026-02-04
fix(adapters): skip JSON schema for DSPy custom types in prompt (#9251)

darinkishore:
skip json schema for inbuilt types

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -12,6 +12,7 @@ from pydantic import TypeAdapter
 from pydantic.fields import FieldInfo
 
 from dspy.adapters.types.base_type import Type as DspyType
+from dspy.adapters.types.code import Code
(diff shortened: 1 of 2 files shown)


darinkishore:
@codex review

arjunguha:
While we are at this, I have a mild preference that dspy.Code not default to Python :) Maybe omit the word "Python" from the prompt if it is not specified? I often have signatures that read `problem:str language:str -> dspy.Code`, so the code language is itself an input.

darinkishore:
@chenmoneygithub 

scoped it to just dspy.Code, sanity check ?

darinkishore:
@arjunguha heard, can do this in a follow up, say more on the API design u think would be good here ?

darinkishore:
merging cuz been a minute

PR #9261, opened 2026-02-06
Thread Local and Async Local history for BaseLM

TyTodd:
I've ran into this use case a couple times now where I need to get the history of a `Predict` but feel uneasy about  using `LM.history` because the same LM is being used in multiple coroutines. The devx for getting a thread safe view of history is a little verbose, you either have to use `dspy.context` with a copy of the global LM. Or run a separate copy of your module if it has its `self.lm`. In some (rare) cases even this is not enough. I recently ran into a scenario where I needed to use the same module across multiple co-routines. The module had its own `self.lm` so using `dspy.context` wasn't a viable solution since it injects the LM into the local `dspy.settings`. 

This short PR adds thread/async-local history to LM via a ContextVar called `local_history`. Allowing users to view the history from the local thread.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -2,7 +2,7 @@ import copy
 import datetime
 import uuid
 from contextvars import ContextVar
-from typing import Any
+from typing import Any, TextIO
 
 from dspy.dsp.utils import settings
 from dspy.utils.callback import with_callbacks
@@ -20,7 +20,7 @@ class BaseLM:
     `forward` method and make sure the return format is identical to the
     [OpenAI response format](https://platform.openai.com/docs/api-reference/responses/object).
 
-    Example:
+    Examples:
 
     ```python
     from openai import OpenAI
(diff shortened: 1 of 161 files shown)


dspy/clients/base_lm.py
```diff
@@ -48,6 +49,11 @@ def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, c
         self.cache = cache
         self.kwargs = dict(temperature=temperature, max_tokens=max_tokens, **kwargs)
         self.history = []
+        # unique per instance
+        self._local_history: ContextVar[list | None] = ContextVar(
```
isaacbmiller, line 54: Will break BaseLM.copy()  because ContextVars arent serializeable

dspy/clients/base_lm.py
```diff
@@ -182,6 +196,9 @@ def update_history(self, entry):
 
         self.history.append(entry)
 
+        # thread-local/async-local history
+        self.local_history.append(entry)
```
isaacbmiller, line 213: This should respect settings.max_history_size as well

TyTodd:
@isaacbmiller thanks for your review. I fixed the serilization issue and max_history_size. Still working on test cases. Should have something pushed this weekend

TyTodd:
@isaacbmiller Sorry for the delay. Added the test cases and the deepcopy protection.

isaacbmiller:
@codex review

PR #9267, opened 2026-02-06
fix(pythoninterpreter): Guard against common install issues

isaacbmiller:
PythonInterpreter runs Python code in a secure sandbox using Deno and Pyodide (Python compiled to WebAssembly). This PR fixes initialization issues where Pyodide couldn't load properly, and improves error messages when things go wrong.

Key changes:

**Fixed Pyodide loading** - The Deno cache directory now has write permission (not just read), and node_modules/ is auto-detected for projects using deno.json with nodeModulesDir
**Better diagnostics** - When the sandbox fails to initialize or Deno isn't found, errors now explain common causes and link to GitHub issues


isaacbmiller:
fixed by #10206

PR #9272, opened 2026-02-08
fix: ColBERTv2RetrieverLocal forward method variable scoping bug

veeceey:
Fixes variable scoping issue in ColBERTv2RetrieverLocal that causes UnboundLocalError when filtered_pids is used.

## Summary

The forward method had a variable scoping bug where filtered_pids was referenced before assignment when the local filtering branch was taken.

## Changes

- Fixed variable initialization order to ensure filtered_pids is defined before use
- Maintains same functionality, just fixes the scoping issue

## Test Plan

- Tested with various combinations of k and filtered_pids
- Existing tests pass
- No breaking changes


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

isaacbmiller:
Thank you for the pr!

PR #9282, opened 2026-02-09
fix(RLM): show head and tail for RLM outputs

isaacbmiller:
Previously, all truncation was head-only for both variables and for REPL outputs. Now, we preserve both the beginning and end. This will also help if there is an error output only towards the end of an output.

**Outputs**
The first truncation happened when an output was created. We only stored up 100k characters per output. Then there was also a cap when it was being shown to the model at 5k characters.

Now the full output is stored in REPLEntry, and truncation only happens before the output is serialized into a prompt.

max_output_chars default unified to 10,000 (was 100,000 in RLM.__init__ and 5,000 in REPLEntry.format()).

**Variables**
The truncation is also moved to be head/tail, but there is no lever to adjust it inside the RLM.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -498,10 +498,9 @@ class RLM(Module):
             output = str(result) if result else ""
 
         output = self._format_output(output)
-        history = history.append(reasoning=pred.reasoning, code=code, output=output)
         if self.verbose:
-            logger.info(REPLEntry.format_output(output, history.max_output_chars))
-        return history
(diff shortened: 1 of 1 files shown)


okhat:
LGTM but one question: is there a kwarg now to uniformly control the lengths of ALL these? Last I checked, some of them may have been hard-coded to short values?

isaacbmiller:
Updated PR description @okhat .

Variable hard coded to 500,  output length can change via RLM init arg

PR #9240, opened 2026-01-26
Add native Dataframe support for RLMs

kmad:
## Summary

This PR adds a `dspy.DataFrame` primitive which is intended for use by the RLM module, enabling LLMs to operate directly on DataFrames in the REPL with full type preservation. Non-RLM modules receive a string summary of the dataframe but will not be able to operate on its content. This PR also adds a `get_variable` function to allow for the extraction of variables from the REPL environment.

Disclosure: Code was produced with the assistance of Claude Opus 4.5 and reviewed manually.

### Core Features

**Parquet Serialization**
- DataFrames are serialized via PyArrow Parquet format and injected into the pyodide environment
- Preserves column dtypes: int64, float64, datetime64, categorical, boolean, etc.
- Binary data is base64-encoded for transport to the Pyodide sandbox

**Rich Metadata in Prompts**
- LLM receives DataFrame shape, memory usage, column dtypes, null counts
- First 3 and last 3 rows shown as preview
- Field descriptions and constraints from Signature are included

**`dspy.DataFrame` Type Annotation**
- New Pydantic-compatible type for use in DSPy Signatures
- Validates that values are pandas DataFrames
- Non-RLM modules (ChainOfThought, Predict) are shown a warning that only a string summary is provided

**New `get_variable` Method**
- Adds ability to retrieve variables from the sandbox after RLM execution
- Useful for extracting computed values not included in SUBMIT (e.g., base64-encoded charts)
- Returns JSON-deserialized values when possible, string representation otherwise

## Usage Example

```python
import dspy
import pandas as pd

class DataAnalysis(dspy.Signature):
    """Analyze the provided data."""
    data: dspy.DataFrame = dspy.InputField(desc="Sales data")
    summary: str = dspy.OutputField(desc="Analysis summary")
    total_revenue: float = dspy.OutputField()


# Configure your LM
lm = dspy.LM("anthropic/claude-sonnet-4-5-20250929", max_tokens=16000)
dspy.configure(lm=lm)

df = pd.DataFrame({
    "product": ["A", "B", "C"],
    "revenue": [100.0, 200.0, 150.0]
})

analyzer = dspy.RLM(DataAnalysis, max_iterations=10)
result = analyzer(data=df)

print(result.summary)
print(f"Total: ${result.total_revenue}")
```

## Files Changed

| File | Description |
|------|-------------|
| `dspy/primitives/repl_types.py` | Added `DataFrame` type with Pydantic schema support |
[...]

PR #9283, opened 2026-02-10
RLM Generic Type Support + Dataframe implementation

kmad:
A new implementation of #9240 by implementing DataFrames as a DSPy-native Type and using Adapters properly, per feedback from @isaacbmiller . This approach introduces a generic interface on `dspy.Type` for any custom type to *opt in* to RLM sandbox support, with `dspy.DataFrame` as the first implementation.

**Additions**

  - Generic RLM type interface — Any `dspy.Type` subclass can now participate in RLM sandbox injection by implementing three methods:
    - `sandbox_setup()` — imports needed inside the sandbox; currently we are restricted by default to what is included with pyodide. A list of packages is [here](https://pyodide.org/en/stable/usage/packages-in-pyodide.html).
    - `to_sandbox(var_name)` — serialize the value for injection
    - `rlm_preview(max_chars)` — structured preview shown to the LLM in the prompt

  - **dspy.DataFrame** `Type` (`dspy/adapters/types/dataframe.py`) — Reference implementation of the interface. Wraps a pandas DataFrame, serializes via `pandas.to_json()`, and provides a structured preview to the LLM.
  - A notebook with a simple example of how to use it.

**Changes**
  - PythonInterpreter (`dspy/primitives/python_interpreter.py`) — The interpreter now checks for to_sandbox via duck typing and delegates serialization to the type itself. No DataFrame-specific or pandas-specific code in the interpreter.

  - RLM (`dspy/predict/rlm.py`) — Two additions:
    - _wrap_rlm_inputs() auto-wraps raw values (e.g. a raw pd.DataFrame) into their annotated `dspy.Type` when the type has RLM support, so users can write `rlm(data=my_df)` without manually wrapping
    - _build_variables() uses rlm_preview() for RLM-supported types instead of generic string serialization

**Design decisions**

  - Duck typing over base class methods: The interface is documented via comments on `dspy.Type`, not default method stubs. This avoids all existing types (Image, Audio, etc.) accidentally matching the RLM support check; it may not be optimal or pythonic but is a way to enable the functionality while keeping it opt-in.
  - Wrapping happens in the RLM, while the interpreter stays generic. The RLM's forward() uses signature annotations to wrap raw values to retain flexibility.
  - JSON serialization: Currently we use pandas' built-in `to_json(orient="records")`. Dtypes may need re-inference in the sandbox (e.g. dates arrive as strings), which the LLM should handle naturally. In the future we should explore using `pyarrow` and parquet types for better serialization performance, preservation of types, etc.


rawwerks:
if types are given the greenlight from the maintainers, i will refactor https://github.com/stanfordnlp/dspy/issues/9289 

cc @okhat @isaacbmiller

rawwerks:
Nice work on the  /  /  /  protocol — this is a really clean pattern for getting custom types into the RLM sandbox.

We adopted this same protocol for Audio and Image in our RLM feature branch ([rawwerks/dspy:feat/rlm-media-types-protocol](https://github.com/rawwerks/dspy/tree/feat/rlm-media-types-protocol)). Previously we had a hardcoded media registry + `llm_query_with_media` approach that was specific to Audio/Image. After seeing this PR, we refactored to use the generic `dspy.Type` sandbox protocol instead — any type that implements `to_sandbox()` + `format()` is now automatically detected as multimodal, injected into the sandbox, and available for `llm_query_with_media()`.

Also adopted the `_has_rlm_support()` check, `_wrap_rlm_inputs()` auto-wrapping, `_inject_pending_vars()`, and the `rlm_preview()` approach for `_build_variables()` — all great patterns.

One small thought: might be worth documenting the protocol methods on the `Type` base class (even as optional/abstract methods or in a docstring) so future type authors know what to implement for RLM support. Right now you have to look at DataFrame to discover the pattern.

Related upstream issue with our full RLM feature set (multimodal, budget, cost, depth>1, LocalInterpreter): https://github.com/stanfordnlp/dspy/issues/9289

dspy/adapters/types/base_type.py
```diff
@@ -129,6 +129,16 @@ def parse_lm_response(cls, response: str | dict[str, Any]) -> Optional["Type"]:
         """
         return None
 
+    # RLM Sandbox Support
```
isaacbmiller, line 132: Lets make this into a protocol, SandboxSerializable, so that we can check type conformance using isInstance rather than checking for the existence of methods.

dspy/predict/rlm.py
```diff
@@ -327,12 +327,49 @@ def _get_output_fields_info(self) -> list[dict]:
             fields.append(field_info)
         return fields
 
+    def _wrap_rlm_inputs(self, input_args: dict[str, Any]) -> dict[str, Any]:
```
isaacbmiller, line 330: Lets not include this method. If we need to throw for methods we don't know how to serialize, then we can

dspy/predict/rlm.py
```diff
@@ -327,12 +327,49 @@ def _get_output_fields_info(self) -> list[dict]:
...
+                wrapped[name] = value
+        return wrapped
+
     def _build_variables(self, **input_args: Any) -> list[REPLVariable]:
         """Build REPLVariable list from input arguments with field metadata."""
         variables = []
         for name, value in input_args.items():
             field_info = self.signature.input_fields.get(name)
-            variables.append(REPLVariable.from_value(name, value, field_info=field_info))
+            if hasattr(value, 'rlm_preview') and callable(getattr(value, 'rlm_preview', None)):
```
isaacbmiller, line 365:
```diff
-if hasattr(value, 'rlm_preview') and callable(getattr(value, 'rlm_preview', None)):
+isInstance(SandboxSerializable)
```

dspy/predict/rlm.py
```diff
@@ -327,12 +327,49 @@ def _get_output_fields_info(self) -> list[dict]:
...
+                # Use rlm_preview() for types with RLM support (gives better LLM context)
+                preview = value.rlm_preview()
```
isaacbmiller, line 367: Lets give SandboxSerializable a derived method that is `.to_REPLVariable`

isaacbmiller:
Thanks for the PR! Sorry for the delay in reviewing.

After thinking about it, it really doesn't make sense to have Dataframe as a DSPy.Type. Most DSPy programs shouldn't use dataframes as inputs, and we don't really want to imply that they might.

Something being a DSPy.Type doesn't mean it works with RLMs and vice-versa for an RLM specific type. A good example is Images. You need a way to put the actual images into a context window. Maybe you add an image parameter to the `llm_query` function, maybe you add a `look` function which loads it into the main RLM context. I am still not sure what the right implementation for an Image with an RLM is, where we have things that meta-affect the context.

(1) We should make a `typing.Protocol` called `SandboxSerializable` with the 4 methods you proposed:
  1. `sandbox_setup() -> str`: (imports needed in sandbox)
  2. `to_sandbox() -> bytes`: (serialize for injection) - should this be str or bytes?
  3. `sandbox_assignment(var_name, data_expr) -> str`: (reconstruction code)
  4. `rlm_preview(max_chars) -> str`: (LLM-friendly preview)

Then inside RLM, we can check conformance using `isInstance(SandboxSerializable)` rather than for the existence of one of the individual methods.

(2) We want RLM to own the gathering, serialization, and deserialization of the types. All PythonInterpreter is responsible for is running the code that is given to it + moving variables into files if they are too big.

So then sandbox_assignment can be called the same way regardless, where we know the loaded str/bytes will be coming as a variable. Its just how that loaded str/bytes gets into the repl is decided by the Interpreter.

(3) Sandbox setup/serialization appears to happen per execute() cycle, not once per RLM session. For dataframe-like inputs this is expensive and unnecessary. We just need to re-serialize if the deno process dies.

(4) I don't want to merge DataFrame into core. I **really** like including it inside of the tutorial, as we don't have great examples of writing your own types for use with DSPy. This can be the canonical RLM tutorial, and people might want to use their own types with RLMs because these types are larger.

This is an awesome PR, and I really appreciate your help!

kmad:
Thanks for the feedback, @isaacbmiller - that all makes sense. Closing this PR and I opened #9411 as a potential implementation. Will finalize and publish for review shortly!

PR #9284, opened 2026-02-10
fix(streaming): propagate exceptions in apply_sync_streaming

adityasingh2400:
## Summary
Fixes #9142 — `apply_sync_streaming` was silently swallowing exceptions raised by the async generator.

## Problem
When using `streamify` with `async_streaming=False`, exceptions raised in the async generator were caught in the background thread but never propagated to the main thread consuming the queue. This made debugging difficult and caused unexpected silent failures.

## Solution
- Added an `exception_sentinel` to signal when an exception occurred
- Wrapped the async iteration in a try-except that catches exceptions and puts them in the queue as a tuple `(exception_sentinel, exception)`
- In the consumer loop, check if the item is an exception tuple and re-raise it

## Testing
All streaming tests pass (34 passed, 3 skipped):
```
python -m pytest tests/streaming/ -v
```

## Reproduction
```python
import dspy
from dspy.streaming import streamify

async def faulty_generator():
    yield 1
    raise ValueError("Something went wrong")
    yield 2

# Before: silently fails
# After: properly raises ValueError
for item in streamify(faulty_generator(), async_streaming=False):
    print(item)
```

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -228,6 +228,7 @@ def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
     """Convert the async streaming generator to a sync generator."""
     queue = Queue()  # Queue to hold items from the async generator
     stop_sentinel = object()  # Sentinel to signal the generator is complete
+    exception_sentinel = object()  # Sentinel to signal an exception occurred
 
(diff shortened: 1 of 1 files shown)


adityasingh2400:
Closing - duplicate of #9145

PR #9285, opened 2026-02-10
fix(deno): use analyzePath for reliable cross-platform directory checks

### LINKED ISSUE #9107: [Bug] On windows, code execution using deno does not work because of error description mismatch inside js
### What happened?

Os: windows 11,
dspy Version: 3.0.4
Deno: 
deno 2.5.6 (stable, release, x86_64-pc-windows-msvc)
v8 14.0.365.5-rusty
typescript 5.9.2

In the deno script runner.js, when creating a file path with mkdir, existing directory are handled using an error catch like this:
              } catch (e) {
                  if (!(e && e.message && e.message.includes('File exists'))) {
                      console.log("[DEBUG] Error creating directory in Pyodide file system:", cur, "|", e.message);
                  }
              }
where the script assumes that the error message would contain 'File exists', which is not the case for my version of deno (message is empty for this error).
A more robust approach would be to just test if the directory exists using a function like this:

function directoryExists(path) {
  try {
    const entry = pyodide.FS.analyzePath(path);
    return entry.exists; // True if it's a directory
  } catch (e) {
    console.log("Error checking directory existence in Pyodide:", e);
    return false; // If an error occurs (e.g., path doesn't exist), return false
  }
}

I have attached the full updated script that works in my case.

[runner.js](https://github.com/user-attachments/files/23888489/runner.js)

### Steps to reproduce

When using the code interpreter with file loading:
ex:
dspy.PythonInterpreter({}, enable_read_paths=files_path).execute(expression)

files_path could be any file path and expression can be any expression.
The problem is on 
Os: windows 11,
Deno: 
deno 2.5.6 (stable, release, x86_64-pc-windows-msvc)
v8 14.0.365.5-rusty
typescript 5.9.2

### DSPy version

3.0.4

adityasingh2400:
## Summary
Fixes #9107 — Deno code execution fails on Windows due to error message mismatch when checking directory existence.

## Problem
The current code in `runner.js` uses try/catch blocks to handle "directory already exists" errors when creating directories in Pyodide's virtual filesystem. However, this relies on matching specific error messages like `'File exists'`, which varies across platforms and Deno versions. On Windows, the error message is empty, causing the check to fail.

## Solution
Replace the try/catch pattern with `pyodide.FS.analyzePath()` which:
- Returns an object with an `exists` property
- Works reliably across all platforms (Windows, Linux, Mac)
- Doesn't depend on error message text

## Changes
- Updated directory creation in the file path setup loop
- Updated `inject_var` method's `/tmp` and `/tmp/dspy_vars` directory creation

## Testing
The fix is JavaScript-only and uses the documented Pyodide FS API. The `analyzePath` approach is:
- Recommended by the issue reporter who tested on Windows
- A standard Emscripten FS method that Pyodide exposes
- More semantically correct (ask "does this exist?" vs "try and catch failure")

Tested that existing Deno-based tests still pass on Mac/Linux.

### DIFF
--- a/dspy/primitives/runner.js
+++ b/dspy/primitives/runner.js
@@ -261,12 +261,10 @@ while (true) {
       let cur = '';
       for (const d of dirs) {
         cur += '/' + d;
-        // Check if directory exists before creating
-        try {
-          pyodide.FS.stat(cur);
-          // Directory exists, continue to next
-        } catch {
-          // Directory doesn't exist, create it
+        // Check if directory exists before creating using analyzePath
+        // This is more reliable than catching exceptions across platforms
+        const pathInfo = pyodide.FS.analyzePath(cur);
+        if (!pathInfo.exists) {
           pyodide.FS.mkdir(cur);
         }
       }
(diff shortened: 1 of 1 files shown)


adityasingh2400:
Closing - duplicate of #9108

PR #9294, opened 2026-02-11
fix: effective_max_tokens when using a `load`'ed OpenAI reasoning model

peterhadlaw:
Fix OpenAI reasoning-model save/load by normalizing token keys in `Predict.load_state()`. It prevents constructor key collisions during reload, enforces clear precedence (`max_completion_tokens` > `max_tokens`), and logs overrides at debug level.

## Problem

OpenAI reasoning models are constructed with `max_tokens`, but persisted LM kwargs use `max_completion_tokens`.

During round-trip loading, `Predict.load_state()` previously reconstructed LMs via:

`LM(**state["lm"])`

If serialized state included `max_completion_tokens` (or mixed token keys), reconstruction could become ambiguous and, in conflict/mixed-key cases, trigger constructor-level key collisions.

## Root Cause

`Predict.load_state()` passed serialized LM token keys directly into `LM(...)` with no normalization layer.

For OpenAI reasoning models, this creates a key-shape mismatch between:
- constructor input (`max_tokens`)
- persisted state (`max_completion_tokens`)

and may introduce key conflicts when both appear.

## Solution

- Add `_normalize_lm_token_state()` in `dspy/predict/predict.py`.
- Normalize serialized LM token keys before `LM(...)` construction:
  - map `max_completion_tokens` -> `max_tokens`
  - if both keys exist, `max_completion_tokens` takes precedence
- Emit `logger.debug(...)` when normalization overrides an existing `max_tokens` value.
- Keep constructor behavior unchanged; this fix is scoped to load-time state reconstruction.

## Tests

Added regressions covering three layers:

1. **Round-trip shape and behavior**
   - `test_predict_load_normalizes_serialized_openai_reasoning_tokens`
   - Verifies persisted state uses `max_completion_tokens` and load preserves equivalent OpenAI reasoning kwargs.

2. **Precedence + debug logging**
   - `test_load_state_prefers_max_completion_tokens_for_openai_reasoning_models_with_debug_log`
   - Verifies mixed/conflicting keys resolve in favor of `max_completion_tokens` and emit one debug override log.

3. **Constructor-call normalization boundary**
   - `test_load_state_normalizes_reasoning_token_keys_before_lm_construction`
   - Verifies `Predict.load_state()` calls `LM(...)` with normalized `max_tokens` only (no `max_completion_tokens` kwarg passed through).

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -40,6 +40,35 @@ def _sanitize_lm_state(lm_state: dict, allow_unsafe_lm_state: bool) -> dict:
     return sanitized_lm_state
 
 
+def _normalize_lm_token_state(lm_state: dict) -> dict:
+    """Normalize serialized LM token fields before reconstructing `LM(...)`.
+
+    OpenAI reasoning models are configured with `max_tokens` at construction time, but
+    their persisted LM kwargs use `max_completion_tokens`. During `Predict.load_state()`,
+    this key mismatch or key conflict can produce ambiguous behavior if state includes
+    one or both keys.
+    This normalization maps `max_completion_tokens` into `max_tokens` and gives
+    `max_completion_tokens` precedence when both are present.
+    """
+    normalized_lm_state = dict(lm_state)
+    if "max_completion_tokens" not in normalized_lm_state:
+        return normalized_lm_state
+
+    max_completion_tokens = normalized_lm_state.pop("max_completion_tokens")
+    has_max_tokens_defined = "max_tokens" in normalized_lm_state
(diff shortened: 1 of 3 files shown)


isaacbmiller:
Let's use a simpler approach than the sentinel.

If we are loading in `max_completion_tokens` from a saved LM, lets combine that into max_tokens. This would be in load_state of predict.py.

peterhadlaw:
@isaacbmiller thank you for the feedback, adjusted, lmk if that's good or not, happy to adjust further

peterhadlaw:
@isaacbmiller bump :)

PR #9301, opened 2026-02-14
fix(best_of_n): prevent fail_count mutation and silent None return

zamalali:
#### 1. `self.fail_count` mutation across calls

`self.fail_count -= 1` permanently decrements the instance attribute on every exception. After repeated use, the same `BestOfN` instance progressively tolerates fewer failures until it crashes on the first error:

**Fix:** Use a local `error_count` counter instead of mutating `self.fail_count`.

#### 2. Silent `None` return when all attempts fail

When `fail_count >= N` and every attempt throws an exception, the loop exits normally and `best_pred` (initialized to `None`) is returned. Callers then hit a confusing `AttributeError: 'NoneType' object has no attribute 'answer'` with no indication that all attempts failed.

**Fix:** Add a post-loop guard that re-raises the last caught exception, or raises a descriptive `RuntimeError` if no exception was recorded.

#### 3. `print()` instead of `logger.warning()`

Error messages were printed via `print()`, inconsistent with the rest of `dspy/predict/` which uses `logging.getLogger(__name__)`.

**Fix:** Replace with `logger.warning()`, matching `react.py`, `rlm.py`, `predict.py`, etc.

#### Tests Added

- `test_best_of_n_reusable_across_calls` — verifies `fail_count` is not mutated between successive calls
- `test_best_of_n_all_failures_raises_error` — verifies an exception is raised instead of returning `None`

All existing tests pass unchanged (129 passed, 48 skipped across `tests/predict/`).

> **Note:** The sibling module `Refine` has identical bugs — PR #9252 already addresses `fail_count` mutation there.
#### 2. Silent `None` return when all attempts fail

When `fail_count >= N` and every attempt throws an exception, the loop exits normally and `best_pred` (initialized to `None`) is returned. Callers then hit a confusing `AttributeError: 'NoneType' object has no attribute 'answer'` with no indication that all attempts failed.

**Fix:** Add a post-loop guard that re-raises the last caught exception, or raises a descriptive `RuntimeError` if no exception was recorded.

#### 3. `print()` instead of `logger.warning()`

Error messages were printed via `print()`, inconsistent with the rest of `dspy/predict/` which uses `logging.getLogger(__name__)`.

**Fix:** Replace with `logger.warning()`, matching `react.py`, `rlm.py`, `predict.py`, etc.

#### Tests Added

- `test_best_of_n_reusable_across_calls` — verifies `fail_count` is not mutated between successive calls
- `test_best_of_n_all_failures_raises_error` — verifies an exception is raised instead of returning `None`

All existing tests pass unchanged (129 passed, 48 skipped across `tests/predict/`).

> **Note:** The sibling module `Refine` has identical bugs — PR #9252 already addresses `fail_count` mutation there.

### DIFF
--- a/dspy/predict/best_of_n.py
+++ b/dspy/predict/best_of_n.py
@@ -50,7 +50,7 @@ class BestOfN(Module):
             ```
         """
         if N <= 0:
-                raise ValueError(f"BestOfN: N must be greater than 0, got {N}")
+            raise ValueError(f"BestOfN: N must be greater than 0, got {N}")
         self.module = module
         self.reward_fn = lambda *args: reward_fn(*args)  # to prevent this from becoming a parameter
         self.threshold = threshold
@@ -93,7 +93,7 @@ class BestOfN(Module):
 
         if best_pred is None:
             if last_exception is not None:
-                raise last_exception.with_traceback(last_exception.__traceback__)
+                raise last_exception
             raise RuntimeError(f"BestOfN: All {self.N} attempts failed. No successful prediction was obtained.")
 
         if best_trace:


kvr06-ai (commented):
(Supersedes my earlier comment — this is a more thorough pass.)

Three observations.

**Semantic shift in fail_count threshold**

The old code compared `idx` (all attempts, including successes) against `self.fail_count`. The new code compares `error_count` (failures only). These produce different results when successes and failures are interleaved.

Example (N=5, fail_count=3): attempts 0, 2, 3 fail while attempt 1 succeeds.
- Old: at attempt 3, `idx=3 > fail_count=1` (after two decrements) raises.
- New: `error_count=3` is not `> 3`, so the loop continues to attempt 4.

The new behavior matches the docstring ("the number of times the module can fail before raising an error"). This is a user-visible change worth noting in the PR description. Consider adding a test that exercises interleaved successes and failures to pin down this behavior explicitly.

**N=0 produces a misleading error**

If `N=0`, `rollout_ids` is empty, the loop never executes, and the post-loop check raises `RuntimeError("All 0 attempts failed...")`. Nothing actually failed. Either validate `N > 0` in `__init__` or adjust the message.

**Pre-existing: raise e resets the traceback**

Both the in-loop `raise e` and post-loop `raise last_exception` reset the traceback to the raise site instead of preserving the original call stack. Bare `raise` inside the `except` block preserves it. Not blocking, but since this code path is being rewritten, it's an opportunity to improve debuggability.

dspy/predict/best_of_n.py
isaacbmiller, line 58: This isnt in this PR, but while I am looking at the code, fail_count=0 doesn't work

dspy/predict/best_of_n.py
```diff
@@ -46,6 +49,8 @@ def one_word_answer(args, pred):
             # Returns: Brussels
             ```
         """
+        if N <= 0:
+                raise ValueError(f"BestOfN: N must be greater than 0, got {N}")
```
isaacbmiller: Too much indentation

dspy/predict/best_of_n.py
```diff
@@ -78,10 +85,16 @@ def forward(self, **kwargs):
...
-                self.fail_count -= 1
+                error_count += 1
+                last_exception = e
+                logger.warning(f"BestOfN: Attempt {idx + 1}/{self.N} failed with rollout id {rid}: {e}")
+                if error_count > self.fail_count:
+                    raise 
+
+        if best_pred is None:
+            if last_exception is not None:
+                raise last_exception.with_traceback(last_exception.__traceback__)
```
isaacbmiller: equivalent to just `last_exception.with_traceback()`

kvr06-ai:
@zamalali I've opened a PR against your branch to address @isaacbmiller's review feedback: https://github.com/zamalali/dspy/pull/1

Changes:
- Fixed indentation on the `raise ValueError` line (was double-indented)
- Simplified `raise last_exception.with_traceback(last_exception.__traceback__)` to `raise last_exception` (redundant since the exception already carries its own traceback)

isaacbmiller:
LGTM once ruff is fixed

PR #9303, opened 2026-02-16
feat(interpreter): use deno pypi package for bundled binary

### LINKED ISSUE #9228: [Feature] Use deno pypi package

https://github.com/denoland/deno_pypi
deno has an official python package that contains the deno binary
This way the dependency can be handled properly in Python instead of relying on the user to manually install deno

### Additional Context

_No response_

adityasingh2400:
## Summary
Uses the official [deno pypi package](https://github.com/denoland/deno_pypi) for the Deno runtime binary instead of requiring manual installation.

## Changes
- Add `_get_deno_binary()` helper that tries the deno package first, falls back to system deno
- Add `deno` as optional dependency: `pip install dspy[deno]`
- Update docstring to reflect new installation option

## Usage
```bash
# Install with bundled deno binary
pip install dspy[deno]

# Or continue using system deno
pip install dspy
```

The interpreter automatically uses the pypi-bundled deno if available, otherwise falls back to system deno. This is fully backwards compatible.

Fixes #9228

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -23,6 +23,17 @@ __all__ = ["PythonInterpreter", "FinalOutput", "CodeInterpreterError"]
 
 logger = logging.getLogger(__name__)
 
+
+# Try to use deno from pypi package if available, otherwise fall back to system deno
+def _get_deno_binary() -> str:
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Thanks for the PR!

As discussed in #9228, we waiting until the Deno pypi package until it is more mature to move over.

Closing for now

PR #9306, opened 2026-02-16
feat(rlm): allow for passing pydantic models inside an RLM

isaacbmiller:
RLM executes LLM-generated Python code in a sandboxed Deno/Pyodide interpreter. Host-side "tools" (Python functions) can be called from sandbox code, but arguments and return values cross a JSON serialization boundary. Previously, only simple types (str, int, float, bool) were handled — complex types like Pydantic models were silently ignored.

This PR adds first-class Pydantic model support across the full pipeline:

**Tool parameters**: dicts from the sandbox are coerced into Pydantic models before the tool is called, via pydantic.TypeAdapter python_interpreter.py:345-347
**Tool returns**: Pydantic model return values are serialized to JSON dicts before crossing back to the sandbox python_interpreter.py:349-350
**Input variables**: Pydantic model instances are serialized via model_dump(mode="json") for injection into sandbox code python_interpreter.py:475-476
**Output fields**: JSON schema metadata is attached to complex output types so the sandbox's SUBMIT() gets proper type hints rlm.py:327-337
**Prompt metadata**: REPLVariable now includes JSON schema for Pydantic inputs so the LLM understands their structure repl_types.py:77-83

Additionally, tool calls are now kwargs-only (no positional args), variadic tool signatures are rejected, and tool bridge errors in runner.js are returned as structured payloads instead of thrown exceptions (preventing Deno crashes from unhandled rejections).

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -21,7 +21,7 @@ import pydantic
 
 import dspy
 from dspy.adapters.types.tool import Tool
-from dspy.adapters.utils import parse_value, translate_field_type
+from dspy.adapters.utils import _get_json_schema, parse_value, translate_field_type
 from dspy.primitives.code_interpreter import SIMPLE_TYPES, CodeInterpreter, CodeInterpreterError, FinalOutput
 from dspy.primitives.module import Module
 from dspy.primitives.prediction import Prediction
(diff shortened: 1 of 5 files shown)


okhat:
Nice :D

isaacbmiller:
Closing in favor of smaller PRs

PR #9309, opened 2026-02-18
refactor(rlm): enhance code fence parsing and REPL entry formatting

isaacbmiller:
The RLM module has the LLM write Python in markdown code fences (```python ... ```), which must be stripped before execution. Previously, _strip_code_fences used a permissive regex that silently passed through anything it couldn't parse — unlabeled fences, multiple blocks, or prose mixed with code would reach the interpreter and fail in confusing ways.

This PR rewrites _strip_code_fences as a strict parser that raises SyntaxError on malformed input rlm.py:69-103. Because it can now raise, a new _execute_action_code method rlm.py:534-550 wraps both fence stripping and interpreter execution with proper error handling, replacing duplicated try/except blocks in the sync and async iteration paths.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -560,7 +560,12 @@ class RLM(Module):
                 f"Reasoning: {action.reasoning}\nCode:\n{action.code}"
             )
 
-        code = _strip_code_fences(action.code)
+        try:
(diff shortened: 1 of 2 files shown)


isaacbmiller:
@codex review

isaacbmiller:
Discussed offline

PR #9313, opened 2026-02-19
feat(signature): add type validation for input fields

michaelisaac-dev:
### Summary

1. Implement type validation for input fields used in signatures - Supports both string and class-based signatures.
2. Implement a check for fields used in a module invocation that are not defined in the signature.
3. Log a warning in case 1 or 2 is violated
4. Skip type validation for cases where the signature has no type defined for the given `InputField`.

For example:

```
    predict_instance = Predict("mode:Literal['auto','manual'] -> result")
    predict_instance(mode="stick", year=2025)
```

Results in the following warnings:
```
WARNING dspy.predict.predict: Input contains fields not in signature. These fields will be ignored: ['year']. Expected fields: ['mode'].
WARNING dspy.predict.predict: Type mismatch for field 'mode': expected Literal['auto', 'manual'] based on given Signature, but the provided value is incompatible: stick.
```

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -19,6 +19,26 @@ from dspy.utils.constants import IS_TYPE_UNDEFINED
 
 logger = logging.getLogger(__name__)
 
+UNSAFE_LM_STATE_KEYS = {"api_base", "base_url", "model_list"}
+
+
+def _sanitize_lm_state(lm_state: dict, allow_unsafe_lm_state: bool) -> dict:
+    if allow_unsafe_lm_state:
+        return lm_state
+
+    unsafe_keys = sorted(UNSAFE_LM_STATE_KEYS.intersection(lm_state))
+
+    if not unsafe_keys:
+        return lm_state
+
+    sanitized_lm_state = {k: v for k, v in lm_state.items() if k not in UNSAFE_LM_STATE_KEYS}
+    logger.warning(
+        "Ignoring unsafe LM config key(s) during state load: %s. "
+        "Pass allow_unsafe_lm_state=True to preserve these keys for trusted files.",
+        unsafe_keys,
+    )
+    return sanitized_lm_state
+
 
 class Predict(Module, Parameter):
     """Basic DSPy module that maps inputs to outputs using a language model.
@@ -69,11 +89,13 @@ class Predict(Module, Parameter):
         state["lm"] = self.lm.dump_state() if self.lm else None
         return state
 
-    def load_state(self, state: dict) -> "Predict":
+    def load_state(self, state: dict, *, allow_unsafe_lm_state: bool = False) -> "Predict":
         """Load the saved state of a `Predict` object.
 
         Args:
             state: The saved state of a `Predict` object.
+            allow_unsafe_lm_state: If True, preserves `api_base`, `base_url`, and `model_list` from
+                serialized LM state. Enable only when loading trusted files.
 
         Returns:
             Self to allow method chaining.
@@ -85,7 +107,8 @@ class Predict(Module, Parameter):
                 setattr(self, name, value)
 
         self.signature = self.signature.load_state(state["signature"])
-        self.lm = LM(**state["lm"]) if state["lm"] else None
+        sanitized_lm_state = _sanitize_lm_state(state["lm"], allow_unsafe_lm_state) if state["lm"] else None
+        self.lm = LM(**sanitized_lm_state) if sanitized_lm_state else None
 
         if "extended_signature" in state:  # legacy, up to and including 2.5, for CoT.
             raise NotImplementedError("Loading extended_signature is no longer supported in DSPy 2.6+")
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,20 +1,16 @@
-default_language_version:
-  python: python3.10
-
 default_stages: [pre-commit]
 default_install_hook_types: [pre-commit]
 
 repos:
-  - repo: local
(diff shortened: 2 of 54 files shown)


okhat:
This makes a ton of sense, but can quickly become annoying for code that works fine. Maybe even some of the tutorials might result in warnings; do we know if they would?

So the important things here are:

- When can we merge / release this? Maybe we wait for a "minor" version upgrade.

- How annoying are the warnings? For example, do the warnings stop after the first time for each signature? Do the warnings offer a short line like `dspy.configure(warn_for_...=False)` that stops these warnings easily?

michaelisaac-dev:
Thanks for the feedback, @okhat!

I've just run the [Signature tutorial](https://dspy.ai/learn/programming/signatures/) and couldn't see any warnings. I've also added a few unit tests to cover custom types validation.

Additionally, I've introduced a flag to suppress the warnings (`dspy.configure(warn_on_type_mismatch=...)`) if needed. The current default is `True`. I've updated `signature.md` with info reg. type validation and how to set the flag.

Since it's a non-breaking change, I'd say we can get this out with a minor version bump.

MaximeRivest:
I have no intention to derail that PR, but I am genuinely curious to understand the advantage of that, the reason I don't 'see' them is that I am thinking that everything is 'rendered' to a string anyways?

michaelisaac-dev:
> I have no intention to derail that PR, but I am genuinely curious to understand the advantage of that, the reason I don't 'see' them is that I am thinking that everything is 'rendered' to a string anyways?

When a user defines types in a Signature, those types are actually communicated to the LLM as part of the system message. For example, if your signature declares:

question: str

…the system message will include something like:

System message:
Your input fields are:

question (str)

So if, in a later call, you invoke a module with that signature but pass something that isn’t a string, you’re effectively giving the model inputs that contradict what it was told to expect, which can “confuse” the LLM.

Also, this PR triggers a warning when you invoke a module with an input field that isn’t part of the signature. Right now, that extra input is silently ignored and not sent to the LLM, which can lead to confusing behavior (especially after a user refactors a Signature - e.g., renames an InputField).

MaximeRivest:
> When a user defines types in a Signature, those types are actually communicated to the LLM as part of the system message. 

yes and I question the impact of telling the input type on the quality of the llm response, it certainly 'cost' tokens, does it improve performance? Cyrus was telling me that adapters had not been benchmarked. 

> Also, this PR triggers a warning when you invoke a module with an input field that isn’t part of the signature. Right now, that extra input is silently ignored and not sent to the LLM, which can lead to confusing behavior (especially after a user refactors a Signature - e.g., renames an InputField).

the advantage of that are clear to me thanks!

---

Did you already design for what happens with custom adapters that don't use the input type? Everybody knows I love custom adapters :)

dspy/predict/predict.py
```diff
@@ -1,8 +1,10 @@
 import logging
 import random
+from typing import Any, Literal, get_args, get_origin
 
 from pydantic import BaseModel
 from pydantic_core import PydanticUndefined
+from typeguard import check_type
```
isaacbmiller: We need to add this to project deps.

`uv add typeguard`

dspy/signatures/signature.py
```diff
@@ -294,7 +296,7 @@ class MySig(dspy.Signature):
         return Signature(cls.fields, instructions)
 
     @classmethod
-    def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dict[str, Any]) -> type["Signature"]:
+    def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dict[str, Any]) -> "Signature":
```
isaacbmiller: The actual return type was correct before. Applies to all of the method return type changes.

`Signature(...) -> SignatureMeta.__call__ -> make_signature -> Signature class`

dspy/predict/predict.py
```diff
@@ -221,6 +253,45 @@ def get_config(self):
...
+        # Handle Ellipsis in tuples (e.g., tuple[int, ...])
+        args_str = ", ".join("..." if arg is ... else _get_type_name(arg) for arg in args)
+        origin_name = getattr(origin, "__name__", str(origin))
+        return f"{origin_name}[{args_str}]"
+
+    return getattr(origin, "__name__", str(origin))
+
+def _is_value_compatible_with_type(value: Any, expected: type) -> bool:
+    """Return True if the value matches the expected type hint."""
+    from typeguard import TypeCheckError
```
isaacbmiller: is this a heavy enough import that we should make it lazy? Or should we just group it with the other import at the top.

dspy/predict/predict.py
michaelisaac-dev: Good catch! It seems it was getting pulled through another package's transitive dependency.

michaelisaac-dev:
> Did you already design for what happens with custom adapters that don't use the input type? Everybody knows I love custom adapters :)

In that case, the underlying Signature should ideally have no types for its InputFields, so no warning would be logged anyway.

Otherwise, it might be a use case for the flag `dspy.configure(warn_on_type_mismatch=False)`

dspy/signatures/signature.py
```diff
@@ -508,7 +508,7 @@ def load_state(cls, state):
         return signature_copy
 
 
-def ensure_signature(signature: str | type[Signature], instructions=None) -> None | Signature | type[Signature]:
+def ensure_signature(signature: str | type[Signature], instructions=None) -> type[Signature]:
```
isaacbmiller: Return should be `None | type[Signature]`

PR #9318, opened 2026-02-22
fix(gepa): preserve list[Type] inputs in make_reflective_dataset

ugne-sak:
## Summary

When `MultiModalInstructionProposer` is used with an input field of type `list[dspy.Image]`, the reflective dataset creation in `make_reflective_dataset()` was stringifying the entire list. This happened because `isinstance(list_of_images, Type)` returns `False` for a list, so it fell through to `str(input_val)` — embedding images as base64 text in the proposer prompt instead of passing them as proper `image_url` content blocks.

A single `dspy.Image` was already handled correctly; this fix extends that to `list[Type]` inputs.

## Changes

- `dspy/teleprompt/gepa/gepa_utils.py`: add an `elif` branch that preserves lists whose first element is a `Type` instance, matching the existing behaviour for single `Type` inputs
- `tests/teleprompt/test_gepa_instruction_proposer.py`: add regression test `test_reflection_lm_gets_structured_images_list` that mirrors the existing `test_reflection_lm_gets_structured_images` test but uses `list[dspy.Image]`

## Reproduction

```python
student = dspy.Predict("images: list[dspy.Image] -> label: str")
# Before fix: reflection LM receives images as stringified base64 text
# After fix: reflection LM receives structured image_url content blocks
```

### DIFF
--- a/dspy/teleprompt/gepa/gepa_utils.py
+++ b/dspy/teleprompt/gepa/gepa_utils.py
@@ -266,6 +266,8 @@ class DspyAdapter(GEPAAdapter[Example, TraceData, Prediction]):
                     if isinstance(input_val, Type) and self.custom_instruction_proposer is not None:
                         # Keep original object - will be properly formatted when sent to reflection LM
                         new_inputs[input_key] = input_val
+                    elif isinstance(input_val, list) and len(input_val) > 0 and isinstance(input_val[0], Type) and self.custom_instruction_proposer is not None:
+                        new_inputs[input_key] = input_val
                     else:
                         new_inputs[input_key] = str(input_val)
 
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Thank you for the pr!

The fix works for the `list[dspy.Image]` case., but will miss more complicated types that aren't List[Image]

`instruction_proposal.py` already handles this pattern recursively in `render_value_with_images` (lines 197-215), where it walks through lists, tuples, and dicts to find `Type` instances at any depth.

Could you replace the two isinstance branches with a small recursive helper, something like:

```python
def _contains_custom_type(val):
    if isinstance(val, Type):
        return True
    if isinstance(val, (list, tuple)):
        return any(_contains_custom_type(item) for item in val)
    if isinstance(val, dict):
        return any(_contains_custom_type(v) for v in val.values())
    return False
```

Then the check becomes:
```python
if _contains_custom_type(input_val) and self.custom_instruction_proposer is not None:
    new_inputs[input_key] = input_val
else:
    new_inputs[input_key] = str(input_val)
```

Ideally reusing the helper if possible.

PR #9319, opened 2026-02-22
docs: clarify max_rounds behavior in BootstrapFewShot

### LINKED ISSUE #9250: [Bug] max_round in bootstrampfewshot is not working as described in document
### What happened?

the DSPY document says: 
in Bootstrapfewshot:
Max_round: 
Number of iterations to attempt generating the required bootstrap examples. If unsuccessful after max_rounds, the program ends. Defaults to 1.

so it would be like the whole train dataset will run up to max_rounds number until it collects the enough number of bootstrapped samples. (if still not enough to max round , it will stop).

but the testing shows:
in sequence, from sample1, if the metric return score is < thread hold, it will re-run this sample immediately and check the metric return score, if still < thread hold, then it will run again,  up to max_rounds number; if in certain round the metric return score is > thread hold, then it will collected it as 'good' one and move to next smaple2. 

so the problem is:
if max_rounds is working like this, when set it > 1, it is making the samples that are placed further back in the order have less chance to be trained.

why? 
for example, if we only want 1 bootstrapped sample from sample1 and sample2:
Based on document say:
If it is working as document says, sample1 failed to pass the thread hold, then sample2 will be the next one, if sample2 also not pass thread hold. and when round 1 is over, it will start round 2, and sample1 will be trained one more time, if still not pass, then sample1 is trained one more time, it is go on and on, until up to the max rounds.
so in the case, sample2 has chance to be trained.

based on testing:
Now it is working like , try up to max_ruonds times on sample1, MAYBE the 2nd try on sample1 it passes the thread hold, then program stops, sample2 is never be trained.

### Steps to reproduce

set max_rounds > 1

### DSPy version

dspy==3.1.2

bledden:
## Change Summary

The docstring for `max_rounds` in `BootstrapFewShot` described it as the number of full iterations over the training set, but the actual behavior is per-example: each training example is retried up to `max_rounds` times (with cache-busting via fresh `rollout_id` at `temperature=1.0`) before moving to the next example.

This was causing confusion (see #9250) since users expected round-robin behavior over the dataset. A contributor [confirmed](https://github.com/stanfordnlp/dspy/issues/9250#issuecomment-2640419903) the implementation is intentional — the docstring just didn't describe it accurately.

Updated the docstring to reflect the actual per-example retry semantics.

## Related issue number

fix #9250

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -62,8 +62,11 @@ class BootstrapFewShot(Teleprompter):
                 Defaults to 4.
             max_labeled_demos (int): Maximum number of labeled demonstrations to include.
                 Defaults to 16.
-            max_rounds (int): Number of iterations to attempt generating the required bootstrap
-                examples. If unsuccessful after `max_rounds`, the program ends. Defaults to 1.
+            max_rounds (int): Maximum number of bootstrap attempts per training example.
+                Each round after the first uses a fresh rollout with ``temperature=1.0``
+                to bypass caches and gather diverse traces. If a successful bootstrap is
+                found on any round, the example is accepted and the optimizer moves to the
+                next one. Defaults to 1.
             max_errors (Optional[int]): Maximum number of errors until program ends.
                 If ``None``, inherits from ``dspy.settings.max_errors``.
         """


isaacbmiller:
Thanks for submitting!

PR #9320, opened 2026-02-22
fix: only suggest reducing valset in GEPA when it is large

### LINKED ISSUE #8829: [Bug] GEPA prints message to reduce valset no matter what
### What happened?

Context: https://discord.com/channels/1161519468141355160/1410016143782973652/1418044465657942138

### Steps to reproduce

Running GEPA anytime causes this issue

### DSPy version

3.0.1+

bledden:
## Change Summary

GEPA unconditionally logs a message suggesting users reduce their valset size to allow more diverse exploration, even when the valset is already small (e.g., 5-10 examples). This is confusing since there's no benefit to reducing it further at that point.

Now the suggestion is only shown when the valset has more than 35 examples, which is the default `minibatch_size` used in `auto_budget()` for Pareto score tracking. Below that threshold, the valset is already small enough that sampling doesn't help, so only the count is logged.

## Related issue number

fix #8829

### DIFF
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -506,6 +506,9 @@ class GEPA(Teleprompter):
                 "No valset provided; Using trainset as valset. This is useful as an inference-time scaling strategy where you want GEPA to find the best solutions for the provided tasks in the trainset, as it makes GEPA overfit prompts to the provided trainset. In order to ensure generalization and perform well on unseen tasks, please provide separate trainset and valset. Provide the smallest valset that is just large enough to match the downstream task distribution, while keeping trainset as large as possible."
             )
         valset = valset or trainset
+        # 35 matches the default minibatch_size in auto_budget(); when the valset is
+        # at or below this size, suggesting further reduction is unhelpful since GEPA
+        # would already evaluate the full valset per step.
         if len(valset) > 35:
             logger.info(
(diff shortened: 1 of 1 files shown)


dspy/teleprompt/gepa/gepa.py
```diff
@@ -506,9 +506,12 @@ def compile(
                 "No valset provided; Using trainset as valset. This is useful as an inference-time scaling strategy where you want GEPA to find the best solutions for the provided tasks in the trainset, as it makes GEPA overfit prompts to the provided trainset. In order to ensure generalization and perform well on unseen tasks, please provide separate trainset and valset. Provide the smallest valset that is just large enough to match the downstream task distribution, while keeping trainset as large as possible."
             )
         valset = valset or trainset
-        logger.info(
-            f"Using {len(valset)} examples for tracking Pareto scores. You can consider using a smaller sample of the valset to allow GEPA to explore more diverse solutions within the same budget. GEPA requires you to provide the smallest valset that is just large enough to match your downstream task distribution, while providing as large trainset as possible."
-        )
+        if len(valset) > 35:
```
isaacbmiller, line 512: Add a comment explaining how 35 was picked

bledden: Done - added a comment. 35 matches the default `minibatch_size` in `auto_budget()`, so if the valset is already at or below that size GEPA is evaluating the whole thing each step anyway and the suggestion to reduce it doesn't make sense.

PR #9321, opened 2026-02-22
feat: add file output support to inspect_history and fix return type

### LINKED ISSUE #9180: [Feature] Improve inspect_history / pretty_print_history API (return + output redirection)

I often run LLM inference and want to save history examples to a file after module optimization.
Currently this is only possible by temporarily redirecting stdout, which feels hacky.

There are two related issues:

1. **Misleading return value in inspect_history(...)**: inspect_history(...) returns the result of pretty_print_history(GLOBAL_HISTORY, n), but this function returns None. This is confusing and hidden by missing type annotations. **Suggestion**: remove the return statement and add proper typing.

2. **No way to redirect history output**: There is no option to write history to a file or other sink. **Suggestion**: add an optional file argument (object with .write(), propagated to print(file=...)). When file is provided, ANSI colors can be disabled automatically. Default behavior remains unchanged.

### Additional Context

_No response_

bledden:
## Change Summary

Two improvements to `inspect_history()` / `pretty_print_history()`:

1. **Fixed misleading return value**: All three `inspect_history` methods (`BaseLM.inspect_history`, global `inspect_history`, and `Module.inspect_history`) used `return pretty_print_history(...)`, but `pretty_print_history` returns `None`. Removed the misleading returns and added explicit `-> None` type annotations.

2. **Added optional `file` parameter**: All three methods now accept an optional `file` argument (any object with a `.write()` method). When provided, output is written to the file and ANSI color codes are automatically disabled. Default behavior (stdout with colors) is unchanged.

```python
# Write history to a file (no ANSI codes)
with open("history.txt", "w") as f:
    dspy.inspect_history(n=5, file=f)

# Or use StringIO for programmatic access
import io
buf = io.StringIO()
dspy.inspect_history(n=1, file=buf)
history_text = buf.getvalue()
```

## Related issue number

fix #9180

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -160,7 +160,7 @@ class BaseLM:
 
         return new_instance
 
-    def inspect_history(self, n: int = 1, file: 'TextIO | None' = None) -> None:
+    def inspect_history(self, n: int = 1, file: "TextIO | None" = None) -> None:
         pretty_print_history(self.history, n, file=file)
 
     def update_history(self, entry):
@@ -283,7 +283,7 @@ class BaseLM:
         return [result]
 
 
-def inspect_history(n: int = 1, file: 'TextIO | None' = None) -> None:
+def inspect_history(n: int = 1, file: "TextIO | None" = None) -> None:
     """The global history shared across all LMs.
 
     Args:
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -250,7 +250,7 @@ class Module(BaseModule, metaclass=ProgramMeta):
             set_attribute_by_name(self, name, func(predictor))
         return self
 
-    def inspect_history(self, n: int = 1, file: 'TextIO | None' = None) -> None:
+    def inspect_history(self, n: int = 1, file: "TextIO | None" = None) -> None:
         """Display the LM call history for this module.
 
         Prints a formatted view of the most recent language model calls


dspy/utils/inspect_history.py
```diff
@@ -1,35 +1,60 @@
...
     for item in history[-n:]:
         messages = item["messages"] or [{"role": "user", "content": item["prompt"]}]
         outputs = item["outputs"]
         timestamp = item.get("timestamp", "Unknown time")
 
-        print("\n\n\n")
-        print("\x1b[34m" + f"[{timestamp}]" + "\x1b[0m" + "\n")
+        print("\n\n\n", file=out)
+        if colors:
+            print("\x1b[34m" + f"[{timestamp}]" + "\x1b[0m" + "\n", file=out)
```
isaacbmiller: this should use the use_colors function

dspy/utils/inspect_history.py
```diff
@@ -1,35 +1,60 @@
...
+
+    Args:
+        history: The history list to print from.
+        n: Number of recent entries to display. Defaults to 1.
+        file: An optional file-like object to write output to (must have a
+            ``.write()`` method). When provided, ANSI color codes are
+            automatically disabled. Defaults to ``None`` (prints to stdout).
+    """
+    out = file or sys.stdout
+    colors = file is None
```
isaacbmiller: Call this use_colors instead of colors

dspy/primitives/module.py
```diff
@@ -1,6 +1,6 @@
 import inspect
 import logging
-from typing import Any
+from typing import IO, Any
```
isaacbmiller: Should this be typing.TextIO instead of IO?

dspy/primitives/module.py
```diff
@@ -260,11 +260,11 @@ def inspect_history(self, n: int = 1):
         Args:
             n: The number of recent history entries to display.
                 Defaults to 1.
-
-        Returns:
-            The formatted history output.
+            file: An optional file-like object to write output to. When
+                provided, ANSI color codes are automatically disabled.
+                Defaults to ``None`` (prints to stdout).
```
isaacbmiller: remove double quotes in favor of single quotes

dspy/clients/base_lm.py
```diff
@@ -1,6 +1,6 @@
 import datetime
 import uuid
-from typing import Any
+from typing import IO, Any
```
isaacbmiller: same as module:3

dspy/utils/inspect_history.py
bledden: Good catch, swapped it to use `_blue()`. Not sure why I didn't just do that in the first place.

dspy/utils/inspect_history.py
bledden: Yep, renamed to `use_colors` to match the kwarg name on the helper functions.

dspy/primitives/module.py
bledden: Yeah you're right, `TextIO` is more accurate here since we're only dealing with text streams. Fixed across all three files.

dspy/primitives/module.py
bledden: Fixed, switched to single quotes.

dspy/clients/base_lm.py
bledden: Fixed here too.

dspy/clients/base_lm.py
```diff
@@ -283,6 +283,6 @@ def _process_response(self, response):
         return [result]
 
 
-def inspect_history(n: int = 1, file: "IO[str] | None" = None) -> None:
+def inspect_history(n: int = 1, file: 'TextIO | None' = None) -> None:
     """The global history shared across all LMs."""
```
isaacbmiller: Lets add the parameter descriptions to this method too.

```diff
+file: An optional file-like object to write output to. When
+                provided, ANSI color codes are automatically disabled.
+                Defaults to ``None`` (prints to stdout).
```

dspy/utils/inspect_history.py
```diff
@@ -22,7 +22,7 @@ def _blue(text: str, end: str = "\n", *, use_colors: bool = True) -> str:
     return str(text) + end
 
 
-def pretty_print_history(history: list[dict[str, Any]], n: int = 1, file: IO[str] | None = None) -> None:
+def pretty_print_history(history: list[dict[str, Any]], n: int = 1, file: TextIO | None = None) -> None:
```
isaacbmiller, line 25: is it possible to change the type to be `TextIO | <whatever sys.stdout is>` so we dont need to do a null check?

dspy/primitives/module.py
isaacbmiller: doesnt seem to be fixed.

Anecdotally Language models are sometimes really bad at this specific style change. Might need to do it manually.

dspy/utils/inspect_history.py
isaacbmiller: Ideally there already exists a type that has:`must have a `.write()` method`.

dspy/utils/inspect_history.py
```diff
@@ -1,35 +1,57 @@
...
+
+
+def pretty_print_history(history: list[dict[str, Any]], n: int = 1, file: TextIO | None = None) -> None:
+    """Print the last n prompts and their completions.
+
+    Args:
+        history: The history list to print from.
+        n: Number of recent entries to display. Defaults to 1.
+        file: An optional file-like object to write output to (must have a
+            ``.write()`` method). When provided, ANSI color codes are
```
isaacbmiller: double quotes again

isaacbmiller (changes requested):

dspy/clients/base_lm.py
bledden: Added — param descriptions now match the other two inspect_history methods.

dspy/utils/inspect_history.py
bledden: sys.stdout is typed as TextIO in typeshed, so `TextIO | None` already covers it. The null check is just for toggling ANSI colors — when file is None we write to stdout with colors on, when it's provided we write to the given file with colors off.

dspy/primitives/module.py
bledden: My bad — the type annotation on the function signature was already switched to single quotes but I missed the RST double backticks in the docstring. Switched those to single backticks now.

dspy/utils/inspect_history.py
bledden: `TextIO` from typing is exactly that — it's defined as `IO[str]` which covers anything with .read()/.write() that operates on str. sys.stdout satisfies it.

dspy/utils/inspect_history.py
bledden: Fixed — switched to single backticks in the docstrings.

isaacbmiller:
LGTM. Thanks for the contribution

PR #9325, opened 2026-02-23
Adding Reason-Plan-ReAct module

OniricApps:
This is an implementation of the Reason-Planner-ReAct agent architecture using DSPy ReAct module (dspy.ReAct) in a nested mode.

The RP-ReAct approach splits complex agent work into two roles to reduce cognitive overload and improve reliability. Alse naturally handles the trajectory truncation because each call to Proxy-Execution Agent has its own trajectory so the trajectory pf the Reasoner-Planner is not growing as fast as a "simple" ReAct.

* A Reasoner-Planner Agent (RPA) handles high-level strategy, breaking a user request into simple sub-questions and
adapting the plan based on results.
* A Proxy-Execution Agent (PEA) handles low-level execution, translating each sub-question into precise tool calls with
correct parameters and syntax.

RP-ReAct Methodology adapted to DSPy from the original paper (Giargiari et al. 2025) https://arxiv.org/pdf/2512.03560
Proof Code from the paper — https://github.com/giargiapower/RP-ReAct

### DIFF
--- a/dspy/predict/__init__.py
+++ b/dspy/predict/__init__.py
@@ -9,6 +9,7 @@ from dspy.predict.predict import Predict
 from dspy.predict.program_of_thought import ProgramOfThought
 from dspy.predict.react import ReAct, Tool
 from dspy.predict.refine import Refine
+from dspy.predict.rpreact import RPReAct
 from dspy.predict.rlm import RLM
 
 __all__ = [
@@ -23,6 +24,7 @@ __all__ = [
     "ReAct",
     "Refine",
     "RLM",
+    "RPReAct",
     "Tool",
     "Parallel",
 ]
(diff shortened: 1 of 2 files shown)


isaacbmiller:
This paper is very new, so it may not be a good time to ship to DSPy directly. 

However, if you can share a github gist/colab notebook to demonstrate it improves the performance significantly compared to existing modules, we can check in the code. Thank you!

OniricApps:
Thanks for the answer @isaacbmiller . I understand you. Unfortunately, I probably won't have time and resources to do the demonstration you ask in the following weeks. So, let's let this PR in standby or whatever state you prefer until I have the time to do it or somebody else do it. Maybe the author of the paper? @giargiapower

Note that I'm actively using this implementation of RP-ReAct and it works fine, better than just ReAct in some cases, same performance the rest. However, I don't have a dataset to do the benchmark you ask and I typically use local, small, LMs, and I can't say if it is the reason to improve results in some cases. A new and large LLM may be self-sufficient in planning, reasoning, and executing while maintaining a large trajectory (context) so, just ReAct is enough.

PR #9328, opened 2026-02-24
fix: run evaluation on main thread when num_threads=1

### LINKED ISSUE #9150: How to use DSPy with stateful agent benchmarks?
How should we use DSPy with interactive, stateful agent benchmarks such as [AppWorld](https://github.com/stonybrooknlp/appworld) or [Terminal Bench](https://github.com/laude-institute/terminal-bench)?

I'm aware that agents like `dspy.ReAct` can invoke tools. The [advanced tool use](https://dspy.ai/tutorials/tool_use/#tutorial-advanced-tool-use) tutorial shows how each task can have a different set of tools. But, these tools seem to be "retrieval-only" and does not mutate the task state in any way. Whereas in a benchmark like AppWorld, the agent can e.g. write to the filesystem, whose effects are persistent and needs to be explicilty reset should you wish to start fresh.

I'm not sure how to use such stateful task environments with DSPy optimizers. For instance, there's no clean hook into GEPA to control when environments are created or reset.

These environments are also usually process-bound, allowing only one active instance per process (thus not easily parallelizable via threads).

1. Is there a recommended or “standard” pattern for using DSPy with stateful benchmarks? Should we write a custom agent subclassing `dspy.Module`, whose `forward` function handles the task environment creation/deletion? This might work but feels like a hack. The agent technically shouldn't handle this.
2. Are there existing tutorials, examples, or design guidelines for integrating DSPy with these interactive agent benchmarks?

Any help is greatly appreciated!

zamal-db:
Closes #9150

## Problem

When using `dspy.Evaluate` with `num_threads=1`, work is still dispatched to a `ThreadPoolExecutor`. This breaks frameworks like [AppWorld](https://github.com/stonybrooknlp/appworld) and [Terminal Bench](https://github.com/laude-institute/terminal-bench) that call `signal.signal()` inside their task execution, because `signal.signal()` can only be called from the main thread:

\\\
ValueError: signal only works in main thread of the main interpreter
\\\

Even setting `num_threads=1` doesn't help, since `ParallelExecutor` unconditionally routes through `ThreadPoolExecutor`.

## Solution

Add a `_execute_sequential()` method to `ParallelExecutor` that runs items directly on the calling thread when `num_threads == 1`. The sequential path preserves all existing behavior:

- Error counting and `max_errors` cancellation
- `failed_indices` and `exceptions_map` tracking
- Progress bar updates (both regular and `compare_results` mode)
- `KeyboardInterrupt` handling

When `num_threads > 1`, the existing `_execute_parallel()` path is used unchanged.

## Changes

- **`dspy/utils/parallelizer.py`**: Add `_execute_sequential()` and route to it from `execute()` when `num_threads == 1`
- **`tests/utils/test_parallelizer.py`**: Add 5 tests for sequential execution (main-thread verification, error handling, max_errors, failed index tracking, compare_results)
- **`tests/evaluate/test_evaluate.py`**: Add 1 end-to-end test verifying `Evaluate(num_threads=1)` runs the metric on the main thread

### DIFF
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -71,12 +71,7 @@ class ParallelExecutor:
         return safe_func
 
     def _execute_sequential(self, function, data):
-        """Execute items sequentially on the main thread.
-
-        This avoids spawning a ThreadPoolExecutor when num_threads == 1,
-        which is necessary for frameworks that require main-thread execution
-        (e.g. those using signal.signal()).
-        """
+        """Execute items sequentially on the main thread."""
         results = [None] * len(data)
 
         pbar = tqdm.tqdm(
@@ -92,22 +87,8 @@ class ParallelExecutor:
                     break
 
                 outcome = function(item)
-
-                if isinstance(outcome, Exception):
-                    self.failed_indices.append(idx)
-                    self.exceptions_map[idx] = outcome
-                else:
-                    results[idx] = outcome
-
-                if self.compare_results:
-                    vals = [r[-1] for r in results if r is not None]
-                    self._update_progress(pbar, sum(vals), len(vals))
-                else:
-                    self._update_progress(
-                        pbar,
-                        len([r for r in results if r is not None]),
-                        len(data),
-                    )
+                self._process_outcome(results, idx, outcome)
+                self._report_progress(pbar, results, len(data))
         except KeyboardInterrupt:
             self.cancel_jobs.set()
             logger.warning("SIGINT received. Cancelling.")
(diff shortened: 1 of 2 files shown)


tests/evaluate/test_evaluate.py
```diff
@@ -56,6 +56,38 @@ def test_evaluate_call():
     assert score.score == 100.0
 
 
+def test_evaluate_single_thread_runs_in_main_thread():
+    """Evaluate with num_threads=1 should run in the main thread (issue #9150).
```
isaacbmiller: unnecessary mention of the issue and usecase for main thread.

dspy/utils/parallelizer.py
```diff
@@ -68,6 +70,57 @@ def safe_func(item):
...
         return safe_func
 
+    def _execute_sequential(self, function, data):
+        """Execute items sequentially on the main thread.
+
+        This avoids spawning a ThreadPoolExecutor when num_threads == 1,
+        which is necessary for frameworks that require main-thread execution
+        (e.g. those using signal.signal()).
+        """
+        results = [None] * len(data)
```
isaacbmiller, line 75: Is there any way to share more logic between `execute_sequential` and `execute_parallel`

zamal-db:
Hey, pushed a fix for both comments. Extracted shared _process_outcome/_report_progress helpers so sequential and parallel use the same logic now, and trimmed the test docstring. Ready for another look when you get a chance!

okhat:
I like this!

isaacbmiller:
Thanks for the contribution @zamal-db !

Issue #8369, opened 2025-06-12
[Bug] Vertex AI location


kkannar-infinitusai:
### What happened?

Is there a way for dspy to honor the location/region being passed to vertex AI? 
 

`lm = dspy.LM('vertex_ai/gemini-2.0-flash',  project="A",  location="us-west1",  cache=False)` does not seem to honor the` location  passed here and uses the gcloud config's location. 

Using `vertexai.init(project="A", location="us-west1")` did not work. We do not want to change the gcloud config. It defaults to us-central1 at the moment. 

### Steps to reproduce

`lm = dspy.LM('vertex_ai/gemini-2.0-flash',  project="A",  location="us-west1",  cache=False)` 

### DSPy version

2.6.27

arnavsinghvi11:
Hi @kkannar-infinitusai , you should be setting `vertex_location`. Feel free to reference the LiteLLM VertexAI config [docs](https://docs.litellm.ai/docs/providers/vertex#authentication---vertex_project-vertex_location-etc) for more info

kkannar-infinitusai:
oh fantastic. Thank you!

Issue #7974, opened 2025-03-17
[Feature] Google GenAI SDK


TheMellyBee:
### What feature would you like to see?

Google has released a unified SDK for both Gemini and Vertex AI that has the newest models. I'm currently evaluating frameworks to move forward with, and we'd love to get going with DSPy. However this LLM is critical for us. 

https://googleapis.github.io/python-genai/

### Would you like to contribute?

- [ ] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

TheMellyBee:
I'm flexible and happy to help. However I'm also low on time.

arnavsinghvi11:
Hi @TheMellyBee , thanks for raising this!

DSPy LM queries go through LiteLLM, which does support both Gemini and VertexAI in their providers (saw some example snippets [here](https://github.com/BerriAI/litellm/issues/8809)). 

Does the newly-released SDK require a different provider? Raising this with LiteLLM would be the best way to integrate with DSPy!

TheMellyBee:
It does! This is a new unified API that will allow API keys from either of those. It also cleaned things up quite a bit. Moving forward they will be removing support for the other SDK. I've gone ahead and requested this feature with LiteLLM

chenmoneygithub:
You can also subclass from `dspy.BaseLM` to make your own LM that under the hood uses `google-genai`: https://github.com/stanfordnlp/dspy/blob/6ee8cdca4bf3283a8bde6b92c628cbbd0851fe9b/dspy/clients/base_lm.py#L28

Issue #1021, opened 2024-05-13
Google Vertex AI returns 503 error


felixgao:
```
Traceback (most recent call last):
  File "/Users/ggao/github/ged/GED_MONOREPO/projects/gemini/.venv/lib/python3.11/site-packages/google/api_core/grpc_helpers.py", line 76, in error_remapped_callable
    return callable_(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ggao/github/ged/GED_MONOREPO/projects/gemini/.venv/lib/python3.11/site-packages/grpc/_channel.py", line 1181, in __call__
    return _end_unary_response_blocking(state, call, False, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ggao/github/ged/GED_MONOREPO/projects/gemini/.venv/lib/python3.11/site-packages/grpc/_channel.py", line 1006, in _end_unary_response_blocking
    raise _InactiveRpcError(state)  # pytype: disable=not-instantiable
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "Getting metadata from plugin failed with error: 'str' object has no attribute 'before_request'"
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"Getting metadata from plugin failed with error: \'str\' object has no attribute \'before_request\'", grpc_status:14, created_time:"2024-05-13T14:23:12.184109-07:00"}"
>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/ggao/github/ged/GED_MONOREPO/projects/gemini/gemini/doc_qa.py", line 178, in <module>
    pred = predictor(document="Test Document")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ggao/github/ged/GED_MONOREPO/projects/gemini/.venv/lib/python3.11/site-packages/dspy/predict/predict.py", line 61, in __call__
    return self.forward(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #9329, opened 2026-02-24
docs: add Vertex AI (GCP) tab to Language Models guide

zamal-db:
Adds a **Vertex AI (GCP)** tab to the Language Models page showing how to use DSPy with Google Cloud service account credentials  both via kwargs and environment variables.

Several issues (#8369, #1021, #7974) show developers struggling with Vertex AI setup in DSPy. The confusion is always the same: wrong model prefix, wrong parameter names, or missing auth config. This tab addresses that directly with tested examples and a short pitfalls note.

Both approaches were tested end-to-end against a real GCP service account.

### DIFF
--- a/docs/docs/learn/programming/language_models.md
+++ b/docs/docs/learn/programming/language_models.md
@@ -41,6 +41,44 @@ dspy.configure(lm=lm)
         dspy.configure(lm=lm)
         ```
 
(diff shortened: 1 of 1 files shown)


isaacbmiller:
Thank you for the contribution!

PR #9330, opened 2026-02-24
Add Exa web search retriever and ReAct tools

10ishq:
## Summary

Adds [Exa](https://exa.ai) as a native integration in DSPy, providing both a retriever module for RAG pipelines and tool classes for ReAct agents.

Exa is a web search API purpose-built for AI — it returns clean, parsed content (not just links) and supports neural, keyword, and auto search modes.

### `ExaRM` — Retriever module (`dspy.Retrieve` subclass)
For use in declarative RAG pipelines. Performs live web search, making it suitable for queries requiring up-to-date information.

```python
from dspy.retrievers.exa import ExaRM

retriever = ExaRM(k=3)
dspy.configure(lm=lm, rm=retriever)
results = dspy.Retrieve(k=3)("what happened in AI this week").passages
```

### `ExaSearchTool` — Web search tool for ReAct agents
```python
from dspy.retrievers.exa import ExaSearchTool

tool = ExaSearchTool(num_results=5)
agent = dspy.ReAct("question -> answer", tools=[tool])
```

### `ExaContentsTool` — URL content extraction tool for ReAct agents
```python
from dspy.retrievers.exa import ExaContentsTool

agent = dspy.ReAct("question -> answer", tools=[ExaSearchTool(), ExaContentsTool()])
```

### `ExaFindSimilarTool` — Find similar pages tool for ReAct agents
```python
from dspy.retrievers.exa import ExaFindSimilarTool

tool = ExaFindSimilarTool()
agent = dspy.ReAct("question -> answer", tools=[tool])
```

## Installation

```bash
pip install dspy[exa]
```

Requires an Exa API key (pass as `exa_api_key` parameter or set `EXA_API_KEY` env var). Get one at https://exa.ai.

## Changes

- `dspy/retrievers/exa.py` — ExaRM retriever + ExaSearchTool, ExaContentsTool, ExaFindSimilarTool
- `tests/retrievers/test_exa.py` — 16 unit tests (all passing)
- `pyproject.toml` — Added `exa` optional dependency (`exa-py>=2.0.0`)

All API calls include `x-exa-integration: exa-dspy` header for usage tracking.

---

[Link to Devin run](https://app.devin.ai/sessions/91a0b735355242bfa8292e0eed2fe2d4)
Requested by Tanishq Jaiswal (@10ishq)

PR #9332, opened 2026-02-25
docs: add Exa web search tutorial for ReAct agents

10ishq:
Adds a tutorial showing how to build a web search research agent using DSPy ReAct with [Exa](https://exa.ai) search tools (via the [`exa-dspy`](https://github.com/exa-labs/exa-dspy) package).

## Changes

- **New tutorial**: `docs/docs/tutorials/exa_web_search/index.md`
  - Covers `ExaSearchTool`, `ExaContentsTool`, and `ExaFindSimilarTool` with ReAct
  - Domain-specific search (date/domain filtering)
  - Multi-source research agent example
  - Tool configuration reference
- **Nav update**: Added to Real-World Examples in `mkdocs.yml`
- **Index update**: Added entry in `tutorials/real_world_examples/index.md`

Per the suggestion in https://github.com/stanfordnlp/dspy/pull/9330#issuecomment-2807299519 to contribute a tutorial rather than a custom retriever integration.

### DIFF
--- a/docs/docs/tutorials/exa_web_search/index.md
+++ b/docs/docs/tutorials/exa_web_search/index.md
@@ -1,20 +1,30 @@
-# Web Research Agent with DSPy and Exa
+# Real-Time Web Research with DSPy and Exa
 
-Most DSPy programs retrieve from static corpora — a vector DB loaded once, or ColBERTv2 over a Wikipedia snapshot. This works for closed-domain QA, but breaks down when your agent needs **live, up-to-date web information**: recent news, current research, real-time market data, or pages that didn't exist when your corpus was built.
+## The Problem: DSPy Programs Can't Access Live Information
 
-[Exa](https://exa.ai) is a search engine built for AI agents. Unlike scraping Google or calling a keyword API, Exa uses a neural search index that understands *meaning* — so the natural-language queries that DSPy modules generate work well out of the box. It also returns clean extracted text, so you don't need a browser or HTML parser in your pipeline.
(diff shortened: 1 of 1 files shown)


isaacbmiller:
Thanks for submitting!

I don't think this tutorial shows the value that the Exa integration could brings to DSPy relative to any other tool.

In order to merge we should heavily revise the tutorial to show the value proposition, give some unique insights about how this can change a user's workflow, and demonstrate when and why a DSPy user should really consider using this integration.

10ishq:
Hey @isaacbmiller, thanks so much for your thoughts. I rewrote the tutorial based on your feedback. Key changes:

- Opens with a concrete problem: a recency question that static corpora fail on, with real agent output showing Exa solving it
- "Why Exa over other search APIs" section with specific technical reasons (neural search handles optimizer-generated NL queries, built-in content extraction, sub-200ms latency, find-similar) + links to [eval benchmarks](https://exa.ai/evals)
- "When to use this (and when not to)" — honest guidance so readers know exactly when this is relevant
- Working optimizer example with a real metric function instead of `...`

The core value prop for DSPy users: optimizers generate natural-language queries that Exa's neural search handles natively, removing the need for query formatting steps that keyword APIs require.

isaacbmiller:
It would be interesting to see how you are able to perform on benchmarks that use the actual dspy programs with Exa tools.

And showing how you can prompt optimize something on one of these benchmarks to get even more juice out of the exa tools would be awesome.

The most compelling(but also most involved) version of this would be comparing how a DSPy program could do even better compared with Exa compared to the other common LLM search APIs as tools.

PR #9333, opened 2026-02-25
Add safe (pickle-free) module serialization

### LINKED ISSUE #9164: [FR] Implement safe save / load methods for DSpy module and DSpy settings

Motivation:

Currently, DSpy uses `cloudpickle` to dump / load DSpy module program, but cloudpickle causes CVE issue.

and for DSpy settings, DSpy library does not provide save / load methods.

In MLflow DSpy flavor code, it needs to save / load DSpy module program and DSpy settings, it has to rely on cloudpickle and can't avoid CVE.

### Additional Context

_No response_

debu-sinha:
Closes #9164

## What changes are proposed

This PR adds a JSON-based save/load path that avoids cloudpickle entirely, addressing CVE concerns with untrusted `.pkl` files. Users can now save and load DSPy programs without any pickle involvement:

```python
# Save
module.save("my_model", save_program=True, safe=True)

# Load (no allow_pickle needed)
loaded = dspy.load("my_model")
```

The safe format stores:
- `metadata.json` with `"format": "safe_v1"` for auto-detection
- `program.json` with the module class, sub-module tree, and full state

On load, modules are reconstructed via `object.__new__()` + `_base_init()` + `load_state()`, following the same pattern as PyTorch's `state_dict`.

### Key changes

**Signature serialization** (`dspy/signatures/signature.py`):
- `dump_state()` now emits field metadata: `name`, `field_type` (input/output), and `type` (serialized Python type annotation)
- `load_state()` has three paths: (1) update in-place when field names match, (2) reconstruct from scratch when fields differ (safe load path), (3) legacy format backward compat
- Added `_type_to_str()` / `_str_to_type()` for round-tripping Python type annotations through JSON

**Predict config** (`dspy/predict/predict.py`):
- `dump_state()` now includes `config` (temperature, max_tokens, etc.)
- `load_state()` restores config with backward compat for older state dicts

**Module tree** (`dspy/primitives/base_module.py`):
- Added `_collect_module_tree()` that walks `__dict__` recursively to record sub-module paths and classes (handles attributes, lists, dicts)
- Added `safe=True` parameter to `save()` which writes `program.json` + `metadata.json` instead of `program.pkl`

**Safe loading** (`dspy/utils/saving.py`):
- `dspy.load()` auto-detects `safe_v1` format from metadata and loads without requiring `allow_pickle`
- `_reconstruct_module()` creates modules via `object.__new__()`, wires the sub-module tree, then calls `load_state()`
- Handles bracketed paths like `layers[0].predict` and `modules['key']`
- Legacy cloudpickle format continues to work unchanged

### Backward compatibility

- Existing `save_program=True` (without `safe=True`) behavior is unchanged
- `load_state()` handles both new enhanced format and legacy format transparently
- `dspy.load()` auto-detects format; legacy `.pkl` models still work with `allow_pickle=True`

## How is this PR tested

27 new tests in `tests/utils/test_safe_saving.py`:
- Type serialization round-trips (str, int, list[str], dict[str, int], etc.)
- Signature dump/load with field metadata (in-place update, reconstruction from scratch, legacy compat)
- Predict config serialization and backward compat
- Safe save file structure validation
- Full round-trip for ChainOfThought, Predict, and nested custom modules
- Version mismatch warnings
- Error cases (missing class, non-directory path)

All 41 existing signature tests pass. All 13 relevant base_module tests pass.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -64,6 +64,7 @@ class Predict(Module, Parameter):
 
         state["signature"] = self.signature.dump_state()
         state["lm"] = self.lm.dump_state() if self.lm else None
+        state["config"] = self.config
         return state
 
     def load_state(self, state: dict) -> "Predict":
@@ -75,7 +76,7 @@ class Predict(Module, Parameter):
         Returns:
             Self to allow method chaining.
         """
-        excluded_keys = ["signature", "extended_signature", "lm"]
+        excluded_keys = ["signature", "extended_signature", "lm", "config"]
         for name, value in state.items():
             # `excluded_keys` are fields that go through special handling.
             if name not in excluded_keys:
@@ -83,6 +84,9 @@ class Predict(Module, Parameter):
 
         self.signature = self.signature.load_state(state["signature"])
         self.lm = LM(**state["lm"]) if state["lm"] else None
+
+        if "config" in state:
+            self.config = state["config"]
 
         if "extended_signature" in state:  # legacy, up to and including 2.5, for CoT.
             raise NotImplementedError("Loading extended_signature is no longer supported in DSPy 2.6+")
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -160,31 +160,79 @@ class BaseModule:
         for name, param in self.named_parameters():
             param.load_state(state[name])
 
-    def save(self, path, save_program=False, modules_to_serialize=None):
+    def _collect_module_tree(self):
+        """Walk the module graph and record each sub-module's path and class.
+
+        Returns a list ordered so that parents appear before their children,
+        which is the order needed for reconstruction.
+        """
+        tree = []
+
+        def _walk(obj, prefix):
+            for attr_name, attr_value in obj.__dict__.items():
+                if isinstance(attr_value, BaseModule):
+                    path = f"{prefix}.{attr_name}" if prefix else attr_name
+                    tree.append(
+                        {
+                            "path": path,
+                            "class": f"{attr_value.__class__.__module__}.{attr_value.__class__.__qualname__}",
+                        }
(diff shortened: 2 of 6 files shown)


isaacbmiller:
We already have a non-pkl saving/loading option by saving to a .json path.

You can do `module.save("path.json")` to save many non-parameter attributes. You will still need to re-instantiate the module yourself, but the DSPy specific components such as an optimized prompt will remain.

Can you explain the desired usecases that the json path doesnt fill yet?

debu-sinha:
The existing `.json` path saves state only. You need to manually instantiate the module class before calling `load()`. That works when you know the class structure, but it doesn't cover cases where the loader needs to reconstruct the full module from scratch (model serving, MLflow integration, sharing across teams).

`save_program=True` solves that by saving the full module so `dspy.load()` can reconstruct everything, but it goes through cloudpickle, which executes arbitrary code on load.

This PR adds `safe=True` to `save_program=True` so you get full reconstruction without cloudpickle. Same `dspy.load()` experience, but the serialization is JSON-only. The module class, sub-module tree, and state all go to `program.json` instead of `program.pkl`.

The three paths after this PR:
- `module.save("path.json")`: state only, manual instantiation (unchanged)
- `module.save("dir", save_program=True)`: full program via cloudpickle (unchanged)
- `module.save("dir", save_program=True, safe=True)`: full program via JSON (new)

migurski:
I tested this and found it to work, very good. In my example the type is a named subclass of `TypedDict`, it comes out with just its name “Place” in the result JSON. I wonder if some deeper introspection on such types would be useful, or to expose them as compatible underlying simple Python types like `dict`?

debu-sinha:
Closing this since I am not pursuing this direction anymore. Thanks for the initial feedback.

PR #9334, opened 2026-02-25
fix(saving): block unsafe LM loading keys

isaacbmiller:
When a DSPy program is saved to JSON, the serialized LM config can include endpoint override keys like api_base, base_url, or model_list. A tampered file could inject a malicious endpoint, silently redirecting API calls — along with API keys sourced from environment variables — to an attacker-controlled server.

This PR strips these unsafe keys by default during deserialization, with an opt-in allow_unsafe_lm_state=True flag for trusted files. The sanitization happens in Predict.load_state predict.py:89 via a new helper predict.py:22-38, and the flag is threaded through the module hierarchy so that module.load("model.json") is safe by default.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -91,7 +91,7 @@ class Predict(Module, Parameter):
 
         Args:
             state: The saved state of a `Predict` object.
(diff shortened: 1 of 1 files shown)


okhat:
LGTM

PR #9335, opened 2026-02-26
fix: dspy.Parallel breaks streaming with RuntimeError from non-AnyIO worker threads

### LINKED ISSUE #9154: [Bug] dspy.Parallel breaks streaming and status adapters
### What happened?

Using dspy.Parallel in a module and trying to use stream with status adapters with it breaks with 

2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when applying callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>'s end handler on function __call__: This function can only be run from an AnyIO worker thread.

### Steps to reproduce

```python
import asyncio

import dspy
from dspy.streaming import StatusMessageProvider, StreamListener


lm = dspy.LM(
    model="openai/gpt4.1"
)

dspy.configure(lm=lm)


class SimpleSignature(dspy.Signature):
    """Answer a simple question."""
    question: str = dspy.InputField(description="Question to answer")
    answer: str = dspy.OutputField(description="Answer to the question")


class SimpleStatusProvider(StatusMessageProvider):
    """Simple status provider that logs events."""
    
    def lm_start_status_message(self, instance, inputs) -> str:
        return "LM started"
    
    def lm_end_status_message(self, outputs) -> str:
        return "LM completed"


class ModuleWithParallelCalls(dspy.Module):
    """Module that makes parallel LM calls using dspy.Parallel."""
    
    def __init__(self):
        super().__init__()
        self.predict = dspy.Predict(SimpleSignature)
    
    def forward(self, questions: list[str]) -> list[str]:
        """Run multiple questions in parallel using dspy.Parallel."""
        
        pairs = [(self.predict, {"question": q}) for q in questions]
        results = dspy.Parallel()(pairs)
        
        return dspy.Prediction(result=results)


# Create the module and streamify it
module = ModuleWithParallelCalls()
stream_listeners = [StreamListener(signature_field_name="answer")]
streamed_module = dspy.streamify(
    module,
    async_streaming=True,
    status_message_provider=SimpleStatusProvider(),
)


def test_non_streamified():
    """Non-streamified version works fine."""
    print("=" * 50)
    print("Testing NON-STREAMIFIED module (should work)")
    print("=" * 50)
    
    questions = [
        "What is 2+2?",
        "What is the capital of France?",
        "What color is the sky?",
    ]
    
    result = module(questions=questions)
    print(f"Result: {result}")
    print()


async def test_streamified():
    """Streamified version fails with AnyIO worker thread error."""
    print("=" * 50)
    print("Testing STREAMIFIED module (will show AnyIO error)")
    print("=" * 50)
    
    questions = [
        "What is 2+2?",
        "What is the capital of France?",
        "What color is the sky?",
    ]
    
    async for chunk in streamed_module(questions=questions):
        print(chunk)


if __name__ == "__main__":
    # This works fine
    test_non_streamified()
    
    # This fails with "This function can only be run from an AnyIO worker thread"
    asyncio.run(test_streamified())
``` 

## Output

```
==================================================
Testing NON-STREAMIFIED module (should work)
==================================================
Processed 3 / 3 examples: 100%|██████████████████████████████████████████████████████████████████████| 3/3 [00:01<00:00,  1.98it/s]
Result: Prediction(
    result=[Prediction(
    answer='4'
), Prediction(
    answer='The capital of France is Paris.'
), Prediction(
    answer='The sky is typically blue.'
)]
)

==================================================
Testing STREAMIFIED module (will show AnyIO error)
==================================================
  0%|                                                                                                        | 0/3 [00:00<?, ?it/s]2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when calling callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>: This function can only be run from an AnyIO worker thread
2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when calling callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>: This function can only be run from an AnyIO worker thread
2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when applying callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>'s end handler on function __call__: This function can only be run from an AnyIO worker thread.
2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when calling callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>: This function can only be run from an AnyIO worker thread
2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when applying callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>'s end handler on function __call__: This function can only be run from an AnyIO worker thread.
Processed 1 / 3 examples:   0%|                                                                              | 0/3 [00:00<?, ?it/s]2026/01/05 16:11:27 WARNING dspy.utils.callback: Error when applying callback <dspy.streaming.messages.StatusStreamingCallback object at 0x12eac4e50>'s end handler on function __call__: This function can only be run from an AnyIO worker thread.
Prediction(
    result=[Prediction(
    answer='4'
), Prediction(
    answer='The capital of France is Paris.'
), Prediction(
    answer='The sky is typically blue.'
)]
)
``` 

### DSPy version

3.1.0b1

zamal-db:
## Problem

I was building a module that runs multiple LM calls in parallel using dspy.Parallel and streams results back to the user via dspy.streamify. Every parallel call crashed with:

`
RuntimeError: This function can only be run from an AnyIO worker thread
`

The single-call (sequential) version worked fine  only the Parallel path broke.

## Root Cause

dspy.Parallel dispatches work to a standard ThreadPoolExecutor. The streaming infrastructure (both sync_send_to_stream in messages.py and sync_stream_completion in lm.py) relied on syncer.syncify() which internally calls nyio.from_thread.run()  but that only works from AnyIO-managed worker threads, not from plain TPE threads.

Two crash sites:

1. **dspy/streaming/messages.py  sync_send_to_stream()**: The StatusStreamingCallback sends status messages from TPE threads via syncify()  crash.
2. **dspy/clients/lm.py  sync_stream_completion()**: The LM streaming path calls syncify(stream_completion) from TPE threads  crash.

## Fix

**messages.py**: Replaced the syncify-based approach with send_nowait() + call_soon_threadsafe(). When called from a thread without a running event loop (the TPE case), we schedule the send_nowait on the consumer's event loop via call_soon_threadsafe, which is the standard thread-safe way to poke an asyncio loop from another thread. The consumer loop reference is cached when streamify() starts.

**lm.py**: Wrapped the existing syncify() call in a try/except. On RuntimeError, falls back to litellm's synchronous completion(stream=True) API and forwards chunks via the fixed sync_send_to_stream.

**streamify.py**: One-line addition to cache the running event loop so background threads can find it.

## Testing

- All 7 existing 	ests/predict/test_parallel.py tests pass
- Verified end-to-end with a real LLM (Gemini via Vertex AI)  Parallel + streamify completes without errors

Closes #9154

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -338,8 +338,28 @@ def _get_stream_completion_fn(
         return litellm.stream_chunk_builder(chunks)
 
     def sync_stream_completion():
-        syncified_stream_completion = syncify(stream_completion)
-        return syncified_stream_completion(request, cache_kwargs)
+        try:
+            syncified_stream_completion = syncify(stream_completion)
+            return syncified_stream_completion(request, cache_kwargs)
+        except RuntimeError:
+            # Not in an AnyIO worker thread (e.g. dspy.Parallel's ThreadPoolExecutor).
+            # Fall back to litellm's synchronous streaming API and forward chunks
+            # via sync_send_to_stream which is safe to call from any thread.
+            from dspy.streaming.messages import sync_send_to_stream
+
+            response = litellm.completion(
+                cache=cache_kwargs,
+                stream=True,
+                headers=headers,
+                **request,
+            )
+            chunks = []
+            for chunk in response:
+                if caller_predict_id:
+                    chunk.predict_id = caller_predict_id
+                chunks.append(chunk)
+                sync_send_to_stream(stream, chunk)
+            return litellm.stream_chunk_builder(chunks)
 
     async def async_stream_completion():
         return await stream_completion(request, cache_kwargs)
--- a/dspy/streaming/messages.py
+++ b/dspy/streaming/messages.py
@@ -1,9 +1,7 @@
 import asyncio
-import concurrent.futures
+import threading
 from dataclasses import dataclass
 from typing import Any
-
-from asyncer import syncify
 
 from dspy.dsp.utils.settings import settings
 from dspy.utils.callback import BaseCallback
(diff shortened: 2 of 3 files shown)


isaacbmiller:
Thanks for the PR. The fix here patches the surface error without addressing the root issue.

The streaming architecture routes chunks to listeners by `predict_id` (the `id()` of the predictor object). When `dspy.Parallel` calls the same predictor from multiple threads, they all share the same `predict_id`, so the listener can't tell which parallel call a chunk belongs to. The `StreamListener` also has internal state (`stream_start`, `stream_end`, buffer state) that isn't thread-safe and will get corrupted with concurrent calls.

As @TomeHirata noted on the original issue, streaming + parallel is fundamentally unsupported right now, not just crashing.

This needs a design-level fix to the streaming architecture to support per-call identity, which is a bigger undertaking. I would encourage a deeper dive into the original thread and issues discussed inside of it.

Going to close this for now, but the issue (#9154) stays open.

PR #9341, opened 2026-02-26
fix(interpreter): Harden JSONRPC communication and tool bridge

isaacbmiller:
This PR makes three changes to the tool bridge — the protocol that lets sandboxed Python code call host-side functions via a JS intermediary:

1. Keyword-only tool calls: The bridge no longer passes positional args, only kwargs. The *args, **kwargs fallback for untyped tools is removed. Tools with *args or *kwargs are not supported, but positional arguments and keyword arguments in the function defn are.

2. Error-as-return-value in JS bridge: toolCallBridge previously used throw for errors, which could trigger Deno's unhandledrejection handler and crash the process. It now returns a structured error object that the generated Python wrapper detects and re-raises as RuntimeError.

3. Resilient stdout parsing: `_send_request` and `execute`now loops to skip non-JSON lines from Deno's stdout (e.g., Pyodide package loading messages) instead of crashing on the first non-JSON line.

### DIFF
--- a/tests/primitives/test_python_interpreter.py
+++ b/tests/primitives/test_python_interpreter.py
@@ -312,6 +312,26 @@ def test_tool_all_positional_args():
         assert result == "60"
 
 
+def test_tool_error_surfaces_as_runtime_error():
+    """Test that exceptions raised by a tool surface as RuntimeError in the sandbox."""
+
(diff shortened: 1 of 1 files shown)


isaacbmiller:
@codex

PR #9342, opened 2026-02-26
feat(interpreter): Add Pydantic type coercion

isaacbmiller:
The deno sandbox for RLM communicates with the host Python process via JSON-RPC, meaning everything crossing the boundary must be JSON-serializable. Previously, passing a pydantic model anywhere in this pipeline would fail.

This PR adds automatic pydantic model conversion at every boundary point:

**Input variables** → model_dump() before injection into sandbox; JSON schema shown to LLM in variable metadata
**Tool arguments** → TypeAdapter.validate_python() coerces raw dicts back into models when calling host-side tools
**Tool return values** → model_dump() before sending back through JSON-RPC
**Output fields (SUBMIT)** → JSON schema propagated to sandbox so SUBMIT() knows expected shapes; model_dump() called before FinalOutput

The core pattern at each boundary is the same — serialize with model_dump(mode="json") on the way out, validate with TypeAdapter on the way in.

We do not do any coercion to a specific type inside the REPL inside this PR. That is to come in a follow on pr.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -20,7 +20,7 @@ import pydantic
 
 import dspy
 from dspy.adapters.types.tool import Tool
-from dspy.adapters.utils import parse_value, translate_field_type
+from dspy.adapters.utils import _get_json_schema, parse_value, translate_field_type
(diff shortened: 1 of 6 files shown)


isaacbmiller:
superseded by #9753

PR #9348, opened 2026-02-27
fix(dspy): Make LiteLLM import lazy

isaacbmiller:
## Summary

Defers all `import litellm` statements from module level to first use across 12 files. All call sites go through a single `get_litellm()` helper that lazily imports and configures litellm on first call.

`litellm` accounts for ~550ms of import time (700+ submodules). We investigated importing specific submodules to avoid the full cost, but every path triggers the full `litellm/__init__.py` due to circular imports:

| Import path | Time |
|---|---|
| `import litellm` | 0.544s |
| `from litellm.exceptions import ContextWindowExceededError` | 0.543s |
| `from litellm.types.utils import ModelResponseStream` | 0.556s |
| `from litellm.main import completion` | 0.545s |

So the only option is deferring the entire import.

### `get_litellm()` helper (`dspy/clients/_litellm.py`)

A module-level sentinel (`_litellm = None`) guards a function that on first call:
1. Sets `LITELLM_LOCAL_MODEL_COST_MAP=True` if not already set (use bundled pricing, skip HTTP fetch)
2. Imports litellm
3. Applies DSPy defaults (`telemetry=False`, `cache=None`, `suppress_debug_info=True`, logging silenced)

The sentinel ensures configuration runs exactly once, so user overrides like `dspy.enable_litellm_logging()` are not clobbered by subsequent `get_litellm()` calls from other code paths.

All litellm configuration is centralized here. Call sites use `litellm = get_litellm()` inside functions. Files inside `dspy/clients/` import `get_litellm` at the top level; files outside the package (adapters, streaming) use inline imports to avoid a circular import chain through `dspy/clients/__init__.py`.

The other option would be to add it at `dspy/_litellm.py`, but we chose the tradeoff of inline imports over adding another top-level file to dspy.

A regression test (`tests/clients/test_litellm_lazy.py`) spawns a subprocess to verify `import dspy` never loads litellm.

### `TYPE_CHECKING` guards

Type-annotation-only imports (`ModelResponseStream`) use `TYPE_CHECKING` guards so they never execute at runtime.

## Bug fix: `LITELLM_LOCAL_MODEL_COST_MAP` ordering

This PR also fixes an import-ordering bug on `main`. In `dspy/clients/__init__.py`, `import litellm` was at the **top** of the file, but `os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"` was set **further down**. This meant litellm never saw the env var during its own init, so it made an HTTP request to GitHub on every `import dspy` to fetch model pricing data ([`get_model_cost_map.py`](https://github.com/BerriAI/litellm/blob/main/litellm/litellm_core_utils/get_model_cost_map.py)) — with a 5s timeout and fallback to a local backup JSON. That round-trip adds ~110ms.

With `get_litellm()`, the env var is set immediately before `import litellm`, so it always reads the bundled local backup.

## Benchmark

Median of 15 runs, Python 3.14, macOS:

```
                        main      this PR     change
import dspy             0.68s     0.26s       -62%
LM() after import       0.00s     0.30s       (deferred cost)
import dspy + LM()      0.67s     0.55s       -19%
```

The combined time is ~140ms faster (not just shifted) because the env var fix eliminates the network round-trip from litellm's init.

<details>
<summary>Reproduction script</summary>

```python
import subprocess, statistics, os

PYTHON = os.path.join(".venv", "bin", "python")

snippets = {
    "import dspy": (
        "import time; t0=time.perf_counter(); import dspy;"
        " print(f'{ time.perf_counter()-t0:.4f}')"
    ),
    "LM() after import": (
        "import time, os; os.environ.setdefault('OPENAI_API_KEY','sk-placeholder');"
        " import dspy; t0=time.perf_counter(); dspy.LM('openai/gpt-4o-mini');"
        " print(f'{ time.perf_counter()-t0:.4f}')"
    ),
    "import dspy + LM()": (
        "import time, os; os.environ.setdefault('OPENAI_API_KEY','sk-placeholder');"
        " t0=time.perf_counter(); import dspy; dspy.LM('openai/gpt-4o-mini');"
        " print(f'{ time.perf_counter()-t0:.4f}')"
    ),
}

def measure(snippet, n=15):
    times = []
    for _ in range(n):
        r = subprocess.run([PYTHON, "-c", snippet], capture_output=True, text=True)
        times.append(float(r.stdout.strip()))
    return statistics.median(times)

for label, snippet in snippets.items():
    print(f"{label:<24} {measure(snippet):.3f}s")
```

</details>

### DIFF
--- a/dspy/clients/_litellm.py
+++ b/dspy/clients/_litellm.py
@@ -1,16 +1,20 @@
-import functools
 import logging
 import os
 
+_litellm = None
 
-@functools.lru_cache(maxsize=1)
+
 def get_litellm():
     """Import and configure litellm.
 
-    Configuration (telemetry, caching, logging) runs once via lru_cache so
-    that user overrides like ``enable_litellm_logging()`` are not clobbered
-    by subsequent calls.
+    Configuration (telemetry, caching, logging) runs once so that user
+    overrides like ``enable_litellm_logging()`` are not clobbered by
+    subsequent calls.
     """
+    global _litellm
+    if _litellm is not None:
+        return _litellm
+
     # Use litellm's bundled model cost map instead of fetching from GitHub
     # on every import (~110ms saved). Must be set before importing litellm.
     if "LITELLM_LOCAL_MODEL_COST_MAP" not in os.environ:
@@ -27,4 +31,5 @@ def get_litellm():
     for h in verbose_logger.handlers:
         h.setLevel(logging.ERROR)
 
-    return litellm
+    _litellm = litellm
+    return _litellm


isaacbmiller:
closing in favor of #9388

Issue #9347, opened 2026-02-27
[Bug] Inefficient data structure used for history


nkaretnikov:
### What happened?

From https://github.com/stanfordnlp/dspy/blob/ab7f8aee86d129c7c336dcdf88e21817631e1547/dspy/clients/base_lm.py

Both global history and local history use `pop(0)` on a list, which is an inefficient operation on Python lists. Consider using a deque if you want to add/remove from both sides.

```py
GLOBAL_HISTORY = []

...
        self.history = []

...

        if len(GLOBAL_HISTORY) >= MAX_HISTORY_SIZE:
            GLOBAL_HISTORY.pop(0)

...

        if len(self.history) >= settings.max_history_size:
            self.history.pop(0)
```

### Steps to reproduce

See the code.

### DSPy version

main

nkaretnikov:
Somewhat related, but probably not worth its own issue: entries can be arbitrary-large strings, right? So it doesn't make sense to use the length of the list as the limit. I would keep track of the total size of bytes inside.

sukhitashvili:
I would also add that IMO it should be disabled by default unless needed; additionally, tracking it in a single place within the self.history list would be better.

isaacbmiller:
The performance gain from popping an array vs dequeing is not worth losing the fact that history is a list for DX purposes.

We lose the ability to slice properly by going with a deque.

isaacbmiller:
closing this issue so people stop generating solutions for it. Feel free to open another issue to discuss shared tracking/disabling by default.

PR #9352, opened 2026-02-28
fix: use collections.deque for O(1) history eviction instead of O(n) list.pop(0)

giulio-leone:
## Problem

`GLOBAL_HISTORY`, `BaseLM.history`, `Module.history`, and `CodeAct.history` all use Python lists with `list.pop(0)` for capacity management. `list.pop(0)` is O(n) because it requires shifting all remaining elements, making each append O(n) when the history is at capacity (default 10,000 entries).

## Solution

Replace all history lists with `collections.deque(maxlen=...)`:

- `deque` with `maxlen` automatically evicts the oldest entry when full — **O(1) amortized**
- Eliminates manual `len(history) >= MAX; history.pop(0)` guard code
- Reduces `update_history()` from 3 conditional blocks to 3 simple `.append()` calls

## Changes

| File | Change |
|------|--------|
| `dspy/clients/base_lm.py` | `GLOBAL_HISTORY = deque(maxlen=MAX_HISTORY_SIZE)`, `self.history = deque(maxlen=...)` |
| `dspy/clients/lm.py` | `self.history = deque(maxlen=...)` |
| `dspy/primitives/module.py` | `self.history = deque(maxlen=...)` in `__init__`, `_base_init`, `__setstate__`, and `ProgramMeta.__call__` |
| `dspy/predict/code_act.py` | `self.history = deque()` |
| `dspy/utils/inspect_history.py` | `list(history)[-n:]` for slice compatibility (deque doesn't support slicing) |
| `tests/clients/test_inspect_global_history.py` | Updated type assertions to `deque` |

## Performance Impact

For a full history of 10,000 entries, each `update_history()` call drops from **O(10,000)** (element shift) to **O(1)** amortized.

Fixes #9347

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,5 +1,6 @@
 import datetime
 import uuid
+from collections import deque
 from typing import Any
 
 from dspy.dsp.utils import settings
@@ -7,7 +8,7 @@ from dspy.utils.callback import with_callbacks
 from dspy.utils.inspect_history import pretty_print_history
 
 MAX_HISTORY_SIZE = 10_000
-GLOBAL_HISTORY = []
+GLOBAL_HISTORY = deque(maxlen=MAX_HISTORY_SIZE)
 
 
 class BaseLM:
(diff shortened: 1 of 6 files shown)


giulio-leone:
Closing in favor of #9359 which has a more complete implementation.

PR #9353, opened 2026-02-28
fix: propagate exceptions from async generator in apply_sync_streaming

giulio-leone:
## Problem

`apply_sync_streaming` silently swallows exceptions raised by the async generator. When an exception occurs during iteration, the `finally` block sends the stop sentinel to the queue, but the exception itself is never communicated to the consumer thread. The caller receives a truncated stream with no indication that an error occurred.

## Root Cause

The producer's `runner()` function wraps the async iteration in `try/finally`. When the generator raises, execution goes straight to `finally` which puts `stop_sentinel` on the queue. The consumer sees the sentinel and breaks out of the loop — the exception is silently discarded.

## Fix

- Add an `error_sentinel` alongside the existing `stop_sentinel`
- Catch `BaseException` in the producer and enqueue a `(error_sentinel, exc)` tuple
- On the consumer side, detect error tuples and re-raise the original exception
- The `finally` block still sends `stop_sentinel` to guarantee clean shutdown

This preserves all items yielded before the exception and re-raises the original exception with its full traceback on the consumer thread.

## Tests

- `test_apply_sync_streaming_propagates_exceptions`: verifies mid-stream `ValueError` propagation with correct partial results
- `test_apply_sync_streaming_propagates_runtime_error`: verifies immediate `RuntimeError` propagation

Fixes #9142

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -228,6 +228,7 @@ def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
     """Convert the async streaming generator to a sync generator."""
     queue = Queue()  # Queue to hold items from the async generator
     stop_sentinel = object()  # Sentinel to signal the generator is complete
+    error_sentinel = object()  # Sentinel to signal an exception occurred
 
     # To propagate prediction request ID context to the child thread
     context = contextvars.copy_context()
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #9357 which has a more complete implementation.

PR #9355, opened 2026-02-28
perf: use deque instead of list for LM history to avoid O(n) pop(0)

giulio-leone:
## Summary

Fixes #9347

Replaces `list`-based history storage with `collections.deque` for both `GLOBAL_HISTORY` and per-LM/per-module history. This eliminates the O(n) cost of `list.pop(0)` when the history reaches its size limit.

## Problem

`GLOBAL_HISTORY`, `self.history`, and `module.history` all use `list.pop(0)` to evict the oldest entry when the history reaches its maximum size. On a Python `list`, `pop(0)` is O(n) because it shifts all remaining elements. With `MAX_HISTORY_SIZE = 10_000`, this means up to 10,000 element shifts on every LM call once the history is full.

## Solution

- **`GLOBAL_HISTORY`**: Changed to `deque(maxlen=MAX_HISTORY_SIZE)`. The `maxlen` parameter automatically evicts the oldest entry in O(1) when the limit is reached, eliminating the manual size check entirely.
- **`self.history`** (per-LM): Changed to `deque()`. Uses `popleft()` (O(1)) instead of `pop(0)` (O(n)).
- **`module.history`** (per-module): Uses `popleft()` when available, with a fallback to `pop(0)` for backward compatibility with any external code that initializes history as a plain list.
- **`pretty_print_history`**: Updated to materialize the deque to a list before slicing, since `deque` does not support slice notation.

## Testing

- All 4 global history tests pass (updated `isinstance` checks from `list` to `deque`)
- All module history and base module tests pass
- 2 consecutive clean test runs with 0 errors

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,5 +1,6 @@
 import datetime
 import uuid
+from collections import deque
 from typing import Any
 
 from dspy.dsp.utils import settings
@@ -7,7 +8,7 @@ from dspy.utils.callback import with_callbacks
 from dspy.utils.inspect_history import pretty_print_history
 
 MAX_HISTORY_SIZE = 10_000
-GLOBAL_HISTORY = []
+GLOBAL_HISTORY: deque = deque(maxlen=MAX_HISTORY_SIZE)
 
 
 class BaseLM:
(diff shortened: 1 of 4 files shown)


giulio-leone:
Closing in favor of #9359 which is the same fix.

PR #9357, opened 2026-02-28
fix: propagate exceptions from apply_sync_streaming instead of swallowing

giulio-leone:
## Summary

Fixes #9142

When using `streamify` with `async_streaming=False`, exceptions raised in the async generator are silently swallowed. The producer thread's `runner()` catches exceptions only in a `finally` block, sending the stop sentinel without propagating the error to the consumer.

### Root Cause

```python
async def runner():
    try:
        async for item in async_generator:
            queue.put(item)
    finally:
        queue.put(stop_sentinel)  # ← exception lost here
```

When the async generator raises, the `finally` block sends the stop sentinel and the consumer exits normally — the exception is lost.

### Fix

Add an explicit `except BaseException` clause that puts an error sentinel followed by the exception into the queue before the stop sentinel. The consumer checks for the error sentinel and re-raises the exception.

## Tests

Added `test_sync_streaming_propagates_exceptions` that verifies exceptions are properly propagated through the queue mechanism.

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -228,6 +228,7 @@ def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
     """Convert the async streaming generator to a sync generator."""
     queue = Queue()  # Queue to hold items from the async generator
     stop_sentinel = object()  # Sentinel to signal the generator is complete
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Superseded by #9366 which has the same fix with cleaner implementation.

PR #9359, opened 2026-02-28
fix: replace list.pop(0) with deque.popleft() for O(1) history eviction

giulio-leone:
## Problem

`GLOBAL_HISTORY`, `BaseLM.history`, `LM.history`, and `Module.history` all use Python lists with `list.pop(0)` for eviction when the history buffer exceeds its size limit. `list.pop(0)` is an **O(n)** operation because it requires shifting all remaining elements left by one position. With `MAX_HISTORY_SIZE = 10_000`, this means up to 10,000 element shifts on every LM call once the buffer is full.

## Solution

Replace `list` with `collections.deque`, which provides **O(1)** `popleft()` operations via a doubly-linked list implementation.

### Key design decisions:

1. **`GLOBAL_HISTORY`**: Uses `deque(maxlen=MAX_HISTORY_SIZE)` — the maxlen is fixed at 10,000 and doesn't change at runtime, so deque's automatic eviction is ideal.

2. **`LM.history` and `Module.history`**: Use `deque()` **without** `maxlen`, because `settings.max_history_size` can change dynamically via `dspy.context(max_history_size=N)`. Instead, explicit `popleft()` calls in `update_history()` enforce the dynamic limit — still O(1) but respecting runtime configuration.

3. **`pretty_print_history()`**: Converts to `list()` for the final slice since `deque` doesn't support slicing. This is a display-only path called infrequently, so the conversion has negligible impact.

## Performance Impact

| Operation | Before (list) | After (deque) |
|-----------|--------------|---------------|
| Eviction (`pop(0)` vs `popleft()`) | O(n) | **O(1)** |
| Append | O(1) amortized | O(1) |
| Index access (`history[i]`) | O(1) | O(1) |
| Iteration | O(n) | O(n) |

For a full 10,000-entry buffer, each eviction goes from ~10,000 element shifts to a single pointer update.

## Changes

- `dspy/clients/base_lm.py`: `GLOBAL_HISTORY` → `deque(maxlen=...)`, `self.history` → `deque()`, `pop(0)` → `popleft()`
- `dspy/clients/lm.py`: `self.history` → `deque()`
- `dspy/primitives/module.py`: All `self.history` → `deque()`
- `dspy/utils/inspect_history.py`: `list(history)[-n:]` for slice compatibility
- `tests/clients/test_inspect_global_history.py`: Accept `deque` type

Fixes #9347

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,5 +1,6 @@
 import datetime
 import uuid
+from collections import deque
 from typing import Any
 
 from dspy.dsp.utils import settings
@@ -7,7 +8,7 @@ from dspy.utils.callback import with_callbacks
 from dspy.utils.inspect_history import pretty_print_history
 
 MAX_HISTORY_SIZE = 10_000
-GLOBAL_HISTORY = []
+GLOBAL_HISTORY: deque[dict[str, Any]] = deque(maxlen=MAX_HISTORY_SIZE)
 
 
 class BaseLM:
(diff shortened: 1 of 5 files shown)


giulio-leone:

giulio-leone:
Superseded by #9365 which includes this fix plus additional improvements (all Module init sites updated).

PR #9360, opened 2026-02-28
fix: share usage_tracker across parallel workers for correct aggregation

### LINKED ISSUE #9201: [Bug] dspy.Parallel inside dspy.Module not working when track_usage is enabled
### What happened?

When I enable `track_usage` and use `dspy.Parallel` for parallel predictions inside the `dspy.Module` `forward` method, only a list of `None` is returned without any error.

It works when I call this code without `dspy.Module` or use the `forward` method of my module directly. Therefore, I assume, there is a bug in the `dspy.Module.__call__` method.

Thank you

### Steps to reproduce

```Python
import dspy
import os

lm = dspy.LM(
    "azure/gpt-4o",
    api_key=os.environ["LM_API_KEY"],
    api_base=os.environ["LM_API_ENDPOINT"],
    api_version="2024-12-01-preview",
    cache=False,
    temperature=1,
)


def code():
    pred = dspy.Predict("q -> a")
    r = dspy.Parallel(num_threads=1, max_errors=0)(
        [
            (pred, {"q": "President of Czechia"}),
            (pred, {"q": "Capital of Vietnam"}),
        ],
    )
    print(r)
    for a in r:
        print(a.get_lm_usage())
    return r


class Mod(dspy.Module):
    def forward(self):
        return code()


print("Without module")
with dspy.settings.context(lm=lm, track_usage=True):
    code()


print("\nWith module, using __call__")
with dspy.settings.context(lm=lm, track_usage=True):
    r = Mod()()
```

Output:
```
Without module
  0%|          | 0/2 [00:00<?, ?it/s]
Processed 2 [/](https://file+.vscode-resource.vscode-cdn.net/) 2 examples: 100%|██████████| 2/2 [00:01<00:00,  1.77it/s]
[Prediction(
    a='The current President of Czechia is Petr Pavel. He assumed office on March 9, 2023.'
), Prediction(
    a='Hanoi'
)]
{'azure/gpt-4o': {'completion_tokens': 33, 'prompt_tokens': 141, 'total_tokens': 174, 'completion_tokens_details': {'accepted_prediction_tokens': 0, 'audio_tokens': 0, 'reasoning_tokens': 0, 'rejected_prediction_tokens': 0, 'text_tokens': None, 'image_tokens': None}, 'prompt_tokens_details': {'audio_tokens': 0, 'cached_tokens': 0, 'text_tokens': None, 'image_tokens': None}}}
{'azure/gpt-4o': {'completion_tokens': 14, 'prompt_tokens': 140, 'total_tokens': 154, 'completion_tokens_details': {'accepted_prediction_tokens': 0, 'audio_tokens': 0, 'reasoning_tokens': 0, 'rejected_prediction_tokens': 0, 'text_tokens': None, 'image_tokens': None}, 'prompt_tokens_details': {'audio_tokens': 0, 'cached_tokens': 0, 'text_tokens': None, 'image_tokens': None}}}

With module, using __call__
  0%|          | 0/2 [00:00<?, ?it/s]
[None, None]

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
...
AttributeError: 'NoneType' object has no attribute 'get_lm_usage'
```

### DSPy version

3.1.0

giulio-leone:
## Problem

When `dspy.Parallel` runs inside a `dspy.Module` with `track_usage=True`, calling the module via `__call__` returns broken results and usage data is lost. This is reported in #9201.

### Root Cause

`ParallelExecutor` deep-copies the `usage_tracker` for each worker thread (line 94-95 in `parallelizer.py`):

```python
if new_overrides.get("usage_tracker"):
    new_overrides["usage_tracker"] = copy.deepcopy(new_overrides["usage_tracker"])
```

This creates independent tracker instances per thread. When workers add usage data, it goes to the thread-local copy and never reaches the parent tracker. After `Parallel` completes, the parent's `usage_tracker.get_total_tokens()` returns empty data.

Additionally, `Module._set_lm_usage()` didn't handle `list` outputs (which is what `Parallel` returns), so even if usage was available, it would log a warning and fail to attach it.

### 1. Thread-safe UsageTracker
Added `threading.Lock` to `UsageTracker.add_usage()` so the same tracker instance can safely be shared across worker threads:

```python
def add_usage(self, lm, usage_entry):
    if len(usage_entry) > 0:
        flattened = self._flatten_usage_entry(usage_entry)
        with self._lock:
            self.usage_data[lm].append(flattened)
```

### 2. Share tracker instead of deep-copying
Removed the `copy.deepcopy(usage_tracker)` in `ParallelExecutor` worker threads. All workers now share the parent's tracker, correctly aggregating usage data.

### 3. Handle list outputs in _set_lm_usage
When `Module.forward()` returns a list of Predictions (from `Parallel`), `_set_lm_usage` now iterates and attaches usage to each Prediction.

## Before vs After

| Scenario | Before | After |
|----------|--------|-------|
| `Mod()()` with track_usage | `[None, None]` or empty usage | Correct Predictions with usage |
| Parent tracker after Parallel | Empty | Aggregated from all workers |
| `_set_lm_usage` with list output | Warning, no usage attached | Usage attached to each Prediction |

Fixes #9201

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -326,6 +326,12 @@ class Module(BaseModule, metaclass=ProgramMeta):
             prediction_in_output = output
         elif isinstance(output, tuple) and len(output) > 0 and isinstance(output[0], Prediction):
             prediction_in_output = output[0]
+        elif isinstance(output, list) and len(output) > 0 and isinstance(output[0], Prediction):
+            # Parallel returns a list of Predictions — attach usage to each
+            for pred in output:
+                if isinstance(pred, Prediction):
+                    pred.set_lm_usage(tokens)
+            return
         if prediction_in_output:
             prediction_in_output.set_lm_usage(tokens)
         else:
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -1,5 +1,4 @@
 import contextlib
-import copy
 import logging
 import signal
 import sys
@@ -90,9 +89,9 @@ class ParallelExecutor:
 
             original = thread_local_overrides.get()
             new_overrides = {**original, **parent_overrides.copy()}
-            if new_overrides.get("usage_tracker"):
-                # Usage tracker needs to be deep copied across threads so that each thread tracks its own usage
-                new_overrides["usage_tracker"] = copy.deepcopy(new_overrides["usage_tracker"])
+            # Share the parent's usage_tracker across threads instead of deep-copying.
+            # UsageTracker.add_usage() is thread-safe, so parallel workers can safely
+            # aggregate usage data into the same tracker instance.
             token = thread_local_overrides.set(new_overrides)
 
             try:
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -1,5 +1,6 @@
 """Usage tracking utilities for DSPy."""
 
+import threading
 from collections import defaultdict
 from contextlib import contextmanager
 from typing import Any, Generator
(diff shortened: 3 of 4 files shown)


giulio-leone:

PR #9361, opened 2026-02-28
fix: support TRL >= 0.16 by using max_length instead of deprecated max_seq_length

### LINKED ISSUE #8762: [Bug] TypeError: SFTConfig.__init__() got an unexpected keyword argument 'max_seq_length'

giulio-leone:
## Summary

Fixes #8762

TRL 0.16+ renamed `SFTConfig.max_seq_length` to `max_length`, causing a `TypeError` when using `BootstrapFinetune` with modern TRL versions:

```
TypeError: SFTConfig.__init__() got an unexpected keyword argument 'max_seq_length'
```

## Root Cause

`train_sft_locally()` hard-codes `max_seq_length=...` in the `SFTConfig()` constructor, which fails with TRL >= 0.16 where the parameter was renamed to `max_length`.

## Fix

- **Version detection**: Uses `inspect.signature(SFTConfig)` to detect which parameter name the installed TRL version accepts (`max_length` vs `max_seq_length`)
- **Backward compatibility**: Normalizes `max_seq_length` → `max_length` in `train_kwargs` so users can pass either name
- **Decoupled tokenization**: Extracts the sequence length into a local variable for the tokenization step, independent of TRL's naming

## Compatibility

| TRL Version | Parameter | Status |
|---|---|---|
| <= 0.15.x | `max_seq_length` | ✅ Works |
| >= 0.16.0 | `max_length` | ✅ Works |

### DIFF
--- a/dspy/clients/lm_local.py
+++ b/dspy/clients/lm_local.py
@@ -175,6 +175,14 @@ class LocalProvider(Provider):
             "output_dir": output_dir,
         }
         train_kwargs = {**default_train_kwargs, **(train_kwargs or {})}
+
+        # Normalize max_seq_length / max_length for TRL >= 0.16 compatibility.
+        # TRL 0.16+ renamed max_seq_length to max_length in SFTConfig.
+        if "max_length" in train_kwargs and "max_seq_length" not in (train_kwargs or {}):
+            pass  # User explicitly set max_length — use it as-is
+        elif "max_seq_length" in train_kwargs:
+            train_kwargs.setdefault("max_length", train_kwargs.pop("max_seq_length"))
+
         output_dir = train_kwargs["output_dir"]  # user might have changed the output_dir
 
         logger.info(f"Starting local training, will save to {output_dir}")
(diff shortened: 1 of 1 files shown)


giulio-leone:

PR #9362, opened 2026-02-28
fix: propagate exceptions from async generator in apply_sync_streaming

giulio-leone:
## Problem

`apply_sync_streaming()` converts an async generator to a sync generator using a background daemon thread. When the async generator raises an exception, the daemon thread silently swallows it — Python does not propagate exceptions from daemon threads to the main thread. The consumer loop receives the `stop_sentinel` and exits normally, with no indication that an error occurred.

## Root Cause

The async `runner()` function has a `try/finally` that always sends `stop_sentinel`, but exceptions raised during async iteration are not captured or forwarded. Since `producer()` runs in a daemon thread, unhandled exceptions are silently lost.

## Fix

- Added an `error_sentinel` alongside the existing `stop_sentinel`
- Exceptions caught in `runner()` are wrapped as `(error_sentinel, exception)` tuples and put on the queue before `stop_sentinel`
- Consumer loop checks for error tuples using identity comparison (`item[0] is error_sentinel`) and re-raises the original exception with its traceback preserved
- The `error_sentinel` is a closure-local `object()`, so user-yielded tuples cannot collide with the error pattern

## Testing

Added `test_apply_sync_streaming_propagates_exceptions` that verifies:
- Items yielded before the exception are delivered normally
- The exception is re-raised in the consumer thread
- The error message matches the original exception

Fixes #9142

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -228,6 +228,7 @@ def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
     """Convert the async streaming generator to a sync generator."""
     queue = Queue()  # Queue to hold items from the async generator
     stop_sentinel = object()  # Sentinel to signal the generator is complete
+    error_sentinel = object()  # Sentinel to signal an exception occurred
 
     # To propagate prediction request ID context to the child thread
     context = contextvars.copy_context()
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Superseded by #9366 which has the same fix with cleaner implementation.

Issue #8799, opened 2025-09-11
[Bug] Bus error on import when cache on network drive (occurs rarely on some machines only)


hrabalm:
### What happened?

This is a rare bug than occurs on some machines on our cluster if the ~/.dspy_cachedir is located on lustre (network filesystem). For some reason, this crash happens only on certain machines.

The crash occurs in https://github.com/stanfordnlp/dspy/blob/e29d054755899aacd09544013df71106a97bf584/dspy/clients/__init__.py when the cache is created during import (it boils down to the creation of diskcache.FanoutCache on the network filesystem, I believe the Bus error itself occurs somewhere in CPython or sqlite C code).

The problem is that the Python interpreter crashes as a whole with no exception raised, which makes pin pointing the source of the crash harder. I had to track down the line where the crash occurs with `gdb`. Since the crash occurs on dspy import, you can't even disable the cache with `dspy.configure_cache`

On one hand, storing the cache on network filesystem can be a bad idea, because of issues with file locking, on the other hand, it would be nice to avoid hard crashes on library import.

Workaround is to explicitly set DSPY_CACHEDIR to a local directory, for example:
```bash
export DSPY_CACHEDIR="$(mktemp -d)"
```

```bash
export DSPY_CACHEDIR="$TMPDIR/dspy_cache"  # if TMPDIR is set, for example by SLURM
```
(Note that in such a case the user should remember to clean up the cache afterwards.)

The machine where this issue occurs has "Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz". I suspect it might be linked to some CPU models as I currently don't get these issues on other machines.

I am not sure how to fix this without deferring cache creation to the point where it is actually needed (and configured), but even then, it would just defer the time of crash, since the disk cache is currently used by default.

At the very least I think it's useful to document this as a GitHub Issue so that the workaround hopefully shows if people google something like "dspy bus error".

#### Prerequisites
- a machine with Intel(R) Xeon(R) CPU E5-2620 v4
- ~/.dspy_cache or DSPY_CACHEDIR is a directory located on lustre filesystem

#### Steps

```bash
python3 -c 'import dspy; print("Hello world!")'
```

Or by directly creating the diskcache.FanoutCache"
```bash
python3 -c 'from diskcache import FanoutCache; cache = FanoutCache("test_cache")'
```

Both result in Python interpreter crash and "Bus error".

### DSPy version

3.0.3, also present in previous versions

[...]

PR #9376, opened 2026-03-03
feat(deps): make diskcache an optional dependency

gittb:
[CVE-2025-69872](https://www.sentinelone.com/vulnerability-database/cve-2025-69872/) (CVSS 9.8) is a remote code execution vulnerability in diskcache via pickle deserialization, affecting all versions through 5.6.3. No patched version exists. Because DSPy declares diskcache as a required dependency, every DSPy installation inherits this critical finding, which blocks deployment in environments with mandatory security scanning.

This PR moves diskcache from required to optional `pip install dspy[diskcache]`. Users who have diskcache installed see no change. Users without it get memory-only caching automatically - a path DSPy already supported via `_get_dspy_cache()`'s existing try/except fallback.

**Changes** - 5 files:

**pyproject.toml** - moved diskcache to optional extra, added to dev extras for CI
**dspy/clients/cache.py** - conditional import with DISKCACHE_AVAILABLE flag (follows existing `SF_AVAILABLE` pattern in `adapters/types/audio.py`), guard in `Cache.__init__`
**dspy/clients/__init__.py** - `configure_cache()` and `_get_dspy_cache()` auto-detect diskcache availability; explicit enable_disk_cache=True without the package raises a clear `ImportError`
**tests/clients/test_cache.py** - conditional import, 2 new tests for the no-diskcache path. All tests pass!

**Precedent:** [FastMCP had the same issue](https://github.com/PrefectHQ/fastmcp/issues/3239) for this CVE and took the same approach. Google ADK avoids the problem entirely by using optional dependency groups for non-core features `pip install google-adk[extensions]`

Related: #8717 (disk caching unconditionally enabled at import time), #8799 (bus error on import from disk cache on network FS)

### DIFF
--- a/dspy/clients/__init__.py
+++ b/dspy/clients/__init__.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import litellm
 
 from dspy.clients.base_lm import BaseLM, inspect_history
-from dspy.clients.cache import Cache
+from dspy.clients.cache import DISKCACHE_AVAILABLE, Cache
 from dspy.clients.embedding import Embedder
 from dspy.clients.lm import LM
 from dspy.clients.provider import Provider, TrainingJob
@@ -17,7 +17,7 @@ DISK_CACHE_LIMIT = int(os.environ.get("DSPY_CACHE_LIMIT", 3e10))  # 30 GB defaul
 
 
 def configure_cache(
-    enable_disk_cache: bool | None = True,
+    enable_disk_cache: bool | None = None,
     enable_memory_cache: bool | None = True,
     disk_cache_dir: str | None = DISK_CACHE_DIR,
     disk_size_limit_bytes: int | None = DISK_CACHE_LIMIT,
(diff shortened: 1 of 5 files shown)


isaacbmiller:
It is much harder for us to make diskcache optional, as we rely on it as our primary caching mechanism.

Notably the actual CVSS score is 5.2, not 9.8. I think 9.8 is a hallucination from SentinelOne. There is no remote aspect to this, it requires local filesystem access to achieve.

The relevant dependabot alert is https://github.com/stanfordnlp/dspy/security/dependabot/181.

A solution to this should maintain the persistent nature of disk caching, as defaulting to memory is not ideal.

The recommended solution from [cve disclosure](https://github.com/EthanKim88/ethan-cve-disclosures/blob/main/CVE-2025-69872-DiskCache-Pickle-Deserialization.md) is to change to JsonSerializer. We would need to add serializers for ModelResponse and potentially some other objects, but it is doable.

Can you say more about the policies of the security scanner you are using? Is this actively blocking deployments?

gittb:
> Can you say more about the policies of the security scanner you are using? Is this actively blocking deployments?

Using [Trivy](https://github.com/aquasecurity/trivy) and the logs say that it's being found by [Grype](https://github.com/anchore/grype). Currently it is blocking deployments.

> Notably the actual CVSS score is 5.2, not 9.8. I think 9.8 is a hallucination from SentinelOne.

I think we could both argue that it could be lower, since you need write access for the attack, not net, but CISA-ADP has classified it as 9.8, which is what security tools use as a source. https://www.cve.org/CVERecord?id=CVE-2025-69872

I don't think SentinelOne is hallucinating - we just have conflicting authoritative sources here.

Even if we switch to JsonDisk, which is technically safe, DiskCache hasn't been updated and likely won't be, given it's dormant for 2 years at this point. Switching to a safer implementation of that package doesn't get around the fact that we depend on a package with this vulnerability that likely won't get updated. Keeping it would add friction for teams using in behind scanners, even if it's technically not using the exploitable methods.

Really what we need to do is pivot away from diskcache entirely, or make it optional. Switching to JsonDisk is a good hardening step regardless, but it doesn't resolve the scanner findings since the CVE is against the package itself, not a specific code path.

gittb:
@isaacbmiller Following up here.

Want to do what's best and unblock deployments.

isaacbmiller:
Thanks for the clarifications, very helpful to understand.

Early this week I will be working on a Diskcache replacement for DSPy, and make Diskcache optional. 

Once that is tested, we will do the release for 3.2, which will unblock your deployments. I will close this PR, and you can track via #9376

PR #9377, opened 2026-03-03
fix(predict): create thread-local interpreter in forward() to fix concurrent Evaluate

### LINKED ISSUE #9082: [Bug] ProgramOfThought and CodeAct fail using Evaluate with num_threads > 1
### What happened?

Running a simple evaluation of the ProgramOfThought module using dspy.Evaluate with num_threads > 1 results in "I/O operation on closed file" errors when it should run to completion. Currently, ProgramOfThought and CodeAct are not usable for me due to this concurrency issue.

### Steps to reproduce

The following is a simple evaluation of ProgramOfThought on AIME-2025 which results in the error. Replacing ProgramOfThought with CodeAct results in the same issue as well.

```python
import dspy
from datasets import load_dataset

def metric(example, prediction, trace=None, pred_name=None, pred_trace=None):
    correct_answer = int(example['answer'])
    try:
        llm_answer = int(prediction.answer.replace('[','').replace(']','').strip())
    except ValueError as e:
        return 0
    return int(correct_answer == llm_answer)

def load_aime25():
    dataset = load_dataset("yentinglin/aime_2025", "default", split="train")
    return [
        dspy.Example(question=sample["problem"], answer=str(sample["answer"]))
        .with_inputs("question")
        for sample in dataset
    ]

devset = load_aime25()
lm = dspy.LM(
    "openai/gpt-4.1-mini",
    api_key=API_KEY,
    temperature=1.0,
    max_tokens=32000,
)
dspy.configure(lm=lm)

evaluator = dspy.Evaluate(devset=devset, metric=metric, num_threads=10, display_progress=True, display_table=True, max_errors=50, provide_traceback=True)
signature = "question -> answer"
agent = dspy.ProgramOfThought(signature, max_iters=10)
evaluation = evaluator(agent)
```

The problem is that both ProgramOfThought and CodeAct initialize a single interpreter in their constructor and then shut down the interpreter at the end of the forward function. When Evaluate with num_threads > 1 is used, I believe this shares the interpreter across threads, leading to issues when one thread shuts down the shared interpreter.

One way to fix this is that there can still be a single interpreter object created in the constructor, but it should be copied  before use in the forward function of CodeAct and ProgramOfThought.

### DSPy version

3.0.4 and HEAD of main

CloveSVG:
Fixes #9082

## Problem

`ProgramOfThought` and `CodeAct` crash with `"I/O operation on closed file"` when used with `dspy.Evaluate(num_threads > 1)`.

## Root Cause

Both modules stored a single interpreter instance as `self.interpreter` (created in `__init__`). When `dspy.Evaluate` runs with multiple threads, all threads share the same module instance. One thread closes the interpreter at the end of `forward()` while another thread is still using it.

## Fix

Move interpreter creation into `forward()` as a local variable, eliminating all shared mutable state between threads. The `try/finally` cleanup pattern is preserved to ensure proper resource release.

```python
# Before (broken with num_threads > 1)
def __init__(self):
    self.interpreter = Interpreter()

def forward(self, ...):
    try:
        result = self.interpreter.execute(code)
    finally:
        self.interpreter.close()

# After (thread-safe)
def forward(self, ...):
    interpreter = Interpreter()
    try:
        result = interpreter.execute(code)
    finally:
        interpreter.close()
```

## Testing

Added concurrent evaluation test that runs `ProgramOfThought` with `dspy.Evaluate(num_threads=4)` to verify no race conditions.

### DIFF
--- a/dspy/predict/code_act.py
+++ b/dspy/predict/code_act.py
@@ -65,8 +65,7 @@ class CodeAct(ReAct, ProgramOfThought):
         self.tools: dict[str, Tool] = tools
         self.codeact = dspy.Predict(codeact_signature)
         self.extractor = dspy.ChainOfThought(extract_signature)
-        # It will raises exception when dspy cannot find available deno instance by now.
-        self.interpreter = interpreter or PythonInterpreter()
+        self._user_interpreter = interpreter
 
     def _build_instructions(self, signature, tools):
         instructions = [f"{signature.instructions}\n"] if signature.instructions else []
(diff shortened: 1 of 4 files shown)


ghost:
Hi\! Just checking in on this PR — let me know if there's anything to adjust. Thanks\!

PR #9382, opened 2026-03-04
fix(adapters): empty LM text response silently returns None instead of raising parse error

### LINKED ISSUE #9378: [Bug] Pydantic typing skipped on empty return value and returns None
### What happened?

When an LM models returns an empty string as a response, dspy will return with no error but output will not be of correct typing it will always be None.

For example if I have this signature
```
class Translate(dspy.Signature):
    """Translate a word from English to Spanish."""
    english: str = dspy.InputField()
    spanish: str = dspy.OutputField()
```

On return I expect the typing of the object to match the schema I have provided so `spanish` field should be a str or error out to indicate typing issues. This works most of the time except if the LLM returns an empty string, this leads to a base case where the values all become null. Breaking my expectation of typed responses.

### Steps to reproduce

```Python
import dspy


class Translate(dspy.Signature):
    """Translate a word from English to Spanish."""
    english: str = dspy.InputField()
    spanish: str = dspy.OutputField()

class FakeLM(dspy.BaseLM):
    def __init__(self, responses):
        self.model = "fake-model"
        self.responses = responses
        self.history = []
        self._call_count = 0
        self.kwargs ={}

    def __call__(self, prompt=None, messages=None, **kwargs):
        resp = self.responses[self._call_count % len(self.responses)]
        self._call_count += 1
        return [resp]

# --- Working case: text is non-empty ---
good_lm = FakeLM([{"text": '[[ ## spanish ## ]]\n"hola"'}])
with dspy.context(lm=good_lm, adapter=ChatAdapter()):
    result = dspy.Predict(Translate)(english="hello")
print("good case:", result.spanish)   # → "hola"


# --- Bug case: text is empty, answer is in reasoning_content ---
buggy_lm = FakeLM([{"text": "", "reasoning_content": 'translating to spanish, translate as "hola"'}])
with dspy.context(lm=buggy_lm, adapter=ChatAdapter()):
    result = dspy.Predict(Translate)(english="hello")
print("bug case:", result.spanish)    # → None  (BUG: should be "hola")

# --- Bug case:  anything but empty will correctly fail validation ---
buggy_lm = FakeLM([{"text": "a", "reasoning_content": 'translating to spanish, translate as "hola"'}])
with dspy.context(lm=buggy_lm, adapter=ChatAdapter()):
    result = dspy.Predict(Translate)(english="hello")
```

### DSPy version

3.1.3

kvr06-ai:
## Summary

Fixes #9378. When an LM returns an empty string as `text`, `_call_postprocess` in `base.py` checks `if text:`, which is `False` for `""`. This skips `self.parse()` entirely and silently sets all output fields to `None`, breaking typed response contracts.

The fix changes `if text:` to `if text is not None:` so empty strings are passed to `parse()`, which then raises `AdapterParseError` (the expected behavior for unparseable responses).

## Changes

- `dspy/adapters/base.py`: Replace truthiness check with explicit `None` check on line 136
- `tests/adapters/test_chat_adapter.py`: Add regression test confirming empty-string responses raise `AdapterParseError`

## Testing

- New test `test_empty_string_response_raises_parse_error_instead_of_returning_none` passes
- All 28 existing `test_chat_adapter.py` tests pass

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -133,7 +133,7 @@ class Adapter:
                 output_logprobs = output.get("logprobs")
                 tool_calls = output.get("tool_calls")
 
-            if text:
+            if text is not None:
                 value = self.parse(processed_signature, text)
                 for field_name in original_signature.output_fields.keys():
                     if field_name not in value:
--- a/tests/adapters/test_chat_adapter.py
+++ b/tests/adapters/test_chat_adapter.py
@@ -743,3 +743,27 @@ All interactions will be structured in the following way, with the appropriate v
 In adhering to this structure, your objective is: 
         Answer the question with multiple answers and scores"""
     assert system_message == expected_system_message
+
+
+def test_empty_string_response_raises_parse_error_instead_of_returning_none():
+    """Regression test for #9378: empty LM text response should raise AdapterParseError,
+    not silently set all output fields to None."""
+
+    class Translate(dspy.Signature):
+        """Translate a word from English to Spanish."""
+
+        english: str = dspy.InputField()
+        spanish: str = dspy.OutputField()
+
+    adapter = dspy.ChatAdapter()
+
+    with mock.patch("litellm.completion") as mock_completion:
+        mock_completion.return_value = ModelResponse(
+            choices=[Choices(message=Message(content=""))],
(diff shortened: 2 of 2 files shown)


isaacbmiller:
Closing as duplicate of #9389

PR #9387, opened 2026-03-04
fix(streaming): remove default tool status messages from StatusMessageProvider for consistency

### LINKED ISSUE #9177: [Bug] Remove default messages in StatusMessageProvider for consistency
### What happened?

Currently, StatusMessageProvider implements default status messages for tool start and tool end. This is inconsistent with other methods in the provider.

The default messages for these tool events should be removed to ensure consistent behavior across the class and avoid unwanted default output.

opened #9176 to fix this

### Steps to reproduce

NA

### DSPy version

3.1.0

adityasingh2400:
## Problem

Fixes #9177.

`StatusMessageProvider` is inconsistent: `tool_start/end_status_message` return hardcoded strings (always emitting messages), while `module_start/end` and `lm_start/end` return `None` (silent by default).

```python
# Before — inconsistent defaults
def tool_start_status_message(self, instance, inputs):
    return f'Calling tool {instance.name}...'   # ← always emitted

def module_start_status_message(self, instance, inputs):
    pass                                          # ← silent by default
```

This meant tool invocations silently produced status messages even when users had not opted in, while all other hooks were silent.

## Fix

Change `tool_start_status_message` and `tool_end_status_message` to `pass` (return `None`), consistent with all other base-class methods.

Users who want tool status messages can subclass `StatusMessageProvider` and override these methods:

```python
class MyProvider(StatusMessageProvider):
    def tool_start_status_message(self, instance, inputs):
        return f'Calling tool {instance.name}...'

    def tool_end_status_message(self, outputs):
        return 'Tool calling finished!'
```

## Tests Updated

- Tests that asserted the default provider emits 2 tool status messages now assert 0 messages.
- Non-blocking timing tests (`test_status_message_non_blocking*`) define an inline `ToolStatusProvider` subclass that replicates the old behavior so the timing logic continues to be exercised.

> **Note:** This is a minor breaking change for users relying on the default tool status messages. If needed, a deprecation warning could be added before removing the defaults in a future release — but since the issue author considers this a bug, we remove the defaults directly.

### DIFF
--- a/dspy/streaming/messages.py
+++ b/dspy/streaming/messages.py
@@ -72,11 +72,11 @@ class StatusMessageProvider:
 
     def tool_start_status_message(self, instance: Any, inputs: dict[str, Any]):
         """Status message before a `dspy.Tool` is called."""
-        return f"Calling tool {instance.name}..."
+        pass
 
     def tool_end_status_message(self, outputs: Any):
         """Status message after a `dspy.Tool` is called."""
-        return "Tool calling finished! Querying the LLM with tool calling results..."
+        pass
 
     def module_start_status_message(self, instance: Any, inputs: dict[str, Any]):
         """Status message before a `dspy.Module` or `dspy.Predict` is called."""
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Thanks for the PR, this is intended behavior.

We want the Tool calling agent to display tool messages in streaming mode.

PR #9388, opened 2026-03-04
fix: set LITELLM_LOCAL_MODEL_COST_MAP before litellm import to avoid HTTP fetch

isaacbmiller:
## Problem

`import dspy` triggers an unnecessary HTTP request to GitHub on every invocation (~100ms), fetching `model_prices_and_context_window.json` from the litellm repository.

The env var `LITELLM_LOCAL_MODEL_COST_MAP` was already being set in `dspy/clients/__init__.py`, but **after** `import litellm`. And litellm was already imported even earlier via the streaming module chain:

```
dspy/__init__ → predict/ → utils/ → streaming/streamify.py → import litellm
```

So litellm never saw the env var during its import and always made the HTTP request.

## Fix

Move `os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")` to the top of `dspy/__init__.py` — the only place guaranteed to run before any module in the package can trigger `import litellm`.

## Measurements

Median of 15 runs, clean subprocess each time, `LITELLM_LOCAL_MODEL_COST_MAP` unset:

| | `import dspy` |
|---|---|
| **Before** (main) | 0.695s |
| **After** (this PR) | 0.600s |
| **Delta** | **-95ms (14%)** |

The ~100ms matches the measured round-trip time of the HTTP GET to `raw.githubusercontent.com` that litellm performs when the env var is not set.

### DIFF
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -1,4 +1,5 @@
-import os; os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")  # noqa: E702
+import os
+os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")
 
 from dspy.predict import *
 from dspy.primitives import *


dspy/__init__.py
```diff
@@ -1,3 +1,5 @@
+import os; os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")  # noqa: E702
```
TomeHirata: Let's not have multiple statements in one line

PR #9395, opened 2026-03-06
perf(history): replace O(n) list.pop(0) with deque for history eviction

adityasingh2400:
Fixes #9347.

## Problem

`GLOBAL_HISTORY`, `self.history`, and per-module histories all evict old entries with `list.pop(0)`, which is O(n) — shifting every remaining element on each LM call. At the default cap of 10,000 entries this is ~10k element moves per request.

## Solution

Replace with `collections.deque`:

- **`GLOBAL_HISTORY`** → `deque(maxlen=MAX_HISTORY_SIZE)`: eviction is automatic and O(1); the manual `len >= MAX_HISTORY_SIZE` check is gone entirely.
- **`self.history` / per-module histories** → `deque()` with explicit `popleft()` (O(1)) instead of `pop(0)` (O(n)).

## Addressing the slicing concern

A previous comment noted that deques lose slice support. This is handled:

- **`pretty_print_history`** (the only internal consumer) already converts to list first: `history_list = list(history)`, so negative-index slicing works.
- Users who need to slice `lm.history` directly can do `list(lm.history)[start:end]` — the same pattern used everywhere internally.
- The deque still supports integer indexing (`lm.history[0]`, `lm.history[-1]`) and iteration, so the most common access patterns are unchanged.

## Backward compatibility

A `__setstate__` migration path is included: unpickling an older saved module that has `history: list` automatically converts it to `deque`, so saved programs continue to load correctly.

### DIFF
--- a/tests/clients/test_inspect_global_history.py
+++ b/tests/clients/test_inspect_global_history.py
@@ -1,5 +1,6 @@
-import pytest
 from collections import deque
+
+import pytest
 
 import dspy
 from dspy.clients.base_lm import GLOBAL_HISTORY


isaacbmiller:
Because users can and do interact with module.history on their own, losing the slicing ergonomics is not ideal.

Will close for now.

PR #9396, opened 2026-03-06
improved and added doc-strings for field and signature

MaximeRivest:
I followed the google style and focused on the public or complex (but internal) elements


dspy/signatures/field.py
```diff
@@ -1,8 +1,34 @@
...
+factories delegate to `pydantic.Field` but first split caller-supplied
+keyword arguments into two groups:
+
+- **DSPy-specific kwargs** (listed in `DSPY_FIELD_ARG_NAMES`) — stored
+  under `json_schema_extra` so DSPy code can inspect them later.
+- **Standard Pydantic kwargs** — forwarded directly to `pydantic.Field`.
+
+The splitting is performed by the internal `move_kwargs` helper. In
+addition, any Pydantic constraint kwargs (`ge`, `le`, `min_length`,
+etc.) are translated into a human-readable `"constraints"` string and
```
isaacbmiller: add that this human-readable constraint is for the llm

dspy/signatures/field.py
```diff
@@ -1,8 +1,34 @@
...
+stored alongside the DSPy metadata in `json_schema_extra`.
+
+Important:
+    In the current codebase, `desc` and translated `constraints` affect how
+    the standard adapters describe fields to the language model. By contrast,
+    field-level `prefix`, `format`, and `parser` **HAVE NO EFFECT** on
```
isaacbmiller: Rewrite now that PR is merged

dspy/signatures/field.py
```diff
@@ -1,8 +1,34 @@
...
+
+This module also contains the legacy `OldField` / `OldInputField` /
+`OldOutputField` classes and a `new_to_old_field` bridge function.
+These legacy types are still publicly exported for backward compatibility.
+"""
+
 import pydantic
 
-# The following arguments can be used in DSPy InputField and OutputField in addition
-# to the standard pydantic.Field arguments. We just hope pydanitc doesn't add these,
```
isaacbmiller, line 8: lets keep at least the second commend about clashing

dspy/signatures/field.py
```diff
@@ -41,8 +62,31 @@ def move_kwargs(**kwargs):
...
-    """Extracts Pydantic constraints and translates them into human-readable format."""
-
+    """Translate Pydantic constraint kwargs into a human-readable string.
+
+    Scans `kwargs` for keys present in `PYDANTIC_CONSTRAINT_MAP` and
+    builds a comma-separated description of all matched constraints. The
+    result is stored by `move_kwargs` under
+    `json_schema_extra["constraints"]`.
+
+    In the current codebase, adapter utilities include this constraints string
```
isaacbmiller:
```diff
-`json_schema_extra["constraints"]`.
-
-In the current codebase, adapter utilities include this constraints string
+`json_schema_extra["constraints"]` and is included inside the final prompt.
```

dspy/signatures/field.py
```diff
@@ -41,8 +62,31 @@ def move_kwargs(**kwargs):
...
+            `PYDANTIC_CONSTRAINT_MAP` are used; all others are ignored.
+
+    Returns:
+        str: Comma-separated human-readable constraint descriptions, or an
+        empty string if no recognized constraint kwargs are present.
+
+    Examples:
+        >>> _translate_pydantic_field_constraints(ge=5, le=10, desc="score")
+        'greater than or equal to: 5, less than or equal to: 10'
+        >>> _translate_pydantic_field_constraints(desc="name")
```
isaacbmiller: I dont get this example

dspy/signatures/field.py
```diff
@@ -51,11 +95,113 @@ def _translate_pydantic_field_constraints(**kwargs):
...
+    This factory is the standard way to declare an input on a DSPy
+    `Signature`. It accepts all `pydantic.Field` keyword arguments
+    **plus** the DSPy-specific kwargs listed in `DSPY_FIELD_ARG_NAMES`:
+
+    - `desc` — description stored in metadata and used by the standard
+      adapters when they describe fields to the language model. Current
+      optimizer flows such as `GEPA` and `MIPROv2` do not optimize this
+      per-field metadata.
+      
+    - `prefix` — **CURRENTLY HAS NO EFFECT.** Intended as a prompt label, but ignored by modern DSPy adapters.
```
isaacbmiller: update

dspy/signatures/field.py
```diff
@@ -51,11 +95,113 @@ def _translate_pydantic_field_constraints(**kwargs):
...
+        `json_schema_extra["__dspy_field_type"]` set to `"input"`.
+
+    Examples:
+        Use `InputField` in the idiomatic context: a `dspy.Signature`
+        subclass.
+
+        >>> import dspy
+        >>> class ClassifySentiment(dspy.Signature):
+        ...     text: str = dspy.InputField(desc="Text to classify", min_length=1)
+        ...     label: str = dspy.OutputField(desc="Sentiment label")
```
isaacbmiller: I think a second example usecase which is more informative than the fields themselves would be appending an input field to a signature, which is probably the main other time that people find themselves writing InputField other than when writing a class based signature:
```python
dspy.Signature({**signature.input_fields}, "\n".join(instr))
            .append("trajectory", dspy.InputField(), type_=str)
```

dspy/signatures/field.py
```diff
@@ -51,11 +95,113 @@ def _translate_pydantic_field_constraints(**kwargs):
...
+    This factory is the standard way to declare an output on a DSPy
+    `Signature`. It accepts all `pydantic.Field` keyword arguments
+    **plus** the DSPy-specific kwargs listed in `DSPY_FIELD_ARG_NAMES`:
+
+    - `desc` — description stored in metadata and used by the standard
+      adapters when they describe fields to the language model. Current
+      optimizer flows such as `GEPA` and `MIPROv2` do not optimize this
+      per-field metadata.
+
+    - `prefix` — **CURRENTLY HAS NO EFFECT.** Intended as a prompt label, but ignored by modern DSPy adapters.
```
isaacbmiller: update

dspy/signatures/field.py
```diff
@@ -51,11 +95,113 @@ def _translate_pydantic_field_constraints(**kwargs):
...
+    Examples:
+        Use `OutputField` in the idiomatic context: a `dspy.Signature`
+        subclass.
+
+        >>> import dspy
+        >>> class ClassifySentiment(dspy.Signature):
+        ...     text: str = dspy.InputField(desc="Text to classify")
+        ...     label: str = dspy.OutputField(desc="Sentiment label", max_length=20)
+        >>> field = ClassifySentiment.output_fields["label"]
+        >>> field.json_schema_extra["__dspy_field_type"]
```
isaacbmiller: I dont think these are helpful examples

dspy/signatures/field.py
```diff
@@ -68,15 +214,13 @@ def new_to_old_field(field):
 
 
 class OldField:
-    """A more ergonomic datatype that infers prefix and desc if omitted."""
```
isaacbmiller, line 98: Lets  clarify that these are legacy

dspy/signatures/signature.py
```diff
@@ -1,18 +1,49 @@
...
+           '''Answer the question.'''
+           question: str = dspy.InputField()
+           answer: str = dspy.OutputField()
+
+2. **String syntax** — call the base class to get a generated subclass::
+
+       QA = dspy.Signature("question -> answer")
+
+In the class syntax the subclass's docstring becomes the instruction text.
+In the string syntax, instructions can be supplied as the second positional
```
isaacbmiller: show this rather than describe it

dspy/signatures/signature.py
```diff
@@ -1,18 +1,49 @@
...
-You can call Signature("input1, input2 -> output1, output2") to create a new signature type.
-You can also include instructions, Signature("input -> output", "This is a test").
-But it's generally better to use the make_signature function.
+A DSPy signature is a Pydantic model whose fields are explicitly tagged as
+inputs or outputs with `InputField` and `OutputField`.  The `Signature` class
+exposes convenience methods for non-mutating field and metadata manipulation
+(`with_instructions`, `with_updated_fields`, `prepend`, `append`,
+`insert`, `delete`) and lightweight state serialization (`dump_state`,
+`load_state`).  All mutating-style methods return a **new** signature class
+and leave the original unchanged.
```
isaacbmiller: I dont know if these are helpful for users or agents reading this class.

dspy/signatures/signature.py
```diff
@@ -212,44 +370,97 @@ def _validate_fields(cls):
 
     @property
     def instructions(cls) -> str:
+        """Return the signature's instruction text with normalized indentation.
+
+        DSPy stores a signature's natural-language instructions in the class
+        docstring. This property exposes that text in a clean form by applying
```
isaacbmiller: Just restates the code

dspy/signatures/signature.py
```diff
@@ -768,13 +1248,36 @@ def resolve_name(type_name: str):
...
+    Args:
+        attribute_name: Raw Python attribute name such as `camelCaseText` or
+            `html_parser2`.
+
+    Returns:
+        str: Readable label suitable for use as a field prefix, without the
+        trailing colon that callers typically add separately.
 
     Examples:
-        "camelCaseText" -> "Camel Case Text"
```
isaacbmiller, line 788: This format is fine

dspy/signatures/signature.py
```diff
@@ -632,31 +1090,53 @@ def _parse_field_string(field_string: str, names=None) -> dict[str, str]:
...
-        - For "x: MyModule.CustomType", it processes attribute access to return the actual type
+    Name resolution proceeds in this order:
+
+    1. The supplied `names` mapping (seeded from `typing.__dict__` when
+       `names` is `None`).
+    2. A small set of common built-in types (`int`, `str`, `float`,
+       `bool`, `list`, `tuple`, `dict`, `set`, `frozenset`,
+       `complex`, `bytes`, `bytearray`).
+    3. Dynamic `importlib.import_module` for the base name of dotted
+       expressions such as `dspy.Image`.
```
isaacbmiller: Too long. In practice these should be annotated lines in the code if necessary.

isaacbmiller (changes requested):
Didn't get all the way through annotating exactly which ones are too long, but a lot of the docstrings do effectively line by line descriptions of what happens inside the method.

The docstring for user facing types and functions should be short representations of what the user can expect or what the function does while the inline code has comments to clarify anything confusing for someone who needs to dive deeper.

A lot of descriptions mention `field.json_schema_extra["__dspy_field_type"]`, which is generally true of what these methods are doing under the hood, while unimportant for a downstream user.

dspy/signatures/signature.py
MaximeRivest: users' are not expected to read the code, I believe restate the code should not be an issue, but I will take that comment as a queue that is was not helpful enough to a human and improve the explanation

MaximeRivest:
Thanks for the thorough review, Isaac. I've addressed all your feedback across 3 new commits + merged upstream to resolve the conflict with #9394.

**What changed**

`signature.py` — full rewrite of every public method docstring (`588da8e`):
- `with_instructions`, `with_updated_fields`, `prepend`, `append`, `delete`, `insert`, `equals`, `dump_state`, `load_state`, `ensure_signature`, `make_signature`, `infer_prefix`
- Every docstring is now brief, user-facing, and follows a consistent structure: one-line summary, short explanation, args, and tested examples showing real patterns (predictor assignment, chaining, optimizer loops, `custom_types`)
- All references to `json_schema_extra` and `__dspy_field_type` removed from user-facing text
- `infer_prefix` marked as deprecated
- Added TODOs for dropping `prefix` serialization in `dump_state`/`load_state`

`field.py` — kept the runtime deprecation warnings from #9394 alongside our docstrings (`876ca72`). The `InputField`/`OutputField` docstrings improved in [3bba64f](https://github.com/stanfordnlp/dspy/pull/9396/commits/3bba64fbf074ee73576a3767d1abd905c86724f7).

**Bonus bug fix** (`d3170e4`): While writing the `make_signature` examples I found that when `instructions` is omitted and `custom_types` is provided, the default-instruction synthesis re-enters `Signature()` without forwarding `custom_types`. This silently relies on stack-frame auto-detection, which breaks when the custom type isn't in the caller's local scope. Fixed by building a temporary signature from the already-parsed fields. 4-line change.

**Why these docstrings matter for a one-liner method**

The Signature mutation API (`with_instructions`, `prepend`, `append`, etc.) is the foundation for anyone building custom modules or optimizers on top of DSPy. These methods are simple, but their contract matters: immutability, the need to assign back, what `desc` vs `type_` does, etc. Good docstrings here mean users can build on the public API without reading source code — they just hover in their IDE and get everything they need. That's the bar we aimed for: every public method should be usable without leaving your editor.

Ready for another look when you get a chance.

dspy/signatures/field.py
```diff
@@ -74,12 +109,99 @@ def _warn_deprecated_field_args(**kwargs):
...
 
 
-def InputField(**kwargs): # noqa: N802
+def InputField(**kwargs):  # noqa: N802
+    """Declare an input field on a `dspy.Signature`.
+
+    Both the type hint and `desc` are optional. When provided,
+    adapters pass them to the language model for additional context.
+
+    >>> class QA(dspy.Signature):  # doctest: +SKIP
```
isaacbmiller: do we use doctests?

dspy/signatures/field.py
```diff
@@ -74,12 +109,99 @@ def _warn_deprecated_field_args(**kwargs):
...
+    """
     _warn_deprecated_field_args(**kwargs)
     return pydantic.Field(**move_kwargs(**kwargs, __dspy_field_type="input"))
 
 
-def OutputField(**kwargs): # noqa: N802
+def OutputField(**kwargs):  # noqa: N802
+    """Declare an output field on a `dspy.Signature`.
+
+    Both the type hint and `desc` are optional. When provided,
```
isaacbmiller: lets include that type hint defaults to str if none is provided

dspy/signatures/field.py
```diff
@@ -74,12 +109,99 @@ def _warn_deprecated_field_args(**kwargs):
...
+        You can also append an output field to an existing signature
+        programmatically:
+
+        >>> WithConfidence = Review.append(
+        ...     "confidence", dspy.OutputField(desc="confidence score"), type_=float
+        ... )
+        >>> list(WithConfidence.output_fields.keys())
+        ['summary', 'rating', 'confidence']
+
+        See [`dspy.LM`][dspy.LM] for supported providers.
```
isaacbmiller: does this linking work in the docs? Or in IDEs? Not familiar with the syntax

dspy/signatures/field.py
```diff
@@ -74,12 +109,99 @@ def _warn_deprecated_field_args(**kwargs):
...
+    to determine how to parse the model's outputs into Python objects.
+
+    >>> class QA(dspy.Signature):  # doctest: +SKIP
+    ...     question: str = dspy.InputField(desc="a factual question")
+    ...     answer: str = dspy.OutputField(desc="a short answer")
+
+    Args:
+        desc (str): Optional plain-language description of the field.
+            Adapters include this when describing the field to the
+            language model. Recommended for output fields.
```
isaacbmiller: I don't agree that they are recommended, as often times just the name of the field is semantically enough for the LLM to understand what is going on.

Also LLMs already add descriptions too much, so I don't want to encourage this.

dspy/signatures/signature.py
```diff
@@ -1,18 +1,33 @@
...
-If you are not sure if your input is a string representation, (like "input1, input2 -> output1, output2"),
-or a signature, you can use the ensure_signature function.
-
-For compatibility with the legacy dsp format, you can use the signature_to_template function.
+"""DSPy `Signature` class and string-based signature parsing.
+
+A signature declares the input/output contract for a DSPy module.
+Two syntaxes:
+
+    # Class syntax - the docstring becomes the instruction text
```
isaacbmiller: remove comment

dspy/signatures/signature.py
```diff
@@ -1,18 +1,33 @@
...
+    >>> import dspy
+    >>> class Review(dspy.Signature):
+    ...     \"\"\"
+    ...     Summarize the review and give it a rating.
+    ...     \"\"\"
```
isaacbmiller: The triple dots before lines look weird.

dspy/signatures/signature.py
```diff
@@ -260,53 +297,136 @@ def __repr__(cls):
 
 class Signature(BaseModel, metaclass=SignatureMeta):
     ""
-
-    # Note: Don't put a docstring here, as it will become the default instructions
-    # for any signature that doesn't define it's own instructions.
+    # Design constraint: Signature cannot carry a standard class docstring.
+    #
+    # DSPy stores each concrete signature's instruction text in `cls.__doc__`.
+    # A non-empty docstring here would be inherited by every subclass that does
```
isaacbmiller: we dont need to repeat

dspy/signatures/signature.py
```diff
@@ -496,17 +709,88 @@ def dump_state(cls):
...
+
+        The original signature is unchanged.  Fields in `state` are
+        matched positionally against the current field order, not by
+        name.
+
+        Args:
+            state: Dict previously produced by `dump_state`.
+
+        Returns:
+            type[Signature]: A new signature class.
```
isaacbmiller: while we are here can we add the return type as type[Signature]

dspy/signatures/field.py
MaximeRivest: I have started to run doctest on all files I create to ensure the example run. Example that don't work can be very annoying and discouraging the users.

dspy/signatures/field.py
MaximeRivest: it shows that no constraints causes no problem and just returns empty string

dspy/signatures/field.py
MaximeRivest: It is the mkdocs syntax, yes. but looking at the screenshot, I will do another round of formatting so its renders very nice on the website.

Image

dspy/signatures/signature.py
MaximeRivest: ok, yeah, i decided to revert that whole comment. and make sure we have a signature api doc using the .md file 

Image

MaximeRivest:
I prefer the repl format because we can run doctest and we can use it to automatically guarantee that example don't drift away from the code.

for examples that I marked as #doctest: SKIP. We could make them code block (but I rather think we should find ways that they can also run! but i was thinking of punting that to later). 

My goal is that a user could make a module or an optimizer from reading the docstrings in their IDE, a while back learning dspy from reading the code for optimizers and modules I was confused by the append, prepend, etc and it took me some fiddling to realize a really had to reassign the signature into a predict. I think in doubt, I would lean toward too many examples instead of not enough. Let me know what you think!

dspy/signatures/field.py
```diff
@@ -56,10 +71,30 @@ def move_kwargs(**kwargs):
     pydantic_kwargs["json_schema_extra"] = json_schema_extra
     return pydantic_kwargs
 
-
 def _translate_pydantic_field_constraints(**kwargs):
-    """Extracts Pydantic constraints and translates them into human-readable format."""
+    """Translate Pydantic constraint kwargs into a human-readable string.
```
isaacbmiller: This docstring is too long for a method that is pretty readable as code on its own.

dspy/signatures/field.py
```diff
@@ -74,12 +109,100 @@ def _warn_deprecated_field_args(**kwargs):
...
+        >>> summarizer.lm = dspy.LM("provider/model-name")  # doctest: +SKIP
+        >>> summarizer(text="DSPy is a framework for...")  # doctest: +SKIP
+        Prediction(...)
+
+        You can also append an input field to an existing signature
+        programmatically:
+
+        >>> WithContext = Summarize.append(
+        ...     "context", dspy.InputField(desc="background info"), type_=str
+        ... )
```
isaacbmiller: This is the useful part of this example

dspy/signatures/field.py
```diff
@@ -74,12 +109,100 @@ def _warn_deprecated_field_args(**kwargs):
...
+            language model.
+        **kwargs: Any additional `pydantic.Field` keyword argument.
+
+    Returns:
+        (pydantic.fields.FieldInfo): A field marked as a DSPy input.
+
+    Examples:
+        >>> import dspy
+        >>> class Summarize(dspy.Signature):
+        ...     text: str = dspy.InputField(desc="text to summarize")
```
isaacbmiller: This line doesn't add anything above 119

dspy/signatures/field.py
isaacbmiller: feel free to resolve comment if this got fixed

dspy/signatures/field.py
```diff
@@ -74,12 +109,100 @@ def _warn_deprecated_field_args(**kwargs):
...
+        ...     rating: float = dspy.OutputField(desc="rating", ge=0, le=5)
+        >>> reviewer = dspy.Predict(Review)
+        >>> reviewer.lm = dspy.LM("provider/model-name")  # doctest: +SKIP
+        >>> reviewer(text="This product is great!")  # doctest: +SKIP
+        Prediction(...)
+
+        You can also append an output field to an existing signature
+        programmatically:
+
+        >>> WithConfidence = Review.append(
```
isaacbmiller: same comments about example usefulness as in InputField

dspy/signatures/signature.py
```diff
@@ -260,53 +297,105 @@ def __repr__(cls):
...
     @classmethod
     def with_instructions(cls, instructions: str) -> type["Signature"]:
-        """Return a new Signature class with identical fields and new instructions.
+        """Return a new signature with different instructions.
+
+        The original signature is unchanged.  Fields, types, and
+        metadata are copied as-is.  The returned class must be
+        assigned to a predictor (or used to create one) to take effect.
 
-        This method does not mutate `cls`. It constructs a fresh Signature
```
isaacbmiller, line 273: Constructing the fresh Signature Class as opposed to copying or deepcopying is a nice bit of info we should keep

dspy/signatures/signature.py
```diff
@@ -260,53 +297,105 @@ def __repr__(cls):
...
-        class using the current fields and the provided `instructions`.
+        This is the primary mechanism DSPy optimizers use to swap in
+        candidate instructions during prompt search.
 
         Args:
-            instructions (str): Instruction text to attach to the new signature.
+            instructions: Task description the language model will see.
 
         Returns:
-            A new Signature class whose fields match `cls.fields`
```
isaacbmiller, line 280: I like this return desc better,

dspy/signatures/signature.py
```diff
@@ -260,53 +297,105 @@ def __repr__(cls):
...
+            ... )
+            >>> enriched is not QA
+            True
         """
         return Signature(cls.fields, instructions)
 
     @classmethod
     def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dict[str, Any]) -> type["Signature"]:
-        """Create a new Signature class with the updated field information.
+        """Return a new signature with updated metadata on one field.
```
isaacbmiller: Lets keep the old wording indicating its a class

dspy/signatures/signature.py
```diff
@@ -260,53 +297,105 @@ def __repr__(cls):
...
 
-        Returns a new Signature class with the field, name, updated
-        with fields[name].json_schema_extra[key] = value.
+        The original signature is unchanged.  Only the named field is
+        modified in the copy.  The returned class must be assigned to
+        a predictor (or used to create one) to take effect.
+
+        The two things you will typically update are `desc` (the
```
isaacbmiller: [No changes] How much do you use this method?

The most common non-optimizer usecases I can think of are updating a Literal to restrict to some other subset that you discover during runtime.

dspy/signatures/signature.py
```diff
@@ -260,53 +297,105 @@ def __repr__(cls):
...
+            type_: New Python type annotation (e.g. `int`,
+                `list[str]`), or `None` to keep the current type.
+            **kwargs: Field metadata updates.  The most common key
+                is `desc` (the description the language model sees).
 
         Returns:
-            A new Signature class (not an instance) with the updated field information.
+            (type[Signature]): A new signature class.
+
+        Examples:
```
isaacbmiller: I don't think these examples add much above reading the method code

dspy/signatures/signature.py
```diff
@@ -321,86 +410,115 @@ def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dic
 
     @classmethod
     def prepend(cls, name, field, type_=None) -> type["Signature"]:
-        """Insert a field at index 0 of the `inputs` or `outputs` section.
```
isaacbmiller, line 326: I like this desc better

dspy/signatures/signature.py
```diff
@@ -321,86 +410,115 @@ def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dic
...
+
+        The field goes into inputs or outputs depending on whether
+        `field` is an `InputField` or `OutputField`.  The original
+        signature is unchanged.  The returned class must be assigned
+        to a predictor (or used to create one) to take effect.
 
         Args:
-            name (str): Field name to add.
-            field: `InputField` or `OutputField` instance to insert.
-            type_ (type | None): Optional explicit type annotation. If `type_` is `None`, the effective type is
```
isaacbmiller, line 331: Lets keep the bit about NoneType here

dspy/signatures/signature.py
```diff
@@ -321,86 +410,115 @@ def with_updated_fields(cls, name: str, type_: type | None = None, **kwargs: dic
...
-            class MySig(dspy.Signature):
-                input_text: str = dspy.InputField(desc="Input sentence")
-                output_text: str = dspy.OutputField(desc="Translated sentence")
-
-            NewSig = MySig.prepend("context", dspy.InputField(desc="Context for translation"))
-            print(list(NewSig.fields.keys()))
-            ```
+            (type[Signature]): A new signature class.
+
+        Examples:
```
isaacbmiller: examples don't add much.

dspy/signatures/signature.py
```diff
@@ -483,10 +641,34 @@ def equals(cls, other) -> bool:
 
     @classmethod
     def dump_state(cls):
```
isaacbmiller, line 421: Add dict as return type

dspy/signatures/signature.py
```diff
@@ -496,17 +678,88 @@ def dump_state(cls):
 
     @classmethod
     def load_state(cls, state):
```
isaacbmiller, line 440: Add Signature as return type

docs/mkdocs-dev.yml
```diff
@@ -0,0 +1,7 @@
+INHERIT: mkdocs.yml
```
isaacbmiller: Is this file supposed to remain?

isaacbmiller:
The reasoning against extra examples is that it can prioritize the end behavior over the actual code itself, which in some cases is quite concise and communicates the end behavior better than the examples.

isaacbmiller:
Also merge conflicts

docs/mkdocs-dev.yml
MaximeRivest: oups, no sorry

dspy/signatures/field.py
MaximeRivest: i can remove this whole, change and focus on public api.

dspy/signatures/field.py
MaximeRivest: line 132 to 139 is a 'full' example. line 118 to 120 is and illustration of how/where that inputfield is used. many many people don't know how to subclass things in python. I myself learned it for dspy. Seeing that 118 to 120 gives me a visual to attach in my mind the description and the args and understand all together

dspy/signatures/field.py
MaximeRivest: As discussed offline, I believe InputField is a special dspy language interface to the user that deserve a complete miniature example of how to use it in the framework.

dspy/signatures/signature.py
MaximeRivest: The semantic, or the format? The format must be (...): .... to render well with mkdocs

Issue #9398, opened 2026-03-07
[docs] Add Google-style docstrings to dspy/evaluate/auto_evaluation.py


stbiadmin:
Part of the docstring coverage initiative (#8926).

I will add Google-style docstrings to the following public APIs in `dspy/evaluate/auto_evaluation.py`:

- `SemanticF1` class (`__init__` and `forward`)
- `CompleteAndGrounded` class (`__init__` and `forward`)
- `f1_score` function

All docstrings will follow the Google Python style guide with Args/Returns sections.

Issue #8926, opened 2025-10-10
[Help wanted] Add docstring to public APIs


chenmoneygithub:
Hi all,

We would like to get your help on adding docstrings to our public APIs. The general rule is every public API should have proper docstrings unless the API name + args is self-explainable.  Currently we probably only have ~50% coverage, and we would like to move towards >90%. 

To get started:

- Find a file that has missing docstrings.
- Create a github issue before working on it, so we know this file is claimed. In the created issue, mention this issue so that we can easily track it.
- Put up a PR following our [contributing guide](https://github.com/stanfordnlp/dspy/blob/main/CONTRIBUTING.md) Make sure in the PR you mention your issue by adding `resolve #....` in the PR description.
- Tag any collaborator as the reviewer. If you don't know who to request review, you can tag me as the reviewer. 

Some rules:

- The docstring must be 100% correct, wrong docstring is way worse than having nothing.
- The docstring follows the [Google python style guide](https://google.github.io/styleguide/pyguide.html)
- It's recommended to include a code example to explain how to use the API/class. 
- We don't enforce type hints, but we don't reject type hints either. 
- Nothing in my head right now, probably more coming later...

You can refer to existing docstrings as reference, and if you feel ambiguous, feel free to tag us for questions.

Thank you ahead for your help.

Cheers,
chenmoney

onel:
I can help with that.
I've been helping several OSS projects with this exact problem. I'll make a PR soon with some contributions.

Should we keep this issue as the reference one?

chenmoneygithub:
@onel  Thank you! Could you create a separate issue that lists out the file you are working on with the PR? thanks!

onel:
Sure thing, created #8930

MaximeRivest:
I would add that griffe is the handler used by mkdocs consult for specific syntax: https://mkdocstrings.github.io/griffe/reference/docstrings/#google-section-modules

Please render on your machine and add screenshots of the proper rendering in the PR.

joaquinhuigomez:
Claiming `dspy/teleprompt/random_search.py` — added Google-style docstrings to `BootstrapFewShotWithRandomSearch` (class, `__init__`, and `compile`). PR: #9469, tracking issue: #9468.

isaacbmiller:
Hi folks, we will be closing this issue. Its too easy for a bot to drive by and add these PRs. Please talk to the maintainers on discord or X to confirm that you are a human working on this.

PR #9399, opened 2026-03-07
docs(evaluate): add docstrings to auto_evaluation.py

stbiadmin:
Part of #8926

Adds class-level docstrings to SemanticF1, CompleteAndGrounded, and f1_score in dspy/evaluate/auto_evaluation.py.

Ref: #9398


isaacbmiller:
LGTM. Thanks for contributing!

PR #9411, opened 2026-03-09
Add SandboxSerializable protocol for custom RLM types

kmad:
Per the discussion in #9283, @isaacbmiller had good feedback regarding the architecture of RLM type support. This PR implements those changes to create an abstraction layer allowing developers to flexibly implement types for use within the RLM REPL.

Specifically this PR covers:

  1. `typing.Protocol` which defines a new `SandboxSerializable` protocol with 4 abstract methods: `sandbox_setup`, `to_sandbox`, `sandbox_assignment`, `rlm_preview`. Types that inherit get a concrete to_repl_variable() for free. Conformance is checked via `isinstance()`. This allows for a standardized way to interact with the REPL while allowing flexibility for defining dependencies (e.g., defining imports that you need for that type) and other serialization logic.
  2. RLM owns serialization (not the interpreter); `_prepare_serializable_vars()` in RLM handles serialization once
  before the iteration loop. The interpreter stays generic as it just runs code and manages variable injection.
  3. `DataFrame` is provided as an initial custom implementation, but not as part of "core" DSPy: `DataFrame` is not a dspy import. An example implementation lives in `docs/docs/tutorials/dataframe_rlm/cohort_analysis/` as a standalone reference implementation, with a tutorial notebook that walks through a cohort churn analysis across three DataFrames.
      - This notebook shows a concrete example of how to process Dataframes natively with RLMs by passing them to the REPL.

Closes #9283.


kmad:
@isaacbmiller bumping for review - does this satisfy the initial concerns regarding custom types?

dspy/predict/rlm.py
```diff
@@ -351,12 +353,39 @@ def _get_output_fields_info(self) -> list[dict]:
...
+    @staticmethod
+    def _is_sandbox_serializable(value: Any) -> bool:
+        """Check if a value implements the SandboxSerializable protocol.
+
+        Checks isinstance first (for types that inherit from SandboxSerializable),
+        then falls back to duck-typing on the 4 core methods for structural conformance.
+        """
+        return (
+            isinstance(value, SandboxSerializable)
+            or (
```
isaacbmiller, line 365: why do we have this fallback?

dspy/primitives/sandbox_serializable.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Implementors define how to serialize their data into the sandbox,
+    what setup code (imports) is needed, how to reconstruct the value
+    from the serialized payload, and how to produce an LLM-friendly preview.
+
+    Example implementation::
+
+        class DataFrame(SandboxSerializable):
+            def sandbox_setup(self) -> str:
+                return "import pandas as pd\\nimport base64\\nimport io"
```
isaacbmiller, line 36: How is the model supposed to know that we import pandas as pd

dspy/primitives/sandbox_serializable.py
```diff
@@ -0,0 +1,94 @@
...
+
+            def sandbox_assignment(self, var_name: str, data_expr: str) -> str:
+                return f"{var_name} = pd.read_parquet(io.BytesIO(base64.b64decode({data_expr})))"
+
+            def rlm_preview(self, max_chars: int = 500) -> str:
+                return f"DataFrame: {self.data.shape[0]} rows x {self.data.shape[1]} columns"
+    """
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type: Any, handler: GetCoreSchemaHandler) -> core_schema.CoreSchema:
```
isaacbmiller, line 49: Why do we have this method if we arent subclassing?

dspy/primitives/sandbox_serializable.py
```diff
@@ -0,0 +1,94 @@
...
+class SandboxSerializable(Protocol):
+    """Protocol for types that support RLM sandbox injection.
+
+    Implementors define how to serialize their data into the sandbox,
+    what setup code (imports) is needed, how to reconstruct the value
+    from the serialized payload, and how to produce an LLM-friendly preview.
+
+    Example implementation::
+
+        class DataFrame(SandboxSerializable):
```
isaacbmiller, line 34: You probably wouldn't subclass the protocol unless we add the mandatory method .to_repl_variable

dspy/predict/rlm.py
```diff
@@ -351,12 +353,39 @@ def _get_output_fields_info(self) -> list[dict]:
...
+        )
+
     def _build_variables(self, **input_args: Any) -> list[REPLVariable]:
         """Build REPLVariable list from input arguments with field metadata."""
         variables = []
         for name, value in input_args.items():
             field_info = self.signature.input_fields.get(name)
-            variables.append(REPLVariable.from_value(name, value, field_info=field_info))
+            if self._is_sandbox_serializable(value):
+                if hasattr(value, "to_repl_variable"):
```
isaacbmiller, line 379: should we add to_repl_variable as an optional method on the protocol with a default implementation?

Make it a mandatory method?

WDYT

dspy/predict/rlm.py
```diff
@@ -370,6 +399,48 @@ def _validate_inputs(self, input_args: dict[str, Any]) -> None:
         if missing:
             raise ValueError(f"Missing required inputs: {sorted(missing)}")
 
+    def _prepare_serializable_vars(
```
isaacbmiller, line 402: How does this interact with large variables?

Can you add a test for this?

isaacbmiller (changes requested):
Awesome, this is looking great!

Some smaller questions before approving, but very excited for this.

Thank you for the contribution and sorry for all the delays.

dspy/predict/rlm.py
kmad: Removed the multiple fallbacks, now relying just on `isinstance(value, SandboxSerializable)`.

dspy/primitives/sandbox_serializable.py
kmad: We now surface `sandbox_setup()` in the variable description via `build_repl_variable()`, so the prompt explicitly tells the model which imports/names are available in the sandbox, (in the DataFrame example, it will include aliases like `pd`).

dspy/primitives/sandbox_serializable.py
kmad: Removed the Pydantic hook from the protocol. `SandboxSerializable` is now a pure 4-method protocol; the schema hook lives only on the optional `SandboxSerializableBase` for cases where the type is used directly as a field annotation in a class-based `dspy.Signature` (the more readable case, imo).

dspy/primitives/sandbox_serializable.py
kmad: Agreed; kept `to_repl_variable` off the protocol and put the ergonomic path on `SandboxSerializableBase` instead, so the protocol stays minimal while subclassing users still get a default helper.

dspy/predict/rlm.py
kmad: I moved this to `SandboxSerializableBase` to keep the protocol itself minimal. RLM still uses the same default logic via `build_repl_variable()`; this just moves the host-side REPLVariable packaging out of the protocol surface. This seemed like a good way to organize it conceptually but open to feedback.

dspy/predict/rlm.py
kmad: Payloads are provided using the sandbox's variables as transport rather than being inlined into the code string, so existing large-variable handling remains intact. Added a unit test asserting that a multi-MB payload is captured as a variable, not embedded in the code, and that it is processed as such in the Deno sandbox.

kmad:
@isaacbmiller my turn to apologize for delays. See changes & responses above, let me know if any more feedback.

dspy/primitives/sandbox_serializable.py
isaacbmiller: Now that we have the full class, I am not sure if the protocol makes as much sense? 1. When would you implement the protocol if you can't use that type inside of a Signature?
2. We can make this class an ABC and explicitly mark the methods that a child class would need to implement.

If all SandboxSerializeables need the pydantic function to work in dspy, then we should just have a base class.

Can you also explain the pydantic function (both to me and in the docs)?

kmad: A class is probably cleaner, this was the route I was initially trying to go down but tried the Protocol per the last PR's comments, so happy to revert (I think this is a more intuitive approach anyway). 

Latest commit implements an ABC and tries to clean up/ simplify the implementation.

If I'm understanding your question about the pydantic function correctly, this is what allows you to use it as a Pydantic type, e.g., `df: Dataframe = dspy.InputField()`, otherwise you can run into serialization or type errors. I've added this to the RLM docs and self-documented in the code.

isaacbmiller:
Awesome, this looks really great!!

Thank you for working with me to get it ready to merge!

PR #9414, opened 2026-03-09
Add opt-in safe disk cache backend (msgspec + restricted pickle)

### LINKED ISSUE #9413: [Feature] Diskcache Replacement to unblock deployments
[CVE-2025-69872](https://www.sentinelone.com/vulnerability-database/cve-2025-69872/) (CVSS 9.8) is a remote code execution vulnerability in diskcache via pickle deserialization, affecting all versions through 5.6.3.

Diskcache is no longer actively maintained, so no patched version exists. 

Because diskcache is a required dependency, every DSPy installation inherits this CVE, which blocks deployment in environments with mandatory security scanning. The risk surface of this exploit is realistically limited to chains where you already have write access.

Precedent: https://github.com/PrefectHQ/fastmcp/issues/3239 for this CVE and took the same approach. Google ADK avoids the problem entirely by using optional dependency groups for non-core features pip install google-adk[extensions]

We cannot follow the exact same fixes as Prefect or Google, as diskcache is a core dependency for us.

Issue created to track #9376 and related PRs.

#9376 is not mergeable because we still need persistent caching. In-memory caching only is not acceptable for users who are used to have persistent caches between runs.

Plan:

1. We will create a diskcache replacement for our usecases, serializing our responses to JSON.
2. We only store the ModelResponse object inside of the cache. This is currently not JSON Serializable.
3. Importantly, we also need to ensure that we can migrate from older caches.

Because this is a larger dependency change, we will release 3.2 alongside it once we are sure that the fix is working well.

isaacbmiller:
## Summary

This PR adds an opt-in safe disk cache backend for DSPy as an alternative to pickle-based disk deserialization. Users can opt in with:

```python
dspy.configure_cache(use_pickle=False)
```

Pickle remains the default in this release, so existing users see no behavior change unless they explicitly switch.

The new backend uses a **routing-byte prefix** (inspired by [CocoIndex](https://cocoindex.io/blogs/type-guided-serde)) to dispatch between serialization engines:

| Byte | Engine | Used for |
|------|--------|----------|
| `0x01` | msgspec (msgpack) | Plain values (dicts, lists, scalars), dataclasses |
| `0x02` | Pydantic | `BaseModel` subclasses (type metadata in msgpack body) |
| `0x80` | Restricted pickle | `numpy.ndarray` only (hardcoded allowlist) |

This addresses the arbitrary-code-execution risk called out in [#9413](https://github.com/stanfordnlp/dspy/issues/9413).

Closes #9413

## Usage

```python
import dspy

# Opt into safe disk serialization
dspy.configure_cache(use_pickle=False)

# Register custom pydantic models or dataclasses for safe serialization
dspy.configure_cache(use_pickle=False, safe_types=[MyCustomModel])

# Explicitly keep pickle (current default)
dspy.configure_cache(use_pickle=True)
```

Safe mode and pickle mode can share a cache directory. Switching between modes is safe in both directions:
- **Pickle -> safe**: Legacy pickle entries are detected (non-bytes values from raw SQLite storage, or legacy diskcache modes) and treated as cache misses, then recomputed in the new format.
- **Safe -> pickle**: Safe-mode entries are tagged in diskcache's `tag` column. Pickle-mode reads detect the tag and transparently decode safe-format entries.

## What Changed

| File | Change |
|------|--------|
| `pyproject.toml` | Adds `msgspec>=0.18.0` dependency. |
| `dspy/clients/disk_serialization.py` | New module. Implements routing-byte serialization: `0x01` msgspec msgpack for plain values/dataclasses, `0x02` pydantic `model_dump` + msgpack for `BaseModel` subclasses (with type metadata, schema hash, and allowlist enforcement), `0x80` restricted pickle for numpy `ndarray` only via `_RestrictedUnpickler` (hardcoded to allow only `numpy.dtype` and `numpy._core.numeric._frombuffer`). Pydantic reconstruction uses `TypeAdapter(cls).validate_python()` which handles `RootModel`, nullable fields, and non-init dataclass fields. `SafeDisk.fetch` guards against legacy non-bytes entries. |
| `dspy/clients/cache.py` | Adds `use_pickle` and `safe_types` cache configuration. Routes disk cache initialization through `SafeDisk` when `use_pickle=False`. `_transform_value` matches main's scope (pydantic + callable only). `Cache.__init__` guards `os.fspath(disk_cache_dir)` for `None` when disk cache is disabled. |
| `dspy/clients/__init__.py` | Exposes `use_pickle` and `safe_types` through `configure_cache()`. |
| `tests/clients/test_disk_serialization.py` | Tests for all three routing paths, RootModel roundtrips, corrupt payload handling, restricted unpickler rejection, schema change detection, allowlist isolation. |
| `tests/clients/test_cache.py` | Tests for safe-mode roundtrips, legacy primitive entry migration, configure_cache with safe_types, bidirectional format compatibility, disk-disabled with None dir. |

## Design Decisions

- **Routing-byte prefix** instead of JSON envelope. Single byte selects the engine; body is engine-specific. Inspired by [CocoIndex's type-guided serde](https://cocoindex.io/blogs/type-guided-serde).
- **msgspec for plain values** -- fast msgpack encode/decode, no type metadata needed.
- **Pydantic path embeds type metadata** (module, class, schema hash) in the msgpack body because the shared cache doesn't know the return type at decode time.
- **Restricted pickle for numpy only** -- `_RestrictedUnpickler.find_class` is hardcoded to allow only numpy array reconstruction classes. Not user-extensible.
- **Allowlist for pydantic/dataclass types** on decode as defense-in-depth. Built-in OpenAI and LiteLLM response types auto-included via `default_allowed_types()`.
- **Tuple->list drift** is a known msgpack limitation (no tuple type on the wire), documented in tests.
- **`_transform_value` kept minimal** -- matches main's scope (pydantic models + callables), no added dataclass/set/tuple branches.

## Tests

```bash
uv run pytest --deno tests/clients/test_cache.py tests/clients/test_disk_serialization.py
```

### DIFF
--- a/dspy/clients/disk_serialization.py
+++ b/dspy/clients/disk_serialization.py
@@ -42,7 +42,7 @@ def _restricted_load(f: Any, allowed: set[tuple[str, str]]) -> Any:
                 return super().find_class(module, name)
             raise DeserializationError(
                 f"Type {module}.{name} is not in the safe_types allowlist. "
-                f"Register it via safe_types=[...] in configure_cache()."
+                f"Register it via dspy.configure_cache(safe_types=[...])."
             )
 
     return _Unpickler(f).load()


dspy/clients/cache.py
```diff
@@ -51,17 +81,62 @@ def __init__(
...
+                    size_limit=disk_size_limit_bytes,
+                )
+            else:
+                # When DSPY_MIGRATE_CACHE=1, read legacy pickle entries BEFORE
+                # creating FanoutCache (which overwrites the shard DB files).
+                pending_entries = None
+                if os.environ.get("DSPY_MIGRATE_CACHE") == "1":
+                    pending_entries = read_legacy_entries(disk_cache_dir)
+                    if pending_entries:
+                        remove_legacy_shard_dbs(disk_cache_dir)
```
chenmoneygithub: Do we want to only remove the pickle-based cache if there is no error during migration? As written, if `_migrate_entries` partially fails, the legacy DBs are already gone and the user ends up with a half-migrated cache and no way back.

dspy/clients/cache.py
```diff
@@ -51,17 +81,62 @@ def __init__(
...
+                    shards=16,
+                    timeout=10,
+                    directory=disk_cache_dir,
+                    size_limit=disk_size_limit_bytes,
+                )
+            else:
+                # When DSPY_MIGRATE_CACHE=1, read legacy pickle entries BEFORE
+                # creating FanoutCache (which overwrites the shard DB files).
+                pending_entries = None
+                if os.environ.get("DSPY_MIGRATE_CACHE") == "1":
```
chenmoneygithub: Can you document `DSPY_MIGRATE_CACHE=1` in the `Cache.__init__` docstring (and ideally `configure_cache` too)? Right now it's only discoverable from the PR description, and we should make this more visible

dspy/clients/cache.py
```diff
@@ -75,56 +150,47 @@ def cache_key(self, request: dict[str, Any], ignored_args_for_cache_key: list[st
...
-        elif self.enable_disk_cache and key in self.disk_cache:
-            # Found on disk but not in memory cache, add to memory cache
-            response = self.disk_cache[key]
+                if key in self.memory_cache:
+                    response = self.memory_cache[key]
+                    memory_hit = True
+
+        if not memory_hit and self.enable_disk_cache:
+            try:
+                response = self.disk_cache.get(key, _CACHE_MISS)
```
chenmoneygithub: Is the sentinel pulling its weight? It only matters if something legitimately caches `None`, and as far as I can tell nothing does — `request_cache` only stores LM responses. `self.disk_cache.get(key)` with an `is None` check gets you the same single-lookup / race-free behavior without the module-level sentinel. Happy to keep it if you're planning to cache `None` somewhere, but then it'd be worth a comment saying so.

dspy/clients/disk_serialization.py
```diff
@@ -0,0 +1,132 @@
...
+            import numpy as np
+
+            return np.asarray(value[_ENCODED_DATA_KEY], dtype=np.dtype(value[_ENCODED_DTYPE_KEY]))
+        return {k: _decode_value(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_decode_value(item) for item in value]
+    return value
+
+
+def _resolve_class(module_name: str, qualname: str) -> type:
```
chenmoneygithub: The PR description says this checks an `allowed_module_prefixes` allowlist and validates `issubclass(cls, BaseModel)`, but neither is here. `cls.model_validate(...)` blocks direct RCE on a non-BaseModel result (AttributeError), but `importlib.import_module(arbitrary_name)` still runs the target module's top-level code on first import — attacker-controlled cache bytes can force-import any installed module, which is the threat this PR is motivated by.

Minimum fix: add an `issubclass(cls, pydantic.BaseModel)` check here and a test that a poisoned `__dspy_cache_module__` (e.g. `"os"` / `"subprocess"`) entry is rejected. The prefix allowlist is nice-to-have on top.

dspy/clients/disk_serialization.py
```diff
@@ -0,0 +1,132 @@
...
+    module = importlib.import_module(module_name)
+    obj = module
+    for attr in qualname.split("."):
+        obj = getattr(obj, attr)
+    return obj
+
+
+def _pydantic_to_json(value: Any) -> Any:
```
chenmoneygithub: Question: why not just `return value.model_dump(mode="json")` here? `model_dump` already produces plain nested dicts (nothing extra to flatten), and the hand-rolled version looks like it would miss `@field_serializer` / `@model_serializer` / `@computed_field` and bypasses pydantic's alias resolution by reading `value.__dict__` directly. Did you hit a case where `model_dump(mode="json")` didn't round-trip through `model_validate`? If so, a comment pointing at it would help; if not, the shorter version seems safer.

tests/clients/test_disk_serialization.py
```diff
@@ -0,0 +1,387 @@
...
+        function={"name": "test", "arguments": "{}"}, id="call_123", type="function",
+    )
+    result = _deserialize(_serialize(tc))
+    assert isinstance(result, ChatCompletionMessageToolCall)
+    assert result.function.name == "test"
+
+
+def test_envelope_like_keys_not_confused(tmp_path):
+    cache = _make_fanout_cache(str(tmp_path))
+    tricky = {"_pydantic": "fake.Module.FakeClass", "_data": {"foo": "bar"}}
```
chenmoneygithub: These aren't the real envelope keys — the code uses `__dspy_cache_type__` / `__dspy_cache_module__` / `__dspy_cache_qualname__` / `__dspy_cache_data__`, so this test passes trivially (neither `_pydantic` nor `_data` is recognized by `_decode_value`).

Please update `tricky` to use the actual sentinel keys, e.g.
```python
tricky = {
    "__dspy_cache_type__": "pydantic",
    "__dspy_cache_module__": "os",
    "__dspy_cache_qualname__": "system",
    "__dspy_cache_data__": {},
}
```
and assert the behavior you want — either it raises, or it round-trips identically. Today, a user dict with those keys at the top level would be misinterpreted as an envelope on fetch.

tests/clients/test_cache.py
```diff
@@ -345,27 +407,242 @@ def test_function(**kwargs):
...
+        )
+
+        # Write garbage bytes to a .pkl file
+        pkl_path = str(tmp_path / "corrupt.pkl")
+        with open(pkl_path, "wb") as f:
+            f.write(b"\x80\x05\x95CORRUPT_GARBAGE_DATA_NOT_VALID_PICKLE")
+
+        # load_memory_cache does not catch deserialization errors;
+        # it raises UnpicklingError from cloudpickle/pickle
+        with pytest.raises((pickle.UnpicklingError, Exception)):
```
chenmoneygithub: The docstring above says _"NOTE: `load_memory_cache()` does NOT catch deserialization errors — corrupt pickle files raise UnpicklingError. This is documented as a discovered issue."_ Pinning known-broken behavior in a test is a weird pattern — it locks the bug in place. Either fix it here (a small try/except around the `cloudpickle.load`) or file a separate issue and link it from the test, rather than asserting the broken behavior.

Also, `pytest.raises((pickle.UnpicklingError, Exception))` reduces to `pytest.raises(Exception)` since `UnpicklingError` subclasses `Exception`, and `Exception` is too broad to mean anything.

isaacbmiller: Removing this test

isaacbmiller:
@codex review

isaacbmiller:
@codex review

isaacbmiller:
@codex review

isaacbmiller:
@codex review

chenmoneygithub (approved):

isaacbmiller:
superceded by #9629

isaacbmiller:
superceded by #9629

PR #9418, opened 2026-03-11
Fix/docstring examples section header

MaximeRivest:
Fixes the rendering problems for `Examples`, it cannot be singular. And fixes the trailing ::: that was in most sections in api ref.

Screenshot from 2026-03-11 07-41-19
Screenshot from 2026-03-11 07-45-17
Screenshot from 2026-03-11 07-43-03

@isaacbmiller can you review?


isaacbmiller:
LGTM

PR #9432, opened 2026-03-12
fix(signature): reject duplicate input and output field names

MaximeRivest:
This fixes a bug where input field did not 'register' when input and output fields had the same name. Now we throw a ValueError.


isaacbmiller (approved):
LGTM. OOC What happened before? Did only one field stay?

MaximeRivest:
the model was working without that input

image

Issue #9425, opened 2026-03-12
[Bug] TypeError: issubclass() arg 1 must be a class when using ReAct with tools


MoNafea01:
### What happened?

When using dspy.ReAct with tool functions, I'm getting a TypeError in the adapter's _call_preprocess method. The error occurs when DSPy tries to check issubclass(field.annotation, Type) but the annotation is not a class (likely a generic type like List[dict]).

Environment:

DSPy version: 3.1.3
Python version: 3.x (miniconda)
OS: Windows

### Steps to reproduce

import dspy
from pydantic import BaseModel

dspy.configure(lm=dspy.LM("groq/qwen/qwen3-32b"))

class Date(BaseModel):
    year: int
    month: int
    day: int
    hour: int

class UserProfile(BaseModel):
    user_id: str
    name: str
    email: str

def get_user_info(name: str):
    """Fetch the user profile from database with given name."""
    return UserProfile(user_id="1", name=name, email=f"{name}@example.com")

class CustomerService(dspy.Signature):
    """Handle user requests."""
    user_request: str = dspy.InputField()
    process_result: str = dspy.OutputField()

react = dspy.ReAct(CustomerService, tools=[get_user_info])
result = react(user_request="Help me, my name is Adam")  # TypeError here

### DSPy version

3.1.3

isaacbmiller:
Thanks for the report. I tried to reproduce this locally.

My Setup:
- uv with CPython 3.10.19
- inspected the generated ReAct signature, especially next_tool_args
- exercised the adapter preprocess path directly, since the reported TypeError should happen before the actual LM call
- also checked the full react(...) path with a stubbed LM response path

I could not reproduce the crash from the example as written.

I think there is still some environment-specific detail missing from the repro, and I don't have a windows machine to check with.

Could you share:
1. the full traceback, not just the final exception line
2. exact Python version, including patch version, from python --version

Thanks for the help!

MoNafea01:
The issue was due to an incompatibility between pydantic and dspy versions
I was using 
pydantic==2.10.3  
dspy-ai==3.1.3

the error was 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File c:\Users\a1mme\miniconda3\envs\nawafiz\lib\site-packages\dspy\adapters\chat_adapter.py:73, in ChatAdapter.__call__(self, lm, lm_kwargs, signature, demos, inputs)
     72 try:
[...]

PR #9433, opened 2026-03-12
fix(adapter): guard annotation subclass checks for ReAct tool args #9425

isaacbmiller:
Adapter code was calling `issubclass()` on arbitrary annotations before confirming the annotation was one of the native DSPy response types. 

On Python 3.10, ReAct's generated `next_tool_args: dict[str, Any]` can hit that path and raise `TypeError`, so agent calls fail before normal formatting or parsing completes.

This only happens on older versions of pydantic, which 3.14 wont install. Reproduced on py 3.10.19 + pydantic 2.10.3

## Change
The fix is small and shared: reorder the native-response check so `issubclass()` only runs after `field.annotation in self.native_response_types`, and use the same TypeError-safe subclass helper for the other adapter annotation checks. I also added an end-to-end ReAct regression for the generic-alias failure mode.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -105,8 +105,8 @@ class Adapter:
         for name, field in signature.output_fields.items():
             if (
                 isinstance(field.annotation, type)
-                and issubclass(field.annotation, Type)
                 and field.annotation in self.native_response_types
+                and issubclass(field.annotation, Type)
             ):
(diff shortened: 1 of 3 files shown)


isaacbmiller:
doesn't work! Codex really convinced me this was a real thing until i pushed it harder to repro the original issue

isaacbmiller:
Actually it does work! Reproduced by setting pydantic version to 2.10.3 with python 3.10.

PR #9434, opened 2026-03-12
Improve Predict Ergonomics

MaximeRivest:
Predict is a major interface for interacting with DSPy. This PR improves that interface by adding:

1. autocompletions, 
2. hover
3. improving docstrings
4. make input fields args instead of kwargs (sooo nice!) 

 for `dspy.Predict`. The autocompletion is nicest for the engines based on runtime information (ipython notebook for example). In this PR, I also make some of the reserved kwargs explicit: `lm`, `config`, `demos` and `signature`. That latter, `signature`, enabled a per call signature override, the pattern was never used in the codebase but seems nonetheless useful and some code may depend on it, it is still usable, but I have added a more explicit `signature_override` for the kwargs.

This is now what users of notebooks will see, it works on class signature:

---

paste-2026-03-12T12-58-49

---

and strings:

---

paste-2026-03-12T13-00-38

---

In vscode notebooks:

---

paste-2026-03-12T17-13-38

---

But not for autocompletion in vscode notebook, we see ... instead:

---

paste-2026-03-12T17-15-43

---

and sometimes, in vscode, we see that:

---

paste-2026-03-12T17-16-33

---

it also works in other notebooks (mrmd :) )

---

paste-2026-03-12T12-56-18

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -1,3 +1,10 @@
+"""Prompt a language model with named inputs and get named outputs back.
+
+`Predict` pairs a DSPy signature with a language model. Call the
+predictor with inputs that match the signature, then read the outputs
+from the returned `Prediction`.
+"""
+
 import inspect
 import logging
 import random
@@ -24,6 +31,7 @@ UNSAFE_LM_STATE_KEYS = {"api_base", "base_url", "model_list"}
 
 
 def _sanitize_lm_state(lm_state: dict, allow_unsafe_lm_state: bool) -> dict:
+    """Strip sensitive keys from a serialized LM config unless explicitly allowed."""
     if allow_unsafe_lm_state:
         return lm_state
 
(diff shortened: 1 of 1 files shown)


dspy/predict/predict.py
```diff
@@ -115,35 +291,94 @@ def load_state(self, state: dict, *, allow_unsafe_lm_state: bool = False) -> "Pr
...
+            inputs=inputs,
+            input_field_names=set(self.signature.input_fields),
+        )
+        effective_signature = ensure_signature(self.signature if signature_override is None else signature_override)
+        demos = self.demos if demos is None else demos
+        config = {**self.config, **({} if config is None else config)}
 
         # Get the right LM to use.
-        lm = kwargs.pop("lm", self.lm) or settings.lm
+        lm = self.lm if lm is None else lm
```
isaacbmiller, line 380: Unclear code

dspy/predict/predict.py
```diff
@@ -40,35 +49,198 @@ def _sanitize_lm_state(lm_state: dict, allow_unsafe_lm_state: bool) -> dict:
...
+
+    if len(args) > len(input_field_names):
+        raise TypeError(
+            "Too many positional arguments when calling `dspy.Predict`. "
+            f"Expected at most {len(input_field_names)} positional input field(s) matching this signature: "
+            f"'{', '.join(input_field_names)}'."
+        )
+
+    bound_inputs = {name: value for name, value in zip(input_field_names, args, strict=False)}
+    duplicate_fields = sorted(set(bound_inputs).intersection(inputs))
```
michaelisaac-dev, line 82: Hmm I think this can diverge from the `_predict_signature_to_call_signature` in some cases. Right now positional args are bound by field order, but the generated signature below may force later params to be keyword-only

For example, the signature of `context` suggests it's a kwarg, however, it's obviously accepted as a positional argument at runtime
Image

dspy/predict/predict.py
```diff
@@ -40,35 +49,198 @@ def _sanitize_lm_state(lm_state: dict, allow_unsafe_lm_state: bool) -> dict:
...
+        if not keyword_only_inputs and default is not inspect.Parameter.empty:
+            seen_positional_default = True
+
+    reserved_params = [
+        ("config", dict[str, Any] | None),
+        ("signature_override", str | type[Signature] | None),
+        ("demos", list | None),
+        ("lm", BaseLM | None),
+    ]
+    for name, annotation in reserved_params:
```
michaelisaac-dev, line 130: How do you deal with collisions in case the Signature has e.g., a `config` input field?

michaelisaac-dev: Image

michaelisaac-dev:
I really like that this PR adds the possibility of using positional args for Signature inputs, however, I’m less certain about the added complexity around constructing and updating a runtime Signature. At least in my testing, neither PyCharm nor VSCode notebooks seem to recognize the typed signature in a meaningful way, so I’m not sure about the payoff there.

To me, the bigger win would be if we could introduce static typing for module inputs and return types based on the signature, since that would help much more consistently with autocomplete and type checking. For the runtime-only input type hints, I’m not fully sure who this feature is helping in practice.

michaelisaac-dev:
For consistency, I think we’d need positional-arg support for other modules too, not just `Predict`. How about moving your implementation to `Module.__call__` instead of `Predict.__call__`?

PR #9444, opened 2026-03-13
Improve dspy.Example docstrings and examples

MaximeRivest:
## Summary

Improve the `dspy.Example` docs so they read better in both MkDocs and IDE help.

This PR:
- rewrites the class docstring around the mental model users actually need: an `Example` is one row of task data
- connects `Example` to familiar concepts like pandas rows and HuggingFace dataset records
- explains how `with_inputs()`, `inputs()`, and `labels()` are used in practice
- adds clearer doctest-style examples, including trainset and metric usage
- documents all public methods shown in the API reference with concrete Google-style docstrings
- keeps `__init__` focused on constructor help so signature help in IDEs is more useful

## Why

`dspy.Example` is one of the first APIs new users encounter, but the previous docs were vague and a bit awkward when rendered in MkDocs. In particular:
- the class overview did not clearly explain what an `Example` is in everyday terms
- the rendered "Key features" section felt odd
- `__init__` help was too abstract for users typing `dspy.Example(...)` in an IDE
- several public methods in the API reference were undocumented

This revision aims to make the docs more concrete, beginner-friendly, and easier to use from both the website and editor hovers.

# Before

image

# After

image

### DIFF
--- a/dspy/primitives/example.py
+++ b/dspy/primitives/example.py
@@ -2,11 +2,11 @@ from pydantic import BaseModel
 
 
 class Example:
-    """Store one DSPy data point with named fields.
+    """A flexible data container for DSPy examples and training data with named fields.
 
-    Think of an `Example` as one row from a HuggingFace dataset or pandas
-    DataFrame. Each field is a column value you can access by name, either with
-    `example.question` or `example["question"]`.
+    An `Example` is roughly one row from a HuggingFace dataset or pandas
+    `DataFrame`. It behaves a lot like a dictionary or dot-access record: you
+    can read fields with `example["question"]` or `example.question`.
 
     In DSPy, lists of `Example` objects are your trainset, devset, and testset.
     Most examples are built from keyword arguments or an existing record, then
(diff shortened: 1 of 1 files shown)


dspy/primitives/example.py
```diff
@@ -2,91 +2,102 @@
...
-
-        # Get with default value
-        city = example.get("city", "Unknown")
-        print(city)  # "Unknown"
-        ```
+    Examples:
+        Build one from keyword arguments:
+
+        >>> import dspy
+        >>> example = dspy.Example(
```
isaacbmiller, line 25: Examples should pretty much never be written without .with_inputs, so the doc examples should reflect that

dspy/primitives/example.py
```diff
@@ -2,91 +2,102 @@
 
 
 class Example:
-    """A flexible data container for DSPy examples and training data.
+    """Store one DSPy data point with named fields.
```
isaacbmiller: Prefer a first line of the docstring saying what it is not what it does.

```suggestion
    """A flexible data container for DSPy examples and training data with named fields.
```

dspy/primitives/example.py
```diff
@@ -2,91 +2,102 @@
...
 
-    The `Example` class is the standard data format used in DSPy evaluation and optimization.
+    Think of an `Example` as one row from a HuggingFace dataset or pandas
```
isaacbmiller:
```suggestion
An `Example` is the equivalent to one row from a HuggingFace dataset or pandas. 
```

Add that it generally acts like a dict/dotdict except it has some keys that are marked as inputs.

MaximeRivest:
updated the opening line to describe what Example is, clarified the dict-like behavior, and revised the examples to use with_inputs(...).

PR #9445, opened 2026-03-13
Add API reference pages for dspy.configure and dspy.context

MaximeRivest:
Add hand-written API reference pages for `dspy.configure` and `dspy.context`. These are bound-method aliases that mkdocstrings/Griffe cannot resolve, so the pages are written manually.

**Changes:**
- New `docs/docs/api/utils/configure.md` with settings reference table, examples, and thread-safety docs
- New `docs/docs/api/utils/context.md` with examples covering LM swap, adapter swap, nesting, and concurrency
- Docstrings added to `Settings.configure` and `Settings.context` in `settings.py`
- Nav entries added to `mkdocs.yml`

### DIFF
--- a/docs/docs/api/utils/configure.md
+++ b/docs/docs/api/utils/configure.md
@@ -1,0 +1,95 @@
+# dspy.configure
+
+Set the default language model, adapter, and other settings for DSPy.
+
+```python
+dspy.configure(**kwargs)
+```
+
+Call `dspy.configure(...)` once near the top of your script or notebook.
+Every DSPy module will use these defaults unless you override them with
+[`dspy.context`](context.md). The values persist until you call
+`dspy.configure(...)` again.
+
+!!! note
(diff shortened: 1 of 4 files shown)


isaacbmiller:
Can you attach screenshots of what these new pages + the api reference look like?

Also can you note why you need to make it manually somewhere in the code.

MaximeRivest:
Yes!   for the note, what would be the job of that note (I am not sure I see it so I struggle guessing where to put it).

fyi: its because `dspy.configure` and `dspy.context` are aliases of the corresponding methods of the `settings` instance.

# Configure

image
image
image


# Context

image
image

isaacbmiller (approved):
A note somewhere in settings.py that says changes need to be reflected in the docs manually. 

If there is a way to automatically link the two then we should do that.

LGTM after note.

PR #9460, opened 2026-03-16
fix: pass num_retries to streaming completion path

### LINKED ISSUE #9459: [Bug] Streaming completion path skips num_retries / retry_strategy — no retry on rate limit errors
## Description

When using `dspy.streamify()` (or any configuration where `dspy.settings.send_stream` is set), all LLM calls go through `_get_stream_completion_fn` in `dspy/clients/lm.py`. This streaming path calls `litellm.acompletion(stream=True, ...)` **without** passing `num_retries` or `retry_strategy`, so rate limit errors (429) crash immediately instead of retrying with exponential backoff.

The non-streaming path correctly passes these parameters:

```python
# Non-streaming (line ~410 in alitellm_completion) — HAS retries ✅
return await litellm.acompletion(
    cache=cache,
    num_retries=num_retries,
    retry_strategy="exponential_backoff_retry",
    headers=headers,
    **request,
)
```

```python
# Streaming (_get_stream_completion_fn, line ~325) — NO retries ❌
response = await litellm.acompletion(
    cache=cache_kwargs,
    stream=True,
    headers=headers,
    **request,  # num_retries and retry_strategy are missing
)
```

## Reproduction

```python
import dspy

# Configure with retries
lm = dspy.LM("anthropic/claude-opus-4-6", num_retries=5)
dspy.configure(lm=lm)

class SimpleQA(dspy.Signature):

module = dspy.ChainOfThought(SimpleQA)

# Without streamify — retries work correctly
result = module(question="What is 2+2?")  # Will retry on 429

# With streamify — retries are skipped
streamed = dspy.streamify(module)
result = streamed(question="What is 2+2?")  # Crashes immediately on 429
```

To trigger the rate limit error, run many concurrent calls against an API with a low rate limit, or temporarily set a very low TPM limit.

## Impact

Any application using `dspy.streamify()` has **zero retry protection** against rate limit errors, even when `num_retries` is explicitly configured. This is particularly severe for:

- Production workloads using streaming for real-time output
- Batch processing with many concurrent LLM calls
- Any deployment hitting API rate limits

## Root Cause

`_get_stream_completion_fn` (line 305) does not accept or forward `num_retries`:

1. `litellm_completion()` and `alitellm_completion()` receive `num_retries` as a parameter
2. They call `_get_stream_completion_fn()` which does not accept `num_retries`
3. The closure returned by `_get_stream_completion_fn` calls `litellm.acompletion()` without retry params
4. The caller functions then call `stream_completion()` directly, bypassing the retry logic that exists in the non-streaming `else` branch

## Suggested Fix

Pass `num_retries` through `_get_stream_completion_fn` and apply it to the `litellm.acompletion` call, or wrap the streaming call with tenacity retry logic matching the non-streaming path:

```python
def _get_stream_completion_fn(
    request: dict[str, Any],
    cache_kwargs: dict[str, Any],
    num_retries: int = 0,  # Add this parameter
    sync=True,
    headers: dict[str, Any] | None = None,
):
    # ... existing code ...

    async def stream_completion(request, cache_kwargs):
        response = await litellm.acompletion(
            cache=cache_kwargs,
            stream=True,
            headers=headers,
            num_retries=num_retries,                    # Add this
            retry_strategy="exponential_backoff_retry",  # Add this
            **request,
        )
        # ... rest unchanged ...
```

And update callers:

```python
# In litellm_completion():
stream_completion = _get_stream_completion_fn(request, cache, num_retries=num_retries, sync=True, headers=headers)

# In alitellm_completion():
stream_completion = _get_stream_completion_fn(request, cache, num_retries=num_retries, sync=False)
```

**Note:** If `litellm.acompletion(stream=True)` doesn't honor `num_retries` internally (since the stream connection is already opened), the alternative is wrapping the entire `stream_completion` closure with `tenacity.retry` matching the same exponential backoff strategy.

## Environment

- DSPy version: 3.1.3 (also confirmed on latest `main` branch)
- Python: 3.13
- litellm: 1.x

saivedant169:
Fixes #9459

## Description

The streaming completion path via `_get_stream_completion_fn` called `litellm.acompletion(stream=True)` **without** `num_retries` or `retry_strategy`, so rate limit errors (429) crashed immediately instead of retrying with exponential backoff. The non-streaming path correctly passes both parameters.

## Changes

1. Added `num_retries` parameter to `_get_stream_completion_fn()` 
2. Pass `num_retries` and `retry_strategy="exponential_backoff_retry"` to `litellm.acompletion(stream=True)` inside the stream closure
3. Updated `litellm_completion()` caller to forward `num_retries`
4. Updated `alitellm_completion()` caller to forward `num_retries` and `headers` (headers were also previously missing from the streaming path inconsistent with the sync caller and the non-streaming path)

## Before vs After

```python
# BEFORE: streaming no retries, crashes on 429
response = await litellm.acompletion(
    cache=cache_kwargs,
    stream=True,
    headers=headers,
    **request,
)

# AFTER: streaming retries match non-streaming path
response = await litellm.acompletion(
    cache=cache_kwargs,
    stream=True,
    num_retries=num_retries,
    retry_strategy="exponential_backoff_retry",
    headers=headers,
    **request,
)
```

## Testing

`ruff check` and `ruff format` pass clean.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -160,12 +160,7 @@ class LM(BaseLM):
 
         return completion_fn, litellm_cache_args
 
-    def forward(
-        self,
-        prompt: str | None = None,
-        messages: list[dict[str, Any]] | None = None,
-        **kwargs
-    ):
+    def forward(self, prompt: str | None = None, messages: list[dict[str, Any]] | None = None, **kwargs):
         # Build the request.
         kwargs = dict(kwargs)
         cache = kwargs.pop("cache", self.cache)
@@ -345,6 +340,7 @@ class LM(BaseLM):
 def _get_stream_completion_fn(
     request: dict[str, Any],
     cache_kwargs: dict[str, Any],
+    num_retries: int = 0,
     sync=True,
     headers: dict[str, Any] | None = None,
 ):
@@ -365,6 +361,8 @@ def _get_stream_completion_fn(
         response = await _get_litellm().acompletion(
             cache=cache_kwargs,
             stream=True,
+            num_retries=num_retries,
+            retry_strategy="exponential_backoff_retry",
             headers=headers,
             **request,
         )
@@ -394,7 +392,7 @@ def litellm_completion(request: dict[str, Any], num_retries: int, cache: dict[st
     request = dict(request)
     request.pop("rollout_id", None)
     headers = _add_dspy_identifier_to_headers(request.pop("headers", None))
-    stream_completion = _get_stream_completion_fn(request, cache, sync=True, headers=headers)
+    stream_completion = _get_stream_completion_fn(request, cache, num_retries=num_retries, sync=True, headers=headers)
     if stream_completion is None:
         return _get_litellm().completion(
             cache=cache,
@@ -442,7 +440,9 @@ async def alitellm_completion(request: dict[str, Any], num_retries: int, cache: 
     request = dict(request)
     request.pop("rollout_id", None)
     headers = _add_dspy_identifier_to_headers(request.pop("headers", None))
-    stream_completion = _get_stream_completion_fn(request, cache, sync=False, headers=headers)
+    stream_completion = _get_stream_completion_fn(
+        request, cache, num_retries=num_retries, sync=False, headers=headers
+    )
     if stream_completion is None:
         return await _get_litellm().acompletion(
             cache=cache,
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -160,7 +160,12 @@ class LM(BaseLM):
 
         return completion_fn, litellm_cache_args
 
-    def forward(self, prompt: str | None = None, messages: list[dict[str, Any]] | None = None, **kwargs):
+    def forward(
+        self,
+        prompt: str | None = None,
+        messages: list[dict[str, Any]] | None = None,
+        **kwargs
+    ):
         # Build the request.
         kwargs = dict(kwargs)
         cache = kwargs.pop("cache", self.cache)
@@ -340,7 +345,7 @@ class LM(BaseLM):
 def _get_stream_completion_fn(
     request: dict[str, Any],
     cache_kwargs: dict[str, Any],
-    num_retries: int = 0,
+    num_retries: int,
     sync=True,
     headers: dict[str, Any] | None = None,
 ):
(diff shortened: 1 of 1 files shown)


dspy/clients/lm.py
```diff
@@ -305,6 +300,7 @@ def _check_truncation(self, results):
 def _get_stream_completion_fn(
     request: dict[str, Any],
     cache_kwargs: dict[str, Any],
+    num_retries: int = 0,
```
isaacbmiller: Let's have num_retries as a required function parameter

dspy/clients/lm.py
```diff
@@ -403,7 +401,9 @@ async def alitellm_completion(request: dict[str, Any], num_retries: int, cache:
     request = dict(request)
     request.pop("rollout_id", None)
     headers = request.pop("headers", None)
-    stream_completion = _get_stream_completion_fn(request, cache, sync=False)
+    stream_completion = _get_stream_completion_fn(
+        request, cache, num_retries=num_retries, sync=False, headers=_add_dspy_identifier_to_headers(headers)
```
isaacbmiller: Make the syntax should match lm:355

dspy/clients/lm.py
```diff
@@ -130,12 +130,7 @@ def _get_cached_completion_fn(self, completion_fn, cache):
 
         return completion_fn, litellm_cache_args
 
-    def forward(
-        self,
-        prompt: str | None = None,
-        messages: list[dict[str, Any]] | None = None,
-        **kwargs
-    ):
+    def forward(self, prompt: str | None = None, messages: list[dict[str, Any]] | None = None, **kwargs):
```
isaacbmiller: undo change unless necessary for ruff

isaacbmiller (changes requested):

saivedant169:
Hey @isaacbmiller , wanted to see if there's anything else you need from me on this. Happy to make changes if something's off.

saivedant169:
Hey @isaacbmiller, all three items from your review are in. num_retries is required, the async caller matches the sync style, and the formatting is reverted. Let me know if anything else needs changing.

saivedant169:
Rebased on top of current main. Conflicts in dspy/clients/lm.py were just my prior identifier-header changes overlapping with the new wrappers on main, resolved both sides cleanly so the streaming path still calls _add_dspy_identifier_to_headers and still passes num_retries through.

All three review items from earlier are still in: num_retries is required on every litellm_*_completion entrypoint, the async call matches the sync caller style, and the unrelated formatting is reverted.

@isaacbmiller please take another look when you have a moment.

saivedant169:
Hi @isaacbmiller, this one has been quiet for a while. All three review items are in: num_retries is now required, the async caller matches the sync caller style, and the formatting is reverted. I also rebased onto main on the 17th and resolved the lm.py conflicts. Anything else you need from me to move it forward?

PR #9458, opened 2026-03-16
docs: add Google-style docstrings to dspy/datasets/dataloader.py

saivedant169:
Resolves #9457
Part of #8926

## Description

Adds comprehensive Google-style docstrings with `Args`, `Returns`, `Raises`, and `Example` sections to all 9 public APIs in `DataLoader`:

- `DataLoader` class docstring
- `from_huggingface()` — includes split behavior (dict vs flat list)
- `from_csv()`
- `from_pandas()`
- `from_json()`
- `from_parquet()`
- `from_rm()`
- `sample()`
- `train_test_split()` — documents float vs int size semantics

Each docstring includes a runnable code example showing typical usage.

`ruff check` and `ruff format` pass clean.

PR #9463, opened 2026-03-16
docs: add docstrings to predict/refine.py and predict/best_of_n.py

### LINKED ISSUE #9462: Add docstrings to dspy/predict/refine.py and dspy/predict/best_of_n.py
Claiming `dspy/predict/refine.py` and `dspy/predict/best_of_n.py` for docstring additions, as described in #8926.

**Files:**
- `dspy/predict/refine.py` — Adding docstrings to `Refine.forward()`, `inspect_modules()`, and `recursive_mask()`
- `dspy/predict/best_of_n.py` — Adding docstring to `BestOfN.forward()`

All docstrings follow the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) with Args, Returns, and Raises sections.

PR incoming shortly.

aviraldua93:
Adds Google-style docstrings to public APIs in `dspy/predict/refine.py` and `dspy/predict/best_of_n.py` that were missing documentation.

Resolves #9462, ref #8926.

### `dspy/predict/refine.py`
- **`Refine.forward()`** — Documents the refinement loop: how rollout IDs work, the feedback mechanism via `OfferFeedback`, early stopping on threshold, and exception behavior.
- **`inspect_modules()`** — Documents this utility that generates human-readable summaries of a program's predictor modules.
- **`recursive_mask()`** — Documents this helper that replaces non-JSON-serializable objects with placeholder strings.

### `dspy/predict/best_of_n.py`
- **`BestOfN.forward()`** — Documents the best-of-N selection loop, its independence from feedback (unlike `Refine`), and exception behavior.

## Style

All docstrings follow the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) with:
- Summary line
- Detailed description where appropriate
- `Args`, `Returns`, and `Raises` sections
- Cross-references between `Refine` and `BestOfN` for discoverability

Pre-commit checks (ruff lint, merge conflicts, debug statements) all pass.

### DIFF
--- a/dspy/predict/best_of_n.py
+++ b/dspy/predict/best_of_n.py
@@ -69,11 +69,13 @@ class BestOfN(Module):
 
         Returns:
             Prediction: The best prediction found across all attempts, determined by the highest
-                reward score. Returns ``None`` if all attempts fail with exceptions.
+                reward score. Returns ``None`` if no attempt succeeded and the ``fail_count``
+                re-raise threshold was not reached (possible when ``N <= 2`` or
+                ``fail_count >= N``).
 
         Raises:
-            Exception: Re-raises the last encountered exception if the number of consecutive
-                failures exceeds ``fail_count``.
+            Exception: Re-raises the last encountered exception when the current attempt
+                index exceeds the remaining ``fail_count`` (decremented after each failure).
         """
         lm = self.module.get_lm() or dspy.settings.lm
         start = lm.kwargs.get("rollout_id", 0)
--- a/dspy/predict/best_of_n.py
+++ b/dspy/predict/best_of_n.py
@@ -53,6 +53,28 @@ class BestOfN(Module):
         self.fail_count = fail_count or N  # default to N if fail_count is not provided
 
     def forward(self, **kwargs):
+        """Execute the best-of-N selection loop, running the module up to ``N`` times.
+
+        Each iteration runs the module with a unique rollout ID at ``temperature=1.0`` and scores
+        the result using the reward function. If a prediction meets or exceeds the threshold, it is
+        returned immediately. Otherwise, the prediction with the highest reward across all attempts
+        is returned.
+
+        Unlike :class:`~dspy.Refine`, this method does not generate feedback between attempts —
+        each attempt is independent.
+
+        Args:
+            **kwargs: Keyword arguments passed directly to the underlying module's forward method.
+                These should match the input fields defined in the module's signature.
+
+        Returns:
+            Prediction: The best prediction found across all attempts, determined by the highest
+                reward score. Returns ``None`` if all attempts fail with exceptions.
+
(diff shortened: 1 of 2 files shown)


saivedant169 (commented):
Nice work — I just went through the same exercise for DataLoader (#9458) so I know the griffe rendering quirks.

The distinction between Refine and BestOfN in the forward docstrings is really helpful — calling out that BestOfN attempts are independent while Refine generates feedback between attempts is the key difference users need to understand. That's not obvious from reading the class names alone.

One small thing: both forward() docstrings say 'Returns None if all attempts fail with exceptions' — might be worth double-checking that behavior. Looking at the code, it seems like it returns the best prediction found before the exception threshold, not None. If fail_count is exceeded it re-raises, but if some attempts succeeded before that, those are tracked in the candidates list.

aviraldua93:
> Nice work — I just went through the same exercise for DataLoader (#9458) so I know the griffe rendering quirks.
> 
> The distinction between Refine and BestOfN in the forward docstrings is really helpful — calling out that BestOfN attempts are independent while Refine generates feedback between attempts is the key difference users need to understand. That's not obvious from reading the class names alone.
> 
> One small thing: both forward() docstrings say 'Returns None if all attempts fail with exceptions' — might be worth double-checking that behavior. Looking at the code, it seems like it returns the best prediction found before the exception threshold, not None. If fail_count is exceeded it re-raises, but if some attempts succeeded before that, those are tracked in the candidates list.



> Nice work — I just went through the same exercise for DataLoader (#9458) so I know the griffe rendering quirks.
> 
> The distinction between Refine and BestOfN in the forward docstrings is really helpful — calling out that BestOfN attempts are independent while Refine generates feedback between attempts is the key difference users need to understand. That's not obvious from reading the class names alone.
> 
> One small thing: both forward() docstrings say 'Returns None if all attempts fail with exceptions' — might be worth double-checking that behavior. Looking at the code, it seems like it returns the best prediction found before the exception threshold, not None. If fail_count is exceeded it re-raises, but if some attempts succeeded before that, those are tracked in the candidates list.

Thanks @saivedant169 Great catch, thank you! You're right — I traced through the code more carefully and the docstrings were inaccurate in two ways:
 
  1. "Returns None if all attempts fail" — the code actually re-raises when idx > self.fail_count. None is only returned in the narrow case where the 
threshold isn't reached (e.g. N <= 2 with default fail_count).
  2. "consecutive failures" — fail_count is decremented on every failure regardless of intervening successes, not just consecutive ones.
 
 I've pushed a fix correcting both the Returns and Raises sections in both files. Appreciate the thorough review!

aviraldua93:
Can this be merged? Any blockers? @saivedant169 / others?

saivedant169:
I guess the maintainers will review it and let you know.

PR #9470, opened 2026-03-18
feat(module): add get_prompt_templates() for exporting prompts

### LINKED ISSUE #7830: [Feature] Print the final prompt of a DSPy program

Hi DSPy team,

I am using DSPy in my project. I want to convert DSPy programs into regular user and system prompts, so that people who are not familiar with DSPy can run LLMs with regular user and system prompts instead.

As far as I know, the only way to get regular user and system prompts is to run the program at least once, then print the history.
However, the history will contain the actual input, and I need to clean up the actual input to get the generic prompts. That is not elegant.
Thus, I would like to request a function to print out the generic user and system prompts.

The following is an example:
## DSPy implementation
```
    __doc__ = """Answer questions with Yes/No at the begining.""" # instruction

    question = dspy.InputField(desc="The Yes/No question to be answered") # remaining attributes: format, prefix
    answer = dspy.OutputField(desc="Only 'Yes' or 'No'")
```
## Text obtained via inspect_history
```
System message:

Your input fields are:
1. `question` (str): The Yes/No question to be answered

Your output fields are:
1. `answer` (str): Only 'Yes' or 'No'

[[ ## question ## ]]
{question}

[[ ## answer ## ]]
{answer}

[[ ## completed ## ]]

        Answer questions with Yes/No at the begining.


User message:

[[ ## question ## ]]
Does the following poem compare the beach to something else at least twice?

Upon the shore where dreams were spun,
The ocean's breath a gentle sigh,
A seashell whispers of the sun,
Beneath the vast and endless sky.

Respond with the corresponding output fields, starting with the field `[[ ## answer ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.


Response:

[[ ## answer ## ]]
Yes

[[ ## completed ## ]]
```
## Equilavent prompts (what I want the requested function to return)
```
system_prompt = """Your input fields are:
1. `question` (str): The Yes/No question to be answered

Your output fields are:
1. `answer` (str): Only 'Yes' or 'No'

[[ ## question ## ]]
{question}

[[ ## answer ## ]]
{answer}

[[ ## completed ## ]]

        Answer questions with Yes/No at the begining.
"""

user_prompt = """[[ ## question ## ]]
{}

Respond with the corresponding output fields, starting with the field `[[ ## answer ## ]]`, and then ending with the marker for `[[ ## completed ## ]]`.
```
## Use the equilavent prompts without DSPy
```
new_question = '...'
openai.chat.completions.create(
    model=model_name,
    temperature=temperature,
    max_tokens=max_tokens,
    messages=[
        {"role": "system", "content": (system_prompt)},
        {"role": "user", "content": user_prompt.format(new_question)},
    ],
)
```

Best.

### Additional Context

_No response_

kvr06-ai:
## Summary

Adds a `get_prompt_templates()` method to `Module` that returns the formatted system/user message templates for each predictor, with placeholder values (e.g. `{question}`) instead of actual inputs. This lets users export DSPy programs as standard prompts usable without DSPy.

Implements the syntactic sugar @okhat proposed in #7830:

```python
templates = program.get_prompt_templates()
# => {"self": [{"role": "system", "content": "..."}, {"role": "user", "content": "..."}]}
```

Under the hood, it calls `adapter.format(signature, demos, inputs={k: f"{{{k}}}" for k in signature.input_fields})` for each predictor. Accepts an optional `adapter` argument; defaults to `dspy.settings.adapter` or `ChatAdapter()`.

Fixes #7830

## Changes

- `dspy/primitives/module.py`: Add `get_prompt_templates(adapter=None)` method to `Module`
- `tests/primitives/test_module.py`: 5 test cases covering single predictor, multi-predictor, nested modules, demos, and empty module

## Testing

- All 20 tests pass (15 existing + 5 new)
- `ruff check` and `ruff format` clean on changed files

### DIFF
--- a/dspy/primitives/module.py
+++ b/dspy/primitives/module.py
@@ -176,6 +176,44 @@ class Module(BaseModule, metaclass=ProgramMeta):
         """
         return [param for _, param in self.named_predictors()]
 
+    def get_prompt_templates(self, adapter=None):
+        """Get the formatted prompt templates for all predictors in this module.
+
+        Returns the system and user message templates that each predictor would
+        send to the LM, with placeholder values (e.g. ``{question}``) in place
+        of actual inputs. Useful for exporting DSPy programs as standard prompts
+        that can be used without DSPy.
+
+        Args:
+            adapter: The adapter to use for formatting. If ``None``, uses the
+                adapter from ``dspy.settings`` or falls back to ``ChatAdapter()``.
+
+        Returns:
+            dict[str, list[dict[str, str]]]: A dict mapping predictor names to
+                their formatted message lists. Each message is a dict with
+                ``"role"`` and ``"content"`` keys.
+
+        Examples:
+            >>> import dspy
+            >>> program = dspy.Predict("question -> answer")
+            >>> templates = program.get_prompt_templates()
+            >>> templates["self"]  # doctest: +SKIP
+            [{"role": "system", "content": "..."}, {"role": "user", "content": "..."}]
+        """
+        from dspy.adapters.chat_adapter import ChatAdapter
+
+        if adapter is None:
+            adapter = settings.adapter or ChatAdapter()
+
+        return {
+            name: adapter.format(
+                p.signature,
+                demos=p.demos,
+                inputs={k: f"{{{k}}}" for k in p.signature.input_fields},
+            )
+            for name, p in self.named_predictors()
+        }
+
     def set_lm(self, lm):
         """Set the language model for all predictors in this module.
 
@@ -329,8 +367,9 @@ class Module(BaseModule, metaclass=ProgramMeta):
         if prediction_in_output:
             prediction_in_output.set_lm_usage(tokens)
         else:
-            logger.warning("Failed to set LM usage. Please return `dspy.Prediction` object from dspy.Module to enable usage tracking.")
-
+            logger.warning(
+                "Failed to set LM usage. Please return `dspy.Prediction` object from dspy.Module to enable usage tracking."
+            )
 
     def __getattribute__(self, name):
         attr = super().__getattribute__(name)
--- a/tests/primitives/test_module.py
+++ b/tests/primitives/test_module.py
@@ -197,3 +197,64 @@ def test_load_dspy_program_cross_version():
 
     assert len(loaded_react.react.demos) == 2
     assert len(loaded_react.extract.predict.demos) == 2
+
+
+def test_get_prompt_templates_single_predictor():
+    program = dspy.Predict("question -> answer")
+    templates = program.get_prompt_templates()
+
+    assert "self" in templates
+    messages = templates["self"]
+    assert len(messages) >= 2
+    assert messages[0]["role"] == "system"
+    assert messages[-1]["role"] == "user"
+    # Placeholder should appear in the user message
+    assert "{question}" in messages[-1]["content"]
+
+
+def test_get_prompt_templates_multi_predictor():
+    module = HopModule()
+    templates = module.get_prompt_templates()
+
+    assert len(templates) == 2
+    assert "predict1" in templates
+    assert "predict2" in templates
+    # Each should have system + user messages at minimum
+    for _name, messages in templates.items():
+        assert len(messages) >= 2
+        assert messages[0]["role"] == "system"
+
+
+def test_get_prompt_templates_nested_module():
+    class Hop2Module(dspy.Module):
+        def __init__(self):
+            super().__init__()
+            self.hop = HopModule()
+
+    module = Hop2Module()
+    templates = module.get_prompt_templates()
+
+    assert len(templates) == 2
+    assert "hop.predict1" in templates
+    assert "hop.predict2" in templates
+
+
+def test_get_prompt_templates_with_demos():
+    program = dspy.Predict("question -> answer")
+    program.demos = [
+        {"question": "Is the sky blue?", "answer": "Yes"},
+    ]
+    templates = program.get_prompt_templates()
+
+    messages = templates["self"]
+    # With one demo, expect: system, demo user, demo assistant, user
+    assert len(messages) >= 4
+    demo_content = " ".join(m["content"] for m in messages if m["role"] == "assistant")
+    assert "Yes" in demo_content
+
+
+def test_get_prompt_templates_empty_module():
+    module = Module()
+    templates = module.get_prompt_templates()
+
+    assert templates == {}


kvr06-ai:
@okhat This implements the `get_prompt_templates()` helper you suggested. Would you mind taking a look when you get a chance?

MaximeRivest:
this is promising!!!

do you mind humouring me with some trade-offs around the design of the placeholder and some alternatives we could consider?

> with placeholder values (e.g. {question}

my biggest worry here is that some adapter shows that vary same {question} to the model, to explain to the model the pattern to come. I fear both can be confused/entangled.

kvr06-ai:
Good catch @MaximeRivest — that's a real collision. The `ChatAdapter` system message already includes `{question}` literally to show the model the expected format:

```
[[ ## question ## ]]
{question}
```

If we also pass `{question}` as the placeholder value, the user message ends up with the same string, and there's no way to distinguish "this is the adapter explaining structure to the model" from "this is where you fill in your actual input."

Three options I see:

**A. Distinct default placeholder (my preference)**

Default to `${question}` or `<<question>>`. Visually distinct from the adapter's `{question}`, zero ambiguity for the user reading the export. Minimal code change — just the format string in the dict comprehension.

**B. Configurable placeholder factory**

Accept an optional `placeholder_fn: Callable[[str], str]` parameter (defaulting to something like `lambda k: f"${{{k}}}"` ). Power users can define their own format; casual users get a safe default.

**C. Structured return with metadata**

Instead of raw message strings, return a small object per predictor that separates the static prompt from the input slot positions. More robust, but heavier API surface for a convenience method.

I'd lean toward **A** for simplicity (this is syntactic sugar, not a core API), with the door open to **B** if users need customization. Happy to push a revision once we agree on direction.

isaacbmiller:
This was added in #9006 

```python
import dspy

signature = dspy.Signature("question -> answer")
system_message = dspy.ChatAdapter().format_system_message(signature)
print(system_message)
```
The output should resemble:

```
Your input fields are:
1. `question` (str):
Your output fields are:
1. `answer` (str):

[[ ## question ## ]]
{question}
[[ ## answer ## ]]
{answer}
[[ ## completed ## ]]
```

MaximeRivest:
@kvr06-ai is this solution satisfying your need? if not why, if yes. I presume you missed because of a documentation issue?

kvr06-ai:
`#9006` added `Adapter.format_system_message()`, which covers extracting the system prompt for a single signature. `#9470` addresses what the original issue still needs:

1. **User message templates with input placeholders.** The core ask in `#7830` is exporting prompts you can plug directly into `openai.chat.completions.create()`. That requires the user message with `{question}` slots, not just the system message.

2. **Module-level export.** `format_system_message()` operates on a single signature. `get_prompt_templates()` walks an entire module (including nested submodules) and returns templates keyed by predictor name.

3. **Single-call convenience.** One call returns the full `[system, user]` message list per predictor, ready for direct use:

```python
templates = program.get_prompt_templates()

openai.chat.completions.create(
    model="gpt-4",
    messages=[
        templates["self"][0],
        {"role": "user", "content": templates["self"][1]["content"].format(question="my question")},
    ],
)
```

Happy to reference `format_system_message()` in the docstring for users who only need the system prompt.

@isaacbmiller — could you reopen this PR? The scope here is additive to `#9006`, not duplicative.

Ar-temis:
Just an add on: I would like to get the token count before sending the final message/prompt to the LLM so this could be super useful.

MaximeRivest:
@kvr06-ai a ping to notify you that Isaac reopened it.

On the design of the placeholder syntax, did you consider Jinja? I feel that a good feature of the output would be that it is easy to bring to more low level sdks but not just openai completion standard, we might want to name it as such and document that clearly.

I see there is no tests on the prompts that would take dspy.Image, dspy.Tool, etc. I am iffy on some of the details, but my intuitions is that some of those would be very hard to make into the template response string because they act elsewhere in the openai completion sdk. I worry about does not working being frustrating to users. What do you think?

Did you mock up what would that flow from a dspy program that has 3 predictor to move to some none completion sdks like: openai responses, anthropic and google genai sdk?

While I think its good we start simple, I feel it could be almost more dissappointing to have the feature without covering 80% of use cases.

isaacbmiller:
Thanks for pinging @kvr06-ai and @MaximeRivest. Exporting both user and system prompts rather than just the system prompt also makes sense here, especially for a single Predictor

I don't understand the value of exporting multiple prompts at once. 

My main concern is what Omar said:
> P.S. You obviously lose the actual control flow logic this way

If you have optimized a program with multiple sequential modules, needing to recreate the control flow on the other end opens you up to brittle pipelines and building your own prompt loaders, where the .json format is what you should use here. I would consider exporting the prompts here an anti-pattern, although I can imagine setups where its not bad.

I 100% think it is an anti-pattern to extract the prompts from a ReAct agent, as then you need to rebuild all of your tool calling + looping logic elsewhere.

My proposed solution here would be to move this method to a Predictor rather than a Module, but I am open to other uses.

kvr06-ai:
@isaacbmiller @MaximeRivest @okhat — addressing both threads here so we can converge on direction before I push a revision.

**1. Predictor-level API (isaacbmiller's concern)**

Agreed. The primary method belongs on `Predictor`, not `Module`. Exporting a multi-step pipeline without control flow is misleading for anything beyond simple chains, and putting it on `Predictor` makes the unit of export explicit.

Proposed structure:
- `Predictor.get_prompt_template(adapter=None)` → returns `list[dict]` (the message list for that single predictor)
- `Module.get_prompt_templates(adapter=None)` → thin wrapper that calls the above per `named_predictors()`, returns `dict[str, list[dict]]`

The Module-level method stays because it's what @okhat originally proposed and it handles the common case (simple single-predictor modules) in one call. But the docstring will carry an explicit warning that control flow, tool-calling logic, and adapter retries are not captured.

**2. Placeholder syntax (MaximeRivest's collision concern)**

Switching to Jinja-style `{{ field_name }}` as default. Visually distinct from the adapter's literal `{field_name}` in the system message, widely recognized, and portable across template engines.

Will accept an optional `placeholder_fn: Callable[[str], str]` for users who need a different format (e.g., Mustache, f-string). Default: `lambda k: "{{ " + k + " }}"`.

**3. Image/Tool/multimodal fields (MaximeRivest's scope concern)**

Scoping these out of this PR. Text-field signatures are the 80% case and what the original issue requests. For `dspy.Image` or `dspy.Tool` fields, the method will raise `NotImplementedError` with a clear message pointing to the tracking issue. This avoids shipping a half-working multimodal export that frustrates users more than having no export at all.

Happy to open a follow-up issue to track multimodal support.

**4. SDK mock-ups (MaximeRivest's request)**

Single predictor, `question -> answer`, with one demo:

```python
tmpl = predictor.get_prompt_template()
# tmpl == [{"role": "system", "content": "..."}, {"role": "user", "content": "...{{ question }}..."}]
```

**OpenAI:**
```python
from openai import OpenAI
msgs = [tmpl[0], {**tmpl[1], "content": tmpl[1]["content"].replace("{{ question }}", my_question)}]
OpenAI().chat.completions.create(model="gpt-4", messages=msgs)
```

**Anthropic:**
```python
import anthropic
anthropic.Anthropic().messages.create(
    model="claude-sonnet-4-20250514",
    system=tmpl[0]["content"],
    messages=[{**tmpl[1], "content": tmpl[1]["content"].replace("{{ question }}", my_question)}],
)
```

**Google GenAI:**
```python
from google import genai
client = genai.Client()
client.models.generate_content(
    model="gemini-2.0-flash",
    config={"system_instruction": tmpl[0]["content"]},
    contents=tmpl[1]["content"].replace("{{ question }}", my_question),
)
```

The messages format (system + user roles) maps directly to all three. No SDK-specific logic needed.

---

If this direction looks right, I'll push the revision. 👍/👎?

Issue #8283, opened 2025-05-26
[Feature] SemanticF1: Consider decomposing LLM scoring from numerical P/R calculation


komikat:
### What feature would you like to see?

Hi DSPy team,

While exploring the SemanticF1 module I noticed that it asks LLMs to directly output `recall: float` and `precision: float` values. While LLMs excel at semantic understanding, they may be less reliable at generating precise, calibrated numerical scores.

**Current approach:**
- LLM directly outputs precision/recall floats (0-1)
- Potential for subjective or "impressionistic" scoring

**Suggestion:**
Consider a more decomposed approach where:
1. LLM extracts and matches key ideas/claims from both texts
2. A separate step (algorithmic or constrained LLM call) calculates P/R from these mappings

This could provide more fine-grained, reliable, and interpretable semantic evaluation.

Would love to hear the team's thoughts on the design philosophy here and whether there's interest in exploring alternative semantic evaluation approaches.

Thanks for the great work!

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

okhat:
Yes that is fair!

Anuj-G-06:
Hi @okhat! I've submitted a PR for this: #9489

The implementation adds `decompositional="computed"` to `SemanticF1` — the LLM enumerates key ideas and identifies which are covered, then P/R are computed programmatically from the counts. This avoids LLM float-calibration issues while keeping the existing modes fully backward compatible.

Happy to incorporate any feedback!

PR #9489, opened 2026-03-22
feat(evaluate): add computed mode to SemanticF1 for programmatic P/R

Anuj-G-06:
## Summary

Adds a `decompositional="computed"` option to `SemanticF1` that **separates LLM semantic judgment from numerical score computation**, addressing the concern raised in #8283 (approved by @okhat).

### Problem

Current `SemanticF1` asks the LLM to directly output `precision: float` and `recall: float`. LLMs are unreliable at generating calibrated numerical scores — they excel at semantic understanding but not at precise float estimation.

### Solution

New `"computed"` mode that:
1. **LLM does what it's good at** — enumerates key ideas in both responses, then identifies which ideas are covered
2. **Code does what it's good at** — computes P/R programmatically: `recall = covered_gt_ideas / total_gt_ideas`, `precision = grounded_sr_ideas / total_sr_ideas`

### Usage

```python
# New computed mode — programmatic P/R from LLM idea matching
metric = SemanticF1(decompositional="computed")

# Existing modes — fully backward compatible
metric = SemanticF1()                        # LLM outputs P/R directly
metric = SemanticF1(decompositional=True)    # LLM enumerates ideas, then outputs P/R
```

### Changes

- **`dspy/evaluate/auto_evaluation.py`**:
  - New `ComputedSemanticRecallPrecision` signature — LLM enumerates numbered key ideas and lists which are covered
  - `_count_numbered_items()` — counts items in a numbered list
  - `_count_mentioned_numbers()` — counts distinct numbers in a comma-separated string
  - `SemanticF1` updated to accept `decompositional="computed"` with programmatic P/R computation

- **`tests/evaluate/test_auto_evaluation.py`**:
  - Tests for parsing utilities (`_count_numbered_items`, `_count_mentioned_numbers`)
  - Computed mode: correct F1 calculation, trace/threshold behavior, zero-match edge case
  - Backward compatibility: existing `False` and `True` modes unchanged

### Design Decisions

- **Additive, not breaking** — new mode is opt-in via `decompositional="computed"`, existing behavior unchanged
- **Single LLM call** — follows the existing decompositional pattern (enumerate + match in one call)
- **Numbered lists + comma-separated numbers** — simple, robust parsing; matches how LLMs naturally enumerate
- **`set()` for dedup** — if LLM repeats a number, it's counted once

Ref #8283

## Test plan

- [x] `ruff check` — all checks passed
- [x] `ruff format --check` — already formatted
- [x] Python syntax validation — both files pass `ast.parse()`
- [x] Unit tests for parsing utilities (numbered lists, comma-separated numbers, edge cases)
- [x] Unit tests for computed mode (correct F1, trace mode, no-match case)
- [x] Backward compatibility tests (decompositional=False and True unchanged)

### DIFF
--- a/dspy/evaluate/auto_evaluation.py
+++ b/dspy/evaluate/auto_evaluation.py
@@ -1,5 +1,3 @@
-import re
-
 from dspy.predict.chain_of_thought import ChainOfThought
 from dspy.primitives import Module
 from dspy.primitives.prediction import Prediction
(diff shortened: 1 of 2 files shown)


themavik (commented):
The computed mode is a good addition — programmatic P/R avoids the LLM float-calibration drift you describe.

Two edge cases with the regex approach: `_count_numbered_items` matches `\d+[.)]\s` which could false-positive on text like "In section 3. we discuss..." inside a key idea's body. And `_count_mentioned_numbers` returns distinct count, so "covered: 1, 2, 3" and "covered: 1 out of 2, plus 3" both return 3 despite different semantics. Probably fine for well-behaved LLM output, but worth noting.

Anuj-G-06:
Hey, saw this got closed with the "mass coding agent" label. I want to be upfront: I did use AI assistance to write these PRs, but I personally verified every change, ran the tests, and made sure the implementation actually makes sense.

This one implements the feature discussed in #8283 (approved by @okhat). The core idea is solid: let code handle the P/R math instead of asking the LLM to output floats, since LLMs are bad at numerical calibration. I verified the logic, tested the parsing utilities, and made sure backward compatibility holds.

I get that the repo is probably drowning in low effort bot PRs and submitting three at once didn't help. But there's a difference between spam and using AI as a tool while actually understanding and verifying what you're submitting. Would you be open to re-evaluating this one on its merits?

isaacbmiller:
Thanks for the responses @Anuj-G-06.

This and your other PRs were closed due to opening 3 at once, which is common for the Claw style PR bots.

isaacbmiller:
Thanks for the PR and clarification. Appreciate you verifying the changes and happy to take a look.

This PR has two main issues:
1. Signature does not take advantage of types: Any time you end up writing string parsing logic, it is a good sign that there is something wrong with the signature. This could be implemented by asking the LLM for a list of strings and returning a list of indexes (ints) which are covered in the example
2. Decomposed as a text argument into what is now a Bool argument doesnt make sense. We would prefer to keep decompositional argument a boolean, and instead add another boolean.

Anuj-G-06:
Haha fair enough, lesson learned the hard way. Thanks for reopening it, appreciate it! I'll go one at a time from now on.

Anuj-G-06:
Hey, pushed the changes based on your feedback:

1. Signature now uses typed fields — `list[str]` for key ideas and `list[int]` for covered indices. No more string parsing or regex, just `len()` on the lists DSPy already handles.

2. Kept `decompositional` as a bool, added a separate `computed` param. So it's `SemanticF1(computed=True)` now instead of the string arg. Also added a ValueError if someone passes both.

Let me know if anything else needs work.

Anuj-G-06:
Any other feedback on the updated approach, or is this looking good to merge?

PR #9495, opened 2026-03-24
[docs] Remove misleading confidence field from Classification example

xieyj17:
The `confidence: float` output field in the Classification example is not a calibrated probability — the LLM simply generates a plausible number. Remove it to avoid misleading newcomers into thinking DSPy produces reliable confidence scores out of the box.

The motivation is that we recently apply DSPy in building sentiment judges in production, and stakeholders want out of the box confidence scores generated from LLM. We need to explain to them that the confidence score obtained directly from LLM is no better than hallucitaion. 

Follow-up: I plan a separate PR to add a principled confidence measure based on consistency across repeated runs.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -199,7 +199,7 @@ DSPy shifts your focus from tinkering with prompt strings to **programming with 
         ```text
         Prediction(
             sentiment='positive',
-            toxicity=0.75
+            toxicity=0.15
         )
         ```
 


isaacbmiller:
Thanks for the PR!

I agree that uncalibrated confidence float is not much better than hallucination. Sometimes discrete categorical confidence can work okay uncalibrated, but you generally do want to optimize. 

I do like this as an example of multiple differently typed output fields. Something like `toxicity: float` could work ?

xieyj17:
Thanks for the comment and yes that makes sense! Will change from
confidence to toxicity later and tag you to review.

xieyj17:
I have added the `toxicity` back. Could you please take another look? Thanks! @isaacbmiller

PR #9509, opened 2026-03-25
Pin CI actions to SHA and replace curl-pipe-bash Deno install

TomeHirata:
## Summary
- Pin all GitHub Actions across all 4 workflows to full commit SHAs to prevent tag-hijacking supply chain attacks (ref: [litellm#24512](https://github.com/BerriAI/litellm/issues/24512))
- Replace `curl -fsSL https://deno.land/install.sh | sh` with the official `denoland/setup-deno` action (also SHA-pinned) to eliminate arbitrary code execution from a remote URL in CI

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install Deno
         uses: denoland/setup-deno@e95548e56dfa95d4e1a28d6f422fafe75c4c26fb # v2.0.3
         with:
-          deno-version: v2.x
+          deno-version: v2.7.7
       - name: Get correct version for TestPyPI release
         id: check_version
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-push.yml
```diff
@@ -16,9 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
       - name: Set up Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@3235b876344d2a9aa001b8d1453c930bba69e210 # v3
```
michaelisaac-dev, line 21: It seems like this hash doesn't exist. Could you please check? 

https://github.com/actions/setup-node/commit/3235b876344d2a9aa001b8d1453c930bba69e210

@isaacbmiller fyi

michaelisaac-dev: I think it might be a typo, should've been 3235b876344d2a9aa001b8d1453c930bba69e**6**10

Issue #9514, opened 2026-03-25
[Epic] Decouple core framework from dspy.LM to restore BaseLM modularity


MaximeRivest:
### Context & Problem Statement
DSPy originally promised a modular architecture where users could easily bring their own custom LLM backends by subclassing `BaseLM`. However, over time, the concrete `dspy.LM` class (which mostly acts as a wrapper for `LiteLLM`) has seeped into the core framework. 

Currently, `dspy.LM` is hardcoded in places where `dspy.BaseLM` should be the standard interface (e.g., adapters, callbacks, and test utilities). This architectural erosion causes several issues:
1. **Tight Coupling & Leaky Abstractions:** Core framework components now implicitly rely on `LM`-specific behaviors rather than the abstract `BaseLM` contract, violating the Dependency Inversion Principle.
2. **Forced Dependencies:** It is impossible to import parts of DSPy without installing `litellm`, litellm makes DSPy one of the slowest library to import in python, complicating the usage of DSPy in production on things like lambda functions, etc. 
3. **Performance Overhead:** The global reliance on `litellm` adds an unnecessary ~5 seconds to the `dspy` import time for users who may want to use a completely different, lightweight backend.

### Proposed Solution
This epic tracks a series of refactoring PRs aimed at untangling `dspy.LM` from the core framework and restoring `BaseLM` as the primary interface. The goal is to make `litellm` a convenient, fallback, optional backend rather than a structural pillar of the codebase.

### Impact
By completing this refactor, DSPy will regain its backend-agnostic modularity, enforce cleaner boundaries between components, and improve start-up performance for custom backend users.

PR #9515, opened 2026-03-25
Refactor: Make `DummyLM` and callbacks depend on `BaseLM` instead of `LM`

MaximeRivest:

**What this PR does**
This PR takes the first step in decoupling the core framework from `litellm` by targeting test utilities and callback dispatchers:
* **`utils/dummies.py`**: Changes `DummyLM` to subclass `BaseLM` instead of the concrete `dspy.LM`.
* **`utils/callback.py`**: Updates `_get_on_start_handler` and `_get_on_end_handler` to check for `isinstance(instance, dspy.BaseLM)` rather than `dspy.LM`.

**Why this is needed**
* `DummyLM` completely overrides `forward()` and never uses `litellm` under the hood. Subclassing `dspy.LM` introduced unnecessary tight coupling to a specific backend.
* Previously, if a user created a custom backend by subclassing `BaseLM`, it would fail to trigger callbacks because the dispatcher was hardcoded to look for `dspy.LM`. This change restores the proper `BaseLM` contract, ensuring callbacks work for *any* valid backend.

### DIFF
--- a/dspy/utils/callback.py
+++ b/dspy/utils/callback.py
@@ -350,7 +350,7 @@ def with_callbacks(fn):
 
 def _get_on_start_handler(callback: BaseCallback, instance: Any, fn: Callable) -> Callable:
     """Selects the appropriate on_start handler of the callback based on the instance and function name."""
-    if isinstance(instance, dspy.LM):
+    if isinstance(instance, dspy.BaseLM):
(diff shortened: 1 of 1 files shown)


isaacbmiller (approved):
🔥 nice!

PR #9516, opened 2026-03-26
Refactor: Move model capability checks to BaseLM to remove litellm from adapters

MaximeRivest:

**What this PR does**
This PR continues the decoupling effort by completely removing `litellm` imports from the `adapters` and `types` modules.
* **`clients/base_lm.py`**: Introduces base capability properties (`supports_function_calling`, `supports_reasoning`, `supports_response_schema`, `supported_params`) with safe defaults (`False` / empty set).
* **`clients/lm.py`**: Overrides these properties in the concrete `dspy.LM` class to actually perform the `litellm` capability checks.
* **`adapters/*` & `types/*`**: Replaces hardcoded `litellm` function calls with polymorphic property checks on the `lm` instance (e.g., `lm.supports_reasoning`). Moves remaining `litellm` type hints under `TYPE_CHECKING`.

**Why this is needed**
* **Separation of Concerns:** Adapters should not need to know *how* to check if a model supports a feature (which requires knowing about `litellm`); they should only ask the LM if it can do it.
* **Custom Backend Support:** Users creating custom `BaseLM` subclasses can now simply override these properties to explicitly declare what their custom backend supports, entirely bypassing `litellm`.
* **Performance:** Prevents the heavy `litellm` library from being imported in the adapter layer, further isolating the dependency.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,6 +1,5 @@
 import datetime
 import uuid
-from functools import cached_property
 from typing import Any, TextIO
 
 from dspy.dsp.utils import settings
@@ -28,6 +27,24 @@ class BaseLM:
 
 
     class MyLM(dspy.BaseLM):
+        @property
+        def supports_function_calling(self) -> bool:
+            return self.model.startswith("openai/gpt-4o")
+
+        @property
+        def supports_reasoning(self) -> bool:
+            return self.model.startswith("anthropic/claude-3-7")
+
+        @property
+        def supports_response_schema(self) -> bool:
+            return self.model.startswith("openai/gpt-4o")
+
+        @property
+        def supported_params(self) -> set[str]:
+            if self.model.startswith("openai/gpt-4o"):
+                return {"response_format"}  # accepts response_format=...
+            return set()
+
         def forward(self, prompt, messages=None, **kwargs):
             client = OpenAI()
             return client.chat.completions.create(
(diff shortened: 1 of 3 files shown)


dspy/clients/base_lm.py
```diff
@@ -49,6 +50,26 @@ def __init__(self, model, model_type="chat", temperature=0.0, max_tokens=1000, c
         self.kwargs = dict(temperature=temperature, max_tokens=max_tokens, **kwargs)
         self.history = []
 
+    @cached_property
```
isaacbmiller: 1. Why are these cached properties?
2. We can't raise a `NotImplemented` because people who built their own BaseLMs need backwards compatibility, but do we want to silently return `False`?

dspy/adapters/base.py
```diff
@@ -84,15 +83,13 @@ def _call_preprocess(
...
 
-            if tool_call_output_field_name and litellm.supports_function_calling(model=lm.model):
+            if tool_call_output_field_name and lm.supports_function_calling:
                 tools = inputs[tool_call_input_field_name]
                 tools = tools if isinstance(tools, list) else [tools]
 
-                litellm_tools = []
-                for tool in tools:
-                    litellm_tools.append(tool.format_as_litellm_function_call())
+                lm_tools = [tool.format_as_litellm_function_call() for tool in tools]
```
isaacbmiller, line 90: tools are still tightly coupled to LiteLLM function calls.

How do you think we should approach this? Should the LM decide how to format dspy.Tools?

MaximeRivest: for this PR, I suggest this is fine because its not actually the litellm format but rather the 'common' openai caht completion format.

**for longer term**:

This is the beginning of the "bigger" question. To make LiteLLM lazy-load throughout and never load for users with their own custom subclass of BaseLM, we need BaseLM to be a complete abstraction of what adapters, modules, and utilities in DSPy expect.

LiteLLM does it by mapping everything to OpenAI's chat completion format, and sometimes adds to it (makes it a superset) when OpenAI did not support it themselves (that's the case for streaming reasoning and citations).

In the above case, I am almost certain that we named it "LiteLLM function call" but it's actually "chat completion OpenAI function call."

We can make BaseLM expect to work with the chat completion OpenAI format, or the LiteLLM superset, or our own protocol — and custom implementations of BaseLM do the mapping to/from those.
Currently, only the context-exceeded warning and the stream chunks "leaked" out of the LM/BaseLM abstraction. I am studying the streamification now.

If we like one of these designs, we can follow them (without necessarily depending on them) and it makes it so that the DSPy community can easily port existing/tested implementations when they want their own custom BaseLM. And we can choose to support the four main ones, namely: OpenAI Responses, Anthropic, Google, and OpenAI Chat Completions.

image

PR #9521, opened 2026-03-26
Refactor: Introduce DSPy-owned `ContextWindowExceededError` to decouple error handling

MaximeRivest:

**Context & Motivation**
DSPy's core modules (like `ReAct` and some adapters) rely on catching `litellm.ContextWindowExceededError` to trigger fallback logic. If a user creates a custom `BaseLM` backend, they are forced to import and raise a `litellm` error, leaking litellm implementation details.

**What this PR does**
This PR creates a general abstraction for that exception:
* **`utils/exceptions.py`**: add `dspy.ContextWindowExceededError`.
* **`clients/base_lm.py`**: Documents the contract for custom backends to raise this error.
* **`clients/lm.py`**: Catches the `litellm` error and re-raises the new exception (preserving the stack trace).
* **Core & Tests**: Replaces `litellm` error imports with the new exception and adds dedicated unit tests for the exception formatting (not sure if 4 tests is overkill?).

**Impact**
Custom backends can now interface with DSPy's truncation and retry logic without `litellm`.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -113,9 +113,18 @@ class BaseLM:
         """Forward pass for the language model.
 
         Subclasses must implement this method, and the response should be identical to either of the following formats:
+
         - [OpenAI response format](https://platform.openai.com/docs/api-reference/responses/object)
         - [OpenAI chat completion format](https://platform.openai.com/docs/api-reference/chat/object)
         - [OpenAI text completion format](https://platform.openai.com/docs/api-reference/completions/object)
+
+        Raises:
+            dspy.ContextWindowExceededError: When the request fails because the
+                input exceeds the model's context window. DSPy adapters and
+                modules rely on this error to trigger fallback behavior (e.g.
+                truncating the prompt and retrying). Each subclass is
+                responsible for catching its provider's native error and
+                re-raising it as `dspy.ContextWindowExceededError`.
         """
         raise NotImplementedError("Subclasses must implement this method.")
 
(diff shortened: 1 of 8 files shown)


dspy/utils/exceptions.py
```diff
@@ -2,6 +2,26 @@
...
+    retry is appropriate.
+
+    Args:
+        model: The model identifier that rejected the request.
+        message: Optional description. Defaults to `"Context window exceeded"`.
+    """
+
+    def __init__(self, *, model: str | None = None, message: str | None = None):
+        self.model = model
+        msg = message or "Context window exceeded"
```
isaacbmiller: This should just be a default value for the parameter

dspy/clients/base_lm.py
```diff
@@ -113,9 +113,18 @@ def forward(
         """Forward pass for the language model.
 
         Subclasses must implement this method, and the response should be identical to either of the following formats:
+
         - [OpenAI response format](https://platform.openai.com/docs/api-reference/responses/object)
         - [OpenAI chat completion format](https://platform.openai.com/docs/api-reference/chat/object)
         - [OpenAI text completion format](https://platform.openai.com/docs/api-reference/completions/object)
+
+        Raises:
+            dspy.ContextWindowExceededError: When the request fails because the
```
isaacbmiller, line 122: Is there anywhere else we should document the requirement for BaseLM Subclasses to wrap their `ContextWindowExceededError`s?

isaacbmiller (changes requested):
LGTM + parameter Nit

dspy/clients/base_lm.py
MaximeRivest: `BaseLM` is only mentionned there in the docs, and its not in the api docs. We should change that when we are done fixing the leaky dspy.'litelLM' abstraction.

image

PR #9522, opened 2026-03-26
refactor: type-hint adapters with BaseLM instead of LM

MaximeRivest:

This PR is made 'easier' by #9516 which abstract the capability checks properly into `BaseLM`.

**What this PR does**
Continues decoupling effort by changing types hints from `LM` to `BaseLM`:
* **Type Hints:** Updates method signatures (`_call_`, `acall`, `_call_preprocess`, etc.) for `BaseAdapter`, `ChatAdapter`, `JSONAdapter`, and custom `Type` handlers to expect `BaseLM` instead of `LM`.
* **Runtime Validation:** Fixes `TwoStepAdapter.__init__`, which was explicitly enforcing `isinstance(..., LM)`. It now correctly validates against `BaseLM` and raises an accurate `ValueError`. Now any custom backend can use `TwoStepAdapter`.
* **Clean Imports:** Removes legacy `TYPE_CHECKING` blocks that imported `LM` and fixes a potential circular import in `utils_finetune.py`. I felt base lm should have precedence of that util.

**Why this is needed**
Adapters handle the translation between DSPy signatures and model-specific formats. They do not require any of the `litellm`-specific logic found in `dspy.LM`; they only need the contract guaranteed by `BaseLM`. Previously, if a user passed a custom backend subclassing `BaseLM` to `TwoStepAdapter`, the hardcoded `isinstance` check would immediately crash. This PR fixes that bug and aligns static type checking across the entire adapter layer to fully support modular custom backends.

### DIFF
--- a/dspy/clients/utils_finetune.py
+++ b/dspy/clients/utils_finetune.py
@@ -1,12 +1,14 @@
 import os
 from enum import Enum
-from typing import Any, Literal, TypedDict
+from typing import TYPE_CHECKING, Any, Literal, TypedDict
 
 import orjson
 
 import dspy
-from dspy.adapters.base import Adapter
 from dspy.utils.caching import DSPY_CACHEDIR
+
+if TYPE_CHECKING:
+    from dspy.adapters.base import Adapter
 
 
 class TrainingStatus(str, Enum):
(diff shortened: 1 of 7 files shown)


dspy/clients/utils_finetune.py
```diff
@@ -53,7 +52,7 @@ class GRPOStatus(TypedDict):
     pending_batch_ids: list[int] = []
 
 
-def infer_data_format(adapter: Adapter) -> str:
+def infer_data_format(adapter: "dspy.Adapter") -> str:
```
isaacbmiller: nit: should be imported from type checking

isaacbmiller (changes requested):
nit but lgtm.

only merge after #9516

PR #9538, opened 2026-03-27
Refactor: Make `litellm`, lazy-loaded dependency

MaximeRivest:
> [!IMPORTANT]
> This does NOT make litellm an optional dependency. It could be the next pr, with the appropriate version bump, release note and warning to install litellm. Or, we could wait that we have other `BaseLM` subclass before doing that change. I am agnostic. 

**Related Epic:** Relates to #9514 (Phase 4)

**What this PR does**
This PR completely isolates `litellm` from DSPy's core import path, it is only loaded if a user explicitly instantiates `dspy.LM`, `dspy.Embedder`, or uses `litellm`-backed streaming.
* **Lazy Loading via `__getattr__`**: Updates `dspy/__init__.py`, `dspy/clients/__init__.py`, and `dspy/streaming/__init__.py` to use module-level `__getattr__`. This defers the importing of `litellm`-dependent classes until they are actually accessed.
* **Type Hint Scrubbing**: Moves all internal imports of `LM`, `Embedder`, and `litellm` behind `if TYPE_CHECKING:` blocks across the `teleprompt`, `predict`, and `streaming` modules.
* **Safe Instance Checking**: Introduces safe helpers like `_is_model_response_stream` and `_is_lm_keyed_dict` that check types without triggering an eager import.
* **Strict Dependency Testing**: Adds `tests/clients/test_lazy_litellm.py`, which uses a custom `importlib` loader to "poison" `litellm` in a subprocess. This guarantees that importing `dspy`, creating signatures, making modules, and running custom `BaseLM` backends all execute perfectly in a zero-`litellm` environment.

**Why this is needed**
Prior to this PR, `litellm` was imported immediately upon `import dspy`. This added  seconds to DSPy's boot time and forced users to install it even if they exclusively used custom `BaseLM` backends.

DSPy approaches dependency inversion in `BaseLM`. 

image

### DIFF
--- a/dspy/__init__.py
+++ b/dspy/__init__.py
@@ -15,7 +15,6 @@ from dspy.utils.logging_utils import configure_dspy_loggers, disable_logging, en
 from dspy.utils.asyncify import asyncify
 from dspy.utils.syncify import syncify
 from dspy.utils.saving import load
-from dspy.streaming.streamify import streamify
 from dspy.utils.usage_tracker import track_usage
 
 from dspy.dsp.utils.settings import settings
@@ -33,3 +32,16 @@ context = settings.context
 BootstrapRS = BootstrapFewShotWithRandomSearch
 
 cache = DSPY_CACHE
+
+
+def __getattr__(name):
+    """Lazy access for litellm-dependent names (LM, Embedder, streamify)."""
+    if name in ("LM", "Embedder"):
+        from dspy.clients import __getattr__ as _clients_getattr
+
+        return _clients_getattr(name)
+    if name == "streamify":
+        from dspy.streaming.streamify import streamify
+
+        return streamify
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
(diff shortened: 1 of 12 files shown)


isaacbmiller:
I spent a good amount of time going back and forth on approaches for this inside of #9348.

any code involving litellm gets a lot messier using any of these approaches. We ultimately decided not to merge that PR because the ROI wasn't there.

I chose not to merge #9348 despite spending a good amount of time on it, but the ROI for the added complexity just wasn't there. IMO it was as close to mergeable as any solution

Do you have any ideas on how to reduce the complexity of this code?

MaximeRivest:
Yes ok, makes sense.

We must finish making baselm what it was meant to be then. Fixing all the implicit contract between lm and the rest of dspy which Baselm do not internalize. Then making dspy.lm a router that callbacks to litellm. By that time litellm will be properly used only in dspy.lm, as it should. That is the best long-term solution anyways.

PR #9553, opened 2026-04-01
Refactor/contain litellm

MaximeRivest:
# Contain litellm, add native backends, open the door to community providers

## The problem

Every `import dspy` paid a 2+ second tax loading litellm — even if you were calling OpenAI. litellm was woven through `lm.py` (32 references), `__init__.py`, `embedding.py`, the streaming module, and the adapter layer. Adding support for a new provider meant either routing through litellm or adding special-case logic to a 594-line god file. There was no way for the community to bring their own backend.

## The approach

This PR tells a story in three acts. Each commit is small, reviewable, and passes the test suite.

### Act 1: Extract and wire (commits 1–2)

We start by *moving* code, not changing it. Every litellm function is lifted out of `lm.py` into `_litellm.py` with generic names (`complete_request`, `acomplete_request`, `supports_function_calling`, …). Then a `_resolve_backend()` / `_get_backend()` dispatch layer is added. In Phase 1 it always returns `_litellm` — zero behavior change, but `lm.py` drops from 594 to 312 lines and no longer mentions litellm in its core logic.

### Act 2: Native backends (commits 3–8)

With the dispatch layer in place, we add first-party backends that talk directly to provider SDKs:

| Backend | SDK | Models |
|---|---|---|
| `_openai.py` | `openai` | `openai/*`, `azure/*`, `ft:*` |
| `_anthropic.py` | `anthropic` | `anthropic/*` |
| `_google.py` | `google-genai` | `google/*`, `gemini/*` |

Each backend implements the same protocol: capability queries, sync/async completion, and streaming — all returning standard `ChatCompletion` objects. Shared utilities (retry logic, `StreamChunk`, request format conversion) live in `_request_utils.py`.

The resolver routes by prefix: `openai/gpt-4o` → `_openai`, `anthropic/claude-sonnet-4-5-20250929` → `_anthropic`, `gemini/gemini-2.5-pro` → `_google`, everything else → `_litellm`. Streaming is unified: every backend produces `StreamChunk` objects, replacing the litellm-specific `ModelResponseStream` throughout the codebase.

### Act 3: Contain litellm, open the ecosystem (commits 9–13)

With the major providers on native SDKs, we can finally contain litellm:

- litellm configuration moves from `__init__.py` (eagerly loaded) into `_litellm.py` (loaded on first use)
- All remaining litellm imports across the codebase are routed through `_litellm.py`
- `import dspy` drops from ~3.5s to ~1.4s because litellm is no longer imported at module load

Then we open the door: a `backend=` parameter on `dspy.LM` lets you inject any module that implements the protocol, and a `dspy.backends` entry-point group lets installable packages register themselves automatically. A community package can `pip install dspy-mistral` and `dspy.LM("mistral/mistral-large")` just works.

## What changes

```
18 files changed, 2758 insertions(+), 1248 deletions(-)
```

**New files:**
- `dspy/clients/_openai.py` — OpenAI SDK backend
- `dspy/clients/_anthropic.py` — Anthropic SDK backend
- `dspy/clients/_google.py` — Google GenAI SDK backend
- `dspy/clients/_litellm.py` — litellm backend (extracted, not new logic)
- `dspy/clients/_request_utils.py` — shared retry, streaming, and format helpers
- `docs/docs/deep-dive/language-model-backends.md` — backend protocol docs

**Refactored:**
- `dspy/clients/lm.py` — 594→527 lines, litellm refs 32→13 (remaining are lazy re-exports for backward compat)
- `dspy/clients/__init__.py` — no longer eagerly imports litellm
- `dspy/clients/embedding.py` — litellm calls behind `_litellm` helpers
- `dspy/streaming/` — consumes `StreamChunk` instead of `ModelResponseStream`
- `dspy/adapters/` — `format_as_function_call` renamed (was `format_as_litellm_function_call`)

## What doesn't change

- Every existing `dspy.LM("openai/gpt-4o")`, `dspy.LM("anthropic/...")`, `dspy.LM("together_ai/...")` call works exactly as before.
- Caching, retries, `dspy.configure`, `dspy.context`, `inspect_history` — all unchanged.
- litellm remains the fallback for any provider without a native backend.
- No new required dependencies (native backends use SDKs users already have installed for their provider).

### DIFF
--- a/docs/docs/deep-dive/language-model-backends.md
+++ b/docs/docs/deep-dive/language-model-backends.md
@@ -1,0 +1,137 @@
+# Language Model Backends
+
+DSPy routes every LM call through a **backend module** — a plain Python
+module (or object) that knows how to talk to a specific provider.
+
+DSPy ships with built-in backends for **OpenAI-compatible APIs** (including
+vLLM, Ollama, Together, etc.), **Anthropic**, **Google GenAI**, and a
+**litellm** catch-all.  We don't plan to add more provider-specific
+backends to the core library.  Instead, providers and community members
+can publish their own backends as standalone packages on
+[dspy-community](https://github.com/dspy-community) and register them
+via entry points (see below).
+
+## The backend protocol
+
+A backend must expose seven attributes / functions:
+
+| Attribute | Type | Purpose |
+|---|---|---|
+| `ContextWindowError` | Exception class | The exception DSPy should catch to detect context-window overflows. |
+| `supports_function_calling(model)` | `(str) → bool` | Does this model support tool / function calling? |
+| `supports_reasoning(model)` | `(str) → bool` | Does this model expose chain-of-thought / reasoning tokens? |
+| `supports_response_schema(model)` | `(str) → bool` | Does this model support structured `response_format` schemas? |
+| `supported_params(model)` | `(str) → set[str]` | The set of request-parameter names the provider understands. |
+| `complete_request(request, model_type, num_retries)` | sync function | Execute a completion and return an OpenAI-shaped `ChatCompletion`. |
+| `acomplete_request(request, model_type, num_retries)` | async function | Async version of the above. |
+
+### Optional: streaming
+
+If your backend also supports streaming, add:
+
+| Attribute | Type | Purpose |
+|---|---|---|
+| `astream_complete(request, num_retries)` | async function | Return an async iterator of `StreamChunk` objects. After exhaustion, the iterator's `.assembled` attribute must hold the reassembled `ChatCompletion`. |
+
+DSPy only uses the streaming path when `dspy.settings.send_stream` is set
+**and** the backend has an `astream_complete` attribute.
+
+---
+
+## Example
+
+The best references are the built-in backends that translate a
+non-OpenAI wire format into the `ChatCompletion` objects DSPy expects:
+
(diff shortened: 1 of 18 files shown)


isaacbmiller:
Thanks for putting this together! Very excited to make litellm optional!

## Design Questions

(1) Could you share more context about these abstractions and why you chose them to be the way you did? Specifically backend I want to know more about.

(1b) We discussed various other LLM frameworks to take inspiration from. Why did you go with this set of functions and what were the alternatives?

(3) There are also a lot of edge cases here - what are you thinking about/concerned about and how are you going to test them?

(4) Are there any breaking changes? How would you go about deploying a change like this?

## Act 1

Litellm should only be inside of _litellm.py and i guess embedder (which should  pull from _litellm?). We could make it so that litellm is an optional dependency for embedder. WDYT?

## Backends

Backends should be structured in some way. 

What is a backend and what does it represent? Is it stateful? Is it an ABC? Protocol? are you passing a class to LM or an instance?

## Streaming

Streaming was a big consideration as well. It’s surprising that you can get away with only the `astream_complete` function, but thats great. How should a backend declare that it can stream?

I am generally not a fan of hasattr for capability checks like this but it might be the lesser evil between making things OOPed and adding a subclass or Mixin.

PR #9558, opened 2026-04-03
feat: add cache_tool_calls parameter to ReAct

### LINKED ISSUE #9419: [Discussion] Add the option to cache tool calls in ReAct and RLM?

DSPy's built in caching is handy when developing and iterating on pipelines. One might have several Modules that fire before the one they're working on, and not having to wait for everything to refresh is very nice.

With ReAct and RLM, caching exists but it often fails in practice once a tool call is made. Consider the scenario where the Module has a web search tool call. The first step will hit the cache and return the last initial call, but if the tool returns a slightly different set of results (even if the metadata (say, the timestamp of the call) is different), the next LLM call will be a cache miss.

An option to cache tool calls could be very handy. It could work like:

``` python
researcher = dspy.ReAct(
    "research_request: str -> report: str",
    tools=[internet_search, read_webpage],
    cache_tool_calls=True
)
```

By default, `cache_tool_calls` would be `False`. There are many scenarios where we want refreshed data (say a database is updated between calls) and expected behavior would be the tool has the most recent information. Caching is handy during dev, not optimal or expected in prod.

### Additional Context

_No response_

shivam2407:
## Summary

Fixes #9419

Adds opt-in tool call caching to `ReAct` via a new `cache_tool_calls` parameter (default `False`). When enabled, identical tool invocations return cached results via the existing `dspy.cache` infrastructure, avoiding redundant external calls during development iterations.

### Usage

```python
react = dspy.ReAct("question -> answer", tools=[search], cache_tool_calls=True)
```

### Design

- **Cache key**: `SHA256({"_fn_identifier": "<module>.<qualname>", ...args})` — uses the tool function's fully qualified name to prevent collisions between different tools with the same name, consistent with how `request_cache` keys LM calls
- **`finish` tool**: Never cached (always returns `"Completed."`)
- **Exceptions**: Not cached — failed calls retry on the next identical invocation
- **`None` returns**: Correctly cached using `key in cache` check instead of `is not None`
- **Callbacks**: Tool callbacks (`@with_callbacks`) do not fire on cache hits (documented in docstring)
- **Thread safety**: Delegates to `dspy.cache` which uses `RLock` for individual operations

### Changes

| File | Change |
|---|---|
| `dspy/predict/react.py` | Added `cache_tool_calls` param, `_build_tool_cache_request()`, `_call_tool()`, `_acall_tool()`. Fixed pre-existing docstring (`max_iters` default 10 → 20) |
| `tests/predict/test_react.py` | 6 new tests in `TestCacheToolCalls` class covering cache hits, misses, finish exclusion, exception handling, disabled mode, and async path |

### Note on RLM

The issue mentions both ReAct and RLM. RLM executes tools through a code interpreter rather than `Tool.__call__`, so the caching strategy differs. I've focused this PR on ReAct and can follow up with RLM support in a separate PR if desired.

## Test plan

- [x] 6 new tests pass (`pytest -k TestCacheToolCalls`)
- [x] All 14 existing ReAct tests still pass (no regressions)
- [x] Cache isolation via fixture that clears memory + disk cache between tests
- [x] Async cache hit path tested
- [ ] Manual testing with real LLM + external tool (e.g. web search)

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -14,7 +14,13 @@ if TYPE_CHECKING:
 
 
 class ReAct(Module):
-    def __init__(self, signature: type["Signature"], tools: list[Callable], max_iters: int = 20):
+    def __init__(
+        self,
+        signature: type["Signature"],
+        tools: list[Callable],
+        max_iters: int = 20,
+        cache_tool_calls: bool = False,
+    ):
         """
         ReAct stands for "Reasoning and Acting," a popular paradigm for building tool-using agents.
         In this approach, the language model is iteratively provided with a list of tools and has
@@ -26,6 +32,8 @@ class ReAct(Module):
             signature: The signature of the module, which defines the input and output of the react module.
             tools (list[Callable]): A list of functions, callable objects, or `dspy.Tool` instances.
             max_iters (Optional[int]): The maximum number of iterations to run. Defaults to 10.
+            cache_tool_calls (bool): If True, cache tool call results via ``dspy.cache`` so identical
+                invocations return cached results. Useful during development iterations. Defaults to False.
 
         Examples:
 
@@ -40,6 +48,7 @@ class ReAct(Module):
         super().__init__()
         self.signature = signature = ensure_signature(signature)
         self.max_iters = max_iters
+        self.cache_tool_calls = cache_tool_calls
 
         tools = [t if isinstance(t, Tool) else Tool(t) for t in tools]
         tools = {tool.name: tool for tool in tools}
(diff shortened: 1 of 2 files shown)


shivam2407:
Hi @chenmoneygithub @isaacbmiller — this PR adds opt-in tool call caching to ReAct (#9419). Would appreciate a review when you get a chance. Happy to address any feedback!

shivam2407:
Friendly ping! This PR has been open for ~2 weeks — happy to rebase or address any feedback if needed. Let me know if there's anything blocking review. 🙏

PR #9579, opened 2026-04-09
fix: tolerate missing or null message content in chat completions

Emre-C:
## Summary

`BaseLM._process_completion` crashes with `AttributeError` when a chat completion choice has a `message` object but no `content` attribute (or `content=None`). This can happen with tool-only responses or certain provider edge cases where the message is present but content is omitted.

The existing adapter layer (`dspy/adapters/base.py`) already handles empty LM responses gracefully via `AdapterParseError`, but the crash in `_process_completion` prevents that error path from ever being reached.

## Reproduction

```python
from dspy.clients.base_lm import BaseLM

class DummyLM(BaseLM):
    pass

lm = DummyLM("dummy", model_type="chat")
response = dotdict(
    choices=[dotdict(message=dotdict(tool_calls=None), finish_reason="stop")],
    model="dummy",
)

# Before fix: AttributeError: 'dotdict' object has no attribute 'content'
# After fix: returns [""]
lm._process_completion(response, {})
```

## Fix

Replace the direct `c.message.content` access with `getattr(c.message, "content", "") or ""`, which handles both missing and `None` content by normalizing to empty string. This preserves the existing architecture where adapters decide how to handle empty responses.

## Changes

- `dspy/clients/base_lm.py`: 3-line fix in `_process_completion`
- `tests/clients/test_lm.py`: Parametrized regression test covering both `content` missing and `content=None`

## Testing

```bash
uv run pytest tests/clients/test_lm.py -k missing_message_content -v
# 2 passed (content_missing, content_none)

uv run ruff check dspy/clients/base_lm.py tests/clients/test_lm.py
# All checks passed

uv run pre-commit run --files dspy/clients/base_lm.py tests/clients/test_lm.py
# All passed
```

Note: `ruff format --check` reports pre-existing formatting drift in `base_lm.py` outside the bugfix hunk. This PR intentionally does not include unrelated formatting changes to keep the diff minimal and reviewable.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -259,7 +259,10 @@ class BaseLM:
         outputs = []
         for c in response.choices:
             output = {}
-            output["text"] = c.message.content if hasattr(c, "message") else c["text"]
+            if hasattr(c, "message"):
+                output["text"] = getattr(c.message, "content", "") or ""
+            else:
+                output["text"] = c["text"]
 
             if hasattr(c, "message") and hasattr(c.message, "reasoning_content") and c.message.reasoning_content:
                 output["reasoning_content"] = c.message.reasoning_content
--- a/tests/clients/test_lm.py
+++ b/tests/clients/test_lm.py
@@ -16,6 +16,8 @@ from openai.types.responses import ResponseOutputMessage, ResponseReasoningItem
 from openai.types.responses.response_reasoning_item import Summary
 
 import dspy
+from dspy.clients.base_lm import BaseLM
+from dspy.dsp.utils.utils import dotdict
 from dspy.utils.usage_tracker import track_usage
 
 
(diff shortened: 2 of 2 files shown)


isaacbmiller:
Thanks for the PR!

It seems like the most likely case to hit this is when tool calls are enabled. Are there valid cases when the provider sends empty content and we aren't in a tool call?

If it is just the tool call case, then can you test it with a tool call and make sure that the flow works correctly?

Emre-C:
Good point. The spec-backed case I was targeting here is the tool-calling chat-completions path. OpenAI’s docs say assistant content is required unless tool_calls or function_call is present, and tool-call responses may therefore arrive without usable text content.
I added test_process_completion_tool_call_with_no_content to cover the regression at the BaseLM._process_completion boundary: with finish_reason="tool_calls", content=None, and a real tool_calls payload, it now returns {"text": "", "tool_calls": [...]} instead of raising.
The downstream tool-call flow was already covered by the adapter tests, which verify that content=None plus tool_calls is handled correctly after _process_completion. So this PR is mainly making the LM-side normalization robust enough to reach that existing path.
I don’t have strong evidence that null/missing content is expected for ordinary non-tool assistant responses, so I’m happy to keep the framing of the fix specifically around tool-call responses.

PR #9580, opened 2026-04-09
fix(lm): preserve per-message structure in Responses API conversion

lawrence3699:
## Summary

`_convert_chat_request_to_responses_request` flattens all Chat API messages into a single Responses API input entry, breaking multi-turn conversations for `model_type="responses"`.

## Bug

`content_blocks` is initialized once before the message loop. Every message's content is appended to the same shared list. After the loop, the code wraps everything into one entry using the loop variable `msg` — which is the *last* message. The result is a single input entry with the last message's role and all content from all messages merged together.

Before (wrong — 4 messages become 1):
```python
content_blocks = []
for msg in request.pop("messages"):
    c = msg.get("content")
    if isinstance(c, str):
        content_blocks.append(...)
    elif isinstance(c, list):
        for item in c:
            content_blocks.append(...)
request["input"] = [{"role": msg.get("role", "user"), "content": content_blocks}]
#                         ^^^ always last message's role
```

For a typical DSPy call with `[system, user, assistant, user]` messages, the Responses API receives a single `user` message containing the system prompt, few-shot demos, and the actual query all merged together.

## Fix

Move `content_blocks` inside the loop so each message builds its own content list and is appended individually:

```python
input_items = []
for msg in request.pop("messages"):
    content_blocks = []
    ...
    input_items.append({"role": msg.get("role", "user"), "content": content_blocks})
request["input"] = input_items
```

## Test

Added `test_responses_api_preserves_multi_message_structure` — verifies that a 4-message conversation (system → user → assistant → user) produces 4 separate input entries with correct roles and content.

All existing responses-related tests still pass (7 passed, 1 skipped).

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -516,8 +516,9 @@ def _convert_chat_request_to_responses_request(request: dict[str, Any]):
     """
     request = dict(request)
     if "messages" in request:
-        content_blocks = []
+        input_items = []
         for msg in request.pop("messages"):
+            content_blocks = []
             c = msg.get("content")
             if isinstance(c, str):
                 content_blocks.append({"type": "input_text", "text": c})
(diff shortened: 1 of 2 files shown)


MaximeRivest (approved):
LGTM

PR #9586, opened 2026-04-10
docs: add ai generated code policy

isaacbmiller:
Adds an ai generated code policy.

Essentially: Use agents but understand and read what you write.


isaacbmiller:
discussed offline

PR #9590, opened 2026-04-13
fix: make Module.load_state transactional to prevent partial corruption on error

### LINKED ISSUE #9589: Module.load_state silently corrupts modules on partial failure (non-transactional)
# `Module.load_state` silently corrupts modules on partial failure (non-transactional)

## Summary

`dspy.Module.load_state` is a non-transactional in-place mutation loop over named parameters. If any parameter's key is missing from the saved state (signature drift, schema migration, manual edit), the exception fires **after** earlier parameters have already been overwritten with saved state. Callers that catch the exception see a module that passes `isinstance(m, dspy.Module)` but is a Frankenstein — some predictors carry saved demos, others still have template defaults — with no way to distinguish it from a clean fresh template.

This silently corrupts inference in production. We found it while debugging a "why are my RAG answers quality-regressing after this deploy" problem where the actual cause was a signature drift between a saved optimization run and the new template.

## Environment

- DSPy **3.0.3** (installed via `uv pip install dspy`)
- Python 3.13.11
- Platform: macOS (darwin), also reproducible on Linux
- Tested against the canonical `program.save(path, save_program=False)` / `Module.load(path)` persistence pattern for JSON state

## Minimal reproducer

```python
import dspy, json, tempfile
from pathlib import Path


class Sig(dspy.Signature):


class Prog(dspy.Module):
    def __init__(self):
        super().__init__()
        self.a = dspy.ChainOfThought(Sig)
        self.b = dspy.ChainOfThought(Sig)


# 1. Save a program with non-default state (injected demos on both predictors)
source = Prog()
sentinel = Example(question="q1", answer="a1").with_inputs("question")
source.a.predict.demos = [sentinel]
source.b.predict.demos = [sentinel]

    path = Path(d) / "state.json"
    source.save(str(path), save_program=False)

    # 2. Simulate a signature drift: saved state has b.predict, but in the
    #    time since the save, someone deleted / renamed b — or equivalently,
    #    the saved file has been hand-edited or truncated mid-key. Here we
    #    remove b.predict's entry from the saved JSON.
    raw = json.loads(path.read_text())
    corrupted = {k: v for k, v in raw.items() if "b." not in k}
    path.write_text(json.dumps(corrupted))

    # 3. Fresh template, load the now-incompatible state
    template = Prog()
    print(f"BEFORE load: template.a.predict.demos = {template.a.predict.demos}")
    try:
        template.load(str(path))
    except Exception as e:
        print(f"load raised: {type(e).__name__}: {e}")
    print(f"AFTER  load: template.a.predict.demos = {[str(d) for d in template.a.predict.demos]}")
```

### Output

```
BEFORE load: template.a.predict.demos = []
load raised: KeyError: 'b.predict'
AFTER  load: template.a.predict.demos = ["{'question': 'q1', 'answer': 'a1'}"]
```

**Despite the exception**, `template.a.predict.demos` was mutated from the fresh-template default `[]` to the saved sentinel before the exception fired on `b.predict`.

## Root cause

`Module.load_state` in `dspy/primitives/base_module.py`:

```python
def load_state(self, state):
    for name, param in self.named_parameters():
        param.load_state(state[name])
```

This iterates named parameters and mutates each in place. There is no transactional boundary: if `state[name]` raises `KeyError` for parameter N, parameters `0..N-1` have already been overwritten on `self`. Any caller that catches the exception and continues using `self` is now serving inference from a half-hydrated module.

## Impact

Downstream projects that persist DSPy state across runs and catch load failures silently ship corrupted optimization state. We hit this in a Graph RAG MCP server where every restart would load optimization state for 6 DSPy programs, and a single saved-state/template mismatch on one program would leave the others in an inconsistent state invisible from a type check or `isinstance` probe.

The failure mode is especially pernicious because:

1. **The symptom is "answers are slightly worse after this deploy"** — not a hard crash. Attributed to "model quality" during triage.
2. **One ERROR log at startup** is the only signal, easily missed after deploys.
3. **`isinstance(m, dspy.Module)` still passes** — there's no structural way to tell a half-loaded module from a fresh one.

### Option A: Two-phase validate-then-apply (preferred)

Validate that every parameter key is present in the saved state **before** starting to mutate:

```python
def load_state(self, state):
    param_items = list(self.named_parameters())
    missing = [name for name, _ in param_items if name not in state]
    if missing:
        raise KeyError(f"Saved state missing required keys: {missing}")
    for name, param in param_items:
        param.load_state(state[name])
```

This catches the failure before any mutation occurs. Still leaves a gap if `param.load_state` can itself fail mid-load on an individual param's internal state, but closes the common case.

### Option B: Deep-copy, apply, swap

Apply the loaded state to a temporary deep copy of self, then swap atomically:

```python
def load_state(self, state):
    import copy
    scratch = copy.deepcopy(self)
    for name, param in scratch.named_parameters():
        param.load_state(state[name])  # may raise on missing key
    # Only swap if the whole load succeeded
    for name, param in self.named_parameters():
        param.load_state(state[name])
```

More expensive but fully transactional regardless of where the failure occurs.

### Option C: Document the non-transactional contract and let callers deepcopy

If neither of the above is desirable, at least document clearly in the `Module.load` docstring that the method is non-transactional and callers must make their own copy if they want failure-safe loading. In our downstream project we ended up working around this by changing our API to take a `template_factory: Callable[[], dspy.Module]` and calling it a second time on failure — which works but is awkward and requires every caller to know about the footgun.

## Workaround we shipped

For reference, this is what we did in `brm444/graph-rag-mcp-server` PR #3:

```python
def load_program(self, name: str, template_factory: Callable[[], dspy.Module]) -> dspy.Module:
    state_file = self.state_dir / f"{name}.json"
    template = template_factory()
    if not state_file.exists():
        return template
    try:
        template.load(str(state_file))
        return template
    except Exception as e:
        logger.error(
            f"Failed to load program '{name}' from {state_file}: {e}. "
            f"Discarding partially-hydrated module and returning fresh template."
        )
        return template_factory()  # <-- second factory call to get a clean instance
```

We also added a regression test locking in "after a mid-load failure, the returned module's demos are the factory defaults, not the sentinel from the failed load" to prevent this regressing if we later forget why the factory signature exists.

## Related

The docstring on `Module.load` says "Load the saved module. You may also want to check out dspy.load, if you want to load an entire program, not just the state for an existing program." The non-transactional nature is not mentioned.

`dspy.load` (module-level, for `save_program=True`) is transactional because it deserializes the whole program at once. The asymmetry with the JSON / `save_program=False` path is surprising.

adityasingh2400:
## Summary

Fixes #9589

`Module.load_state` was a non-transactional mutation loop. If any parameter key was missing from the saved state (e.g. due to signature drift, schema migration, or a manual edit to the JSON file), the `KeyError` would fire **after** earlier parameters had already been overwritten. Callers that caught the exception would continue using a partially-hydrated module with no way to detect the corruption.

## Root cause

```python
# Before fix:
def load_state(self, state, *, allow_unsafe_lm_state=False):
    for name, param in self.named_parameters():  # mutates as it goes
        param.load_state(state[name])  # KeyError here leaves earlier params dirty
```

## Fix: two-phase validate-then-apply

Collect all parameter names first, then validate **all** keys are present before mutating **any** of them:

```python
def load_state(self, state, *, allow_unsafe_lm_state=False):
    param_items = list(self.named_parameters())

    # Phase 1: validate — fail before touching anything
    missing = [name for name, _ in param_items if name not in state]
    if missing:
        raise KeyError(
            f"Saved state is missing required keys for the following parameters: {missing}. "
            "This can happen when the saved state was created from a different version of the "
            "module (e.g. a parameter was added or renamed). No parameters were modified."
        )

    # Phase 2: apply — only runs if all keys are present
    for name, param in param_items:
        ...
```

This is **Option A** from the issue (preferred by the reporter) — it handles the common case (missing keys) at zero performance cost. If `param.load_state` itself can fail on malformed values, Option B (deep-copy + swap) can be layered on top later, but that's a separate concern.

## What changed

- `dspy/primitives/base_module.py`: two-phase validate-then-apply in `load_state`
- `tests/primitives/test_module.py`: regression test verifying that when a load fails due to a missing key, all parameters remain at their pre-load (clean template) values

## Test

```
PASSED tests/primitives/test_module.py::test_load_state_is_transactional_on_missing_key
```

Reproducer from the issue now raises cleanly with no partial mutation:
```
BEFORE load: template.a.predict.demos = []
load raised: KeyError: "Saved state is missing required keys..."
AFTER  load: template.a.predict.demos = []  # ✅ unchanged
```
Previously `AFTER load` would show the sentinel demo from the partial load.

### DIFF
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -159,7 +159,20 @@ class BaseModule:
     def load_state(self, state, *, allow_unsafe_lm_state=False):
         from dspy.predict.predict import Predict
 
-        for name, param in self.named_parameters():
+        param_items = list(self.named_parameters())
+
+        # Two-phase load: validate all keys are present before mutating any parameter.
+        # Without this check, a missing key would raise after earlier parameters have already
+        # been overwritten, leaving the module in a partially-hydrated (corrupted) state.
+        missing = [name for name, _ in param_items if name not in state]
+        if missing:
+            raise KeyError(
+                f"Saved state is missing required keys for the following parameters: {missing}. "
+                "This can happen when the saved state was created from a different version of the "
+                "module (e.g. a parameter was added or renamed). No parameters were modified."
+            )
+
+        for name, param in param_items:
             if isinstance(param, Predict):
                 param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
             else:
--- a/tests/primitives/test_module.py
+++ b/tests/primitives/test_module.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import pytest
 import dspy
 from dspy.primitives.module import Module, set_attribute_by_name  # Adjust the import based on your file structure
 from dspy.utils import DummyLM
@@ -176,6 +177,59 @@ def test_named_parameters_duplicate_references():
     # Only testing for whether exceptions are thrown or not
     # As Module.named_parameters() is recursive, this is mainly for catching infinite recursion
     module.named_parameters()
+
+
+def test_load_state_is_transactional_on_missing_key():
+    """
+    Regression test for: Module.load_state silently corrupts modules on partial failure.
+
+    When a saved state is missing a key (e.g. due to schema/signature drift), load_state
+    must raise BEFORE mutating any parameters. A module whose load fails should still have
+    all parameters at their original (pre-load) values.
+    """
+    import json
+    import tempfile
+    from pathlib import Path as _Path
+
+    from dspy.primitives.example import Example
+
+    class Sig(dspy.Signature):
+        question: str = dspy.InputField()
+        answer: str = dspy.OutputField()
+
+    class Prog(dspy.Module):
+        def __init__(self):
+            super().__init__()
+            self.a = dspy.ChainOfThought(Sig)
+            self.b = dspy.ChainOfThought(Sig)
+
+    # 1. Save a program with non-default demo state on both predictors
+    source = Prog()
+    sentinel = Example(question="q1", answer="a1").with_inputs("question")
+    source.a.predict.demos = [sentinel]
+    source.b.predict.demos = [sentinel]
+
+    with tempfile.TemporaryDirectory() as d:
+        path = _Path(d) / "state.json"
+        source.save(str(path), save_program=False)
+
+        # 2. Corrupt the saved state by removing b.predict's entry (simulate signature drift)
+        raw = json.loads(path.read_text())
+        corrupted = {k: v for k, v in raw.items() if "b." not in k}
+        path.write_text(json.dumps(corrupted))
+
+        # 3. Load should raise before mutating any parameters
+        template = Prog()
+        assert template.a.predict.demos == [], "Pre-condition: a.predict.demos should be empty before load"
+
+        with pytest.raises(KeyError, match="missing required keys"):
+            template.load(str(path))
+
+        # 4. The critical assertion: a.predict.demos must still be empty (not corrupted)
+        assert template.a.predict.demos == [], (
+            "load_state must not partially mutate the module when it fails — "
+            "template.a.predict.demos was corrupted despite the load raising an exception"
+        )
 
 
 def test_load_dspy_program_cross_version():
--- a/tests/primitives/test_module.py
+++ b/tests/primitives/test_module.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 import pytest
+
 import dspy
 from dspy.primitives.module import Module, set_attribute_by_name  # Adjust the import based on your file structure
 from dspy.utils import DummyLM
@@ -26,7 +27,7 @@ def test_named_predictors():
     module = HopModule()
     named_preds = module.named_predictors()
     assert len(named_preds) == 2, "Should identify correct number of Predict instances"
-    names, preds = zip(*named_preds, strict=False)
+    names, _preds = zip(*named_preds, strict=False)
     assert "predict1" in names and "predict2" in names, "Named predictors should include 'predict1' and 'predict2'"
 
 


MaximeRivest:
Hello Aditya,

Do you mind succinctly explaining you test design and you fix and way this is the best fix? Something extremely, brief would be most helpful.

Additionally, a few words related edge cases, changes to user experience, test robustness would be appreciated.

Best,
Maxime

adityasingh2400:
Hi Maxime,

**Test design.** `test_load_state_is_transactional_on_missing_key` builds a module, deep-copies it as the pre-load snapshot, drops one key from the saved state, expects `load_state` to raise `KeyError`, then asserts every parameter on the module is structurally equal to the snapshot. The structural-equality check is what makes it a transactionality test rather than a "did the right exception fire" test: without the fix, earlier parameters in the iteration order have already been overwritten by the time the `KeyError` is raised, so the equality assertion fails on those parameters even though the exception itself is correct.

**Fix.** Two-phase validate-then-apply: collect `named_parameters()` into a list, raise one `KeyError` listing every missing name if any are absent from `state`, only then loop and call `param.load_state(...)`.

**Why this is the best fix.** Zero performance cost (one extra pass over names, no copies of the actual state values), and it solves the failure mode the linked issue is reporting (missing keys from version drift / signature edits), which is the common case. It deliberately does not try to handle `param.load_state` itself raising mid-loop on individual values; that needs copy-and-swap, pays a real cost on every load, and is straightforward to layer on later if it ever shows up in practice. The single error names every missing key at once, so callers get the full diff on the first failed load instead of one key per retry.

**Edge cases.** Empty state, arbitrary subset of keys missing, all keys present (no behaviour change versus prior implementation), `allow_unsafe_lm_state=False` with LM keys present (the existing safety check still runs after validation, ordering preserved). Explicitly out of scope: `param.load_state` raising on individual malformed values mid-apply, which is a separate failure mode.

**UX.** Same exception type as before (`KeyError`). The message now lists every missing parameter name and ends with `No parameters were modified.` so callers know retrying or rolling back is safe rather than having to guess what state the module is in.

**Test robustness.** Uses the public API only, no private helpers imported. The pre/post comparison is structural via `dump_state()`, so it survives field reordering on the parameters. It does not depend on the iteration order of `named_parameters()`; the assertion is "no parameter was mutated", which is the actual contract being tested, so a future change to that ordering wouldn't make the test flaky or vacuous.

Happy to inline a shorter version of this into the docstring on `load_state` itself if you'd rather the explanation lives in the code.

adityasingh2400:
Superseded by #9741 which landed on 2026-05-13 with a cleaner approach (deepcopy trial run that exercises the full load path before mutating `self`, rather than just an upfront key check). The transactional guarantee my PR was after is now in main. Closing this one. Thanks @ashishSoni1234 for the cleaner fix.

PR #9616, opened 2026-04-16
fix: cloudpickle serialization of Signature on Python 3.14

### LINKED ISSUE #9615: [Bug] `cloudpickle.dumps` fails on dspy.Signature subclasses on Python 3.14: cannot pickle `_abc._abc_data`
### What happened?

🤖 This bug report assisted by Claude Opus 4.6

---

On Python 3.14, `cloudpickle.dumps()` fails when serializing a class-based `dspy.Signature` subclass (or any `dspy.Predict` wrapping one). The error is `TypeError: cannot pickle '_abc._abc_data' object`. String-form signatures work fine.

- Python 3.14.3
- cloudpickle 3.1.2
- dspy 3.1.3
- pydantic 2.12.5

#### Possible cause

The serialization path is:

1. `SignatureMeta` (dspy's metaclass for Signature) has a `__pydantic_parent_namespace__` dict
2. That dict contains an `annotate_func` key (a `_PydanticWeakRef` wrapping a function)
3. The function's `__closure__` captures cell objects
4. One of those cells contains a class with an `_abc_impl` attribute
5. `_abc._abc_data` objects are not picklable on Python 3.14

The `_abc._abc_data` type has never been picklable, but on earlier Python versions this path wasn't reached because `__pydantic_parent_namespace__` either didn't exist or didn't capture the same closure chain. Python 3.14 introduced PEP 749 (deferred evaluation of annotations) which changes how Pydantic captures the parent namespace, introducing the `annotate_func` that carries the problematic closure.

#### Minimal repro

```python
import cloudpickle
import dspy

class Foo(dspy.Signature):
    """Test signature."""
    x: str = dspy.InputField()
    y: str = dspy.OutputField()

cloudpickle.dumps(Foo)  # TypeError: cannot pickle '_abc._abc_data' object
```

String-form equivalent is a workaround:

```python
import cloudpickle
import dspy

sig = dspy.Signature("x -> y")
cloudpickle.dumps(dspy.Predict(sig))  # OK
```

#### Traceback

```
TypeError: cannot pickle '_abc._abc_data' object
when serializing dict item '_abc_impl'
when serializing tuple item 0
when serializing cell reconstructor arguments
when serializing cell object
when serializing tuple item 0
when serializing dict item '__closure__'
when serializing tuple item 1
when serializing function state
when serializing function object
when serializing tuple item 0
when serializing pydantic._internal._model_construction._PydanticWeakRef reconstructor arguments
when serializing pydantic._internal._model_construction._PydanticWeakRef object
when serializing dict item 'annotate_func'
when serializing dict item '__pydantic_parent_namespace__'
when serializing tuple item 0
when serializing dspy.signatures.signature.SignatureMeta state
when serializing dspy.signatures.signature.SignatureMeta object
```

### DSPy version

3.1.3

isaacbmiller:
## Problem

On Python 3.14, `cloudpickle.dumps()` fails on `dspy.Signature` subclasses (and `Predict` modules wrapping them) with:

```
TypeError: cannot pickle '_abc._abc_data' object
```

Two things changed in Python 3.14 that combine to cause this:

### 1. PEP 649 added `__annotate_func__` to every class

PEP 649 (deferred evaluation of annotations) stores a lazy `__annotate__` closure on every class dict. This closure captures the class namespace, which includes `_abc._abc_data` -- an unpicklable C-level object from the ABC machinery. cloudpickle tries to serialize it and fails.

**Fixed upstream** in [cloudpipe/cloudpickle#578](https://github.com/cloudpipe/cloudpickle/pull/578), released as cloudpickle 3.1.2.

### 2. Pydantic captures `SignatureMeta.__new__` locals via frame introspection

Pydantic's `ModelMetaclass.__new__` calls `parent_frame_namespace()`, which uses `sys._getframe(2)` to grab the local variables of whatever code called the metaclass. This is a [hack to support forward references](https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_typing_extra.py#L213-L274) in local scopes (e.g. a model defined inside a function that references a local type alias).

For a plain `BaseModel` subclass, frame depth 2 lands on the user's code (module scope or function body) -- harmless. But DSPy's `SignatureMeta.__new__` calls `super().__new__()`, which inserts an extra frame. So depth 2 lands on `SignatureMeta.__new__` instead, and Pydantic captures *its* locals into `cls.__pydantic_parent_namespace__`. On Python 3.14, those locals now include:

- `annotate_func` -- the PEP 649 `__annotate__` closure (which references `_abc._abc_data`)  
- `field` -- a cell variable holding the class namespace dict (which also contains `_abc._abc_data`)

cloudpickle tries to serialize `__pydantic_parent_namespace__`, hits these, and fails.

Pydantic itself never hits this because `ModelMetaclass` isn't wrapped by another metaclass, so `sys._getframe(2)` always lands on user code. This is DSPy-specific.

## Fix

1. **Bump `cloudpickle>=3.1.2`** -- handles issue 1 (same fix [Pydantic shipped](https://github.com/pydantic/pydantic/pull/12473)).
2. **Pass `__pydantic_reset_parent_namespace__=False`** on Python 3.14+ -- tells Pydantic to skip the frame introspection entirely. This is safe because `SignatureMeta.__new__` eagerly resolves all annotations before calling `super()`, so the captured namespace would never be used.

## Reproduction

```python
import cloudpickle
import dspy

class Foo(dspy.Signature):
    """Test signature."""
    x: str = dspy.InputField()
    y: str = dspy.OutputField()

cloudpickle.dumps(Foo)  # TypeError: cannot pickle '_abc._abc_data' object
```

## References

- [cloudpipe/cloudpickle#572](https://github.com/cloudpipe/cloudpickle/issues/572) -- cloudpickle issue
- [python/cpython#140820](https://github.com/python/cpython/issues/140820) -- CPython issue (closed, deferred to cloudpickle)
- [cloudpipe/cloudpickle#578](https://github.com/cloudpipe/cloudpickle/pull/578) -- cloudpickle fix (released in 3.1.2)
- [pydantic/pydantic#11991](https://github.com/pydantic/pydantic/pull/11991) -- Pydantic Python 3.14 support (xfailed cloudpickle tests)
- [pydantic/pydantic#12473](https://github.com/pydantic/pydantic/pull/12473) -- Pydantic bumps cloudpickle to 3.1.2

Fixes #9615

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -173,6 +173,14 @@ class SignatureMeta(type(BaseModel)):
         ordered_annotations.update({k: v for k, v in raw_annotations.items() if k not in ordered_annotations})
         namespace["__annotations__"] = ordered_annotations
 
+        # On Python 3.14+, prevent Pydantic from capturing this frame's locals via
+        # parent_frame_namespace(). Those locals include references to the __annotate__
+        # closure and the class namespace dict, which contain unpicklable _abc._abc_data
+        # and break cloudpickle. This is safe because DSPy eagerly resolves all
+        # annotations above.
+        if sys.version_info >= (3, 14):
+            kwargs["__pydantic_reset_parent_namespace__"] = False
+
         # Let Pydantic do its thing
         cls = super().__new__(mcs, signature_name, bases, namespace, **kwargs)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "anyio",
     "asyncer==0.0.8",
     "cachetools>=5.5.0",
-    "cloudpickle>=3.0.0",
+    "cloudpickle>=3.1.2",
     "numpy>=1.26.0",
     "xxhash>=3.5.0",
     "gepa[dspy]==0.0.27",
--- a/tests/signatures/test_signature.py
+++ b/tests/signatures/test_signature.py
@@ -1,6 +1,8 @@
+import pickle
 from types import UnionType
 from typing import Any, Optional, Union
 
+import cloudpickle
 import pydantic
 import pytest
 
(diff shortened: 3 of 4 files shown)


isaacbmiller:
@codex review

PR #9617, opened 2026-04-16
chore: add uv exclude-newer to limit dependency resolution window

isaacbmiller:
Adds `exclude-newer = "1 week"` to `[tool.uv]` in `pyproject.toml`. This tells uv to only resolve packages published within the last week when updating the lockfile, reducing exposure to supply chain attacks from newly published malicious versions.

The corresponding `[options]` block in `uv.lock` is updated automatically by uv.


isaacbmiller:
Closing -- the `exclude-newer` setting causes the absolute timestamp in `uv.lock` to change on every resolution, creating merge conflicts in multi-contributor repos. See astral-sh/uv#18708 and astral-sh/uv#18775. Keeping this in personal global config or as a per-command flag instead.

PR #9577, opened 2026-04-08
fix: correct demo index assignment in MIPROv2 optimizer

Ricardo-M-L:
## Summary

- Fix a copy-paste bug in `dspy/teleprompt/mipro_optimizer_v2.py` where `raw_chosen_params[f"{i}_predictor_demos"]` was incorrectly assigned `instruction_idx` instead of `demos_idx`
- Line 776 correctly records `instruction_idx` for instructions, and line 781 correctly uses `demos_idx` to set `predictor.demos`, but line 783 mistakenly reuses `instruction_idx` when recording the chosen demo index
- This causes `raw_chosen_params` to feed wrong demo indices back to optuna's Bayesian optimizer, corrupting its parameter history and degrading MIPROv2 optimization quality

## Changes

```diff
- raw_chosen_params[f"{i}_predictor_demos"] = instruction_idx
+ raw_chosen_params[f"{i}_predictor_demos"] = demos_idx
```

## Test plan

- [x] Verified the surrounding code context confirms the correct variable is `demos_idx` (lines 779-782 all use `demos_idx`)
- [ ] Run existing MIPROv2 optimizer tests to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9627, opened 2026-04-16
fix: correct demo index assignment in MIPROv2 raw_chosen_params

### LINKED ISSUE #9626: MIPROv2: raw_chosen_params stores wrong index for predictor demos
## Summary

In `MIPROv2._select_and_insert_instructions_and_demos`, the `raw_chosen_params` dictionary stores `instruction_idx` under the `f"{i}_predictor_demos"` key instead of `demos_idx`. This causes the logged/returned raw parameters to misreport which few-shot demo set was selected for each predictor.

## Location

`dspy/teleprompt/mipro_optimizer_v2.py`, line 783 (on `main`):

```python
for i, predictor in enumerate(candidate_program.predictors()):
    # Select instruction
    instruction_idx = trial.suggest_categorical(...)
    ...
    raw_chosen_params[f"{i}_predictor_instruction"] = instruction_idx   # ✓ correct
    # Select demos if available
    if demo_candidates:
        demos_idx = trial.suggest_categorical(f"{i}_predictor_demos", ...)
        predictor.demos = demo_candidates[i][demos_idx]
        trial_logs[trial_num][f"{i}_predictor_demos"] = demos_idx        # ✓ correct
        chosen_params.append(f"Predictor {i}: Few-Shot Set {demos_idx}") # ✓ correct
        raw_chosen_params[f"{i}_predictor_demos"] = instruction_idx      # ✗ BUG — should be demos_idx
```

The symmetry with the three surrounding lines — `trial_logs`, `chosen_params`, and the instruction branch — shows this is a copy-paste error: the value should be `demos_idx`, matching the key name `_predictor_demos`.

## Impact

`raw_chosen_params` is returned alongside `chosen_params` and is used by downstream code that inspects raw numerical indices (e.g. reconstructing configurations from logs, hyperparameter analysis). When the number of instruction candidates differs from the number of demo candidates, the stored index can be out of range for the demo axis and will silently point to the wrong demo set even when it happens to be in range.

## Proposed Fix

One-character change — replace `instruction_idx` with `demos_idx` on line 783. Happy to submit a PR.

## Environment

- `main` branch (verified on commit `fcb648de`)

Ricardo-M-L:
## Summary

Fixes #9626.

In `MIPROv2._select_and_insert_instructions_and_demos`, `raw_chosen_params[f\"{i}_predictor_demos\"]` was being assigned `instruction_idx` instead of `demos_idx` — a copy-paste error where the key correctly refers to the demo axis but the value came from the instruction axis.

## Change

One-line fix in `dspy/teleprompt/mipro_optimizer_v2.py`:

```diff
-raw_chosen_params[f\"{i}_predictor_demos\"] = instruction_idx
+raw_chosen_params[f\"{i}_predictor_demos\"] = demos_idx
```

The three surrounding lines (`trial_logs`, `chosen_params`, and the `_predictor_instruction` branch immediately above) all already use the correct index — this line was the outlier.

## Why this matters

`raw_chosen_params` is returned by this method and used by callers that inspect raw numerical indices (e.g. reconstructing a trial's configuration from logs). When `len(instruction_candidates[i]) != len(demo_candidates[i])`, the stored index can be out of range for the demo axis; otherwise it silently points to the wrong demo set.

## Notes

- Re-submitting after previously closing #9577 without an issue. This time the bug is documented in #9626 first per the preferred workflow.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -780,7 +780,7 @@ class MIPROv2(Teleprompter):
                 predictor.demos = demo_candidates[i][demos_idx]
                 trial_logs[trial_num][f"{i}_predictor_demos"] = demos_idx
                 chosen_params.append(f"Predictor {i}: Few-Shot Set {demos_idx}")
-                raw_chosen_params[f"{i}_predictor_demos"] = instruction_idx
+                raw_chosen_params[f"{i}_predictor_demos"] = demos_idx
 
         return chosen_params, raw_chosen_params
 


isaacbmiller:
it is a mass agent pr but its good!

thanks for the pr! See CONTRIBUTING.md for our policy on future AI generated prs.

PR #9628, opened 2026-04-17
Fix cache bugs: os.fspath(None) crash, double disk lookups, lazy logging

isaacbmiller:
**Bug fixes and improvements to the cache layer, no new features.**

### Fixes
- **`os.fspath(None)` crash**: `Cache.__init__` called `os.fspath(disk_cache_dir)` unconditionally, crashing when `disk_cache_dir=None`. Fixed the type annotation to `str | None` to match how callers actually use it, and guard the `os.fspath` call.
- **Double disk lookups**: `get()` did `key in disk_cache` then `disk_cache[key]` -- two separate SQLite queries. Now uses `disk_cache.get(key)`.
- **`disk_cache.set()` over `__setitem__`**: Allows catching write exceptions (disk full, unpicklable values) instead of propagating them.
- **`load_memory_cache` ValueError string**: Backslash line continuation produced a string with embedded whitespace.

### Improvements
- Extract `_transform_value` to module level (was rebuilt as a closure on every `cache_key` call)
- Extract `_prepare_cached_response` (response-clearing logic was duplicated inline)
- Switch f-string logging to lazy `%s` formatting (avoids evaluating args when debug is off)
- Store `disk_cache_dir` as instance attribute

### Tests
- Add `test_cache_init_with_disk_disabled_and_none_dir` for the os.fspath fix
- Split `test_cache_miss` and `test_unserializable_key` into separate tests (were previously one combined test on main)

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -1,6 +1,7 @@
 import copy
 import inspect
 import logging
+import os
 import threading
 from functools import wraps
 from hashlib import sha256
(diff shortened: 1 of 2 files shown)


MaximeRivest (approved):
LGTM

PR #9629, opened 2026-04-17
Add restrict_pickle option for safe disk cache deserialization

isaacbmiller:
## Problem

DSPy uses `diskcache` with unrestricted `pickle.load` for disk cache deserialization. A corrupted or malicious cache file can execute arbitrary code on load.

## Solution

Add an opt-in `restrict_pickle` parameter that swaps `pickle.load` with a restricted unpickler on the read path.

**Depends on #9628.**

```python
dspy.configure_cache(restrict_pickle=True)

# For custom types:
dspy.configure_cache(restrict_pickle=True, safe_types=[MyDataclass, MyOtherClass])
```

### msgspec / pydantic serialization

We initially explored replacing pickle entirely with msgspec (binary) or pydantic JSON serialization. This failed because:

- **diskcache is tightly coupled to pickle.** `Disk.store` and `Disk.fetch` use `pickle.dumps`/`pickle.load` for anything that is not a raw string or bytes value (`MODE_PICKLE`). Replacing the serialization format means reimplementing the entire Disk layer.
- **The cached object graph is complex.** DSPy caches litellm response objects (`ModelResponse`, `ResponsesAPIResponse`), which are deeply nested pydantic models with optional fields, unions, and library-internal types. msgspec requires struct definitions for every type in the graph. Pydantic JSON serialization loses type identity on round-trip (you get dicts back, not model instances).
- **numpy arrays are not JSON-serializable.** Embedding caches store `numpy.ndarray`, which requires a binary format. msgspec can handle this with custom encoders, but adds complexity for no security benefit over restricted pickle.

### Routing-byte envelope (msgspec + pydantic + pickle)

We tried a hybrid approach: a routing byte prefix (`0x01` = msgspec, `0x02` = pydantic, `0x80` = restricted pickle) to select the best serializer per value type. This was over-engineered -- three code paths, three failure modes, and the routing byte format is not a standard that any other tool understands. Restricted pickle alone handles all the types DSPy caches.

### Exact (module, qualname) allowlist for all types

We tried enumerating every exact class that appears in the pickle stream for DSPy cache payloads. This covered the simple cases (9 litellm classes + 5 numpy helpers), but broke immediately for:

- **Responses API** -- `ResponsesAPIResponse` contains nested `openai.types.responses.*` classes (`ResponseOutputMessage`, `ResponseOutputText`, etc.) that were not in the allowlist.
- **Tool-calling responses** -- `ModelResponse` with tool calls pulls in `ChatCompletionMessageToolCall`, `Function`, etc.
- **Forward compatibility** -- any new nested type added by a litellm/openai upgrade silently breaks cache reads (returns `None` instead of the cached value).

## Trust model

The restricted unpickler uses three tiers:

| Tier | Scope | Rationale |
|------|-------|-----------|
| **Module prefix** | `litellm.types.*`, `openai.types.*` | These packages contain only pydantic data models with no side effects. The object graph changes across library versions (new nested types for tool calls, responses API, etc.), so exact pairs are not maintainable. Module prefix is forward-compatible. |
| **Exact pairs** | 5 numpy reconstruction helpers (`dtype`, `_frombuffer`, `_reconstruct` across `numpy.core` / `numpy._core`) | The `numpy` namespace contains dangerous functions (e.g. `numpy.ctypeslib.load_library` enables arbitrary shared library loading). Only the specific helpers needed for `ndarray` deserialization are allowed. |
| **User-registered** | `safe_types=[...]` by exact `(module, qualname)` | Users must register all types in their object graph, including nested ones. If a nested type is missing, the error message names the exact class to add. Validated with `isinstance(t, type)` at init time. |

### Why not prefix for numpy?

`numpy.ctypeslib.load_library` is reachable via `numpy.*` prefix and can load arbitrary `.so`/`.dylib` files -- a direct RCE vector. We verified this by constructing a pickle payload that calls it.

### Write-side enforcement

`Cache.put()` trial-deserializes through the restricted unpickler before writing to disk. If the value contains types not in the allowlist, the disk write is skipped with a `logger.warning` and memory cache still works. This prevents the "writes succeed, reads silently return None on next restart" problem.

## How it works

- `_RestrictedDisk` subclasses `diskcache.Disk`, overrides `fetch()` to use `_RestrictedUnpickler` for `MODE_PICKLE` entries
- `restricted_disk(allowed)` returns a `_RestrictedDisk` subclass with the allowlist baked in as a class attribute (diskcache instantiates the Disk class itself, so parameterization must happen via subclass)
- Write path is untouched -- both modes use identical pickle wire format, entries are cross-compatible
- Corrupt pickle data raises `DeserializationError` (wraps `UnpicklingError`, `EOFError`, etc.), caught by `Cache.get()` as a cache miss

## Files

| File | Lines | Description |
|------|-------|-------------|
| `dspy/clients/disk_serialization.py` | +96 (new) | `_RestrictedUnpickler`, `_RestrictedDisk`, `restricted_disk()`, `DeserializationError` |
| `dspy/clients/cache.py` | +50 | `restrict_pickle` and `safe_types` params, write-side preflight, `DeserializationError` handling |
| `dspy/clients/__init__.py` | +8 | Pass-through in `configure_cache()` |
| `tests/clients/test_disk_serialization.py` | +286 (new) | Roundtrip tests (plain, pydantic, dataclass, numpy, litellm, responses API, tool calls), allowlist enforcement, corruption handling, nested class support |
| `tests/clients/test_cache.py` | +163 | Integration tests: restricted fixture, model response roundtrip, registered dataclass, configure_cache, real SQLite corruption, wire format compatibility, write-skip for unlisted types, safe_types validation |
| `tests/clients/test_embedding.py` | +32 | numpy embedding through restricted mode |

### DIFF
--- a/docs/docs/tutorials/cache/index.md
+++ b/docs/docs/tutorials/cache/index.md
@@ -81,6 +81,51 @@ This is especially beneficial when:
 - Working with long system prompts that remain constant
 - Making multiple requests with similar context
 
+## Restricting Pickle Deserialization
+
+By default, DSPy's on-disk cache uses Python's `pickle` for serialization. While this handles arbitrary Python objects, `pickle.load` can execute arbitrary code -- meaning a corrupted or malicious cache file could be dangerous.
+
+DSPy provides an opt-in `restrict_pickle` mode that restricts which types the cache is allowed to deserialize:
+
+```python
+dspy.configure_cache(restrict_pickle=True)
+```
+
+When enabled, the cache only allows:
+
+- **LiteLLM and OpenAI response types** (`litellm.types.*`, `openai.types.*`) -- the pydantic data models that DSPy caches for LM calls, embeddings, and the Responses API.
+- **NumPy array reconstruction helpers** -- the specific internal functions needed to deserialize `numpy.ndarray` (used by embedding caches).
+- **User-registered types** via `safe_types` -- any additional types you explicitly trust.
+
+If you cache custom types (dataclasses, pydantic models, etc.), register them:
+
+```python
+from dataclasses import dataclass
+
+@dataclass
+class MyResult:
+    score: float
+    label: str
+
+dspy.configure_cache(restrict_pickle=True, safe_types=[MyResult])
+```
+
+If a type is missing from the allowlist, the cache treats it as a miss and returns `None`. The log message will name the exact type that was rejected:
+
+```
+WARNING dspy.clients.cache: Failed to deserialize disk cache entry <key>
+```
+
+### Nested types
+
+If your registered type contains nested custom types, you must register all of them. For example, if `MyResult` contains a `Metadata` field, register both:
+
+```python
+dspy.configure_cache(restrict_pickle=True, safe_types=[MyResult, Metadata])
+```
+
+The error message will tell you exactly which nested type is missing.
+
 ## Disabling/Enabling DSPy Cache
 
 There are scenarios where you might need to disable caching, either entirely or selectively for in-memory or on-disk caches. For instance:
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -13,6 +13,11 @@ import pydantic
 from cachetools import LRUCache
 from diskcache import FanoutCache
 
+from dspy.clients.disk_serialization import (
+    DeserializationError,
+    restricted_disk,
+)
+
 logger = logging.getLogger(__name__)
 
 
@@ -47,6 +52,8 @@ class Cache:
         disk_cache_dir: str | None,
         disk_size_limit_bytes: int | None = 1024 * 1024 * 10,
         memory_max_entries: int = 1000000,
+        restrict_pickle: bool = False,
+        safe_types: list[type[Any]] | None = None,
     ):
         """
         Args:
@@ -55,6 +62,9 @@ class Cache:
             disk_cache_dir: The directory where the disk cache is stored.
             disk_size_limit_bytes: The maximum size of the disk cache (in bytes).
             memory_max_entries: The maximum size of the in-memory cache (in number of items).
+            restrict_pickle: When True, restrict pickle deserialization to a known-safe
+                set of types. When False (default), use unrestricted pickle.
+            safe_types: Additional types to allow when restrict_pickle is True.
         """
 
         self.enable_disk_cache = enable_disk_cache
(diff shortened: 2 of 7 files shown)


docs/docs/tutorials/cache/index.md
```diff
@@ -81,6 +81,51 @@ This is especially beneficial when:
...
+
+@dataclass
+class MyResult:
+    score: float
+    label: str
+
+dspy.configure_cache(restrict_pickle=True, safe_types=[MyResult])
+```
+
+If a type is missing from the allowlist, the cache treats it as a miss and returns `None`. The log message will name the exact type that was rejected:
```
michaelisaac-dev, line 113: The log msg/warning currently doesn't include the type.

dspy/clients/cache.py
```diff
@@ -112,7 +129,12 @@ def get(self, request: dict[str, Any], ignored_args_for_cache_key: list[str] | N
                 return self._prepare_cached_response(response)
 
         if self.enable_disk_cache:
-            response = self.disk_cache.get(key)
+            try:
+                response = self.disk_cache.get(key)
+            except DeserializationError:
+                logger.debug("Failed to deserialize disk cache entry %s", key)
```
michaelisaac-dev, line 135: Do we want to include the type here as well? It would be part of the exception object.

dspy/clients/disk_serialization.py
```diff
@@ -0,0 +1,98 @@
...
+    except Exception as e:
+        raise DeserializationError(f"Corrupt cache entry: {e}") from e
+
+
+class _RestrictedDisk(Disk):
+    """Disk subclass that restricts pickle deserialization to an allowlist."""
+
+    _allowed: frozenset[tuple[str, str]]
+
+    def _load(self, f: Any) -> Any:
```
michaelisaac-dev: This method seems to duplicate `_restricted_load`, which is only used in the tests. I'm not sure the tests add much value in this case.

Maybe we could remove `_load`, and instead call `_restricted_load` from the `fetch` method with `_restricted_load(f, self._allowed)`?

dspy/clients/cache.py
```diff
@@ -112,7 +129,12 @@ def get(self, request: dict[str, Any], ignored_args_for_cache_key: list[str] | N
...
+                return None
```
michaelisaac-dev, line 137: I would consider adding `self.disk_cache.delete(key)` here. Otherwise, the same key re-fails on every get() and re-logs the debug msg

dspy/clients/cache.py
isaacbmiller: it will have the type from when _restricted_load throws

isaacbmiller:
reviewed offline with @chenmoneygithub as well via #9414

PR #9633, opened 2026-04-18
fix(rlm): handle None code in _strip_code_fences gracefully

### LINKED ISSUE #9632: [Bug] RLM: AttributeError: 'NoneType' object has no attribute 'strip'
### What happened?

Getting this error quite a bit when using RLM with a local qwen3.6:35b-a3b model for a variety of use cases:
"AttributeError: 'NoneType' object has no attribute 'strip'"

It appears that some iterations of the RLM loop just fail to produce any reasoning or code, which seems to be causing the error. My logging shows:
"Reasoning: None
Code: None "

Unsure if I am causing this or RLM is really this brittle in practice.

### Steps to reproduce

Run an RLM module with a local qwen3.6:35b-a3b model

### DSPy version

dspy 3.1.3

bobo-xxx:
## Fix: Handle None code in _strip_code_fences

When the LLM fails to generate code, `action.code` and `pred.code` can be `None`.
The `_strip_code_fences()` function called `code.strip()` without checking if `code` is `None` first, causing an `AttributeError`.

### Fix

Added a None check at the start of `_strip_code_fences()`:

```python
def _strip_code_fences(code: str) -> str:
    """Extract Python code from markdown fences, or return as-is if no fences."""
    if code is None:
        return ""
    code = code.strip()
    # ... rest unchanged
```

Fixes #9632.

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -66,6 +66,8 @@ _PYTHON_FENCE_LANGS = {"python", "py", "python3", "py3", ""}
 
 def _strip_code_fences(code: str) -> str:
     """Extract Python code from markdown fences, or return as-is if no fences."""
+    if code is None:
+        return ""
     code = code.strip()
     if "```" not in code:
         return code


tryingET:
Thanks for putting this up — I reproduced the exact `action.code is None` failure locally on stock `dspy.RLM` with Qwen3.6 via vLLM, so I agree the crash needs to be fixed.

I wanted to suggest a slightly stronger variant than `None -> ""`.

What I tested locally was:

1. `_strip_code_fences(code: str | None)` explicitly handles malformed outputs:
   - `code is None` -> raise `SyntaxError("No code returned by the model...")`
   - `code == ""` -> raise `SyntaxError("Empty code block returned by the model...")`
2. In both sync + async iteration paths, that `SyntaxError` is caught the same way other fence-format issues are caught.
3. The loop then records a recoverable `[Error] ...` REPL step and can continue to the next iteration instead of crashing the whole RLM run.
4. I added a unit test for the `code=None` case and the local `tests/predict/test_rlm.py` suite still passes.

Why I think this is better than `None -> ""`:
- it preserves the fact that the model emitted a malformed action
- it keeps the failure visible in the trajectory
- it still prevents the hard crash
- it gives the next iteration a chance to recover

So I think the right behavior is “recoverable malformed action” rather than “quietly execute empty code”. If that direction sounds good, I’d be happy to adapt the patch accordingly.

PR #9652, opened 2026-04-22
fix: type assistant content as output_text in Responses API conversion

therewasaguy:
#9580 fixed message flattening in `_convert_chat_request_to_responses_request` and began emitting each chat message as its own Responses API `input` item with its original role. That fix is a prerequisite for this one — it is what first caused assistant content to actually reach OpenAI with its role intact.

With flattening fixed, the next layer of the bug surfaces: the converter still types every message's content as `input_text` regardless of role. The OpenAI Responses API requires `output_text` (or `refusal`) for `assistant` input items, so any chat request containing an assistant message — including DSPy's own few-shot demos — now fails with:

    OpenAIException: Invalid value: 'input_text'. Supported values are:
    'output_text' and 'refusal'.
    param: input[2].content[0]

This commit makes the content type role-aware: `assistant` -> `output_text`, everything else (`user`, `system`, `developer`, `tool`) -> `input_text`. List-form content goes through the same role-aware mapping. Image and file items are unchanged. `tool`-role handling is unchanged and remains a separate known gap (Responses API wants `function_call_output` items, not a message).

Unblocks any DSPy module that uses few-shot assistant demos via the Responses API, and restores the behavior #9580 intended end-to-end.

Tests: updates `test_responses_api_preserves_multi_message_structure` and adds a list-form assistant content case.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -518,15 +518,22 @@ def _convert_chat_request_to_responses_request(request: dict[str, Any]):
     if "messages" in request:
         input_items = []
         for msg in request.pop("messages"):
+            role = msg.get("role", "user")
+            # Responses API requires assistant text to be typed "output_text";
+            # user/system/developer text must be "input_text". Using "input_text"
+            # for an assistant message causes OpenAI to reject the request with
+            # `Invalid value: 'input_text'. Supported values are: 'output_text'
+            # and 'refusal'.`
+            text_type = "output_text" if role == "assistant" else "input_text"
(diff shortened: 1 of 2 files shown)


therewasaguy:
cc @MaximeRivest

therewasaguy:
hi @MaximeRivest @christianbundy just want to see if you're up for a review, thanks for your time!

christianbundy:
@therewasaguy lgtm, but I'm not a contributor on this repo so my review isn't too helpful 🫡

PR #9659, opened 2026-04-22
feat: make numpy an optional dependency

isaacbmiller:
## Summary

Move `numpy` from a required runtime dependency to an optional extra. 

numpy-dependent features (KNN, Embeddings, SIMBA, MIPROv2, InferRules, DummyVectorizer, Embedder post-processing, etc.) raise a clear `ImportError` pointing at `pip install dspy[numpy]` only when actually invoked.

## Notes

- `import dspy` working without numpy is preserved to avoid a breaking change in the import contract.
- Users who want numpy-backed features install via `pip install dspy[numpy]`.
- Static type checkers (mypy, pyright, ty) will report `numpy` as unresolved if it's not in their environment; the standard fix is to install `dspy[numpy]` in the type-checking env, or scope-ignore per tool (e.g., `[[tool.mypy.overrides]] module = ["numpy.*"]; ignore_missing_imports = true`). No change to repo config needed — our own `uv sync --all-extras` resolves numpy.

Has a minor label, as this is a breaking change and should be grouped with other larger changes rather than in a patch.


isaacbmiller:
Reviewed by @MaximeRivest in [fork pr](https://github.com/cmpnd-ai/dspy/pull/34)

kennethwolters:
Nice work!
Making numpy optional is a good direction for DSPy. I had the same idea when I created dspy-lite in March
Keep core DSPy import/runtime lean, and only require numpy for embeddings/vector-math paths. Glad to see upstream moving this way too :)

PR #9670, opened 2026-04-23
feat(json_adapter): Adds escape hatch to force structured outputs on vLLM and other inference engines 

cezarc1:
_tl;dr: `JSONAdapter` now has explicit response-format control, so users can force strict `json_schema` for self-hosted OpenAI-compatible engines without subclassing `LM` or modifying LiteLLM's registry._

Today, `JSONAdapter`'s default behavior is tied to LiteLLM capability metadata. When LiteLLM does not recognize a model name, `lm.supports_response_schema` can report `False`, so `JSONAdapter` falls back to `{"type": "json_object"}`. For self-hosted model names like `openai/Qwen/...`, `hosted_vllm/...`, or `openai/meta-llama/...`, that means vLLM/SGLang users can get loose JSON mode even when their inference engine supports strict `json_schema` / [structured output mode](https://docs.vllm.ai/en/latest/features/structured_outputs/#experimental-automatic-parsing-openai-api).

This PR adds a `schema_enforcement_mode` kwarg on `JSONAdapter`:

- `"auto"` keeps the existing default behavior: DSPy uses structured outputs when the model is known to support them, otherwise falls back to `json_object`. Automatic structured-output attempts may still fall back to `json_object` if schema construction or the structured call fails.
- `"json_schema"` forces the strict Pydantic response format even when LiteLLM metadata says the model is unsupported. This mode is intentionally strict: schema-build, LiteLLM/provider, and parse errors surface to the caller, and DSPy does not retry with `json_object`.
- `"json_object"` forces `{"type": "json_object"}` for users who want the prompt-guided JSON path even when structured outputs are available.

The PR also tightens schema feasibility checks for open-ended mappings:

- `dict`, `Mapping`, `MutableMapping`, `OrderedDict`, and similar mapping types are treated as open-ended because strict structured outputs require explicit object properties.
- `TypedDict` is excluded because it has a fixed key set and can be represented as a schema.
- `_get_structured_outputs_response_format` uses the same helper, so `Mapping[str, Any]` cannot silently become an incorrect closed empty-object schema.

Warnings still name fields for risky combinations, such as open-ended mappings or `dspy.ToolCalls` with non-native function calling.

Tested locally:

```bash
uv run --extra dev pytest tests/adapters/test_json_adapter.py
uv run --extra dev ruff check dspy/adapters/json_adapter.py
```

Manual validation: tested against vLLM 0.19.x.

## Why not just subclass `LM`?

Mechanically it works:

```python
class VLLMLM(dspy.LM):
    @property
    def supports_response_schema(self) -> bool:
        return True
```

But `ChatAdapter`, DSPy's default adapter, never reads `supports_response_schema`; only `JSONAdapter` does. A user who subclasses `LM` but forgets to also configure `JSONAdapter` gets no structured output and no signal that the override did nothing.

This kwarg puts the opt-in/opt-out at the adapter layer the user configured. Default behavior remains unchanged, and the capability check stays behind `lm.supports_response_schema`, so this should not complicate future LiteLLM decoupling.

### DIFF
--- a/docs/docs/api/adapters/JSONAdapter.md
+++ b/docs/docs/api/adapters/JSONAdapter.md
@@ -19,6 +19,7 @@
             - format_task_description
             - format_user_message_content
             - parse
+            - response_format
             - user_message_output_requirements
         show_source: true
         show_root_heading: true
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -93,8 +93,8 @@ class JSONAdapter(ChatAdapter):
                 except Exception as e:
                     logger.warning(f"Failed to build structured output format, falling back to JSON mode: {e}")
                     return {"type": "json_object"}
-            case _:
-                raise ValueError(f"Invalid schema enforcement mode: {self.schema_enforcement_mode}")
+            case unknown:
+                raise ValueError(f"Invalid schema enforcement mode: {unknown}")
 
     def _effective_enforcement_mode(self, lm: BaseLM, signature: SignatureMeta) -> EffectiveEnforcement:
         open_ended_mapping_fields = _open_ended_mapping_field_names(signature)
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -54,12 +54,14 @@ def _non_native_tool_call_field_names(signature: SignatureMeta) -> Generator[str
             yield name
 
 
-SchemaEnforcement = Literal["auto", "strict", "prompted"]
-"""Which ``response_format`` payload ``JSONAdapter`` ships.
+EffectiveEnforcement = Literal["json_schema", "json_object"]
+
+SchemaEnforcement = Literal["auto"] | EffectiveEnforcement
+"""Which ``response_format`` payload ``JSONAdapter`` ships. Values name the OpenAI-compatible ``response_format.type`` that reaches the wire.
 
 ``"auto"`` delegates to ``lm.supports_response_schema`` unless the signature has open-ended mapping fields or non-native tool calls.
-``"strict"`` forces a Pydantic class (serialized to strict ``json_schema``) -- use for known openai-compatible endpoints.
-``"prompted"`` forces ``{"type": "json_object"}`` with schema hints in the prompt.
+``"json_schema"`` forces the Pydantic-class response_format, shipped as ``{"type": "json_schema", "strict": true, "schema": {...}}``; use for known openai-compatible endpoints that can use structured outputs.
+``"json_object"`` forces ``{"type": "json_object"}`` with schema hints carried in the prompt; escape hatch for LM-known models whose strict-schema backend is broken.
 """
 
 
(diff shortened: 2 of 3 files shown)


isaacbmiller:
Are there any ways to get this behavior without modifying json adapter?

Can you either:
1. Subclass LM and override the lm.supports_response_schema for the models you care about
2. Add them into the litellm registry while you run? Via modifying their [model_prices_and_context dict](https://docs.litellm.ai/docs/provider_registration/add_model_pricing#sample-spec)?

cezarc1:
> Are there any ways to get this behavior without modifying json adapter?
> 
> Can you either:
> 
> 1. Subclass LM and override the lm.supports_response_schema for the models you care about
> 2. Add them into the litellm registry while you run? Via modifying their [model_prices_and_context dict](https://docs.litellm.ai/docs/provider_registration/add_model_pricing#sample-spec)?

I did consider those and both options should work, _but they move the fix to layers the user didn't configure_. The root cause here is that `JSONAdapter` exhibits a pretty clear [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/) for the encoding details.

Today, fixing it means descending a layer (subclass `LM`) or _two_ (modify LiteLLM's registry) _below_ the JSONAdapter the user actually configured. Both options require them to learn/know about abstractions below the one they chose, exactly the shape of leak a framework-level kwarg is meant to seal.

I also want to point out a subtle UX trap with the LM-subclass route specifically: `ChatAdapter` is DSPy's default adapter, and it **doesn't consult lm.supports_response_schema** at all (nor should it!), only JSONAdapter does. So a user who subclasses LM but doesn't also swap in JSONAdapter gets a silent no-op:
```python
class MyFancyLM(dspy.LM):
    @property
    def supports_response_schema(self) -> bool:
        return True  # trust me, my vLLM supports it 🤞

dspy.configure(lm=MyLM("openai/my-vllm-model", api_base="..."))
# Default ChatAdapter is in use, the override is never read!
# No error, no warning, no structured outputs. Configuration does nothing.
# The only way to tell is if by chance pull out the traces (and we all know how much engineers like looking at traces) and be perplexed.
# I think coding agents would be able to spot and fix it, but imo, they also get this wrong as well on if  they 1-shotted it.
```

This new kwarg gives users direct opt-in/opt-out at the adapter layer. Default behavior is unchanged, no new LiteLLM coupling is introduced. One extra init kwarg is a small price for removing the leak 😊.

updated the PR description to account for your suggestions.

isaacbmiller (changes requested):
Thanks for working through this use case, Cezar, and apologies that this sat without a complete disposition for so long. The problem is accepted: users of self-hosted OpenAI-compatible backends need a supported way to override incorrect or missing LiteLLM capability metadata. The current patch needs redesign before merge, rather than a mechanical conflict resolution.

The capability override belongs on the LM/backend boundary, where DSPy already exposes `BaseLM.supports_response_schema` and `supported_params`, rather than as a second source of truth on `JSONAdapter`. Please rebase and narrow this toward an optional per-`dspy.LM` capability override, for example `supports_response_schema: bool | None = None`: `None` delegates to LiteLLM, while `True` also makes `response_format` available to adapter planning. The override must be DSPy configuration, not forwarded as a provider request kwarg. `JSONAdapter` can then keep consuming the LM capability normally.

A current-main rework also needs to:

- preserve #9826's `except LMError: raise` behavior in sync and async paths; provider/backend failures must make exactly one request, not retry in `json_object` mode;
- retain the current `parallel_tool_calls` constructor argument;
- drop the open-mapping detection/warnings from this PR, since #10009 handles nested mappings from the generated JSON Schema (the actual provider boundary);
- avoid making `response_format()` a new public adapter method;
- use a real runtime exception rather than `assert` for any public argument validation.

Focused sync/async coverage showing that the LM override sends the Pydantic response schema despite false LiteLLM metadata, plus one-call propagation on provider failure, should be enough. This keeps the useful escape hatch while preserving one owner for backend capabilities and current error semantics.

PR #9671, opened 2026-04-23
feat(json_adapter): make auto-fallback visible and raise structured-output failures to ERROR

cezarc1:
Two small observability improvements to `JSONAdapter` w/ no public changes.

1) _Warnings on fixable auto-fallback causes._`JSONAdapter` silently drops to `{"type": "json_object"}` when the signature has a `dict[...]` output or uses `ToolCalls` without native function calling. Both are user-fixable — simplify the signature, or flip `use_native_function_calling=True`. 

**This PR adds `logger.warning` at both sites explaining why and what to change.** The third fallback case (`supports_response_schema=False`) stays silent.

2) _Bump structured-output-failure log to ERROR._ When the structured-output call raises at the wire (e.g., provider rejects the schema), `__call__`/`acall` catch, log, and retry with `{"type": "json_object"}`. Today that log is `warning`; 

**This PR bumps it to `error` and includes the exception reason.**

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Any, get_origin
+from typing import Any, Generator, get_origin
 
 import json_repair
 import pydantic
@@ -24,17 +24,21 @@ from dspy.utils.exceptions import AdapterParseError
 logger = logging.getLogger(__name__)
 
 
-def _has_open_ended_mapping(signature: SignatureMeta) -> bool:
+def _open_ended_mapping_field_names(signature: SignatureMeta) -> Generator[str, None, None]:
+    """Yield names of output fields typed as ``dict[...]``. Structured Outputs require
+    explicit properties, so such fields are incompatible and force a fallback to
+    ``{"type": "json_object"}``.
     """
-    Check whether any output field in the signature has an open-ended mapping type,
-    such as dict[str, Any]. Structured Outputs require explicit properties, so such fields
-    are incompatible.
-    """
-    for field in signature.output_fields.values():
-        annotation = field.annotation
-        if get_origin(annotation) is dict:
-            return True
-    return False
+    for name, field in signature.output_fields.items():
+        if get_origin(field.annotation) is dict:
+            yield name
+
+
+def _non_native_tool_call_field_names(signature: SignatureMeta) -> Generator[str, None, None]:
+    """Yield names of output fields with ``dspy.ToolCalls`` annotations."""
+    for name, field in signature.output_fields.items():
+        if field.annotation == ToolCalls:
+            yield name
 
 
 class JSONAdapter(ChatAdapter):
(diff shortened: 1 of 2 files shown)


dspy/adapters/json_adapter.py
```diff
@@ -24,17 +24,21 @@
 logger = logging.getLogger(__name__)
 
 
-def _has_open_ended_mapping(signature: SignatureMeta) -> bool:
+def _open_ended_mapping_field_names(signature: SignatureMeta) -> Generator[str, None, None]:
```
isaacbmiller: Can these two functions just return lists of names instead of generators?

Also can we name it with a verb like `_get_open_ended_mapping_field_names`

dspy/adapters/json_adapter.py
```diff
@@ -24,17 +24,21 @@
...
+    """Yield names of output fields typed as ``dict[...]``. Structured Outputs require
+    explicit properties, so such fields are incompatible and force a fallback to
+    ``{"type": "json_object"}``.
```
isaacbmiller: Prefer single backticks inside of doc strings

isaacbmiller (changes requested):
Two nits, otherwise LGTM!

Thank you for the PR!

dspy/adapters/json_adapter.py
cezarc1: agreed! done.

dspy/adapters/json_adapter.py
cezarc1: done.

isaacbmiller (changes requested):
Thanks for the observability work, Cezar, and apologies that the follow-up sat for so long after the earlier nits were addressed. Current `main` has changed enough that this cannot be merged as-is, but one correctness fix here remains valuable.

Since #9826, provider/backend failures are normalized to `LMError` and propagated by `JSONAdapter` with exactly one provider request. The ERROR-log tests in this branch expect the old behavior—a wire failure being caught and retried with `json_object`—so that portion is now superseded and must be dropped. A propagated failure already retains its cause for the caller to log; a locally recovered schema/setup fallback should remain at most a warning.

The new signature warnings also execute on every adapter call. In evaluation and optimization loops that can emit thousands of identical warnings for the same signature. Please either omit them from this PR or make diagnostics non-repeating per stable signature/cause; the open-mapping diagnostic should also follow #10009's generated-schema detection rather than the direct top-level `dict` check here.

The independently useful fix is `if result is not None:` in both sync and async paths. Current `main` still uses a truthiness check, so a legitimate empty-list result can fall through and issue an unintended second call. A mergeable rework would be:

1. rebase on current `main` and preserve the `LMError` guards and `parallel_tool_calls` support;
2. remove the ERROR-level provider-failure/retry changes and their old-contract tests;
3. keep the two sentinel changes with one focused regression test asserting that an empty result is returned with exactly one LM call;
4. either defer the fallback warnings or make them non-repeating and compatible with #10009.

If preferred, the sentinel fix alone would make an excellent fresh, very small PR and should not be blocked on the observability design.

Issue #9574, opened 2026-04-07
[Bug] ParallelEvaluate `causes cannot schedule new futures after shutdown on long evals`


isaacbmiller:
dspy.Evaluate doesn't expose timeout. ParallelExecutor's 120s straggler timeout causes cannot schedule new futures after shutdown on long evals.

source: https://x.com/kargarisaac/status/2041596855172424159

isaacbmiller:
Help wanted on this one!

If you could investigate a root cause before opening a PR, that would be the best way

NishchayMahor:
Hey @isaacbmiller :), digging into this one. Sharing what I found so the analysis is in the open:

**Where the symptom comes from.** `ParallelExecutor._execute_parallel` submits all `N` items into a `ThreadPoolExecutor(max_workers=num_threads)` and then runs its main wait loop. The straggler-resubmit pass only fires when `0 < self.timeout and len(not_done) <= self.straggler_limit` (so by default: only when 3 or fewer items remain). For each straggler whose `start_time` is older than 120s, it calls `executor.submit(...)` to resubmit the same work item, leaving the original future alive in `futures_set` too (the first one to finish wins via the `results[index] is None` guard).

**Why long evals trip it.** The 120s default is fine for most assertion-style metrics, but it is a structural assumption that individual eval items should complete inside roughly two minutes. For reasoning-heavy models, long-context prompts, slow tool calls, or rate-limited providers, that assumption breaks. Concretely:

1. Most items finish in normal time, leaving 1-3 stragglers (legitimately slow, not hung).
2. The straggler pass fires at the 120s mark and resubmits them.
3. Worker slots are already saturated by the still-running originals, so the resubmits queue.
4. Whichever future finishes first wins; the loser still occupies a worker slot until it returns or the executor shuts down.
5. When the wait loop finally exits and the `finally` block hits `executor.shutdown(wait=False)`, any straggler resubmit that landed in the queue without being picked up is now in an executor that no longer accepts work, hence `cannot schedule new futures after shutdown` surfacing on the user side.

So the underlying bug is not really "`ParallelExecutor` is broken." It is "`ParallelExecutor`'s straggler defaults are workload-specific and `Evaluate` does not let the user override them." A deeper rewrite (per-resubmit cap, smarter backoff, drain-before-shutdown handshake) is reasonable, but it is a much bigger change and the right value for the timeout would still depend on the workload.

[...]

PR #9588, opened 2026-04-11
fix(evaluate): expose timeout and straggler_limit params in Evaluate

vedaant00:
## Summary

`Evaluate` doesn't expose `ParallelExecutor`'s `timeout` and `straggler_limit` params. The hardcoded 120s straggler timeout causes `cannot schedule new futures after shutdown` errors on long-running evaluations (e.g., LLM-as-judge metrics that take >2 minutes per example).

This follows the same pattern as #9199, which exposed these params in `Parallel` and `Module.batch()`.

## Changes

- Add `timeout` and `straggler_limit` params to `Evaluate.__init__()` and `Evaluate.__call__()`
- Pass them through to `ParallelExecutor` only when explicitly set (preserves existing defaults)
- Setting `timeout=0` disables straggler resubmission entirely

## Usage

```python
# Disable straggler resubmission for long evals
evaluator = dspy.Evaluate(devset=devset, metric=metric, timeout=0)

# Custom timeout (e.g., 10 minutes)
evaluator = dspy.Evaluate(devset=devset, metric=metric, timeout=600)

# Override per-call
evaluator(program, timeout=0, straggler_limit=5)
```

## Tests

3 new tests:
- `test_evaluate_timeout_params_passed_to_executor` — verifies params reach `ParallelExecutor`
- `test_evaluate_timeout_defaults_omit_params` — verifies defaults don't override `ParallelExecutor` defaults
- `test_evaluate_timeout_zero_disables_straggler_resubmission` — end-to-end with slow LM

Closes #9574

PR #9675, opened 2026-04-25
fix(parallel): guard straggler resubmission against executor shutdown

weiguangli-io:
## Summary

- Guard `executor.submit()` in `ParallelExecutor`'s straggler resubmission path with `try/except RuntimeError` to prevent `"cannot schedule new futures after shutdown"` crashes on long-running evaluations
- When the executor is already shut down (e.g. due to cancellation, signal handling, or interpreter exit), the resubmission is gracefully skipped instead of raising

## Context

When `ParallelExecutor`'s straggler timeout (default 120s) fires and attempts to resubmit slow futures, the `ThreadPoolExecutor` may already be shut down, causing a `RuntimeError`. This is a defensive fix at the executor level — complementary to #9588 which exposes `timeout`/`straggler_limit` params so users can tune or disable straggler resubmission.

## Test plan

- [ ] Existing `tests/utils/test_parallelizer.py` tests pass
- [ ] Verify that long-running evaluations no longer crash with `RuntimeError: cannot schedule new futures after shutdown`

Closes #9574

### DIFF
--- a/dspy/utils/parallelizer.py
+++ b/dspy/utils/parallelizer.py
@@ -209,13 +209,17 @@ class ParallelExecutor:
                                     st = start_time_map.get(sid, None)
                                 if st and (now - st) >= self.timeout:
                                     resubmitted.add(f)
-                                    nf = executor.submit(
-                                        worker,
-                                        parent_overrides,
-                                        submission_counter,
-                                        idx,
-                                        item,
-                                    )
+                                    try:
+                                        nf = executor.submit(
+                                            worker,
+                                            parent_overrides,
+                                            submission_counter,
+                                            idx,
+                                            item,
+                                        )
+                                    except RuntimeError:
+                                        # Executor already shut down; skip resubmission
+                                        break
                                     futures_map[nf] = (submission_counter, idx, item)
                                     futures_set.add(nf)
                                     submission_counter += 1


WatchTree-19:
useful catch on the resubmission path. three observations:

1. **bare `except RuntimeError`** swallows any `RuntimeError` from `executor.submit`, not just the shutdown one. if a future Python release raises a different `RuntimeError` from this path, the eval would silently break out of straggler resubmission without flagging. matching the specific message keeps the safety net narrow:

```python
except RuntimeError as exc:
    if "cannot schedule new futures after shutdown" not in str(exc):
        raise
    # executor already shut down; skip resubmission
    break
```

2. **after the `break`**, the outer `while futures_set and not self.cancel_jobs.is_set()` loop continues. since the executor is shut down, no new futures will be accepted and `wait(timeout=1)` will spin until the existing in-flight futures resolve on their own (or never). setting `self.cancel_jobs.set()` alongside the `break` would let the outer loop exit cleanly and surface the cancellation to the caller via the existing "Execution cancelled" path:

```python
except RuntimeError as exc:
    if "cannot schedule new futures after shutdown" not in str(exc):
        raise
    logger.warning("executor shut down before straggler resubmission; cancelling")
    self.cancel_jobs.set()
    break
```

3. **test plan checkboxes are unchecked** in the PR body. a small unit test that mocks `ThreadPoolExecutor.submit` to raise `RuntimeError("cannot schedule new futures after shutdown")` on the resubmit branch would lock in the regression and be cheap to add. happy to suggest a sketch if useful.

minor nits, fix is the right shape.

isaacbmiller:

PR #9689, opened 2026-04-28
fix: strip x-* vendor extensions from schemas sent to LM providers

### LINKED ISSUE #9686: JSONAdapter leaks Pydantic `x-*` vendor extensions to LM providers (rejected by strict schemas)
# JSONAdapter leaks Pydantic `x-*` vendor extensions to LM providers (rejected by strict schemas)

## Summary

When a `dspy.Signature` uses a Pydantic model as an output field, `JSONAdapter` emits the model's JSON schema to the LM provider (as `response_format` / `toolSpec.inputSchema`) and embeds a schema snippet in the prompt text. If any Pydantic field in the tree has `json_schema_extra={"x-…": …}`, Pydantic merges those keys into the schema as peer keys (e.g. alongside `anyOf`, `type`). Strict-schema providers reject them.

On AWS Bedrock Converse:

```
BedrockException - 400:
  {"message":"The model returned the following errors: For 'anyOf', 'x-comparison' is not supported"}
```

The adapter then falls back to plain JSON mode (`{"type": "json_object"}`), the model returns `{}`, and every downstream prediction is empty.

`x-*` is the OpenAPI / JSON Schema convention for vendor-only, non-standard extensions. Libraries commonly use it for observability, scoring, or validation metadata that should never reach an LM.

## Environment

- `dspy` 3.2.0
- `pydantic` 2.12.4
- Python 3.12.9
- Provider tested: AWS Bedrock Converse (Claude Haiku 4.5)
- Adapter: `dspy.JSONAdapter(use_native_function_calling=True)`

## Minimal repro (offline — no provider call needed)

This drives DSPy's own schema-assembly path and shows the leaked key
without needing a live LM:

```python
from typing import Optional
import pydantic
import dspy

class Form(pydantic.BaseModel):
    name: Optional[str] = pydantic.Field(
        default=None,
        json_schema_extra={"x-comparison": {"comparator": "exact"}},
    )

class Extract(dspy.Signature):
    text: str = dspy.InputField()
    form: Form = dspy.OutputField()

model = _get_structured_outputs_response_format(
    Extract, use_native_function_calling=True
)
schema = model.model_json_schema()
assert "x-comparison" in str(schema)  # leaks into $defs.Form.properties.name
```

When this schema is handed to LiteLLM as `response_format`, Bedrock rejects the request with the 400 shown above.

The same leak also appears in the **prompt text** via `translate_field_type`, which calls `pydantic.TypeAdapter(field_type).json_schema()`:

https://github.com/stanfordnlp/dspy/blob/da4ae1941d551fdc09d7d1bfbb6f7c01b96063a8/dspy/adapters/utils.py#L109-L134

## Root cause

Pydantic merges `json_schema_extra` dict entries into the generated schema as peer keys (not as a nested `"json_schema_extra"` object). Two DSPy call sites produce schemas that carry those peer keys through to the LM:

1. `pydantic.TypeAdapter(field_type).json_schema()` — reached by `translate_field_type` → embedded in prompt text for every adapter (Chat/XML/JSON) and by `dspy/predict/rlm.py`:

   https://github.com/stanfordnlp/dspy/blob/da4ae1941d551fdc09d7d1bfbb6f7c01b96063a8/dspy/adapters/utils.py#L88

2. `pydantic.create_model(...).model_json_schema()` — handed to LiteLLM as `response_format`. This is the one Bedrock rejects:
   https://github.com/stanfordnlp/dspy/blob/da4ae1941d551fdc09d7d1bfbb6f7c01b96063a8/dspy/adapters/json_adapter.py#L243-L254

### Existing dead code

The scrub already attempted before the schema is sent:

https://github.com/stanfordnlp/dspy/blob/da4ae1941d551fdc09d7d1bfbb6f7c01b96063a8/dspy/adapters/json_adapter.py#L252-L254

On current Pydantic (≥ 2.x, confirmed 2.12.4) this is a no-op — Pydantic does not emit a literal `"json_schema_extra"` key; it merges the dict's contents into the property as siblings. The comment documents the correct intent; the implementation no longer matches. Any `BaseModel.model_json_schema()` with a `json_schema_extra={...}` field demonstrates this:

```python
import pydantic

class A(pydantic.BaseModel):
    x: str = pydantic.Field(json_schema_extra={"x-foo": True, "desc": "hi"})

schema = A.model_json_schema()
prop = schema["properties"]["x"]
assert "json_schema_extra" not in prop  # nothing to pop
assert "x-foo" in prop                  # but this is what's actually there
assert "desc"  in prop
```

## Proposed fix

PR: https://github.com/stanfordnlp/dspy/pull/9692

fanxing11:
Fixes #9686

Pydantic 2.x merges `json_schema_extra` dict entries as sibling keys (e.g. `x-comparison`) rather than nesting them under a `json_schema_extra` key. The existing `pop('json_schema_extra')` in `json_adapter.py` was a no-op, so vendor extensions leaked into both the `response_format` schema and the prompt text.

Strict-schema providers like AWS Bedrock Converse reject these unknown keys:

```
BedrockException - 400: For 'anyOf', 'x-comparison' is not supported
```

The adapter then falls back to plain JSON mode and returns empty predictions.

### Changes

- Added `_strip_vendor_extensions()` recursive helper in `utils.py` that removes all `x-*` keys from a schema dict
- Applied it in `_get_json_schema()` (prompt text path)
- Replaced the dead `pop()` with `_strip_vendor_extensions()` in `_get_structured_outputs_response_format()` (response_format path)
- Two regression tests confirming both paths are clean

All existing tests pass (`pytest tests/adapters/test_json_adapter.py` — 30 passed).

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -10,6 +10,7 @@ from pydantic.fields import FieldInfo
 from dspy.adapters.chat_adapter import ChatAdapter, FieldInfoWithName
 from dspy.adapters.types.tool import ToolCalls
 from dspy.adapters.utils import (
+    _strip_vendor_extensions,
     format_field_value,
     get_annotation_name,
     parse_value,
@@ -249,9 +250,9 @@ def _get_structured_outputs_response_format(
     # Generate the initial schema.
     schema = pydantic_model.model_json_schema()
 
-    # Remove any DSPy-specific metadata.
-    for prop in schema.get("properties", {}).values():
-        prop.pop("json_schema_extra", None)
+    # Strip vendor-extension keys (x-*) that Pydantic 2.x merges from
+    # json_schema_extra.  Strict-schema providers (e.g. Bedrock) reject them.
+    schema = _strip_vendor_extensions(schema)
 
     def enforce_required(schema_part: dict):
         """
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -74,6 +74,22 @@ def format_field_value(field_info: FieldInfo, value: Any, assume_text=True) -> s
         return {"type": "text", "text": string_value}
 
 
+def _strip_vendor_extensions(schema):
+    """Recursively remove ``x-*`` vendor-extension keys from a JSON schema.
+
+    Pydantic 2.x merges ``json_schema_extra`` dict entries as sibling keys
+    (e.g. ``x-comparison``) rather than nesting them under a
+    ``json_schema_extra`` key.  These vendor extensions are not part of the
+    JSON Schema standard that LM providers validate against, so strict-schema
+    providers (e.g. AWS Bedrock) reject them.
+    """
+    if isinstance(schema, dict):
+        return {k: _strip_vendor_extensions(v) for k, v in schema.items() if not (isinstance(k, str) and k.startswith("x-"))}
+    elif isinstance(schema, list):
+        return [_strip_vendor_extensions(item) for item in schema]
+    return schema
+
+
 def _get_json_schema(field_type):
     def move_type_to_front(d):
         # Move the 'type' key to the front of the dictionary, recursively, for LLM readability/adherence.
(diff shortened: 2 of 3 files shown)


shabie:
I've created the PR for my own issue so this one can be closed: https://github.com/stanfordnlp/dspy/pull/9692

fanxing11:
Closing in favor of #9692 — @shabie's version covers the same fix with tests. Thanks!

PR #9705, opened 2026-05-01
fix: raise ValueError instead of UnboundLocalError when no valid program found

Ricardo-M-L:
## Summary

`get_program_with_highest_avg_score` would fall through to a return statement referencing undefined variables (`program`, `mean`, `key`, `params`) when `param_score_dict` was empty. This causes an `UnboundLocalError` instead of a clear error.

Fix: replace the ambiguous fallback return with `raise ValueError("No valid program found in param_score_dict")`, giving callers clear feedback when no valid program exists.

## Test plan
- [ ] Verify optimizer handles empty param_score_dict with ValueError
- [ ] Existing optimizer tests should pass

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -780,7 +780,7 @@ class MIPROv2(Teleprompter):
                 predictor.demos = demo_candidates[i][demos_idx]
                 trial_logs[trial_num][f"{i}_predictor_demos"] = demos_idx
                 chosen_params.append(f"Predictor {i}: Few-Shot Set {demos_idx}")
-                raw_chosen_params[f"{i}_predictor_demos"] = instruction_idx
+                raw_chosen_params[f"{i}_predictor_demos"] = demos_idx
 
         return chosen_params, raw_chosen_params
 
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Looks like this is on the wrong branch?/mixed with another change.

The MIPRO bug is crazy lol. I am not sure how that one crept through for so long!

iamsharduld:
Thanks for catching this — the empty-dict `UnboundLocalError` is real.

One thing before the `raise` lands: I think that fallthrough is also hit when `param_score_dict` is non-empty but all its combos are already in `fully_evaled_param_combos`. Full evals fire on a fixed schedule (`trial_num % (minibatch_full_eval_steps + 1) == 0`, plus the last trial) regardless of whether a new combo exists, so once the minibatches keep reusing the same few combos, a later full eval finds them all evaluated. Today that falls through to the last (lowest-mean) combo — already a bug — but `raise ValueError` would turn it into a crash mid-run:

```python
from dspy.teleprompt.utils import get_program_with_highest_avg_score
d = {"A": [(100.0, "pA", {})], "B": [(50.0, "pB", {})], "C": [(10.0, "pC", {})]}
print(get_program_with_highest_avg_score(d, {"A": {}, "B": {}, "C": {}}))  # ('pC', 10.0, 'C', {})
```

Could we distinguish the two — error on a genuinely empty dict, but treat "all combos already evaluated" as a no-op (skip the redundant full eval)? Happy to send a fix + test. Thanks for digging into this!

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

isaacbmiller:
@greptile

PR #9708, opened 2026-05-01
fix(clients): honor caching=False per-call in dspy.Embedder

spjosyula:
`Embedder._preprocess` used `caching or self.caching`, silently
  ignoring `caching=False` when the instance default was True. Switched
   to an explicit None-sentinel check. Applies to both `__call__` and
  `acall`. Introduced incidentally in #8080.

  Added 4 tests for the override matrix; inherit cases already covered
  by `test_litellm_embedding`.

  ## Warnings
  Used Claude (Anthropic) for code archaeology and test scaffolding.
  Verified by running tests pre- and post-fix.

### DIFF
--- a/dspy/clients/embedding.py
+++ b/dspy/clients/embedding.py
@@ -93,7 +93,7 @@ class Embedder:
             raise ValueError("All inputs must be strings.")
 
         batch_size = batch_size or self.batch_size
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Looks great, thank you for the pr!

PR #9733, opened 2026-05-07
refactor: remove asyncer dependency, inline with direct anyio calls

isaacbmiller:
Remove `asyncer==0.0.8` as a dependency. The library is a thin wrapper around `anyio` that adds kwargs support via `functools.partial`. DSPy already depends on `anyio` directly.

The entire runtime value of asyncer is two functions:

- [`asyncify`](https://github.com/fastapi/asyncer/blob/32fb323c9c4b097f84d52ff251fe88d0f79f85f0/asyncer/_main.py#L370-L377) -- `functools.partial(fn, *args, **kwargs)` then `await anyio.to_thread.run_sync(partial_f, ...)`
- [`syncify`](https://github.com/fastapi/asyncer/blob/32fb323c9c4b097f84d52ff251fe88d0f79f85f0/asyncer/_main.py#L297-L310) -- `functools.partial(fn, *args, **kwargs)` then `anyio.from_thread.run(partial_f)`

This also eliminates the exact-version pin (`==0.0.8`), which was a version conflict risk for downstream users.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import os
 import re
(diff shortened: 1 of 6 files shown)


MaximeRivest (approved):

MaximeRivest:
LGTM

PR #9735, opened 2026-05-07
refactor: replace typeguard with stdlib type checking

isaacbmiller:
## Summary

Remove `typeguard==4.5.1` by replacing its single call site with a ~40-line stdlib `_check_type()` function.

## Context

typeguard was used in exactly one place: `_is_value_compatible_with_type()` in `dspy/predict/predict.py`, which emits a warning when input field types don't match the signature. It used `check_type()` and caught `TypeCheckError`.

## Changes

- **`dspy/predict/predict.py`**: Replace `typeguard.check_type` with `_check_type()` using `get_origin`/`get_args` from `typing`. Handles:
  - `Union` / `X | None` (Optional)
  - `Literal[...]`
  - `list[T]` with element checking
  - `dict[K, V]` with key/value checking
  - `tuple[T1, T2, ...]` (fixed-length) and `tuple[T, ...]` (variadic)
  - `set[T]` / `frozenset[T]`
  - Plain types (`int`, `str`, BaseModel subclasses, etc.)
  - `Any` (always passes)
- **`pyproject.toml`**: Remove `typeguard==4.5.1` from dependencies

## Testing

All 75 predict tests pass, including all 21 type-validation tests covering: basic types, `list[str]`, `list[int]`, `dict[str, int]`, `tuple[str, int, bool]`, `tuple[int, ...]`, `Literal`, `Literal | None`, string signatures.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -1,10 +1,10 @@
 import logging
 import random
-from typing import Any, Literal, get_args, get_origin
+import types
+from typing import Any, Literal, Union, get_args, get_origin
 
 from pydantic import BaseModel
 from pydantic_core import PydanticUndefined
-from typeguard import TypeCheckError, check_type
 
 from dspy.adapters.chat_adapter import ChatAdapter
 from dspy.clients.base_lm import BaseLM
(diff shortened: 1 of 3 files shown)


isaacbmiller:
reviewed by @michaelisaac-dev and discussed offline

PR #9741, opened 2026-05-12
fix: make load_state transactional — validate then deepcopy (#9589)

### LINKED ISSUE #9589: Module.load_state silently corrupts modules on partial failure (non-transactional)

ashishSoni1234:
Fixes #9589

I have been following this thread since the issue was filed and reviewed all three closed PRs before writing this.

**What was wrong**

load_state walked named_parameters() and mutated each one in place.  If state[name] raised on parameter N, everything before N was already  overwritten. The module still passed isinstance checks, looked  structurally valid, but was serving inference from a mix of saved demos and fresh defaults — with no way to tell it apart from a clean template. brm444 tracked this down after days of debugging a RAG quality regression. The error was logged at startup but the module looked fine 
afterward, so it got attributed to model drift.

**What this PR does**

Two things, in order:
First, collect all parameter names and check that every one exists in state before touching anything. If a key is missing the exception fires immediately and self is completely unchanged. This closes the common case — signature drift, truncated save file, manual edit — at zero cost.

Second, per @isaacbmiller's preference, apply the full load to a  throwaway deepcopy of self before mutating the original. If 
param.load_state() blows up on a malformed value mid-loop, self is still in its original state. The validate step alone can't catch this because the key is present but the value is corrupt.

**Why not the three closed PRs**

#9590 and #9657 did validate-first but skipped the deepcopy, so a corrupt value mid-loop still left self half-hydrated.

#9655 did snapshot-and-rollback via dump_state(), but the rollback itself calls load_state() again — which is exactly the function that just failed. deepcopy sidesteps this entirely.

**Reproducer from the issue**

Before:
BEFORE: template.a.predict.demos = []
load raised: KeyError: 'b.predict'
AFTER:  template.a.predict.demos = [sentinel]  ← corrupted

After:
BEFORE: template.a.predict.demos = []
load raised: KeyError: 'b.predict'
AFTER:  template.a.predict.demos = []  ← clean ✅

**Tests**
47 passed, 45 skipped, 11 warnings

Added one regression test: when load fails on a missing key, every parameter stays at its pre-load value.

### DIFF
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -155,7 +155,7 @@ class BaseModule:
 
     def dump_state(self, json_mode=True):
         return {name: param.dump_state(json_mode=json_mode) for name, param in self.named_parameters()}
-    
+
     def load_state(self, state, *, allow_unsafe_lm_state=False):
         from dspy.predict.predict import Predict
 
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -159,34 +159,15 @@ class BaseModule:
     def load_state(self, state, *, allow_unsafe_lm_state=False):
         from dspy.predict.predict import Predict
 
-        # Collect params upfront so we iterate the same list in all three passes
-        param_items = list(self.named_parameters())
+        def _apply(module):
+            for name, param in module.named_parameters():
+                if isinstance(param, Predict):
+                    param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
+                else:
+                    param.load_state(state[name])
 
-        # Reject early if any key is absent — nothing has been touched yet
-        missing = [name for name, _ in param_items if name not in state]
-        if missing:
-            raise KeyError(
-                f"load_state aborted — saved state is missing keys: {missing}. "
-                "No parameters were modified. This usually indicates signature "
-                "drift between the saved state and the current module."
-            )
-
-        # Try the whole load on a throwaway copy first.
-        # If param.load_state() blows up on a malformed value,
-        # self is still in its original state.
-        trial_copy = self.deepcopy()
-        for name, param in trial_copy.named_parameters():
-            if isinstance(param, Predict):
-                param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
-            else:
-                param.load_state(state[name])
-
-        # Trial copy completed without error — now apply to self
-        for name, param in self.named_parameters():
-            if isinstance(param, Predict):
-                param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
-            else:
-                param.load_state(state[name])
+        _apply(self.deepcopy())  # trial run raises before self is touched
+        _apply(self)
     def save(self, path, save_program=False, modules_to_serialize=None):
         """Save the module.
 
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -155,16 +155,38 @@ class BaseModule:
 
     def dump_state(self, json_mode=True):
         return {name: param.dump_state(json_mode=json_mode) for name, param in self.named_parameters()}
-
+    
     def load_state(self, state, *, allow_unsafe_lm_state=False):
         from dspy.predict.predict import Predict
 
-        for name, param in self.named_parameters():
+        # Collect params upfront so we iterate the same list in all three passes
+        param_items = list(self.named_parameters())
+
+        # Reject early if any key is absent — nothing has been touched yet
+        missing = [name for name, _ in param_items if name not in state]
+        if missing:
+            raise KeyError(
+                f"load_state aborted — saved state is missing keys: {missing}. "
+                "No parameters were modified. This usually indicates signature "
+                "drift between the saved state and the current module."
+            )
+
+        # Try the whole load on a throwaway copy first.
+        # If param.load_state() blows up on a malformed value,
+        # self is still in its original state.
+        trial_copy = copy.deepcopy(self)
+        for name, param in trial_copy.named_parameters():
             if isinstance(param, Predict):
                 param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
             else:
                 param.load_state(state[name])
 
+        # Trial copy completed without error — now apply to self
+        for name, param in self.named_parameters():
+            if isinstance(param, Predict):
+                param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
+            else:
+                param.load_state(state[name])
     def save(self, path, save_program=False, modules_to_serialize=None):
         """Save the module.
 
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -174,7 +174,7 @@ class BaseModule:
         # Try the whole load on a throwaway copy first.
         # If param.load_state() blows up on a malformed value,
         # self is still in its original state.
-        trial_copy = copy.deepcopy(self)
+        trial_copy = self.deepcopy()
         for name, param in trial_copy.named_parameters():
             if isinstance(param, Predict):
                 param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
--- a/tests/primitives/test_module.py
+++ b/tests/primitives/test_module.py
@@ -197,3 +197,49 @@ def test_load_dspy_program_cross_version():
 
     assert len(loaded_react.react.demos) == 2
     assert len(loaded_react.extract.predict.demos) == 2
+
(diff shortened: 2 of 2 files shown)


dspy/primitives/base_module.py
```diff
@@ -155,16 +155,38 @@ def reset_copy(self):
 
     def dump_state(self, json_mode=True):
         return {name: param.dump_state(json_mode=json_mode) for name, param in self.named_parameters()}
-
+    
     def load_state(self, state, *, allow_unsafe_lm_state=False):
```
isaacbmiller, line 159: You can combine both of these checks:
```
def load_state(self, state, *, allow_unsafe_lm_state=False):
    from dspy.predict.predict import Predict

    def _apply(module):
        for name, param in module.named_parameters():
            if isinstance(param, Predict):
                param.load_state(state[name], allow_unsafe_lm_state=allow_unsafe_lm_state)
            else:
                param.load_state(state[name])

    _apply(self.deepcopy())  # trial run — raises before self is touched
    _apply(self)
```

tests/primitives/test_module.py
```diff
@@ -197,3 +197,49 @@ def test_load_dspy_program_cross_version():
...
     assert len(loaded_react.extract.predict.demos) == 2
+
+def test_load_state_is_transactional():
+    """
+    Regression test for https://github.com/stanfordnlp/dspy/issues/9589
+
+    load_state must be all-or-nothing. If it fails mid-load (missing key
+    or malformed value), the module must be completely unchanged.
+    """
+    import json
```
isaacbmiller: Move imports to top of file

isaacbmiller (changes requested):
Thanks for the Pr!

Left some comments

ashishSoni1234:
@isaacbmiller Both changes addressed  extracted _apply helper and moved test imports to top. Ready for another look when you get a chance.

dspy/primitives/base_module.py
ashishSoni1234: Done — extracted _apply helper as suggested.

tests/primitives/test_module.py
ashishSoni1234: Done — moved imports to top of file.

isaacbmiller:
Great, thanks for the help!

isaacbmiller:
Once you pass linting lgtm

ashishSoni1234:
@isaacbmiller Linting fixed and pushed all ruff checks passing now !!

ashishSoni1234:
@isaacbmiller Thank you for the thorough review and guidance  learned a lot!

isaacbmiller:
Thank you for looking at the closed issues and trying to understand what was going on before making a PR!

Looking forward to seeing future issues/prs. Feel free to raise something if you see it in your own dspy use.

PR #9744, opened 2026-05-12
feat(inspect_history): render tool_calls and tool-role messages

isaacbmiller:
## What

`pretty_print_history` previously dropped any `tool_calls` array on assistant messages and could not render `tool` role messages at all, making debugging native-function-calling traces impossible. This PR adds rendering for both.

## Changes

- **`_get(obj, key, default)`** helper accepts both dicts and pydantic objects so the renderer works with raw OpenAI wire-format payloads *and* DSPy-native `ToolCall` instances.
- **`_format_tool_call`** handles both wire shapes:
  - OpenAI Chat Completions: `{function: {name, arguments}}`
  - Flat: `{name, args}` (DSPy-native) and `{name, arguments}` (Responses API)
- **Assistant messages with `tool_calls`** now render the calls under a `Tool calls:` header (both inside the messages loop and on the response side of the entry).
- **`tool` role messages** render with their `tool_call_id` so it is obvious which call a tool result corresponds to.

## Tests

`tests/utils/test_inspect_history.py` — 6 new tests covering:
- OpenAI Chat Completions wire shape (`{function: {name, arguments}}`)
- OpenAI Responses API wire shape (`{name, arguments}`)
- DSPy-native `ToolCall` pydantic object (uses `args` dict)
- `pretty_print_history` end-to-end with a DSPy-native ToolCall on the response side
- `pretty_print_history` end-to-end with an assistant message carrying `tool_calls` in the messages list (`{role: "assistant", content: None, tool_calls: [...]}`)
- `pretty_print_history` end-to-end with a `tool` role message and `tool_call_id` in the role label

### DIFF
--- a/dspy/utils/inspect_history.py
+++ b/dspy/utils/inspect_history.py
@@ -22,6 +22,27 @@ def _blue(text: str, end: str = "\n", *, use_colors: bool = True) -> str:
     return str(text) + end
 
 
+def _get(obj: Any, key: str, default: Any = None) -> Any:
+    if isinstance(obj, dict):
+        return obj.get(key, default)
+    return getattr(obj, key, default)
+
+
+def _format_tool_call(tool_call: Any) -> str:
+    # OpenAI Chat Completions wire shape: {"function": {"name", "arguments"}}.
+    function = _get(tool_call, "function")
+    if function is not None:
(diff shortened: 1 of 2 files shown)


isaacbmiller:
superseded by #9835.

PR #9748, opened 2026-05-13
fix(python_interpreter): resolve symlinks for Deno --allow-read paths (#9501)

### LINKED ISSUE #9501: [Bug] PythonInterpreter fails with default setup due to missing Deno read permissions for Pyodide
### What happened?

In the RLM module documentation, it already states "Users have reported issues with the Deno cache not being found by DSPy. We are actively investigating these issues, and your feedback is greatly appreciated." and I take this is the same issue as I experience using the default `PythonInterpreter`: I find it does not work because of missing permissions for Pyodide. By adding explicit permissions via the `enable_read_paths` input parameter, I was able to get the following `hello-pythoninterpreter.py` file to work.

```python
from pathlib import Path
from dspy import PythonInterpreter

PYODIDE_DIR = Path.home() / "node_modules/.deno/pyodide@0.29.3/node_modules/pyodide"

read_paths = [
    PYODIDE_DIR / "python_stdlib.zip",
    PYODIDE_DIR / "pyodide.asm.wasm",
    PYODIDE_DIR / "pyodide-lock.json",
]

with PythonInterpreter(enable_read_paths=[str(p) for p in read_paths]) as interp:
    result = interp("print(1 + 2)")
    print(result)
```
When running a JavaScript program that uses Pyodide I also need an `--allow-read` such as 
```bash
deno run --allow-read=$HOME/node_modules/.deno/ hello-pyodide.js
```
It is unclear to me whether Deno and Pyodide are appropriately installed on my system. I tested another module `lodash` and using this module does not require `--allow-read`.

The following does not work
```python
from pathlib import Path
from dspy import PythonInterpreter

# PYODIDE_DIR = Path.home() / "node_modules/.deno/pyodide@0.29.3/node_modules/pyodide"
PYODIDE_DIR = Path.home() / "node_modules/pyodide"

read_paths = [
    PYODIDE_DIR / "python_stdlib.zip",
    PYODIDE_DIR / "pyodide.asm.wasm",
    PYODIDE_DIR / "pyodide-lock.json",
]

with PythonInterpreter(enable_read_paths=[str(p) for p in read_paths]) as interp:
    result = interp("print(1 + 2)")
    print(result.strip())
```
Even though there is a symbolic link link from `$HOME/node_modules/pyodide` to `$HOME/node_modules/.deno/pyodide@0.29.3/node_modules/pyodide`. Other modules in  `$HOME/node_modules/` do not have symbolic links. 
My `deno info` gives 
```
Deno version: 2.7.7
DENO_DIR location: $HOME/.cache/deno
...
```
That is, not `$HOME/node_modules/.deno`. If the I set `DENO_DIR` explicitly with `export DENO_DIR=$HOME/node_modules/.deno` I can get the `PythonInterpreter` to work without the specification of `enable_read_path`.

It is not clear to me what has gone wrong (with my installation). From the DSPy side an improved documentation of `DENO_DIR` may help. Perhaps a `.resolve` on paths are needed.

### Steps to reproduce

I am not sure how to reproduce it. It requires a specific set up of Deno/Pyodide and DSPy and this is probably the issue.

### DSPy version

3.1.3

npow:
Fixes #9501.

Deno's `--allow-read` does string-prefix matching against the realpath of the file being read (denoland/deno#9607). If `DENO_DIR` or any path in `enable_read_paths` goes through a symlink, the allow-list entry doesn't match and the read is denied — which is what breaks `PythonInterpreter()` on default Pyodide installs (`~/node_modules/pyodide` is a symlink).

Repro without DSPy:

```bash
D=/tmp/repro && rm -rf $D && mkdir -p $D/real && ln -s $D/real $D/link
echo hi > $D/real/data.txt
echo 'console.log(Deno.readTextFileSync(Deno.args[0]))' > $D/read.js

deno run --no-prompt --allow-read=$D/link $D/read.js $D/real/data.txt
# NotCapable: Requires read access to "/tmp/repro/real/data.txt"

deno run --no-prompt --allow-read=$(realpath $D/link) $D/read.js $D/real/data.txt
# hi
```

Fix is to `os.path.realpath()` everything we put into `--allow-read` / `--allow-write` and the `host_path` we hand to runner.js. The virtual path inside the sandbox keeps the user's original basename so user code still opens `/sandbox/<name>` by the name passed in.

Regression test added (`test_enable_read_paths_symlink`) — fails on main, passes here.

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -151,15 +151,15 @@ class PythonInterpreter:
         else:
             args = ["deno", "run"]
 
-            allowed_read_paths = [_canonicalize_path(self._get_runner_path())]
-
             # Also allow reading Deno's cache directory so Pyodide can load its files
             deno_dir = self._get_deno_dir()
-            if deno_dir:
-                allowed_read_paths.append(_canonicalize_path(deno_dir))
-
-            allowed_read_paths.extend(_canonicalize_path(p) for p in self.enable_read_paths)
-            allowed_read_paths.extend(_canonicalize_path(p) for p in self.enable_write_paths)
+            raw_read_paths = [
+                self._get_runner_path(),
+                *([deno_dir] if deno_dir else []),
+                *self.enable_read_paths,
+                *self.enable_write_paths,
+            ]
+            allowed_read_paths = [_canonicalize_path(p) for p in raw_read_paths]
             args.append(f"--allow-read={','.join(allowed_read_paths)}")
 
             self._env_arg = ""
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -46,6 +46,16 @@ JSONRPC_APP_ERRORS = {
 }
 
 
+def _canonicalize_path(path: PathLike | str) -> str:
+    """Resolve symlinks so the path matches what Deno's permission check sees.
+
+    Deno does string-prefix matching against the realpath of the accessed file
+    (denoland/deno#9607), so --allow-read / --allow-write entries must be
+    realpath'd or reads through a symlink (including DENO_DIR) are denied.
+    """
+    return os.path.realpath(os.path.expanduser(os.fspath(path)))
+
+
 def _jsonrpc_request(method: str, params: dict, id: int | str) -> str:
     """Create a JSON-RPC 2.0 request (expects response)."""
     return json.dumps({"jsonrpc": "2.0", "method": method, "params": params, "id": id})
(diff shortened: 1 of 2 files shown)


dspy/primitives/python_interpreter.py
```diff
@@ -141,18 +151,15 @@ def __init__(
...
-                allowed_read_paths.append(deno_dir)
+                allowed_read_paths.append(_canonicalize_path(deno_dir))
 
-            if self.enable_read_paths:
-                allowed_read_paths.extend(str(p) for p in self.enable_read_paths)
-            if self.enable_write_paths:
-                allowed_read_paths.extend(str(p) for p in self.enable_write_paths)
+            allowed_read_paths.extend(_canonicalize_path(p) for p in self.enable_read_paths)
+            allowed_read_paths.extend(_canonicalize_path(p) for p in self.enable_write_paths)
             args.append(f"--allow-read={','.join(allowed_read_paths)}")
```
isaacbmiller, line 163: Can you group all of the canonical read paths together before we add it to the string so that if we add another path in the future, it is automatically canonicalized?

isaacbmiller (changes requested):
Thanks for the PR @npow! Great fix.

One nit then LGTM!

dspy/primitives/python_interpreter.py
npow: updated

npow: the 2 calls with `virtual_path` is mildly annoying but not sure if it's worth making it DRY

isaacbmiller:
LGTM. Thank you for the PR!

PR #9750, opened 2026-05-14
docs(community): add Dynamic Model Router to Built with DSPy

manthan9891994:
Adds Dynamic Model Router to the Open-Source Projects table. It provides a get_lm(task) helper that returns a dspy.LM pinned to the cheapest model tier for the given prompt (3-layer cascade: keywords -> frozen MiniLM -> LLM fallback, classifies in <15ms).

PyPI: https://pypi.org/project/dynamic-model-router/
Repo: https://github.com/manthan9891994/agents-multi-model-support


isaacbmiller:
Hey this seems interesting!

Do you train a model router using dspy?

manthan9891994:
Honest answer: no,  the router itself is independent of DSPy. The package provides a `get_lm(task)` helper that returns a configured `dspy.LM` pinned to the routed model, so DSPy users can adopt it as a drop-in. If "Built with DSPy" is meant for projects that use DSPy internally (e.g., to optimize prompts or train modules), I'm happy to withdraw let me know which fits your editorial line.

isaacbmiller:
Got it. Would love to see something like this built in dspy!

If you decide to do something like that, let me know!

PR #9751, opened 2026-05-14
fix: handle cached native tool calls

### LINKED ISSUE #9737: [Bug] `TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'` when processing cached native tool calls
### What happened?

### Steps to reproduce

### DSPy version

3.2.1

he-yufeng:
﻿## Summary
- normalize native OpenAI/LiteLLM tool-call objects in one `ToolCalls` helper
- avoid calling `model_dump()` on cached Pydantic objects whose serializer may be a stale `MockValSer`
- reuse the helper from both adapter paths and cover the cached-object case in tests

Fixes #9737.

## To verify
- `.venv\Scripts\python.exe -m py_compile dspy\adapters\types\tool.py dspy\adapters\base.py dspy\adapters\two_step_adapter.py tests\adapters\test_tool.py`
- `.venv\Scripts\python.exe -m ruff check dspy\adapters\types\tool.py dspy\adapters\base.py dspy\adapters\two_step_adapter.py`
- `.venv\Scripts\python.exe -m pytest tests\adapters\test_tool.py::test_tool_calls_from_openai_tool_calls_avoids_model_dump tests\adapters\test_tool.py::test_tool_calls_format_from_dict_list tests\adapters\test_chat_adapter.py::test_tool_call_with_null_content_does_not_raise -q --basetemp .tmp\pytest -p no:cacheprovider`
- `git diff --check`

Note: checking all of `tests/adapters/test_tool.py` with ruff currently reports an existing `B011` on an older `assert False` in the file. I left that unrelated cleanup out of this fix.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -2,8 +2,6 @@ import json
 import logging
 from typing import Any, get_origin
 
-import json_repair
-
 from dspy.adapters._legacy_type_markers import (
     _expand_legacy_custom_type_markers_in_chat_message,
     _expand_legacy_custom_type_markers_in_lm_message,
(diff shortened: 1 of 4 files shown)


he-yufeng:
Rebased this onto current upstream/main and resolved the adapter/tool-call refactor conflicts.

The PR behavior is preserved in `ToolCalls.from_openai_tool_calls()`: cached/provider tool-call objects are normalized without calling `model_dump()`. The base and two-step adapters now use that shared normalizer, while preserving the newer upstream tool-call result/id handling.

Validated locally on Windows:
- python -m pytest tests/adapters/test_tool.py -q -k "from_openai_tool_calls or tool_call_results or vague_match" (5 passed)
- python -m pytest tests/adapters -q -k "tool or native or function_call" (80 passed)
- python -m ruff check dspy/adapters/base.py dspy/adapters/two_step_adapter.py dspy/adapters/types/tool.py tests/adapters/test_tool.py
- python -m ruff format --check dspy/adapters/base.py dspy/adapters/two_step_adapter.py dspy/adapters/types/tool.py tests/adapters/test_tool.py
- python -m py_compile dspy/adapters/base.py dspy/adapters/two_step_adapter.py dspy/adapters/types/tool.py tests/adapters/test_tool.py
- git diff --check upstream/main..HEAD

he-yufeng:
Removed the two dead helpers (`_provider_tool_call_to_tool_call_dict`, `_provider_value`) and the now-unused `json_repair` import from base.py. They were only reachable through the old call site that `from_openai_tool_calls` replaced, so nothing else changes. py_compile and ruff check/format pass.

he-yufeng:
Gentle nudge — cached native tool calls are dropped on the replay path here, so a cache hit loses them. Would appreciate a look when there is time.

he-yufeng:
Superseded by the maintainer's own #9830 (merged 2026-05-26) with a regression test on main; issue #9737 closed. Closing.

PR #9752, opened 2026-05-14
Make LiteLLM imports lazy

MaximeRivest:
## Summary
- lazy-load LiteLLM so importing DSPy works without importing the dependency
- configure LiteLLM defaults on first actual use via lazy import on-load hooks
- avoid LiteLLM imports in streaming type checks and annotations
- add tests for lazy import callbacks and DSPy import behavior without LiteLLM

## Tests
- python3 -m pytest tests/clients/test_lazy_litellm_import.py tests/utils/test_lazy_import.py -q

### DIFF
--- a/dspy/clients/embedding.py
+++ b/dspy/clients/embedding.py
@@ -2,12 +2,22 @@ from __future__ import annotations
 
 from typing import Any, Callable
 
-import litellm
-
 from dspy.clients.cache import request_cache
 from dspy.utils.lazy_import import require
 
 np = require("numpy")
+
+
+def _configure_litellm_defaults(litellm):
+    litellm.telemetry = False
+    litellm.cache = None  # By default we disable LiteLLM cache and use DSPy on-disk cache.
+    if not getattr(litellm, "_dspy_logging_configured", False):
+        litellm.suppress_debug_info = True
+        litellm._dspy_logging_configured = True
+
+
+litellm = require("litellm", extra="litellm", feature="dspy.Embedder", on_load=_configure_litellm_defaults)
+
 
 class Embedder:
     """DSPy embedding class.
--- a/dspy/clients/embedding.py
+++ b/dspy/clients/embedding.py
@@ -2,21 +2,15 @@ from __future__ import annotations
 
 from typing import Any, Callable
 
+from dspy.clients._litellm import get_litellm
 from dspy.clients.cache import request_cache
 from dspy.utils.lazy_import import require
 
 np = require("numpy")
 
 
-def _configure_litellm_defaults(litellm):
-    litellm.telemetry = False
-    litellm.cache = None  # By default we disable LiteLLM cache and use DSPy on-disk cache.
-    if not getattr(litellm, "_dspy_logging_configured", False):
-        litellm.suppress_debug_info = True
-        litellm._dspy_logging_configured = True
-
-
-litellm = require("litellm", extra="litellm", feature="dspy.Embedder", on_load=_configure_litellm_defaults)
+def _get_litellm():
+    return get_litellm(feature="dspy.Embedder")
 
 
 class Embedder:
@@ -162,8 +156,8 @@ class Embedder:
 
 def _compute_embeddings(model, batch_inputs, caching=False, **kwargs):
     if isinstance(model, str):
-        caching = caching and litellm.cache is not None
-        embedding_response = litellm.embedding(model=model, input=batch_inputs, caching=caching, **kwargs)
+        caching = caching and _get_litellm().cache is not None
+        embedding_response = _get_litellm().embedding(model=model, input=batch_inputs, caching=caching, **kwargs)
         return [data["embedding"] for data in embedding_response.data]
     elif callable(model):
         return model(batch_inputs, **kwargs)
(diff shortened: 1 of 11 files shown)


dspy/clients/embedding.py
```diff
@@ -2,13 +2,23 @@
...
 
-import litellm
-
 from dspy.clients.cache import request_cache
 from dspy.utils.lazy_import import require
 
 np = require("numpy")
 
+
+def _configure_litellm_defaults(litellm):
```
isaacbmiller: Where should we keep this shared function? I don't think the answer is to rewrite it in every file.

Maybe we extend require with extras to run on first import for specific modules? And to run once we do `@functools.cache`

dspy/clients/lm.py
```diff
@@ -189,8 +204,10 @@ def forward(
                 num_retries=self.num_retries,
                 cache=litellm_cache_args,
             )
-        except LitellmContextWindowExceededError as e:
-            raise ContextWindowExceededError(model=self.model) from e
+        except Exception as e:
```
isaacbmiller, line 195: we are catching a broader range of exceptions here and in line 250. Is this intended?

dspy/streaming/streamify.py
```diff
@@ -20,6 +18,12 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _is_litellm_model_response_stream(value: Any) -> bool:
+    cls = type(value)
+    return cls.__name__ == "ModelResponseStream" and cls.__module__.startswith("litellm")
```
isaacbmiller, line 24: this seems hacky.

Can we require litellm here for now? Will this change once your other LM changes hit to make litellm optional?

dspy/streaming/streaming_listener.py
```diff
@@ -112,7 +113,7 @@ def _could_form_end_identifier(self, concat_message: str, adapter_name: str) ->
 
         return False
 
-    def receive(self, chunk: ModelResponseStream):
+    def receive(self, chunk: "ModelResponseStream"):
```
isaacbmiller: why do we need to comment this out if we are importing it for type checking?

dspy/utils/lazy_import.py
```diff
@@ -42,7 +43,30 @@ def _detect_dspy_dist() -> str:
     "weaviate": "weaviate",
     "anthropic": "anthropic",
     "numpy": "numpy",
+    "litellm": "litellm",
 }
+_ON_LOAD_CALLBACKS: dict[str, list[Callable[[types.ModuleType], None]]] = {}
+_LAZY_MODULES: set[str] = set()
+
+
+class _OnLoadLoader:
```
isaacbmiller: Are there any other ways to accomplish this?

`functools.cache` can accomplish this callback/run once pattern specifically for litellm with a function we write in this file (or in something like `clients/_litellm.py`. 

LiteLLM is the only module that we do this for, and its not a common pattern, so I am fine building one off logic here and refactoring later if we end up doing this with other modules. (we shouldnt!)

tests/clients/test_lazy_litellm_import.py
```diff
@@ -0,0 +1,97 @@
...
+def run_python(code: str) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        [sys.executable, "-c", textwrap.dedent(code)],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+
+
+def test_import_dspy_succeeds_without_litellm():
```
isaacbmiller: I am not a fan of tests that run python in a subprocess if we can avoid it. Here I would just import dspy normally, and check if litellm is in sys.modules yet.

tests/clients/test_lazy_litellm_import.py
```diff
@@ -0,0 +1,97 @@
...
+        import dspy
+
+        _ = dspy.LM
+        _ = dspy.Embedder
+        _ = dspy.streamify
+        """
+    )
+
+
+def test_lm_litellm_use_raises_helpful_error_without_litellm():
```
isaacbmiller: Same comment about running in a subprocess here.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ mcp = ["mcp; python_version >= '3.10'"]
 langchain = ["langchain_core"]
 optuna = ["optuna>=3.4.0"]
 numpy = ["numpy>=1.26.0"]
+litellm = ["litellm>=1.64.0"]
```
isaacbmiller, line 47: we will also need to update uv.lock accordingly (run `uv sync`)

dspy/clients/lm.py
```diff
@@ -20,12 +18,29 @@
...
+    litellm.cache = None  # By default we disable LiteLLM cache and use DSPy on-disk cache.
+    if not getattr(litellm, "_dspy_logging_configured", False):
+        litellm.suppress_debug_info = True
+        litellm._dspy_logging_configured = True
+
+
+litellm = require("litellm", extra="litellm", feature="dspy.LM", on_load=_configure_litellm_defaults)
+
+
+def _is_litellm_context_window_error(error: Exception) -> bool:
```
isaacbmiller: Seems hacky - any other way to do this?

isaacbmiller (changes requested):
Nice job!

My big question is about centralizing all of the custom import logic. We can do this in `lazy_import.py` or in a new `clients/_litellm.py`. Ideally this will be in LiteLLMLM in the future.

Even if we have LiteLLMLM, we will still need a way to only call this configuration logic once, and `functools.cache` solves that problem for us.

dspy/streaming/streamify.py
MaximeRivest: I would keep that one like that, otherwise with pay the 5 sec tax of loading litellm and complain to install it just to check if we should are not use the litelllm streaming path.

dspy/streaming/streaming_listener.py
MaximeRivest: on if we add: from __future__ import annotations. I am doing that now

dspy/utils/lazy_import.py
MaximeRivest: we are loosing the benefits of require if we do that, but given the LanguageModel changes coming, its fine to make it a bit one of as you say.

PR #9755, opened 2026-05-14
refactor(tool-call): canonical ToolCalls with parallel batch support

### LINKED ISSUE #9737: [Bug] `TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'` when processing cached native tool calls

isaacbmiller:
## Summary

Replaces ad-hoc tool-call normalization scattered across adapters and LM clients with explicit, named boundary methods and a canonical `ToolCalls.ToolCall` shape. Subsumes the fix in #9751 (cached native tool calls / MockValSer crash) and extends `ToolCalls` to preserve one or more tool calls from a single LM action turn.

### What changed

**Boundary methods:**
- **`Tool.to_lm_tool_spec(model_type)`** — outbound: serializes a `Tool` definition for the LiteLLM `tools=` payload (chat vs responses shape).
- **`ToolCalls.ToolCall.to_lm_tool_call(model_type)`** — outbound: serializes a canonical tool call for assistant message replay.
- **`from_lm_tool_call(item)`** — inbound: normalizes a LiteLLM tool-call (dict or pydantic object) into a canonical `ToolCalls.ToolCall`. Auto-detects Chat Completions vs Responses API from the item's `type` field.

**Structural changes:**
- `ToolCall` nested inside `ToolCalls` — single canonical type, no module-level alias.
- `ToolCall.format()` returns canonical `{name, args}` shape, decoupled from wire dialect.
- `ToolCall.id` field for round-tripping provider call IDs.
- `ToolCalls` is treated as a batch for one LM action turn and validates/preserves multiple calls.
- `validate_input` type guard rejects raw wire-format dicts with a clear error.
- Tool names are sanitized at `Tool` initialization so outbound schemas and execution lookup stay consistent.
- Guard in `base.py` skips the native tool path for `model_type="text"`.
- Removed `from_dict_list`, inline `json_repair` normalization in adapters.

**Parallel tool-call support:**
- Adds `allow_parallel_tool_calls` to `Adapter`, `ChatAdapter`, and `JSONAdapter`.
- Native function calling only sends `lm_kwargs["parallel_tool_calls"]` when explicitly configured.
- Non-native parsing remains naturally batch-capable through `ToolCalls` without a provider knob.
- Chat Completions and Responses API client paths now preserve multiple function calls.

**inspect_history improvements:**
- Displays `role: "tool"` messages (tool results in multi-turn loops) with tool name and call ID.
- Displays assistant `tool_calls` in message history.
- Handles `text: None` outputs without crashing.

**Transition TODOs** toward `LMToolSpec` / `LMToolCallPart` / `LMToolCallResultPart`:
- `to_lm_tool_spec` will return `LMToolSpec` instead of dict; wire-format serialization moves to the LM.
- `to_lm_tool_call` will return `LMToolCallPart` instead of dict; same.
- `from_lm_tool_call` item will become `LMToolCallResultPart`, owned by clients.
- Tool-name sanitization will move into the LM layer with the provider-specific wire-format logic.

### Validation

- `uv run ruff check ...`
- `uv run pytest --deno tests/adapters/test_tool.py tests/adapters/test_chat_adapter.py tests/adapters/test_json_adapter.py tests/clients/test_lm.py -q --deselect tests/clients/test_lm.py::test_retry_number_set_correctly`

Note: running the same focused pytest command without the deselect still fails only on the known pre-existing `test_retry_number_set_correctly` mock issue.

### Fixes

- Fixes #9737 (cached native tool calls crash)


immu4989:
Pulled the branch and ran a round-trip probe against the new tool-call
boundary (`ToolCall.to_lm_tool_call` ↔ `from_lm_tool_call`). All clean
on the cases I covered:

- Chat dialect: `{type: function, function: {name, arguments}, id}` —
  round-trips with `id` preserved.
- Responses dialect: `{type: function_call, name, arguments, call_id}` —
  round-trips with `call_id` mapped back into `ToolCall.id`.
- `id=None` correctly omits the field (no `"id": null` leaking out).
- Parallel `ToolCalls` preserves order and per-call IDs.
- Args serialized as JSON strings in both dialects.
- Garbage input rejected with a clear `ValueError`.

One small observation, not a blocker: it'd be useful to ship a short
example showing how to reliably exercise `allow_parallel_tool_calls=True`
from a ReAct loop , the wire level mechanics here are solid, but
triggering the parallel path from a real model takes a fairly explicit
prompt and isn't obvious from the docs.

Happy to contribute that example as a follow up if useful.

isaacbmiller:
Closing as superseded by the normalized LM/Responses architecture and ReActV2 work now on main. Canonical LMToolCallPart boundaries, provider conversion, tool-call IDs/results, native history replay, and parallel tool calls are now implemented through the merged #9786/#9802/#9823/#9824 stack; the cached-provider regression was fixed separately by #9830.

PR #9763, opened 2026-05-14
fix: reject comma paths for PythonInterpreter

### LINKED ISSUE #9749: PythonInterpreter: paths containing commas are silently misparsed by Deno's --allow-read
`PythonInterpreter` builds the `--allow-read` / `--allow-write` flags by joining paths with `,`:

```python
# dspy/primitives/python_interpreter.py
args.append(f"--allow-read={','.join(allowed_read_paths)}")
args.append(f"--allow-write={','.join(_canonicalize_path(x) for x in self.enable_write_paths)}")
```

Deno splits the flag on every comma and has no escape syntax. If any path in `enable_read_paths` / `enable_write_paths` contains a literal comma, the flag becomes ambiguous.

Repro:

```python
from pathlib import Path
from dspy import PythonInterpreter

p = Path("/tmp/sales,2025.csv")
p.write_text("hi")

with PythonInterpreter(enable_read_paths=[str(p)]) as interp:
    print(interp(f"open({str(p)!r}).read()"))
# NotCapable: Requires read access to "/tmp/sales,2025.csv"
```

Deno sees the allow-list as `[..., "/tmp/sales", "2025.csv"]` — neither of which is the actual file. The read is denied even though the user explicitly granted it.

This is independent of the `--allow-read` symlink issue (#9501).

Two viable fixes, both reasonable:
1. Reject paths containing commas at construction time with a clear `ValueError`.
2. Detect commas at construction time and refuse to start, suggesting the user rename the file.

Option 1 is probably right — there's no way to express comma-containing paths to Deno today.

he-yufeng:
﻿## Changes Description

Fixes #9749.

`PythonInterpreter` currently joins `enable_read_paths` and `enable_write_paths` with commas when building Deno permission flags. Deno has no escaping for commas in those allow lists, so a path like `sales,2025.csv` is split into two unrelated entries and the sandbox later denies access to the real file.

This patch rejects comma-containing read/write paths when the interpreter is constructed. That fails early with a clear error instead of launching Deno with a permission set that cannot work.

## Contributor Checklist

- [x] Title of the PR follows the repo's existing fix-style convention
- [x] Commit message follows the repo's existing fix-style convention

## Tested

- `python -m pytest tests\primitives\test_python_interpreter_flags.py -q --basetemp .tmp\pytest -p no:cacheprovider`
- `ruff check dspy\primitives\python_interpreter.py tests\primitives\test_python_interpreter_flags.py`
- `ruff format --check tests\primitives\test_python_interpreter_flags.py`
- `python -m py_compile dspy\primitives\python_interpreter.py tests\primitives\test_python_interpreter_flags.py`
- `git diff --check`

Note: `ruff format --check dspy\primitives\python_interpreter.py` would reformat unrelated existing lines in that file, so I did not run formatting over the whole file for this small fix.

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -94,6 +94,15 @@ def _await_in_sync(coroutine: Any) -> Any:
     return loop.run_until_complete(coroutine)
 
 
+def _reject_comma_paths(option: str, paths: list[PathLike | str]) -> None:
+    for path in paths:
+        if "," in str(path):
+            raise ValueError(
+                f"{option} contains a path with a comma, which Deno cannot represent "
+                f"in comma-separated permission flags: {path!r}"
+            )
+
+
 class PythonInterpreter:
     """Local interpreter for secure Python execution using Deno and Pyodide.
 
@@ -150,6 +159,8 @@ class PythonInterpreter:
 
         self.enable_read_paths = enable_read_paths or []
         self.enable_write_paths = enable_write_paths or []
+        _reject_comma_paths("enable_read_paths", self.enable_read_paths)
+        _reject_comma_paths("enable_write_paths", self.enable_write_paths)
         self.enable_env_vars = enable_env_vars or []
         self.enable_network_access = enable_network_access or []
         self.sync_files = sync_files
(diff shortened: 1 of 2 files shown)


he-yufeng:
Still relevant — comma-separated paths slip through the PythonInterpreter path guard. Minor hardening, rebased and green.

PR #9764, opened 2026-05-14
feat(evaluate): expose timeout and straggler_limit on Evaluate

onk3sh:
## Summary

Fixes #9574.

`ParallelExecutor` already accepts `timeout` and `straggler_limit` parameters, but `Evaluate` always passed neither — falling back to the hardcoded 120-second straggler timeout. Any eval where a task runs longer than 2 minutes would trigger the straggler resubmit loop, and if the executor had already shut down, would crash with:

```
```

### Changes

- Add `timeout: int = 120` and `straggler_limit: int = 3` to `Evaluate.__init__`
- Store as `self.timeout` / `self.straggler_limit`
- Pass both to `ParallelExecutor` in `__call__`
- Default values match the existing `ParallelExecutor` defaults — **no behaviour change for existing callers**

### Usage

```python
# Disable straggler resubmit entirely for long-running evals
evaluate = dspy.Evaluate(devset=devset, metric=metric, timeout=0)

# Or extend the timeout
evaluate = dspy.Evaluate(devset=devset, metric=metric, timeout=600)
```

## Test plan

- [ ] `tests/utils/test_parallelizer.py::test_configurable_timeout_accepted` — verifies custom values are stored
- [ ] `tests/utils/test_parallelizer.py::test_timeout_zero_disables_straggler_resubmit` — verifies `timeout=0` completes cleanly
- [ ] All 12 existing parallelizer tests pass

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -81,6 +81,8 @@ class Evaluate:
         failure_score: float = 0.0,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: int = 120,
+        straggler_limit: int = 3,
         **kwargs,
     ):
         """
@@ -97,6 +99,10 @@ class Evaluate:
             failure_score (float): The default score to use if evaluation fails due to an exception.
             save_as_csv (Optional[str]): The file name where the csv will be saved.
             save_as_json (Optional[str]): The file name where the json will be saved.
+            timeout (int): Seconds before a straggler task is resubmitted. Set to 0 to disable.
+                Defaults to 120.
+            straggler_limit (int): Number of remaining tasks below which straggler detection
+                activates. Defaults to 3.
 
         """
         self.devset = devset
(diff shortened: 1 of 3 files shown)


onk3sh:
Good catch — you're right that both original tests were only exercising pre-existing `ParallelExecutor` behaviour.

Pushed a follow-up commit that replaces them with two `Evaluate`-level tests:

- **`test_evaluate_stores_timeout_and_straggler_limit`** — asserts that `Evaluate.__init__` stores `timeout` and `straggler_limit` on the instance
- **`test_evaluate_forwards_timeout_to_parallel_executor`** — patches `ParallelExecutor` and asserts that `Evaluate.__call__` passes the correct `timeout` and `straggler_limit` kwargs through

The old `test_configurable_timeout_accepted` (which tested `ParallelExecutor` directly) has been removed since it didn't cover any new code path.

onk3sh:
any update?

isaacbmiller:

PR #9767, opened 2026-05-16
fix(avatar): remove deprecated prefix args from internal signatures

nullhack:
### Summary
Remove `prefix=` from `InputField`/`OutputField` calls in two avatar-related files that still use the deprecated argument. The `prefix` kwarg is a no-op in the adapter-based architecture and produces 11 `DeprecationWarning`s on every `import dspy`.

### Affected files
- `dspy/predict/avatar/signatures.py` — `Actor` class (3 occurrences)
- `dspy/teleprompt/avatar_optimizer.py` — `Comparator` and `FeedbackBasedInstruction` classes (8 occurrences)

### Verification
```
import warnings
warnings.filterwarnings("default", category=DeprecationWarning)
import dspy  # no warnings after this change
```


isaacbmiller:
LGTM. Thanks for the PR!

PR #9772, opened 2026-05-17
fix(evaluate): expose timeout param and fix straggler shutdown race (#9574)

Godzilaa:
Closes #9574

## Summary

Two related fixes for DSPy's evaluation parallel executor:

### 1. Add `timeout` parameter to `Evaluate`

Users can now control the straggler timeout (previously hardcoded to 120s via `ParallelExecutor`'s default):

```python
ev = Evaluate(devset=..., metric=..., timeout=300)
ev(program, timeout=600)  # or override at call time
```

### 2. Fix RuntimeError on straggler resubmission after shutdown

`_execute_parallel`'s straggler resubmission calls `executor.submit()` while the executor might be shutting down (e.g. after `KeyboardInterrupt` or during long evals), producing:

> RuntimeError: cannot schedule new futures after shutdown

The fix wraps the resubmission in a try-except `RuntimeError` and skips gracefully when the executor is no longer accepting work.

Also guards the straggler check against `None`/zero `timeout` values to prevent `TypeError`.

## Tests

- `test_evaluate_with_custom_timeout` — timeout set at construction propagates correctly
- `test_evaluate_call_with_timeout_override` — per-call timeout overrides constructor value
- `test_timeout_disables_straggler_resubmission` — `timeout=None` skips straggler logic
- `test_zero_timeout_disables_straggler_resubmission` — `timeout=0` also skips

All 24 tests pass.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -81,6 +81,7 @@ class Evaluate:
         failure_score: float = 0.0,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: int | None = None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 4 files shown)


isaacbmiller:

PR #9773, opened 2026-05-18
feat(tool-call): support parallel tool call requests

isaacbmiller:
## Summary

Encapsulates native `ToolCalls` handling inside the `ToolCalls` type and makes adapter native-response handling generic across `dspy.Type` subclasses.

## What changed

- Added `ToolCalls` to the default native response types while preserving explicit `native_response_types` as an exact override.
- Moved native tool-call setup/parsing into `ToolCalls.adapt_to_native_lm_feature()` and `ToolCalls.parse_lm_response()`.
- Added ToolCalls normalization for:
  - OpenAI chat-completions-style `function` tool calls
  - OpenAI Responses API `function_call` items
  - cached LiteLLM tool-call objects
  - multiple / parallel tool calls
- Added `allow_parallel_tool_calls` adapter option and forwards it as `parallel_tool_calls` when explicitly set.
- Updated `JSONAdapter` structured-output handling so unsupported native-output types like `ToolCalls` are excluded from structured schemas when native function calling is active.
- Preserved compatibility for existing custom `Type.adapt_to_native_lm_feature()` hooks.
- Added tests covering native/non-native ToolCalls behavior across `ChatAdapter` and `JSONAdapter`.

## Why

Native tool-call behavior was previously a special side channel in `Adapter`. This change makes `ToolCalls` own its native adaptation/parsing logic, aligning it with the existing native `Type` pattern used by reasoning/citations and making adapter behavior cleaner and easier to extend.

## Validation

- `uv run pytest --deno tests/adapters/test_parallel_tool_calls.py tests/adapters/test_tool.py -q`
  - `43 passed`

## Review notes / boundaries

- This PR is focused on native ToolCalls encapsulation, ToolCalls parsing/normalization, structured-output compatibility, and parallel tool-call request wiring.
- It does not change broader adapter fallback semantics beyond what is needed for native ToolCalls and JSON structured-output behavior.

PR #9775, opened 2026-05-19
fix(adapter): preserve config in ChatAdapter JSON fallback

isaacbmiller:
When ChatAdapter falls back to JSONAdapter on parse failure, forward `callbacks`, `native_response_types`, `use_native_function_calling`, and `allow_parallel_tool_calls` so the fallback adapter behaves consistently.

Stacked on #9773.


isaacbmiller:
superceded by #9825

PR #9786, opened 2026-05-21
Add core language model types

MaximeRivest:
This PR adds `dspy/core/types.py`, containing the normalized request, response, and stream types to be used as interface between adapters and baselm.

### DIFF
--- a/dspy/core/types.py
+++ b/dspy/core/types.py
@@ -5,7 +5,8 @@ from __future__ import annotations
 import json
 import mimetypes
 from collections.abc import AsyncIterator, Callable, Iterator, Mapping
-from dataclasses import dataclass, field as dataclass_field
+from dataclasses import dataclass
+from dataclasses import field as dataclass_field
 from pprint import pformat
 from typing import Annotated, Any, Literal
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


dspy/core/types.py
```diff
@@ -338,7 +374,7 @@ class LMConfig(BaseModel):
     temperature: float | None = None
     max_tokens: int | None = None
     top_p: float | None = None
-    stop: list[str] | None = Field(default_factory=list)
+    stop: list[str] | None = None
```
isaacbmiller, line 437: was this intentionally not None @MaximeRivest?

MaximeRivest: no its better like that

dspy/core/types.py
```diff
@@ -1662,20 +1683,12 @@ def _coerce_part(value: Any) -> LMPart:
         return LMTextPart(text=value)
     if isinstance(value, dict) and "type" in value:
         return pydantic.TypeAdapter(LMPart).validate_python(value)
-    if _is_dspy_image(value):
```
isaacbmiller: Does removing these break anything you're planning @MaximeRivest ? Why would an LMResponse have any of these parts?

PR #9791, opened 2026-05-21
Add exact adapter format message tests

MaximeRivest:
We did not have complete end-to-end coverage for the three adapters to ensure their rendering behavior stays stable over time.

This PR adds exact-message regression tests for the chat, JSON, and XML adapters to help prevent adapter rendering drift.


isaacbmiller (approved):
LGTM - pydantic tests discussed offline

PR #9800, opened 2026-05-22
fix: parse inline ChatAdapter field headers

### LINKED ISSUE #8379: [Bug] ChatAdapter cannot parse response without newlines correctly
### What happened?

When using the Gemini 2.5 pro model, the response sometimes doesn't adhere to the specific newline requirements for parsing the correctly.

<details>
<summary>Error log:</summary>

```
Adapter ChatAdapter failed to parse the LM response. 

LM Response: [[ ## next_thought ## ]]
The user wants me to ...snip...transactions.[[ ## next_tool_name ## ]]
redacted[[ ## next_tool_args ## ]]
{
    "query": "redacted"
}
[[ ## completed ## ]] 

Expected to find output fields in the LM response: [next_thought, next_tool_name, next_tool_args] 

Actual output fields parsed from the LM response: [next_thought] 

```
</details>

To fix this, I propose parsing this without involving newlines to both simplify logic, and prevent this kind of parsing errors like in #8374.

Alternatively, maybe the prompt could also be optimized to be more specific about the formatting?

### Steps to reproduce

Since it's a model hallucination (as it doesn't include the newline correctly), it doesn't happen consistently.

I was using an agent with `dspy.ReAct` and tools is how I got this error. It doesn't happen all the time, but rarely on some queries.

### DSPy version

2.6.27

he-yufeng:
﻿## What
- parse ChatAdapter field headers wherever they appear in the completion, not only at the start of a line
- add a regression test for a model response that glues the next field marker onto the previous field text

## To verify
- `python -m pytest tests\adapters\test_chat_adapter.py -q -k "inline_field_headers or sync_call"`
- `python -m pytest tests\adapters\test_chat_adapter.py -q`
- `python -m ruff check dspy\adapters\chat_adapter.py tests\adapters\test_chat_adapter.py`
- `python -m py_compile dspy\adapters\chat_adapter.py tests\adapters\test_chat_adapter.py`
- `git diff --check -- dspy\adapters\chat_adapter.py tests\adapters\test_chat_adapter.py`

Fixes #8379

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -209,19 +209,15 @@ class ChatAdapter(Adapter):
         return assistant_message_content
 
     def parse(self, signature: type[Signature], completion: str) -> dict[str, Any]:
-        sections = [(None, [])]
-
-        for line in completion.splitlines():
-            match = field_header_pattern.match(line.strip())
-            if match:
-                # If the header pattern is found, split the rest of the line as content
+        sections = [(None, completion)]
+        matches = list(field_header_pattern.finditer(completion))
+        if matches:
+            sections = [(None, completion[: matches[0].start()].strip())]
+            for idx, match in enumerate(matches):
                 header = match.group(1)
-                remaining_content = line[match.end() :].strip()
-                sections.append((header, [remaining_content] if remaining_content else []))
-            else:
-                sections[-1][1].append(line)
-
-        sections = [(k, "\n".join(v).strip()) for k, v in sections]
+                start = match.end()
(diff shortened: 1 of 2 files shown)


he-yufeng:
Quick follow-up: this still has a Greptile 5/5 review and is scoped to parsing inline ChatAdapter field headers without swallowing them into the previous value.

Could a maintainer take a look when convenient?

he-yufeng:
Second nudge here. ChatAdapter's field_header_pattern on main is still the strict form, so an inline field header (marker and content on the same line) fails to parse the same way it did when this was opened. The change loosens the pattern to accept inline headers without touching the strict layout, with coverage for both shapes.

PR #9802, opened 2026-05-22
Introduce normalized LM boundary for adapters

MaximeRivest:
## Summary

This is the first PR in a stack to move adapters toward normalized LM requests and responses.

Adapters still format prompts the same way, and current `BaseLM` calls still receive OpenAI/LiteLLM-shaped kwargs. The difference is that the adapter call path now passes through a normalized boundary internally:

```text
adapter messages
-> LMRequest
-> OpenAI/LiteLLM kwargs
-> current BaseLM
-> LMResponse
-> existing adapter postprocess path
```

## Why

Today adapters talk directly in provider-shaped messages and legacy output objects. That makes it hard to move tools, reasoning, citations, multimodal inputs, and future adapter types onto one LM type system.

This PR adds the seam without changing parser behavior yet. We normalize before and after the LM call, then convert back to the existing postprocess shape so this can stay behavior-preserving.

## Changes

- Add `dspy.clients.openai_format` for normalized `LMRequest` / `LMResponse` conversion to and from OpenAI/LiteLLM-shaped data.
- Add `dspy.adapters._legacy_type_markers` for the old custom type marker expansion.
- Update `Adapter.__call__()` and `acall()` to route through `LMRequest` and `LMResponse`.
- Preserve unknown legacy content blocks during round-trip conversion.
- Preserve text block metadata when coercing OpenAI-style content into `LMTextPart`.
- Add TODOs in `base.py` marking the follow-up adapter plan and direct `LMResponse` parsing work.

## Notes

`dspy.clients.openai_format` is used instead of an adapter-local bridge because provider-format conversion belongs in the client/LM layer. Later LM backends can reuse the same formatter.

The conversion back from `LMResponse` to legacy outputs is temporary. A follow-up PR should replace `_call_preprocess()` / `_call_postprocess()` with an explicit adapter plan and direct `LMResponse` parsing.

## Tests

```bash
uv run ruff check dspy/adapters/base.py dspy/adapters/_legacy_type_markers.py dspy/clients/openai_format.py dspy/core/types.py
uv run pytest -q tests/adapters/test_chat_adapter.py tests/adapters/test_json_adapter.py tests/adapters/test_xml_adapter.py tests/adapters/test_audio.py tests/adapters/test_base_type.py tests/adapters/test_document.py tests/adapters/test_citation.py tests/adapters/test_reasoning.py tests/adapters/test_tool.py tests/adapters/test_two_step_adapter.py
```

Targeted adapter suite: 169 passed.

### DIFF
--- a/dspy/adapters/_legacy_type_markers.py
+++ b/dspy/adapters/_legacy_type_markers.py
@@ -1,0 +1,145 @@
+"""Private compatibility helpers for legacy `dspy.Type` marker serialization."""
+
+from __future__ import annotations
+
+import json
+import re
+from typing import Any
+
+import json_repair
+
+from dspy.adapters.types.base_type import CUSTOM_TYPE_END_IDENTIFIER, CUSTOM_TYPE_START_IDENTIFIER
+from dspy.core.types import LMAudioPart, LMBinaryPart, LMDocumentPart, LMImagePart, LMMessage, LMPart, LMTextPart
+
+
+def _expand_legacy_custom_type_markers_in_chat_message(message: dict[str, Any]) -> dict[str, Any]:
+    """Expand legacy marker payloads in an OpenAI-chat-shaped user message."""
+    if message.get("role") != "user" or not isinstance(message.get("content"), str):
+        return message
+    content = message["content"]
+    if CUSTOM_TYPE_START_IDENTIFIER not in content:
+        return message
+    return {**message, "content": _split_legacy_custom_type_text_to_blocks(content)}
+
+
+def _expand_legacy_custom_type_markers_in_lm_message(message: LMMessage) -> LMMessage:
+    """Expand legacy marker payloads in user `LMMessage`s into normalized parts."""
+    if message.role != "user":
+        return message
+    expanded_parts: list[LMPart] = []
+    changed = False
+    for part in message.parts:
+        if not isinstance(part, LMTextPart) or CUSTOM_TYPE_START_IDENTIFIER not in part.text:
+            expanded_parts.append(part)
+            continue
+        changed = True
+        expanded_parts.extend(_split_legacy_custom_type_text_to_parts(part.text))
+    return message.model_copy(update={"parts": expanded_parts}) if changed else message
+
+
+def _split_legacy_custom_type_text_to_blocks(text: str) -> list[dict[str, Any]]:
+    blocks: list[dict[str, Any]] = []
+    last_end = 0
+    for match in re.finditer(_marker_pattern(), text, re.DOTALL):
+        start, end = match.span()
+        if start > last_end:
+            blocks.append({"type": "text", "text": text[last_end:start]})
+        blocks.extend(_legacy_custom_type_payload_to_blocks(match.group(1).strip()))
+        last_end = end
+    if last_end < len(text):
+        blocks.append({"type": "text", "text": text[last_end:]})
+    return blocks
+
+
+def _split_legacy_custom_type_text_to_parts(text: str) -> list[LMPart]:
+    parts: list[LMPart] = []
+    last_end = 0
+    for match in re.finditer(_marker_pattern(), text, re.DOTALL):
+        start, end = match.span()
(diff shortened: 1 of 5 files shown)


dspy/adapters/_legacy_type_markers.py
```diff
@@ -0,0 +1,140 @@
...
+def _parse_legacy_payload(payload: str) -> Any:
+    for parse_fn in (json.loads, _parse_doubly_quoted_json, json_repair.loads):
+        try:
+            return parse_fn(payload)
+        except Exception:
+            continue
+    return None
+
+
+def _parse_doubly_quoted_json(value: str) -> Any:
```
isaacbmiller, line 91: uh

dspy/adapters/_legacy_type_markers.py
```diff
@@ -0,0 +1,140 @@
...
+    if not isinstance(block, dict):
+        return LMTextPart(text=str(block))
+
+    block_type = block.get("type")
+    if block_type == "text":
+        return LMTextPart(text=block.get("text", ""))
+    if block_type == "image_url":
+        image_url = block.get("image_url", {})
+        source = image_url.get("url") if isinstance(image_url, dict) else image_url
+        from dspy.core.types import LMImagePart
```
isaacbmiller: inline import?

dspy/adapters/_legacy_type_markers.py
```diff
@@ -0,0 +1,140 @@
...
+    if block_type == "document":
+        source = block.get("source", {})
+        return LMDocumentPart(
+            source=source if isinstance(source, dict) else {"type": "text", "data": str(source)},
+            citations=block.get("citations") or {"enabled": True},
+            title=block.get("title"),
+            context=block.get("context"),
+        )
+
+    return LMTextPart(text="", metadata={"legacy_content_block": block})
```
isaacbmiller, line 134: Should this raise instead of being a generic fallback? Do we read this `legacy_content_block` anywhere?

dspy/adapters/base.py
```diff
@@ -175,6 +202,111 @@ def _call_postprocess(
...
+        messages: list[LMMessage | dict[str, Any]],
+    ) -> LMRequest:
+        """Build the normalized LM request for the current adapter call path.
+
+        TODO(adapters-plan): This currently receives already-rendered messages.
+        Once planning lands, this should render from `_AdapterPlan` and apply
+        planned message/part insertions before creating `LMRequest`.
+        """
+        return LMRequest.from_call(
+            model=getattr(lm, "model", ""),
```
isaacbmiller: nit: does LM ever not have model?

I have seen this particular LLMism about BaseLM and solely using getattr many times.

dspy/adapters/base.py
```diff
@@ -175,6 +202,111 @@ def _call_postprocess(
...
+            messages=self._coerce_lm_messages(messages),
+            **lm_kwargs,
+        )
+
+    def _call_lm(self, lm: BaseLM, request: LMRequest) -> LMResponse:
+        """Call current `BaseLM` through the normalized request/response boundary.
+
+        TODO(language-models): When `BaseLM` is replaced by/updated to the
```
isaacbmiller, line 226: nit: this method should probably disappear rather than becoming a direct call

isaacbmiller (approved):
LGTM. Some questions + one inline import.

Some edge cases that agent found which seem reasonable:
Image

MaximeRivest:
Thanks, fixed the small nits:

- added a comment explaining the doubly-encoded JSON legacy parser;
- moved `LMImagePart` to the top-level import;
- added a comment explaining the `legacy_content_block` fallback and where it is rehydrated;
- changed `getattr(lm, "model", "")` to `lm.model`;
- reworded the `_call_lm()` TODO to say this compatibility shim should disappear.

I also fixed the provider-shaped `tool_choice` compatibility case so OpenAI/LiteLLM-shaped values like:

```python
{"type": "function", "function": {"name": "get_weather"}}

PR #9806, opened 2026-05-23
Sandbox serializable output

kmad:
## Summary

Follows up on https://github.com/stanfordnlp/dspy/pull/9411, which added `SandboxSerializable` for moving custom host objects into the RLM sandbox.

This PR completes the round trip: `SandboxSerializable` subclasses can now also be used as `OutputField` annotations.
For example, this is now possible:
```python
class CohortAnalysis(dspy.Signature):
    users: DataFrame = dspy.InputField() # Dataframe is a user-defined SandboxSerializable class

    cohort: DataFrame = dspy.OutputField() # <--- New!
    summary: str = dspy.OutputField()
    count: int = dspy.OutputField()
```

The model still works with the bare sandbox value, e.g. a pandas `DataFrame`, and calls `SUBMIT(cohort=..., summary=..., count=...)`. RLM handles the serializable output field by wrapping it before `FinalOutput` leaves the sandbox, then reconstructing the host-side `SandboxSerializable` instance from the returned payload. Plain outputs continue through the existing parsing path.

Agent code works with the native sandbox value, passes it to `SUBMIT(...)`, and RLM serializes it back across the sandbox boundary before reconstructing the wrapper on the host.

## Changes

- Add sandbox-to-host hooks:
  - `sandbox_serialize_code(var_name)` for converting a bare sandbox value into a wire payload.
  - `from_sandbox(payload)` for reconstructing the wrapper on the host.
- Make `sandbox_setup()` class-level shared setup, used for both input values and output annotations.
- Run setup once for the union of serializable inputs and serializable outputs before the REPL loop.
- Teach `SUBMIT(...)` generation to wrap only the serializable output fields while leaving plain outputs on the normal path.
- Update the cohort DataFrame tutorial to show returning a `DataFrame` output alongside plain fields.
- Add focused unit and Deno-backed coverage for mixed outputs, output-only setup, and setup deduping.

## Why

Before this, `SandboxSerializable` was one-way: useful for giving RLM rich objects to inspect, but not for returning rich objects it creates or transforms.

This enables workflows like passing DataFrames into the sandbox, transforming them with normal pandas code, and returning a typed DataFrame result with dtypes preserved.

## Commentary
These changes add some complexity to `runner.js` because `SUBMIT(...)` now needs to wrap only the output fields annotated with `SandboxSerializable` before raising `FinalOutput`. The implementation keeps that logic field-local: ordinary outputs continue through the existing path, while serializable outputs provide a sandbox-side expression for converting the bare value into a wire payload.

Long term, the right shape still seems to be a small bidirectional interface on `SandboxSerializable`, where a shared sandbox setup runs once per RLM session, sandbox serialization happens inside the generated `SUBMIT(...)` wrapper, and host-side reconstruction happens through `from_sandbox(...)`. This keeps `PythonInterpreter` generic, avoids special-casing `DataFrames` in core, and gives custom types one place to define how they move across the RLM boundary.

## Tests

- `uv run --extra dev pytest tests/primitives/test_sandbox_serializable.py tests/predict/test_rlm.py::TestPrepareSerializableVars tests/predict/test_rlm.py::TestSerializableOutputs -q`
- `uv run --extra dev pytest tests/predict/test_rlm.py::TestSerializableOutputRoundTripThroughSandbox tests/predict/test_rlm.py::TestLargeSerializableRoundTrip -q --deno`

### DIFF
--- a/docs/docs/api/modules/RLM.md
+++ b/docs/docs/api/modules/RLM.md
@@ -185,14 +185,26 @@ rlm = dspy.RLM(
 )
 ```
 
-### Custom Sandbox-Serializable Inputs
+### Custom Sandbox-Serializable Values
 
-For inputs that should be loaded into the sandbox differently from normal Python values, subclass `dspy.SandboxSerializable`. RLM detects these inputs, sends their serialized payload into the interpreter, runs their setup code, and exposes the reconstructed value under the original input name.
+For values that should cross the sandbox boundary differently from normal Python values, subclass `dspy.SandboxSerializable`. RLM detects serializable inputs, sends their payloads into the interpreter, runs shared setup code, and exposes the reconstructed value under the original input name. The same subclass can also be used as an output annotation: the agent passes the bare sandbox value to `SUBMIT(...)`, and RLM serializes it back to the host before reconstructing the wrapper.
 
 ```python
+import base64
+import io
+
+import pandas as pd
+
+import dspy
+
+
 class DataFrame(dspy.SandboxSerializable):
-    def sandbox_setup(self) -> str:
-        return "import pandas as pd\nimport base64\nimport io"
+    def __init__(self, data):
+        self.data = data
+
+    @classmethod
+    def sandbox_setup(cls) -> str:
+        return "import pandas as pd\nimport pyarrow\nimport base64\nimport io"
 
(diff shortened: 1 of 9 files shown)


isaacbmiller:
Very cool! Excited to take a look sometime next week

PR #9807, opened 2026-05-23
fix(evaluate): expose timeout parameter and cap straggler resubmissions per item

ppradyoth:
## Summary

Fixes #9574

Two related bugs caused `cannot schedule new futures after shutdown` errors and
evaluation hangs on long-running LLM calls.

### Bug 1 — `dspy.Evaluate` hardcodes the 120-second straggler timeout

`ParallelExecutor` accepts a `timeout` parameter that controls when a straggler
task is resubmitted, but `Evaluate` never passed it through.  Every evaluation
ran with the hardcoded 120-second default, giving users no way to:

- **increase** the threshold for slow models (large context, low-throughput endpoints, Bedrock cross-region profiles), or
- **disable** straggler resubmission entirely by setting `timeout=0`.

### Bug 2 — resubmitted stragglers can themselves be resubmitted (unbounded chain)

`ParallelExecutor._execute_parallel` tracked already-resubmitted tasks by
**future object**, not by **data index**.  When the resubmitted future also
exceeded the straggler timeout, it was treated as a fresh straggler and
resubmitted again, spawning another future.  This chain repeated on every
timeout interval until the executor was shut down, at which point any in-flight
`executor.submit()` call raised:

```
```

## Changes

| File | Change |
|---|---|
| `dspy/evaluate/evaluate.py` | Add `timeout: int = 120` to `__init__` and `__call__`; forward to `ParallelExecutor` |
| `dspy/utils/parallelizer.py` | Track resubmitted items by **data index** (not future) so each item is resubmitted at most once |
| `tests/evaluate/test_evaluate.py` | 3 new tests: default timeout stored, custom timeout forwarded, `timeout=0` disables resubmission |
| `tests/utils/test_parallelizer.py` | 1 new test: stalling task is resubmitted at most once, not in an unbounded chain |

## How to reproduce the bugs (before this fix)

```python
import dspy

# Bug 1: no way to raise the straggler threshold
# Evaluate does not accept a timeout keyword
ev = dspy.Evaluate(devset=devset, metric=metric, timeout=300)

# Bug 2: with a slow LM and enough time the executor can be overwhelmed
# by an ever-growing chain of straggler resubmissions, eventually raising
# RuntimeError: cannot schedule new futures after shutdown
```

## How to verify the fix

```python
import dspy

# Raise the threshold to 10 minutes for slow LLMs
ev = dspy.Evaluate(devset=devset, metric=metric, timeout=600)

# Disable straggler resubmission entirely
ev = dspy.Evaluate(devset=devset, metric=metric, timeout=0)

# Override per-call
result = ev(program, timeout=600)
```

All existing tests pass unchanged.  New tests cover both fixes without requiring
slow real-model calls.

Fixes #9574

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -81,6 +81,7 @@ class Evaluate:
         failure_score: float = 0.0,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: int = 120,
         **kwargs,
     ):
         """
@@ -97,6 +98,9 @@ class Evaluate:
             failure_score (float): The default score to use if evaluation fails due to an exception.
             save_as_csv (Optional[str]): The file name where the csv will be saved.
             save_as_json (Optional[str]): The file name where the json will be saved.
+            timeout (int): Seconds before a straggler task is resubmitted. Set to 0 to disable
+                straggler resubmission entirely. Increase this when individual LLM calls are
+                expected to take longer than the default 120 s. Defaults to 120.
 
         """
         self.devset = devset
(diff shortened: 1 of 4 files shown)


isaacbmiller:

ppradyoth:
Thanks, no worries. @isaacbmiller

PR #9821, opened 2026-05-24
Make BaseLM own shared LM runtime state

MaximeRivest:
## Summary

This PR makes `BaseLM` a more complete shared runtime base for DSPy LMs, without introducing the new normalized LM request/response path.

It moves common LM runtime state into `BaseLM`, makes `dspy.LM` delegate that setup to `super().__init__()`, and changes `BaseLM.copy()` from a recursive deep copy to an explicit shallow runtime copy.

## Changes

- Add shared runtime fields to `BaseLM.__init__()`:
  - `callbacks`
  - `num_retries`
  - `_warned_zero_temp_rollout`
- Align `BaseLM` defaults with existing `dspy.LM` defaults:
  - `temperature=None`
  - `max_tokens=None`
  - `num_retries=3`
- Make `dspy.LM.__init__()` call `BaseLM.__init__()` for shared setup.
- Move usage tracking into `BaseLM._process_lm_response()` so custom `BaseLM` subclasses benefit from `track_usage()`.
- Make `LM.dump_state()` build on `BaseLM.dump_state()` and only add LM-specific fields.
- Change `BaseLM.copy()` from `copy.deepcopy()` to a shallow runtime copy:
  - preserves provider/client/session/local-model handles by reference
  - isolates DSPy-owned mutable state: `history`, `callbacks` list, and `kwargs`
- Document copy semantics for custom LMs.
- Add focused tests for `BaseLM` defaults, callback list isolation, custom usage tracking, and shallow runtime copy semantics.

## Why

`BaseLM` is the extension point for custom LM implementations, but several pieces of runtime behavior lived only in `dspy.LM` or relied on deep-copy behavior that is brittle for real provider clients.

Deep-copying an LM can fail or produce invalid duplicated runtime resources when the LM owns clients, sessions, locks, local model handles, or other provider state. The new copy behavior treats `copy()` as a runtime copy: provider resources are shared intentionally, while DSPy-owned per-copy state is isolated.

This better matches how `lm.copy(...)` is used in DSPy: same backend/runtime, different request defaults.

## Compatibility notes

- `BaseLM.__init__()` follows the existing `dspy.LM` constructor order for shared arguments.
- Custom `BaseLM` subclasses that relied on `copy()` deep-copying arbitrary subclass-specific mutable attributes should override `copy()` and isolate that state explicitly.
- Callback lists are copied, but callback objects themselves are shared.

## Out of scope

This PR intentionally does not add the new normalized LM type system, including:

- `LMRequest`
- `LMResponse`
- normalized streaming events
- `LMCapabilities`
- new direct-call message constructors

Those belong in the follow-up new-LM PR stack.

## Validation

- `ruff check dspy/clients/base_lm.py dspy/clients/lm.py tests/clients/test_lm.py`
- `pytest -q tests/clients/test_lm.py::test_base_lm_init_uses_lm_defaults_and_isolates_callback_list tests/clients/test_lm.py::test_base_lm_tracks_usage_for_custom_subclasses tests/clients/test_lm.py::test_base_lm_copy_is_shallow_runtime_copy_with_isolated_dspy_state tests/clients/test_lm.py::test_rollout_id_bypasses_cache tests/clients/test_lm.py::test_dump_state tests/clients/test_lm.py::test_load_state`

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,3 +1,4 @@
+import copy as copy_module
 import datetime
 import importlib
 import inspect
@@ -5,7 +6,7 @@ import uuid
 from typing import Any, TextIO
 
 from dspy.dsp.utils import settings
-from dspy.utils.callback import with_callbacks
+from dspy.utils.callback import BaseCallback, with_callbacks
 from dspy.utils.inspect_history import pretty_print_history
 
 MAX_HISTORY_SIZE = 10_000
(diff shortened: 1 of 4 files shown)


dspy/clients/base_lm.py
```diff
@@ -99,18 +100,35 @@ def __init__(
...
         model,
         model_type="chat",
-        temperature=0.0,
-        max_tokens=1000,
+        temperature=None,
+        max_tokens=None,
         cache=True,
-        num_retries=0,
+        callbacks: list[BaseCallback] | None = None,
+        num_retries: int = 3,
```
MaximeRivest, line 107: added and changed to be the same as LM. I think LMs defaults are more THE defaults, I am reflecting that here.

dspy/clients/base_lm.py
```diff
@@ -305,22 +326,34 @@ def load_state(cls, state: dict[str, Any], *, allow_custom_lm_class: bool = Fals
...
+                copy. For example, `lm.copy(rollout_id=1, temperature=1.0)`
+                returns an LM whose requests use a different rollout ID at
+                non-zero temperature to bypass cache collisions.
+
+        Returns:
+            A copied LM instance.
+        """
 
-        new_instance = copy.deepcopy(self)
+        new_instance = copy_module.copy(self)
```
MaximeRivest, line 351: I have researched the subject a good bit, it does not seam to mess with our threading and matches with how copy is used in dspy. This is more friendly to custom lm.

isaacbmiller (approved):
LGTM.

PR #9823, opened 2026-05-25
feat(adapter): preserve tool call ids and add tool results

isaacbmiller:
## Summary

Adds the minimal typed data needed to connect native provider tool calls to later observations.

- Preserves provider tool call IDs on `ToolCalls.ToolCall` while keeping `ToolCalls.format()` free of IDs for prompt/JSON rendering.
- Adds `ToolCallResults` with call IDs, tool names, values, and error flags.
- Exports `ToolCallResults` alongside the existing adapter/tool types.
- Adds exact prompt coverage for `ToolCallResults` stored inside `History`.

## Validation

- `uv run pytest -q tests/adapters tests/predict/test_react.py tests/predict/test_react_v2.py`

Stack: PR 1 of 3 for the minimal ReActV2 implementation.

PR #9824, opened 2026-05-25
feat(adapter): replay native tool call history

isaacbmiller:
## Summary

Teaches adapters to replay known tool-loop history as native LM messages when native function calling is enabled.

- Renders prior assistant tool calls as native assistant `tool_calls` with preserved call IDs.
- Renders `ToolCallResults` as native `tool` role messages with matching `tool_call_id`.
- Keeps non-native adapter history text rendering unchanged.
- Prevents absent `tool_call_results` from leaking into current-turn user text.

## Validation

- `uv run pytest -q tests/adapters tests/predict/test_react.py tests/predict/test_react_v2.py`

Stack: PR 2 of 3 for the minimal ReActV2 implementation. Stacked on #9823.

### DIFF
--- a/dspy/adapters/base.py
+++ b/dspy/adapters/base.py
@@ -10,7 +10,8 @@ from dspy.adapters._legacy_type_markers import (
 )
 from dspy.adapters.types import History, Type
 from dspy.adapters.types.reasoning import Reasoning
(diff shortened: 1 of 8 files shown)


MaximeRivest:
lgtm

PR #9825, opened 2026-05-25
feat(predict): add ReActV2

isaacbmiller:
## Summary

Adds a minimal `ReActV2` predictor that uses the native tool loop when supported while still working with text/mock LMs.

- Converts callables to `dspy.Tool` and adds an internal `submit` tool for final outputs.
- Tracks tool calls and tool observations through `History` using `ToolCalls` and `ToolCallResults`.
- Preserves provider tool call IDs and synthesizes local IDs when needed.
- Handles unknown tools, tool exceptions, serialized history input, and forced-submit fallback.

## Validation

- `uv run pytest -q tests/adapters tests/predict/test_react.py tests/predict/test_react_v2.py`

Stack: PR 3 of 3 for the minimal ReActV2 implementation. Stacked on #9824.

### DIFF
--- a/dspy/predict/__init__.py
+++ b/dspy/predict/__init__.py
@@ -8,6 +8,7 @@ from dspy.predict.parallel import Parallel
 from dspy.predict.predict import Predict
 from dspy.predict.program_of_thought import ProgramOfThought
 from dspy.predict.react import ReAct, Tool
+from dspy.predict.react_v2 import ReActV2
 from dspy.predict.refine import Refine
 from dspy.predict.rlm import RLM
 
@@ -21,6 +22,7 @@ __all__ = [
     "Predict",
     "ProgramOfThought",
     "ReAct",
+    "ReActV2",
     "Refine",
     "RLM",
     "Tool",
--- a/dspy/predict/react_v2.py
+++ b/dspy/predict/react_v2.py
@@ -1,0 +1,250 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any, Callable, get_args
+
+import pydantic
+
+import dspy
+from dspy.adapters.types.tool import Tool, ToolCallResults, ToolCalls
+from dspy.primitives.module import Module
+from dspy.primitives.prediction import Prediction
+from dspy.signatures.signature import ensure_signature
+from dspy.utils.exceptions import AdapterParseError, ContextWindowExceededError
+
+logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from dspy.signatures.signature import Signature
+
+
+class ReActV2(Module):
+    def __init__(self, signature: type[Signature], tools: list[Callable | Tool], max_iters: int = 20):
+        super().__init__()
+        self.signature = ensure_signature(signature)
+        self.max_iters = max_iters
+
+        user_tools = [tool if isinstance(tool, Tool) else Tool(tool) for tool in tools]
+        self.tools = {tool.name: tool for tool in user_tools}
+        if "submit" in self.tools:
+            raise ValueError("`submit` is reserved by ReActV2 as the final-output tool.")
+        self.tools["submit"] = self._make_submit_tool()
+
+        self.react = dspy.Predict(self._make_react_signature())
+
+    def _make_submit_tool(self) -> Tool:
+        output_fields = self.signature.output_fields
+        output_names = list(output_fields)
+
+        def submit(**kwargs):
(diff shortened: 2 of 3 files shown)


isaacbmiller:
BrowseComp ReAct sweep rerun from the split stack top (`fix/inspect-history-tool-calls`, commit `743064123`). Full local artifacts remain untracked under `tmp/browsecomp_react_sweep_n25_results.json` and `tmp/browsecomp_react_sweep_n25_summary.md`.

| Setting | Done | Crash | Final recall | Gathered recall | Submit | Avg iter | LM calls | Parse errors |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `react_v1_nonnative` | 22/25 | 3 | 0.222 | 0.307 | 68.18% | 6.59 | 246 | 3 |
| `react_v2_nonnative` | 24/25 | 1 | 0.240 | 0.441 | 87.50% | 6.75 | 167 | 0 |
| `react_v2_native` | 23/25 | 2 | 0.241 | 0.454 | 100.00% | 5.13 | 136 | 0 |

Notes:
- Non-native fallback contamination: 0 examples with native-looking non-native calls.
- Native crashed on qids 769 and 770 with `TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'` in this run.
- cmpnd preflight status: ok; trace `d6a19f60-b1bd-4c80-b3e2-1b59177a718f` retrieved with HTTP 200.

immu4989:
Pulled main and ran a 4 test probe focused on the multi-turn-with-tools
use case from #9115. Headline first, then the one finding.

**All 3 happy path probes pass.** Behaviors verified end to end:

- Parallel tool calls in a single turn — 2 distinct IDs (`call_0_0`,
  `call_0_1`), each with a matching `ToolCallResult.call_id`.
- Cross-call history replay — `agent(question=..., history=prior_pred.history)`
  works cleanly, prior events preserved.
- Non-string tool returns (dict) — `ToolCallResult.value` retains the
  dict structure rather than stringifying.

Big upgrade over V1 for the optimizable-agent-loop use case.

**One real concern , `_ensure_tool_call_ids` collides IDs across `forward()`
calls, and the collision surfaces in the wire payload.**

`_ensure_tool_call_ids` generates IDs as `call_{turn_index}_{call_index}`
where `turn_index` is the loop-local iteration counter, which resets each
`forward()` call. So when you replay via `history=prior_pred.history` and
call 2 iterates ≥2 times, the trajectory contains duplicate IDs across
the conversation:

call 1 trajectory: [('lookup', 'call_0_0'), ('submit', 'call_1_0')]
call 2 trajectory: [('lookup', 'call_0_0'), ('submit', 'call_1_0')]


collision



This isn't just a trajectory level artifact,  it reaches the LM payload.
Inspecting `lm.history` for call 2's iteration-2 request:

msg[3] (user): {"tool_call_results": [{"call_id": "call_0_0", "name": "lookup",
"value": "answer to cats", ...}]}
msg[8] (user): {"tool_call_results": [{"call_id": "call_0_0", "name": "lookup",
"value": "answer to dogs", ...}]}



Same conversation, same `call_id`, two different tool results. The model
seeing this has no way to disambiguate which result belongs to which call.

In `ChatAdapter` default (text adapter) the collision is in the
JSON-in-text `tool_call_results` blocks (shown above). With native
function calling against a real provider (which PR #9824 enables), the
same collision would surface as duplicate
`assistant.tool_calls[].id` / `tool` role `tool_call_id` values — and
OpenAI / Anthropic both validate those as conversation-unique.

**Easiest fix:** in `_ensure_tool_call_ids`, derive the prefix from
`len(history.messages)` rather than the loop-local `turn_index`, e.g.:

```python
def _ensure_tool_call_ids(tool_calls, turn_index, history_offset=0):
    ensured = []
    for call_index, tool_call in enumerate(tool_calls.tool_calls):
        if tool_call.id is None:
            new_id = f"call_{history_offset + turn_index}_{call_index}"
            tool_call = tool_call.model_copy(update={"id": new_id})
        ensured.append(tool_call)
    return ToolCalls(tool_calls=ensured)
```
with history_offset = len(history.messages) at the top of forward().

Alternative: switch to uuid4().hex[:8] per call. Less debuggable but
collision-proof regardless of replay depth.

Happy to send a small follow up PR with the fix + a regression test that
asserts ID uniqueness across a 2-call replay with ≥2 iterations per call.

@ryanh-ai , this is what I found from your end of the test request.
Probe scripts live in /tmp/probe_react_v2.py and
/tmp/probe_react_v2_wire.py if you want them gisted.

PR #9826, opened 2026-05-25
Normalize DSPy LM errors

MaximeRivest:
## Summary

This PR introduces a structured DSPy error hierarchy for LM failures and wires it into `dspy.LM` and adapter fallback behavior.

Key changes:

- Adds `DSPyError` and `LMError` subclasses for common LM failure modes:
  - auth
  - billing/quota
  - rate limits
  - invalid requests
  - context-window errors
  - unsupported models/features
  - timeouts
  - server errors
  - transport errors
  - unexpected LM-boundary failures
- Wraps LiteLLM/provider exceptions in `dspy.LM.forward()` and `dspy.LM.aforward()`.
- Preserves structured metadata when available:
  - `model`
  - `provider`
  - `provider_code`
  - `status`
  - `request_id`
  - `retry_after`
- Updates adapter fallback behavior so LM errors are propagated instead of being swallowed by JSON fallback.
- Makes `AdapterParseError` inherit from `DSPyError`.
- Exports the new errors from top-level `dspy`.
- Documents the error hierarchy and common handling patterns.

## User impact

This is primarily an exception-normalization change. Successful calls are unchanged, and existing `ContextWindowExceededError` handling still works.

Users who catch broad `Exception` are unaffected. Users who catch specific raw LiteLLM/OpenAI exceptions should now catch DSPy exceptions such as `dspy.LMRateLimitError`, `dspy.LMAuthError`, or the common base `dspy.LMError`.

Adapter fallback behavior is tightened so provider/backend failures are not mistaken for parse failures. For example, an auth error or rate limit will now propagate as an `LMError` instead of causing a retry with another adapter. This makes failures more predictable and easier to handle in production.

More specifically:

- `ChatAdapter` still falls back to `JSONAdapter` for adapter/parsing failures, but no longer retries with another adapter for LM/backend failures such as auth errors, rate limits, timeouts, unsupported models, or context-window errors.
- `JSONAdapter` still falls back from structured outputs to JSON mode for local structured-output setup/schema issues, but no longer hides provider/backend failures behind JSON-mode retry.
- `TwoStepAdapter` now distinguishes extraction parse failures from extraction model LM failures.

## Validation

Unit/local validation:

```bash
pytest -q tests/utils/test_exceptions.py tests/adapters/test_chat_adapter.py tests/adapters/test_json_adapter.py tests/adapters/test_two_step_adapter.py tests/clients/test_lm.py::test_retry_number_set_correctly tests/clients/test_lm.py::test_retry_made_on_system_errors tests/clients/test_lm.py::test_reasoning_model_requirements tests/clients/test_lm.py::test_exponential_backoff_retry tests/clients/test_lm.py::test_lm_wraps_unknown_boundary_error_as_unexpected_error tests/clients/test_lm.py::test_lm_wraps_litellm_errors_with_metadata tests/clients/test_lm.py::test_lm_wraps_litellm_context_window_error
# 110 passed
```

```bash
pytest -q tests/clients/test_lm.py -k 'not can_be_queried and not dspy_cache and not lm_calls_support_callables and not lm_calls_support_pydantic_models and not responses_api_tool_calls'
# 37 passed, 6 deselected
```

```bash
pytest -q tests/docs/test_mkdocs_links.py
# 1 passed
```

```bash
python3 -m ruff check ...
# All checks passed
```

Live OpenAI smoke checks with current checkout via `uv`:

- successful LM call
- successful JSONAdapter structured-output call
- invalid API key → `LMAuthError`
- unknown model → `LMUnsupportedModelError`
- invalid request → `LMInvalidRequestError`
- tiny timeout → `LMTimeoutError`
- adapter propagation of `LMAuthError`

## Notes

`RETRYABLE_LM_ERRORS` is advisory. DSPy's built-in `LM` still delegates provider retries to LiteLLM; callers should respect provider policy and `retry_after` when present.

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -74,6 +74,8 @@ class JSONAdapter(ChatAdapter):
             lm_kwargs["response_format"] = structured_output_model
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
         except LMError:
+            # Provider/backend failures should propagate; the fallback below is only for local structured-output
+            # setup/schema failures where retrying in JSON mode is appropriate.
             raise
         except Exception:
             logger.warning("Failed to use structured output format, falling back to JSON mode.")
@@ -99,6 +101,8 @@ class JSONAdapter(ChatAdapter):
             lm_kwargs["response_format"] = structured_output_model
             return await super().acall(lm, lm_kwargs, signature, demos, inputs)
         except LMError:
+            # Provider/backend failures should propagate; the fallback below is only for local structured-output
+            # setup/schema failures where retrying in JSON mode is appropriate.
             raise
         except Exception:
             logger.warning("Failed to use structured output format, falling back to JSON mode.")
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -19,7 +19,7 @@ from dspy.adapters.utils import (
 from dspy.clients.base_lm import BaseLM
 from dspy.signatures.signature import Signature, SignatureMeta
 from dspy.utils.callback import BaseCallback
-from dspy.utils.exceptions import AdapterParseError
+from dspy.utils.exceptions import AdapterParseError, LMError
 
 logger = logging.getLogger(__name__)
 
@@ -73,6 +73,8 @@ class JSONAdapter(ChatAdapter):
             )
             lm_kwargs["response_format"] = structured_output_model
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
+        except LMError:
+            raise
         except Exception:
             logger.warning("Failed to use structured output format, falling back to JSON mode.")
             lm_kwargs["response_format"] = {"type": "json_object"}
@@ -96,6 +98,8 @@ class JSONAdapter(ChatAdapter):
             )
             lm_kwargs["response_format"] = structured_output_model
             return await super().acall(lm, lm_kwargs, signature, demos, inputs)
+        except LMError:
+            raise
         except Exception:
             logger.warning("Failed to use structured output format, falling back to JSON mode.")
             lm_kwargs["response_format"] = {"type": "json_object"}
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -255,6 +255,8 @@ class LM(BaseLM):
                 cache=litellm_cache_args,
             )
         except Exception as e:
+            if isinstance(e, LMError):
+                raise
             raise self._wrap_litellm_exception(e) from e
 
         self._check_truncation(results)
@@ -310,6 +312,8 @@ class LM(BaseLM):
                 cache=litellm_cache_args,
             )
         except Exception as e:
+            if isinstance(e, LMError):
+                raise
             raise self._wrap_litellm_exception(e) from e
 
         self._check_truncation(results)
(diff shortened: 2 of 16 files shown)


dspy/adapters/json_adapter.py
```diff
@@ -73,6 +73,8 @@ def __call__(
             )
             lm_kwargs["response_format"] = structured_output_model
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
+        except LMError:
```
isaacbmiller, line 76: Why do we catch and immediately re-raise this error?

dspy/clients/lm.py
```diff
@@ -164,12 +182,52 @@ def _get_cached_completion_fn(self, completion_fn, cache):
...
+
+        Args:
+            prompt: Optional prompt text. Ignored when `messages` is provided.
+            messages: Optional chat messages to send to the LM.
+            **kwargs: Per-call LM parameters that override defaults from `LM(...)`.
+
+        Raises:
+            dspy.LMError: Base class for wrapped LM configuration, transport,
+                provider, and unsupported-feature failures.
+            dspy.ContextWindowExceededError: The prompt or messages exceed the
```
isaacbmiller: Should dspy.contextwindowexceedederror be an lm error?

dspy/clients/lm.py
```diff
@@ -606,3 +684,125 @@ def _add_dspy_identifier_to_headers(headers: dict[str, Any] | None = None):
...
+
+
+def _exception_message(exc: Exception) -> str:
+    message = getattr(exc, "message", None)
+    if message is None:
+        message = str(exc)
+    return str(message)
+
+
+def _exception_headers(exc: Exception):
```
isaacbmiller, line 779: These methods smell like they wish something was typed but it isn't? Is there a different way to get this info when we know we have it or just accept that they are untyped?

dspy/utils/exceptions.py
```diff
@@ -1,29 +1,229 @@
...
-class AdapterParseError(Exception):
-    """Exception raised when adapter cannot parse the LM response."""
+
+class LMServerError(LMProviderError):
+    """The provider failed while handling the request."""
+
+    default_code = "server"
+
+
+RETRYABLE_LM_ERRORS = (LMRateLimitError, LMTimeoutError, LMServerError, LMTransportError)
```
isaacbmiller: Can we have another way to configure this tuple rather than publicly exposing a constant

tests/adapters/test_json_adapter.py
```diff
@@ -1290,7 +1274,7 @@ class TestSignature(dspy.Signature):
         assert result.answer.result == "Paris"
 
 
-def test_json_adapter_fallback_to_json_mode_on_structured_output_failure():
+def test_json_adapter_does_not_fallback_to_json_mode_on_structured_output_lm_error():
```
isaacbmiller, line 1277: Is this behavior change intended?

isaacbmiller (commented):
LGTM after CI and confirming test behavior

dspy/adapters/json_adapter.py
MaximeRivest: The fallback to JSON mode is only appropriate for local structured-output setup/schema failures. If the LM/backend failed with an LMError (auth, rate limit, timeout, unsupported model, context window, etc.), retrying with JSON mode would not help and would hide the provider failure.

tests/adapters/test_json_adapter.py
MaximeRivest: yes this feels better (right) to me, some LMerrors can cause a litellm retry but should not cause a output mode change. just parsing/formating should trigger that.

dspy/clients/lm.py
MaximeRivest: in litellm exception metadata is not exposed through a single stable typed shape. although these metadata are not all strictly necessary, the likely will help users in production. its best effort parsing but feels worth it.

PR #9830, opened 2026-05-26
[codex] Fix cached provider Pydantic serializers

### LINKED ISSUE #9737: [Bug] `TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'` when processing cached native tool calls

isaacbmiller:
## Summary

Fixes #9737 by making DSPy's provider-object normalization resilient to Pydantic `MockValSer` serializers that can appear after cached provider responses are read in a fresh Python process.

## Problem

LiteLLM and OpenAI provider response objects are Pydantic models with deferred serializer construction. A response can be written to DSPy's restricted disk cache after the writer process has built the serializer, then read back in a fresh process where the unpickled provider object and nested provider objects still have `__pydantic_serializer__ = MockValSer`.

DSPy's OpenAI-format normalization path calls `.model_dump(exclude_none=True)` for provider tool calls, Responses function calls, citations, usage, image/audio/file outputs, and related provider objects. For cached native tool calls, this produced:

```text
TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'
```

Rebuilding only the outer Pydantic model is not sufficient because nested provider objects, such as a chat-completion tool call's `function`, can retain their own stale serializer.

Related upstream reports/context:

- https://github.com/pydantic/pydantic/issues/7713
- https://github.com/pydantic/pydantic/discussions/7710
- https://github.com/BerriAI/litellm/issues/9345
- https://github.com/run-llama/llama_index/issues/16405

## Changes

- Centralize the `MockValSer` repair in `dspy.clients.openai_format.model_dump`.
- Catch only the expected `TypeError` containing `MockValSer`; unrelated `TypeError`s still raise.
- Recursively rebuild serializers for Pydantic provider objects and nested dict/list/tuple values before retrying the dump.
- Route citation and usage provider dumps through the same helper.
- Add deterministic fresh-process cache regression tests without inline `python -c` strings.

## Validation

- `uv run --frozen pytest --deno tests/clients/test_disk_serialization.py`
- `uv run --frozen pytest --deno tests/clients/test_lm.py::test_dspy_cache tests/clients/test_lm.py::test_rollout_id_bypasses_cache`
- `uv run --frozen ruff check dspy/clients/openai_format.py tests/clients/test_disk_serialization.py`
- `git diff --check`

Note: `ruff format --check` was inspected but not applied because it would reformat older long lines elsewhere in the touched files, creating unrelated churn.

### DIFF
--- a/dspy/clients/openai_format.py
+++ b/dspy/clients/openai_format.py
@@ -854,7 +854,11 @@ def get_value(value: Any, key: str, default: Any = None) -> Any:
 
 def _rebuild_pydantic_serializers(value: Any) -> None:
     stack = [value]
+    seen = set()
     for item in stack:
+        if id(item) in seen:
+            continue
+        seen.add(id(item))
         if isinstance(item, pydantic.BaseModel):
             type(item).model_rebuild(force=True, raise_errors=False)
             stack.extend([*item.__dict__.values(), *((item.__pydantic_extra__ or {}).values())])
--- a/dspy/clients/openai_format.py
+++ b/dspy/clients/openai_format.py
@@ -674,7 +674,7 @@ def responses_function_call_to_part(output_item: Any) -> LMToolCallPart:
 
 def citation_to_part(citation: Any) -> LMCitationPart:
     if hasattr(citation, "model_dump"):
-        citation = citation.model_dump(exclude_none=True)
+        citation = model_dump(citation)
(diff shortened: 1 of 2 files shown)


MaximeRivest:
lgtm

PR #9832, opened 2026-05-26
fix(adapters): preserve Python literals in ChatAdapter dict parsing

### LINKED ISSUE #8820: [Bug] Conflict between ChatAdapter.user_message_output_requirements and ChatAdapter.parse leads to high tool call failure rate

madhavcodez:
## 📝 Changes Description

`ChatAdapter` instructs the LM to format dict output fields as a "valid Python dict" (via `ChatAdapter.user_message_output_requirements`). So for a `Dict[str, Any]` field the model correctly returns Python literals, e.g.:

```
[[ ## arguments ## ]] {"title": "...", "revision_id": None}
```

But `parse_value` in `dspy/adapters/utils.py` runs `json_repair.loads()` first, which coerces the Python literal `None` into the string `"None"`. Tool-call arguments are silently corrupted, which is the high failure rate reported in #8820.

Rather than flip the parsing order globally (which would also affect `JSONAdapter`), this makes the behavior **adapter-aware**, matching the direction discussed on the issue — keep `ChatAdapter` parsing Python literals while leaving `JSONAdapter` untouched:

- `parse_value` gains a keyword-only `prefer_python_literal: bool = False`.
- When `True`, it tries `ast.literal_eval` first and falls back to the existing `json_repair`-first logic, so JSON-style payloads (`null`) still parse.
- `ChatAdapter` passes `prefer_python_literal=True`. `JSONAdapter` and `XMLAdapter` are unchanged — the flag defaults to `False`, so their current behavior is preserved exactly.

### Tests
- `tests/adapters/test_chat_adapter.py`: literal `None` stays real `None` (the bug), `True`/`False`/`None` preserved, JSON `null` still parses via the fallback, and a quoted `'None'` stays the string `"None"`.
- `tests/adapters/test_json_adapter.py`: regression guard confirming `JSONAdapter` `Dict[str, Any]` parsing is unaffected.

Closes #8820

## ✅ Contributor Checklist

- [x] Pre-Commit checks are passing (locally and remotely) — `ruff check` is clean on the changed files

## ⚠️ Warnings

Backward compatible: the new flag defaults to `False`, so only the `ChatAdapter` path changes behavior. `JSONAdapter` and `XMLAdapter` parsing is byte-for-byte unchanged.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -219,7 +219,7 @@ class ChatAdapter(Adapter):
         for k, v in sections:
             if (k not in fields) and (k in signature.output_fields):
                 try:
-                    fields[k] = parse_value(v, signature.output_fields[k].annotation)
+                    fields[k] = parse_value(v, signature.output_fields[k].annotation, prefer_python_literal=True)
                 except Exception as e:
                     raise AdapterParseError(
                         adapter_name="ChatAdapter",
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -146,7 +146,33 @@ def find_enum_member(enum, identifier):
     raise ValueError(f"{identifier} is not a valid name or value for the enum {enum.__name__}")
 
 
-def parse_value(value, annotation):
+def _parse_string_candidate(value: str, prefer_python_literal: bool):
+    """Parse a raw string into a Python object, choosing the parser ordering by adapter.
+
+    ChatAdapter instructs the LM (via ``user_message_output_requirements``) to emit a "valid
+    Python dict", so its responses contain Python literals such as ``None``, ``True``, ``False``,
+    and single-quoted strings. Running ``json_repair.loads`` first would silently coerce these
+    (e.g. ``None`` -> the string ``"None"``), corrupting the value. When ``prefer_python_literal``
+    is True we therefore try ``ast.literal_eval`` first and fall back to the json-first logic,
+    which still handles JSON-style payloads (e.g. ``null``). JSONAdapter and XMLAdapter keep the
+    default json-first behavior unchanged.
+    """
+    if prefer_python_literal:
+        try:
+            return ast.literal_eval(value)
+        except (ValueError, SyntaxError):
+            pass
+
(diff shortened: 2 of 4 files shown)


madhavcodez:
Good catch on the redundant `ast.literal_eval` — fixed in 0744b7e. When `prefer_python_literal` is True and the literal parse fails, it now falls back to `json_repair` directly instead of retrying `ast.literal_eval` in the json-first block. Behavior is unchanged (verified the both-parsers-fail path still returns the raw string).

PR #9839, opened 2026-05-27
fix: serialize LiteLLM lazy materialization

### LINKED ISSUE #9833: [Bug] LiteLLM Lazy Import crashes in concurrent settings
## Concurrent first use of `dspy.LM` can fail in LiteLLM lazy import

  DSPy’s LiteLLM lazy import path appears to be unsafe under concurrent first
  use. If multiple threads create/use `dspy.LM` before LiteLLM has been
  materialized, most threads can fail with:

  ```text
  AttributeError: module 'litellm' has no attribute 'completion'

  This happens before any provider/API call is made.

### Observed result

  On my machine this reproduces consistently:

  failures: 7
  AttributeError: module 'litellm' has no attribute 'completion'
    dspy/clients/lm.py:132 in supports_function_calling
    dspy/clients/_litellm.py:33 in get_litellm
    dspy/clients/_litellm.py:26 in _materialize_litellm

  ### Expected result

  Concurrent first use of dspy.LM should either serialize LiteLLM
  materialization or use a thread-safe import path, so all threads see a fully
  initialized LiteLLM module.

  ### Suspected cause

  dspy.utils.lazy_import.require("litellm") installs an
  importlib.util.LazyLoader module in sys.modules. Under concurrent first
  access, one thread begins materializing LiteLLM while other threads retrieve
  the same partially initialized module from sys.modules. Those threads then
  access litellm.completion before the attribute exists.

  A simple workaround is to call get_litellm(...) once on the main thread
  before starting workers. A library fix might be to guard get_litellm() / lazy
  materialization with a lock, or avoid LazyLoader for LiteLLM.

### Steps to reproduce

```python
import threading
import traceback
from concurrent.futures import ThreadPoolExecutor, as_completed


def worker(barrier):
    barrier.wait()

    import dspy

    lm = dspy.LM("openai/gpt-4o-mini", cache=False, num_retries=0)
    return lm.supports_function_calling


threads = 8
barrier = threading.Barrier(threads)
failures = []

with ThreadPoolExecutor(max_workers=threads) as executor:
    futures = [executor.submit(worker, barrier) for _ in range(threads)]
    for future in as_completed(futures):
        try:
            future.result()
        except Exception:  # noqa: BLE001
            failures.append(traceback.format_exc())

print(f"failures: {len(failures)}")
if failures:
    print(failures[0])
```

### DSPy version

3.2.1 - main

he-yufeng:
## Summary
- serialize LiteLLM lazy materialization during first use
- keep DSPy's lazy import behavior unchanged for normal imports and missing LiteLLM errors
- add a concurrent first-use regression test that fails if another thread observes the partially materialized module

Fixes #9833

## To verify
- python -m pytest -p no:cacheprovider tests/clients/test_lazy_litellm_import.py -q
- python -m ruff check dspy/clients/_litellm.py tests/clients/test_lazy_litellm_import.py
- python -m ruff format --check dspy/clients/_litellm.py tests/clients/test_lazy_litellm_import.py
- python -m py_compile dspy/clients/_litellm.py tests/clients/test_lazy_litellm_import.py
- git diff --check

Note: on Windows I used a repo-local TMP/TEMP directory because the default pytest temp/cache paths had permission errors unrelated to this patch.

### DIFF
--- a/dspy/clients/_litellm.py
+++ b/dspy/clients/_litellm.py
@@ -2,10 +2,13 @@ from __future__ import annotations
 
 import functools
 import sys
+import threading
 import types
 from typing import Any
 
 from dspy.utils.lazy_import import require
+
+_LITELLM_IMPORT_LOCK = threading.RLock()
 
 
 @functools.cache
@@ -29,10 +32,11 @@ def _materialize_litellm(litellm: types.ModuleType) -> None:
 @functools.cache
 def get_litellm(*, feature: str) -> Any:
     """Import LiteLLM, apply DSPy's defaults once, and return the module."""
-    litellm = require("litellm", extra="litellm", feature=feature)
-    _materialize_litellm(litellm)
-    _configure_litellm_defaults(litellm)
-    return litellm
+    with _LITELLM_IMPORT_LOCK:
+        litellm = require("litellm", extra="litellm", feature=feature)
+        _materialize_litellm(litellm)
+        _configure_litellm_defaults(litellm)
+        return litellm
 
 
 def is_litellm_context_window_error(error: Exception) -> bool:
--- a/tests/clients/test_lazy_litellm_import.py
+++ b/tests/clients/test_lazy_litellm_import.py
@@ -1,7 +1,9 @@
 import importlib.util
 import sys
 import threading
-from concurrent.futures import ThreadPoolExecutor
+import time
+import types
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


he-yufeng:
Rebased onto current main and addressed Greptile's teardown note. The concurrent fake-LiteLLM test now clears both caches in a finally block, so the fake module cannot leak into later tests in the same pytest session. Local validation: python -m pytest tests\clients\test_lazy_litellm_import.py -q --basetemp .tmp\pytest-9839-final -> 5 passed; python -m compileall -q dspy\clients\_litellm.py tests\clients\test_lazy_litellm_import.py -> passed.

he-yufeng:
Still applicable: `dspy/clients/_litellm.py` has been quiet since the lazy-import rework landed, and the serialization gap this patches (state lost across lazy re-materialization) is still there on main. Should apply cleanly as is; happy to rebase if anything shifted.

he-yufeng:
Superseded by #9840 (merged 2026-05-27): per-module lazy-import locking landed, and the linked issue #9833 is closed. Closing.

PR #9846, opened 2026-05-28
feat: add append_instructions to Signature

### LINKED ISSUE #9829: [Feature]  Add append_instructions to Signature

Currently, Signature provides `with_instructions(instructions)` which replaces a signature's docstring instructions entirely. AFAIK there's no clean way to add to docstring signature instructions without reading them out manually and reconstructing the signature yourself, which can be brittle IMO.

The proposal is to add another method called `append_instructions(instructions)` that appends this supplemental instructions to the doc strings rather than replacing it.

Usage will look something like:

```python
# Read dynamic instructions from detailed_instruction_file.md
with open("detailed_specific_instruction_file.md", "r") as f:
    detailed_specific_instructions = f.read()
ExtractInformationInformation = ExtractInformation.append_instructions(detailed_specific_instructions)
extractor = dspy.Predict(ExtractInformationInformation)
```

### Additional Context

I believe it can be useful in scenarios where the signatures are the same, but the instructions can be different. For example, we want to extract information about equipments. The instructions can change depending on the equipment, but the signature + common instructions remain the same.

dhruviervu:
Closes #9829

## Summary
Adds `append_instructions` method to `Signature` class, complementing the existing `with_instructions` method.

`with_instructions` replaces instructions entirely. `append_instructions` appends to existing instructions without replacing them, which is useful when base instructions need to stay intact while adding context-specific ones dynamically.

## Changes
- Added `append_instructions` classmethod to `Signature` in `dspy/signatures/signature.py`
- Added `append_instructions` to the Signature API docs
- Added `test_append_instructions` in `tests/signatures/test_signature.py`

## Usage
```python
class ExtractInfo(dspy.Signature):
    """Extract information from text."""
    text: str = dspy.InputField()
    info: str = dspy.OutputField()

EquipmentSig = ExtractInfo.append_instructions("Focus on equipment details only.")
```

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -288,18 +288,17 @@ class Signature(BaseModel, metaclass=SignatureMeta):
             A new Signature class whose fields match `cls.fields`
             and whose instructions equal `instructions`.
 
-Examples:
+        Examples:
 ```python
-        import dspy
+            import dspy
 
-        class MySig(dspy.Signature):
-            input_text: str = dspy.InputField(desc="Input text")
-            output_text: str = dspy.OutputField(desc="Output text")
+            class MySig(dspy.Signature):
+                input_text: str = dspy.InputField(desc="Input text")
+                output_text: str = dspy.OutputField(desc="Output text")
 
-        NewSig = MySig.append_instructions("Always respond in French.")
-        assert NewSig is not MySig
-        assert NewSig.instructions == MySig.instructions + "\n" + "Always respond in French."
-        assert MySig.instructions == original_instructions  # original unchanged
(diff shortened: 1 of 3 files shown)


dhruviervu:
@isaacbmiller what do you think?

dspy/signatures/signature.py
```diff
@@ -289,7 +289,7 @@ class using the current fields and the provided `instructions`.
             and whose instructions equal `instructions`.
 
         Examples:
-            ```python
+```python
```
isaacbmiller, line 292: Why is the indenting edited?

isaacbmiller (commented):
Small fix on indentation, otherwise LGTM.

Thanks for the PR!

PR #9848, opened 2026-05-28
Expose timeout / straggler_limit on dspy.Evaluate (#9574)

NishchayMahor:
Closes #9574.

## What I found

\`ParallelExecutor\` (\`dspy/utils/parallelizer.py\`) already takes \`timeout\` and \`straggler_limit\` kwargs — they control the resubmission window for slow items in \`_execute_parallel\`. But \`dspy.Evaluate\` builds the executor with the hardcoded defaults (120 seconds, 3 stragglers) and never lets the caller tune them:

\`\`\`python
# dspy/evaluate/evaluate.py
executor = ParallelExecutor(
    num_threads=num_threads,
    disable_progress_bar=not display_progress,
    max_errors=...,
    provide_traceback=self.provide_traceback,
    compare_results=True,
)
\`\`\`

For long evals (slow tools, long reasoning chains, larger devsets), the 120s straggler timer fires, items get resubmitted, and on bigger runs you start hitting "cannot schedule new futures after shutdown" once the executor tears down with the resubmitted items still in flight.

## What's in this PR

Surface \`timeout\` and \`straggler_limit\` on \`Evaluate.__init__\` and \`Evaluate.__call__\`, and pass them through to \`ParallelExecutor\`. Defaults keep matching the executor's own defaults so this is a no-op for anyone not opting in:

\`\`\`python
ev = dspy.Evaluate(
    devset=devset,
    metric=my_metric,
    num_threads=8,
    timeout=600,        # bump for slow tools / long reasoning
    straggler_limit=5,
)
ev(program, timeout=1200)   # override per call too
\`\`\`

Setting \`timeout=0\` disables straggler resubmission entirely, which is the simplest way to keep one slow item from cascading on a very long eval.

## Notes / follow-ups

This PR is intentionally just the API surface — it gives the caller a way to keep the executor from re-arming on long-running items, which is the lever you most often want. The deeper question of why the executor sometimes shuts down with submissions still in flight is worth a separate look (the straggler resubmission path can race against the finally-block \`executor.shutdown(wait=False)\` in pathological cases), but that fix is independent of the user-facing knob and I didn't want to bundle them.

## Tests

\`tests/evaluate/test_evaluate.py\`:

- \`test_evaluate_initialization\` — extended to assert the new defaults (120 / 3).
- \`test_evaluate_threads_through_executor_timeout\` — spies on \`ParallelExecutor\` to confirm \`timeout\` and \`straggler_limit\` reach it.
- \`test_evaluate_call_overrides_timeout\` — confirms the per-call override beats the constructor default.

\`\`\`
$ pytest tests/evaluate/test_evaluate.py -q
12 passed, 25 skipped in 1.94s

$ ruff check dspy/evaluate/evaluate.py tests/evaluate/test_evaluate.py
All checks passed!
\`\`\`

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -81,6 +81,8 @@ class Evaluate:
         failure_score: float = 0.0,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: int = 120,
+        straggler_limit: int = 3,
         **kwargs,
     ):
         """
@@ -97,6 +99,12 @@ class Evaluate:
             failure_score (float): The default score to use if evaluation fails due to an exception.
             save_as_csv (Optional[str]): The file name where the csv will be saved.
             save_as_json (Optional[str]): The file name where the json will be saved.
+            timeout (int): Per-item timeout in seconds for the parallel executor's straggler
+                resubmission. Set higher when individual examples take longer than the
+                default 120 seconds (e.g., long reasoning chains or slow tool calls).
+                Pass 0 to disable straggler resubmission entirely.
+            straggler_limit (int): How many remaining stragglers trigger the resubmission
+                check. Defaults to 3 to match the executor.
 
         """
         self.devset = devset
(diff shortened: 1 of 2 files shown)


isaacbmiller:

PR #9850, opened 2026-05-28
fix: remove unused MIPROv2 call estimator

### LINKED ISSUE #9849: [Bug] _estimate_lm_calls() method in MIPROv2 is never called
### What happened?

The` _estimate_lm_calls()` method in the MIPROv2 class (line 355 in `mipro_optimizer_v2.py` appears to be unused. It's seems to be a well-implemented method that estimates and formats LM call counts for both prompt and task models, but it's never invoked anywhere in the codebase.

Is this dead code that should be removed, or should this be called during compilation? 

### Steps to reproduce

This is dead code, and as such, can't reproduce it. I was doubting about reporting this as a bug but it seemed to be the best option.

### DSPy version

3.2.1

he-yufeng:
﻿## Summary
- remove the unused `MIPROv2._estimate_lm_calls()` helper
- remove the color constants that only existed for that helper

Fixes #9849.

## To verify
- `python -m py_compile dspy/teleprompt/mipro_optimizer_v2.py`
- `git diff --check`

I also ran `python -m ruff check dspy/teleprompt/mipro_optimizer_v2.py`; it still reports pre-existing style issues in this file (`requires_permission_to_run == False/True` and `CategoricalDistribution` naming) that are unrelated to this dead-code removal.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -49,11 +49,7 @@ AUTO_RUN_SETTINGS = {
     "heavy": {"n": 18, "val_size": 1000},
 }
 
-# ANSI escape codes for colors
-YELLOW = "\033[93m"
 GREEN = "\033[92m"
-BLUE = "\033[94m"
-BOLD = "\033[1m"
 ENDC = "\033[0m"  # Resets the color to default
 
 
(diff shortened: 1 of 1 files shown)


he-yufeng:
Friendly nudge. The MIPROv2 call estimator is dead code — nothing calls it anymore and it still shows up in searches as if it were live. This removes it so nobody wires against it by mistake. Would appreciate a quick look.

PR #9852, opened 2026-05-28
reject reserved output names in ReActV2 signatures

### LINKED ISSUE #9853: [Bug] ReActV2 raises TypeError when user signature output is named `history` or `termination_reason`
### Summary
  
`dspy.ReActV2` attaches `history` and `termination_reason` to every returned `Prediction`. If the user's signature declares an output field with either name, the keyword expansion at the `Prediction(**final_outputs, ...)` call sites collides and raises `TypeError: got multiple values for keyword  argument`. The failure happens mid-`forward()` rather than at construction.
  
Related

ReAct v1 has the same bug class with a different reserved key:
dspy/predict/react.py:118 and :143 do dspy.Prediction(trajectory=trajectory, **extract), 

which collides if a user signature declares an output field named trajectory.

### Steps to reproduce

```python
  import dspy
  
      return f"found {query}"
      
  lm = dspy.utils.DummyLM([
      {
          "next_thought": "ready",
          "tool_calls": dspy.ToolCalls.from_dict_list(
              [{"name": "submit", "args": {"answer": "x", "history": "y"}}]
          ),  
      },  
  ])  
  
      # Signature declares an output named `history` -- collides with the
      # `history=...` kwarg ReActV2 adds to every Prediction.
      dspy.ReActV2("question -> answer, history: str", tools=[lookup])(question="q")

Result:
  
TypeError: dspy.Prediction() got multiple values for keyword argument 'history'

Same outcome for an output field named termination_reason.

### DSPy version

3.3.0b1

tjdharamsi:
ReActV2 attaches `history` and `termination_reason` to every returned Prediction. A user signature declaring either as an output collided at `Prediction(**final_outputs, history=..., termination_reason=...)` with `TypeError: got multiple values for keyword argument`. Validate in __init__ against a reserved-key set, mirroring the existing `submit` tool reservation, so the conflict surfaces at construction with a clear error.
 
ReAct v1 has the same bug class with a different reserved key — dspy.Prediction(trajectory=trajectory, **extract) at dspy/predict/react.py:118 and :143. A user signature with an output field named trajectory would hit the identical TypeError. Not addressed here since it predates 3.3.0 and isn't a regression, but worth tracking as a follow-up.

Fixes #9853 

- [x] added a unit test.

### DIFF
--- a/dspy/predict/react_v2.py
+++ b/dspy/predict/react_v2.py
@@ -15,6 +15,11 @@ from dspy.utils.exceptions import AdapterParseError, ContextWindowExceededError
 
 logger = logging.getLogger(__name__)
 
+# Keys ReActV2 attaches to every returned `Prediction` alongside the user's output fields.
+# A user signature that declares any of these as an output would collide with the keyword
+# expansion at the `Prediction(**final_outputs, history=..., termination_reason=...)` call sites.
+_RESERVED_PREDICTION_KEYS = frozenset({"history", "termination_reason"})
+
 if TYPE_CHECKING:
     from dspy.signatures.signature import Signature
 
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Thanks for the PR! LGTM

PR #9864, opened 2026-05-31
fix: propagate sync streaming errors

he-yufeng:
﻿## Summary

- propagate exceptions raised by the async stream back to the sync `apply_sync_streaming()` caller
- keep normal stream completion unchanged
- add regression coverage for an async generator that yields once and then fails

Fixes #9142.

## To verify

- `python -m py_compile dspy/streaming/streamify.py tests/streaming/test_streaming.py`
- `uv run pytest tests/streaming/test_streaming.py -q -k "apply_sync_streaming_propagates_generator_exceptions or sync_status_streaming"`
- `uv run ruff check dspy/streaming/streamify.py tests/streaming/test_streaming.py`
- `git diff --check`


he-yufeng:
Gentle ping. When a streaming call fails synchronously, the error currently gets swallowed and the run looks like it just ended early. This propagates the error so callers see the real failure. Small, with a test. Would love a review.

isaacbmiller:
lgtm

PR #9865, opened 2026-05-31
fix: mark dspy as typed

### LINKED ISSUE #8873: [Bug] package is missing py.typed

he-yufeng:
﻿## Summary

Fixes stanfordnlp/dspy#8873.

DSPy ships inline type annotations but the wheel does not include the PEP 561 marker, so type checkers treat `dspy` as untyped/missing when it is imported as a dependency. This adds `dspy/py.typed` and includes it in package data alongside the existing `primitives/*.js` asset.

## To verify

- `uv run ruff check pyproject.toml`
- `python -m py_compile dspy\__init__.py`
- `uv run python -m build --wheel --outdir %TEMP%\dspy-pytyped-wheel`
- checked the built wheel contains `dspy/py.typed`
- `git diff --check`


he-yufeng:
Bump on this one. `dspy/py.typed` still does not exist on main today, so downstream type checkers continue to treat the package as untyped and inline every annotation instead of reading them from the wheel. The PR adds only the marker and the packaging entry, so there is no runtime surface at all.

he-yufeng:
Two-line packaging change, conflicting and idle for 112 days with a same-goal duplicate (#9888) also idle. Closing to clear the backlog; easy to redo if the project wants PEP 561 later.

PR #9874, opened 2026-06-01
fix: avoid ReActV2 metadata field collisions

### LINKED ISSUE #9853: [Bug] ReActV2 raises TypeError when user signature output is named `history` or `termination_reason`

he-yufeng:
﻿## Summary
- fix ReActV2 returning `Prediction(**final_outputs, history=..., termination_reason=...)` when user output fields use those names
- preserve user-declared output values by adding ReActV2 runtime metadata only when the names are not already present
- add a regression test for outputs named `history` and `termination_reason`

Fixes #9853.

## To verify
- `uv run --frozen --no-sync pytest tests\predict\test_react_v2.py -q --basetemp .tmp\pytest`
- `uv run --frozen --no-sync ruff check dspy\predict\react_v2.py tests\predict\test_react_v2.py`
- `python -m py_compile dspy\predict\react_v2.py tests\predict\test_react_v2.py`
- `git diff --check`

Note: local `ruff format --check` wants to reflow existing unrelated lines in `tests/predict/test_react_v2.py`; I left those alone to keep this PR focused.


he-yufeng:
Gentle nudge. ReActV2 still lets a user metadata field silently shadow the framework's own keys, which corrupts the trajectory. The patch rejects the collision up front with a clear error instead of letting it through.

PR #9884, opened 2026-06-03
fix: exclude args with defaults from required in Tool function-callin…

### LINKED ISSUE #9882: [Bug] Tool args with defaults are emitted as `required` in native function-calling schema
### What happened?

`dspy.Tool.format_as_litellm_function_call` marks every tool argument as
`required`, despite also advertising default values on argument properties.

### Steps to reproduce

```python
import json
from typing import Any, Literal

import dspy
from dspy.utils import DummyLM


def look_up(query: str, state_optional: Literal["fresh", "cached"] = "fresh") -> str:
    """Look something up."""
    return f"{query}:{state_optional}"


class _FCDummyLM(DummyLM):
    @property
    def supports_function_calling(self) -> bool:
        return True

    def forward(self, prompt: Any = None, messages: Any = None, **kwargs: Any) -> Any:
        message = dotdict(
            content=self._format_answer_fields({"next_thought": "ready"}),
            tool_calls=[
                dotdict(
                    id="call_0",
                    type="function",
                    function=dotdict(name="submit", arguments=json.dumps({"answer": "x"})),
                )
            ],
        )
        return dotdict(
            choices=[dotdict(message=message, finish_reason="stop")],
            model="dummy",
        )


lm = _FCDummyLM([])
with dspy.context(lm=lm, adapter=dspy.ChatAdapter(use_native_function_calling=True)):
    pred = dspy.ReActV2("question -> answer", tools=[dspy.Tool(look_up)], max_iters=1)(question="hi")

print(f"termination_reason={pred.termination_reason!r} answer={pred.answer!r} lm_calls={len(lm.history)}\n")

sent_tools = lm.history[-1]["kwargs"]["tools"]
params = next(t for t in sent_tools if t["function"]["name"] == "look_up")["function"]["parameters"]
print(json.dumps(params, indent=2))
```

Output — `state_optional` is wrongly listed in `required` despite its `default`:

```
termination_reason='submit' answer='x' lm_calls=1

{
  "type": "object",
  "properties": {
    "query": {
      "type": "string"
    },
    "state_optional": {
      "enum": [
        "fresh",
        "cached"
      ],
      "type": "string",
      "default": "fresh"
    }
  },
  "required": [
    "query",
    "state_optional"
  ]
}
```

### DSPy version

3.3.0b1

arnav-144p:
## Summary

Fixes a bug in `Tool.format_as_litellm_function_call` where arguments 
with default values were incorrectly included in the `required` array 
of the generated JSON Schema.

## Root Cause

In `dspy/adapters/types/tool.py`, the `format_as_litellm_function_call` 
method builds the function schema like this:

```python
"required": list(self.args.keys()),
```

This blindly marks **every** argument as required, even when the tool 
construction code (lines 108–109) already correctly stores `default` 
values on the argument properties:

```python
if default_values[k] is not inspect.Parameter.empty:
    args[k]["default"] = default_values[k]
```

So the `properties` object correctly advertises defaults, but `required` 
contradicts it — which confuses LLMs and violates the JSON Schema spec.

## Fix

Filter `required` to only include args that do not have a `default` key 
in their schema property:

```python
# Before
"required": list(self.args.keys()),

# After
"required": [k for k, v in self.args.items() if "default" not in v],
```

## Before / After

Given this tool:
```python
def look_up(query: str, state_optional: str = "fresh") -> str:
    """Look something up."""
    return f"{query}:{state_optional}"
```

**Before:**
```json
{
  "required": ["query", "state_optional"]
}
```

**After:**
```json
{
  "required": ["query"]
}
```

## Testing

Verified locally by calling `Tool(look_up).format_as_litellm_function_call()` 
and confirming only `query` appears in `required`.

Existing tests pass. A targeted test for this behavior can be added if 
the maintainers prefer it in this PR.

## References
- Fixes #9882
- JSON Schema spec: properties with `default` values should not be `required`

### DIFF
--- a/dspy/adapters/types/tool.py
+++ b/dspy/adapters/types/tool.py
@@ -158,7 +158,7 @@ class Tool(Type):
                 "parameters": {
                     "type": "object",
                     "properties": self.args,
-                    "required": list(self.args.keys()),
+                    "required": [k for k, v in self.args.items() if "default" not in v],
                 },
             },
         }
--- a/tests/adapters/test_tool.py
+++ b/tests/adapters/test_tool.py
@@ -146,6 +146,19 @@ def test_tool_from_function():
     assert tool.args["y"]["default"] == "hello"
 
 
+def test_tool_required_excludes_defaults():
+    def look_up(query: str, state_optional: str = "fresh") -> str:
+        """Look something up."""
+        return f"{query}:{state_optional}"
+
+    tool = Tool(look_up)
+    schema = tool.format_as_litellm_function_call()
+    required = schema["function"]["parameters"]["required"]
+
+    assert "query" in required
+    assert "state_optional" not in required  # has default, must not be required
+
+
 def test_tool_from_class():
     class Foo:
         def __init__(self, user_id: str):


arnav-144p:
Hi! Could one of you take a look at this when you get a chance? @rpgoldman @xdite @qhoxie @vincentkoc  — it's a minimal one-line fix, greptile reviewed it and rated 5/5 confidence. Happy to add a test or anything else needed. Thanks!

PR #9906, opened 2026-06-10
fix: honor zero values in Dataset.reset_seeds

alvinttang:
`Dataset.reset_seeds` updates each seed/size with `value or self.value`. Since `0` is falsy, passing a valid `0` is silently dropped and the previous value is kept.

Repro:
```python
from dspy.datasets.dataset import Dataset
d = Dataset(train_seed=5, train_size=10)
d.reset_seeds(train_seed=0, train_size=0)
print(d.train_seed, d.train_size)  # prints "5 10", expected "0 0"
```

This affects anyone resetting to seed 0 (a common default) or requesting an empty split with size 0.

Fix: replace the `or` fallbacks with explicit `is not None` checks so `0` is applied while omitted (`None`) arguments still leave the existing value untouched. Added tests covering both the zero case and the omitted-argument behavior.

### DIFF
--- a/dspy/datasets/dataset.py
+++ b/dspy/datasets/dataset.py
@@ -39,12 +39,12 @@ class Dataset:
         dev_size: int | None = None,
         test_size: int | None = None,
     ) -> None:
-        self.train_size = train_size or self.train_size
-        self.train_seed = train_seed or self.train_seed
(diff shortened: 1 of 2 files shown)


isaacbmiller:
lgtm

PR #9912, opened 2026-06-13
fix(evaluate): expose timeout/straggler_limit and harden straggler resubmission (#9574)

santino18727-debug:
## 📝 Changes Description

Fixes #9574.

`dspy.Evaluate` instantiated `ParallelExecutor` with the hard-coded default `timeout=120`, with no way to override it (`evaluate.py` did not pass `timeout`/`straggler_limit` through). On long-running evaluations whose examples legitimately take longer than 120s, the straggler logic resubmits work and can crash the whole evaluation with `RuntimeError: cannot schedule new futures after shutdown`.

This PR:

1. **Exposes `timeout` and `straggler_limit` on `dspy.Evaluate`** (`__init__`), forwarded to `ParallelExecutor`. Defaults are unchanged (`120` / `3`), so behavior is backwards-compatible. Setting `timeout=0` disables straggler resubmission entirely — the recommended setting for long evals (the executor already guards this path with `if 0 < self.timeout`).

2. **Hardens `ParallelExecutor._execute_parallel`:**
   - Wraps the straggler `executor.submit(...)` in `try/except RuntimeError`, so a pool that is shutting down no longer crashes the evaluation; resubmission is treated as best-effort.
   - Adds the resubmitted future to `resubmitted`, honoring the existing `# We resubmit at most once per item.` invariant even if the resubmission itself later becomes a straggler.

3. **Tests** (all passing locally):
   - `tests/utils/test_parallelizer.py`: completion under a tight straggler timeout, at-most-once resubmission, and `timeout=0` disabling resubmission.
   - `tests/evaluate/test_evaluate.py`: default values and that `timeout`/`straggler_limit` are forwarded to the executor.

### Reproduction

The straggler resubmission mechanism (and the "resubmit at most once" violation when a resubmission itself becomes a straggler) was reproduced in isolation before fixing. With `timeout=0`, the straggler path is skipped entirely and long examples are no longer resubmitted.

## ✅ Contributor Checklist

- [x] Pre-Commit checks are passing (locally) — `ruff check` is clean on the changed lines; new tests pass

## ⚠️ Warnings

The new straggler tests use short `time.sleep`-based timing (run 3× locally with no flakiness). Defaults are preserved, so this is a non-breaking change.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -79,7 +79,7 @@ class Evaluate:
         max_errors: int | None = None,
         provide_traceback: bool | None = None,
         failure_score: float = 0.0,
-        timeout: int = 120,
+        timeout: float = 120,
         straggler_limit: int = 3,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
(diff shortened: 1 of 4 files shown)


isaacbmiller:
Selecting this as the canonical fix for #9574. It covers both user-facing timeout/straggler configuration and the executor's per-item resubmission invariant, with focused tests. I re-ran the relevant tests against current main: the new Evaluate tests passed (2/2) and the parallelizer suite passed (13/13).

PR #9923, opened 2026-06-16
Feat/add append instruct to signatures

### LINKED ISSUE #9829: [Feature]  Add append_instructions to Signature

mathurk1:
## Summary
Add `Signature.append_instructions(instructions: str)` so users can layer extra task guidance on top of an existing signature's instructions without losing the original. Closes #9829.

## Motivation
`Signature.with_instructions(text)` *replaces* the docstring/instructions entirely. There's no clean way to *add* to the existing instructions — users have to read `cls.instructions`, concatenate, and feed it back through `with_instructions`, which is brittle and repetitive. The motivating use case is this helps keep the instructions DRY and composable - we can keep common engineer maintained instructions in the docstring/code base and supplement it with business team maintained behavior instructions

```python
with open("detailed_specific_instruction_file.md") as f:
    extra = f.read()
NewSig = ExtractInformation.append_instructions(extra)
extractor = dspy.Predict(NewSig)
```

## Changes

- dspy/signatures/signature.py — added Signature.append_instructions directly after with_instructions. Returns a new immutableSignature class via the same Signature(cls.fields, ...) construction path, joining the existing instructions to the new text with a blank line `("\n\n")`. Mirrors with_instructions's style: classmethod, hints, no field mutation.
- tests/signatures/test_signature.py — three regression tests covering
    - docstring-based signatures
    - chained append_instructions calls
    - signatures that only have the auto-generated default instr
- docs/docs/api/signatures/Signature.md — added append_instructions to the alphabetical members: list so the API reference page renders it.
- docs/docs/diving-deeper/signatures-in-depth.md — one line instructions in the "Mutating a signature" section.

#### Unit testing:
`pytest tests/`

#### Plus a minimal manual smoke:

```python
"""Minimal demo of Signature.with_instructions vs Signature.append_instructions."""

import dspy


class Summarize(dspy.Signature):
    """Summarize the input text in one sentence."""

    text: str = dspy.InputField()
    summary: str = dspy.OutputField()


def show(label: str, sig) -> None:
    print(f"\n=== {label} ===")
    print(sig.instructions)


show("original", Summarize)

ReplacedSig = Summarize.with_instructions(
    "Reply ONLY with a single haiku (5-7-5 syllables) that captures the input."
)
show("with_instructions", ReplacedSig)

AppendedSig = Summarize.append_instructions(
    "Additionally, end the summary with the literal token
)
show("append_instructions", AppendedSig)

Output:
=== original ===
Summarize the input text in one sentence.

=== with_instructions ===
Reply ONLY with a single haiku (5-7-5 syllables) that captures the input.

=== append_instructions ===
Summarize the input text in one sentence.

Additionally, end the summary with the literal token <END>
```

### DIFF
--- a/docs/docs/api/signatures/Signature.md
+++ b/docs/docs/api/signatures/Signature.md
@@ -6,6 +6,7 @@
     options:
         members:
             - append
+            - append_instructions
             - delete
             - dump_state
             - equals
--- a/docs/docs/diving-deeper/signatures-in-depth.md
+++ b/docs/docs/diving-deeper/signatures-in-depth.md
@@ -105,6 +105,9 @@ Every method here returns a new Signature class via deep-copy.
 **`Signature.with_instructions(instructions: str)`**  
 New class, same fields, replaced docstring.
 
+**`Signature.append_instructions(instructions: str)`**  
+New class, same fields, docstring extended by `instructions` (joined with a blank line). Use when you want to layer extra guidance on top of an existing signature (aka docstring) without restating its instructions.
+
 **`Signature.with_updated_fields(name, type_=None, **json_schema_extra)`**  
 Replaces the field’s metadata by merging the keyword arguments onto its existing `json_schema_extra`. Pass `desc=...` to change the description, `prefix=...` to change the label, or any custom key your adapter or optimizer reads. If you change `type_`, the new annotation replaces the old; if you also pass `IS_TYPE_UNDEFINED=False`, you clear the “treat None as empty string” behavior Predict relies on for unannotated string-form fields.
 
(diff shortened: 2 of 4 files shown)


isaacbmiller:
Looks great. Thanks for the PR!

PR #9924, opened 2026-06-16
fix: improve structured output fallback in JSONAdapter

Tkay6677:
Improves structured output fallback logic in JSONAdapter.
Ensures structured outputs are not prematurely downgraded to JSON mode.


isaacbmiller:
Thanks for the PR!

This is intended behavior. Closing for now, but if there is something broken feel free to give more context in an issue.

Issue #1870, opened 2024-11-28
Error occurs when using List[Literal[1, 0]]


mejessie:
```
import dspy
from typing import List, Literal

# Agent setup

# A signature
class ItemDetectionSig(dspy.Signature):
    '''Given a list of product names, evaluate each item to determine whether it represents a valid product name.
    Return a list of the same length as the input length, where each item is labeled 'Yes' if it is a valid product name and 'No' otherwise.
    For instance, if the input list contains five items, the output list should also have five items, each corresponding directly to the validity of the input item.'''
    
    items: list[str] = dspy.InputField(prefix='Items:', desc='list of comma-separated product names')
    length: int = dspy.InputField(prefix='Length:', desc='length of the list')
    responses: List[Literal[1, 0]] = dspy.OutputField()

# A module
ItemDetectionModule = dspy.Predict(ItemDetectionSig)
```

From the above code, I need the output in a list of yes or no.
The error occurs on line `List[Literal[1, 0]] = dspy.OutputField()`

```
Traceback (most recent call last):
  File "C:\Users\Jedsada\anaconda3\envs\ai01\Lib\site-packages\dspy\adapters\chat_adapter.py", line 78, in parse
    fields[k] = parse_value(v, signature.output_fields[k].annotation) if _parse_values else v
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Jedsada\anaconda3\envs\ai01\Lib\site-packages\dspy\adapters\chat_adapter.py", line 164, in parse_value
    return TypeAdapter(annotation).validate_python(parsed_value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Jedsada\anaconda3\envs\ai01\Lib\site-packages\pydantic\type_adapter.py", line 412, in validate_python
    return self.validator.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for list[literal[1,0]]
  Input should be a valid list [type=list_type, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/list_type

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #9932, opened 2026-06-18
Fix parse_value for int-valued Literal fields in ChatAdapter

Wassbdr:
## What

`parse_value` raised `ValueError` for output fields typed as an int-valued `Literal`, e.g. `Literal[1, 2, 3]`.

Adapters that surface every field as text (notably `ChatAdapter`) hand the field value to `parse_value` as a **string** (`"1"`). For a `Literal`, `parse_value` only compared the raw/stripped string against the allowed members, so a non-string member (`1`) never matched `"1"` and the call raised:

```
ValueError: '1' is not one of (1, 2, 3)
```

This is the scalar form of the long-standing problem reported in #1870 (`List[Literal[1, 0]]`).

## Fix

In the `Literal` branch of `parse_value`, after the existing exact-match attempts, fall back to matching the stripped string against the **stringified form** of each non-string member, returning the original typed member:

```python
for member in allowed:
    if not isinstance(member, str) and v == str(member):
        return member
```

- Exact string matches still take precedence (`Literal[1, "bar"]` + `"bar"` → `"bar"`).
- Out-of-set values still raise (`"4"` for `Literal[1, 2, 3]`).
- Only non-string members are stringified, so string literals are unaffected.

## Tests

Added `test_parse_value_literal_non_string_members` covering int-only and mixed int/str literals, including the out-of-set raise.

```
# new test fails on main (proves the bug):
tests/adapters/test_adapter_utils.py::test_parse_value_literal_non_string_members
  ValueError: '1' is not one of (1, 2, 3)

# with the fix:
$ pytest tests/adapters/test_adapter_utils.py -q
7 passed

# regression check:
$ pytest tests/adapters/test_chat_adapter.py tests/adapters/test_adapter_utils.py -q
76 passed
```

`ruff format` clean; no new `ruff check` findings on the changed lines.

Related: #1870

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -170,6 +170,13 @@ def parse_value(value, annotation):
             if v in allowed:
                 return v
 
+            # Non-string members (e.g. `Literal[1, 2, 3]`) never equal the raw
+            # text emitted by adapters that surface every field as a string, so
+            # fall back to matching on the stringified member.
+            for member in allowed:
+                if not isinstance(member, str) and v == str(member):
+                    return member
+
         raise ValueError(f"{value!r} is not one of {allowed!r}")
 
     if not isinstance(value, str):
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Superseded by #10010

PR #9941, opened 2026-06-23
fix(propose): handle None parent namespace for dynamic signatures

### LINKED ISSUE #9937: [Bug] TypeError: 'NoneType' object is not subscriptable in propose/utils.py with dynamic signatures
### What happened?

When trying to optimize a DSPy program containing dynamically generated signatures, the prompt proposer crashes with `TypeError: 'NoneType' object is not subscriptable`.

In `dspy/propose/utils.py`:
```python
if hasattr(item, "signature") and item.signature is not None and item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig" not in completed_set:
```
Dynamic signatures do not have a `__pydantic_parent_namespace__` (it is `None`), causing the lookup to fail.

### Steps to reproduce

1. Define a dynamic signature at runtime.
2. Use it in a predictor.
3. Attempt to run MIPROv2 prompt optimization.

```python
import dspy
from dspy.teleprompt import MIPROv2

# Create dynamic signature
dynamic_sig = dspy.Signature("question -> answer")

class DynamicProgram(dspy.Module):
    def __init__(self):
        super().__init__()
        self.predict = dspy.Predict(dynamic_sig)
        
    def forward(self, question):
        return self.predict(question=question)

# Setup optimizer and compile (which triggers the grounded proposer)
teleprompter = MIPROv2(metric=your_metric)
optimized_program = teleprompter.compile(
    DynamicProgram(),
    trainset=trainset,
    valset=valset
)
```

### DSPy version

3.2.1

robbiebusinessacc:
Closes #9937.

  ## What
  `get_dspy_source_code` (in `dspy/propose/utils.py`) builds a dedup key from
  `item.signature.__pydantic_parent_namespace__["signature_name"]`. For dynamically
  created signatures — e.g. `dspy.Signature("question -> answer")` — that attribute can
  be `None`, so the subscript raises `TypeError: 'NoneType' object is not subscriptable`
  and crashes `MIPROv2` optimization for any program containing a dynamic signature.

  ## Root cause
  Dynamic signatures are built via `create_model(__base__=Signature)`. On Python 3.14,
  `SignatureMeta` sets `__pydantic_reset_parent_namespace__ = False` (to avoid a
  cloudpickle issue), and pydantic leaves `__pydantic_parent_namespace__` as `None` for
  these classes. The grounded proposer then subscripts `None` when collecting source code.

  ## Fix
  Read the parent namespace defensively and fall back to the signature class `__name__`
  when it is unavailable. For class-defined signatures `__name__` equals the existing
  `signature_name` value, so the dedup key — and therefore behavior — is unchanged on
  that path; only the previously-crashing dynamic-signature path changes.

  ## Tests
  Adds `tests/propose/test_utils.py`:
  - `test_get_dspy_source_code_with_dynamic_signature` reproduces the `None`
    parent-namespace state and asserts a string is returned instead of raising. It fails
    on `main` with the reported `TypeError` and passes with this change.
  - `test_get_dspy_source_code_with_class_signature` confirms the class-defined path is
    unaffected.

### DIFF
--- a/dspy/propose/utils.py
+++ b/dspy/propose/utils.py
@@ -172,13 +172,21 @@ def get_dspy_source_code(module):
             except TypeError:
                 continue
             if isinstance(item, Parameter):
-                if hasattr(item, "signature") and item.signature is not None and item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig" not in completed_set:
-                    try:
-                        header.append(inspect.getsource(item.signature))
-                        print(inspect.getsource(item.signature))
-                    except (TypeError, OSError):
-                        header.append(str(item.signature))
-                    completed_set.add(item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig")
+                if hasattr(item, "signature") and item.signature is not None:
+                    # Dynamically created signatures (e.g. dspy.Signature("q -> a")) may have
+                    # __pydantic_parent_namespace__ set to None, so fall back to the class name.
+                    parent_namespace = getattr(item.signature, "__pydantic_parent_namespace__", None)
+                    if parent_namespace and "signature_name" in parent_namespace:
+                        signature_key = parent_namespace["signature_name"] + "_sig"
+                    else:
+                        signature_key = item.signature.__name__ + "_sig"
+                    if signature_key not in completed_set:
+                        try:
+                            header.append(inspect.getsource(item.signature))
+                            print(inspect.getsource(item.signature))
+                        except (TypeError, OSError):
+                            header.append(str(item.signature))
+                        completed_set.add(signature_key)
             if isinstance(item, dspy.Module):
                 code = get_dspy_source_code(item).strip()
                 if code not in completed_set:
--- a/tests/propose/test_utils.py
+++ b/tests/propose/test_utils.py
@@ -1,0 +1,37 @@
+import dspy
+from dspy.propose.utils import get_dspy_source_code
+
+
+def test_get_dspy_source_code_with_dynamic_signature():
+    """Dynamically created signatures can leave ``__pydantic_parent_namespace__`` set
+    to ``None`` (for example on Python 3.14), and extracting their source code must not
+    crash with ``TypeError: 'NoneType' object is not subscriptable``."""
+
+    class DynamicProgram(dspy.Module):
+        def __init__(self):
+            super().__init__()
+            self.predict = dspy.Predict(dspy.Signature("question -> answer"))
+
+    program = DynamicProgram()
+    # Reproduce the state where pydantic leaves the parent namespace unset.
+    program.predict.signature.__pydantic_parent_namespace__ = None
+
+    source = get_dspy_source_code(program)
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Ran this on 3.10 and then simulated the 3.14 state, and both branches of the new key end up producing the same string for every dynamic signature, so the crash goes away but the drop it was masking stays.

**≤3.13** — `__pydantic_reset_parent_namespace__` is not set, so `__pydantic_parent_namespace__` is a populated dict and the `if parent_namespace and "signature_name" in parent_namespace` branch is taken. The value under that key is a leaked local from `SignatureMeta.__new__` (`signature_name` is the `mcs` argument on line 138 of `signatures/signature.py`), and for anything built through `make_signature` it is the literal `"Signature"`:

| construction | `__name__` | `parent_ns["signature_name"]` | key |
|---|---|---|---|
| `dspy.Signature("question -> answer")` | `StringSignature` | `'Signature'` | `Signature_sig` |
| `dspy.Signature("text -> summary")` | `StringSignature` | `'Signature'` | `Signature_sig` |
| `ChainOfThought(QA).predict.signature` | `StringSignature` | `'Signature'` | `Signature_sig` |

**3.14** — the fallback fires, but `__name__` is `StringSignature` for all three, so the key is `StringSignature_sig` across the board.

Either way `completed_set` is hit on the second signature and it never reaches `program_code_string`:

```python

    def __init__(self):
        super().__init__()
        self.qa = dspy.Predict(QA)

get_dspy_source_code(TwoDynamicSignatures())
```

| | `-> answer` in header | `-> summary` in header |
|---|---|---|
| main | yes | no |
| this PR (3.10 path) | yes | no |
| this PR (3.14 path, `__pydantic_parent_namespace__` forced to `None`) | yes | no |
| keyed on the emitted source | yes | yes |

`grounded_proposer` hands that string to the instruction LM as `program_code_string`, so on the current behavior the summarizer's signature is simply not visible to it.

The `dspy.Module` branch four lines below already keys on the emitted source. Doing the same here also drops the `__pydantic_parent_namespace__` read entirely, which is what makes #9937 version-dependent in the first place:

```python
if getattr(item, "signature", None) is not None:
    try:
        signature_code = inspect.getsource(item.signature)
    except (TypeError, OSError):
        signature_code = str(item.signature)
    if signature_code not in completed_set:
        header.append(signature_code)
        completed_set.add(signature_code)
```

I ran your two tests against that unchanged and they pass; `tests/propose` is 6 passed. Two `dspy.Predict(QA)` sharing one class signature still emit `QA` once, so the dedup isn't lost. I also confirmed `test_get_dspy_source_code_with_dynamic_signature` fails on main (`1 failed, 1 passed`), so it is pinning the reported `TypeError`.

One leftover: `print(inspect.getsource(item.signature))` right after `header.append(...)` is carried over from main. It dumps every signature's source to stdout on each proposer call.

Measured on `dspy` @ 67bf5eb, Python 3.10.12, pydantic 2.12.5.

PR #9942, opened 2026-06-23
fix(ChatAdapter): keep field-header offset on the stripped line in parse

Srivatsa03:
## What

`ChatAdapter.parse` matches the field-header pattern against `line.strip()` but then slices the original, unstripped `line` with `match.end()`:

```python
match = field_header_pattern.match(line.strip())   # offset is into the STRIPPED line
if match:
    header = match.group(1)
    remaining_content = line[match.end() :].strip()  # but slices the UNSTRIPPED line
```

When a header line has leading whitespace, `match.end()` (an offset into the stripped string) is applied to the longer unstripped string, so the slice starts inside the trailing `## ]]` marker and leaks marker characters into the value.

## Repro (before this change)

```python
import dspy

sig = dspy.Signature("question -> answer")
ad = ChatAdapter()

ad.parse(sig, "[[ ## answer ## ]] Paris")     # {'answer': 'Paris'}   (ok, no indent)
ad.parse(sig, "  [[ ## answer ## ]] Paris")   # {'answer': ']] Paris'} (leaked marker)
ad.parse(sig, "  [[ ## answer ## ]]\nParis")  # {'answer': ']]\nParis'}
```

LMs commonly indent markers (markdown lists, quoted/nested blocks), and `ChatAdapter` is the default adapter, so this is silent value corruption on the common path. For non-string fields it instead raises a spurious `AdapterParseError`.

## Fix

Match and slice the same stripped line so the offset is consistent. Non-header content lines are still appended unchanged, so multi-line values are unaffected.

## Tests

Added `test_chat_adapter_parse_handles_indented_field_header`, which fails on `main` and passes with this change. It covers a non-indented control, an indented header, and an indented header with the value on the next line. Found via code inspection; no existing test exercised an indented header.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -219,11 +219,12 @@ class ChatAdapter(Adapter):
         sections = [(None, [])]
 
         for line in completion.splitlines():
-            match = field_header_pattern.match(line.strip())
+            stripped_line = line.strip()
+            match = field_header_pattern.match(stripped_line)
             if match:
                 # If the header pattern is found, split the rest of the line as content
                 header = match.group(1)
-                remaining_content = line[match.end() :].strip()
+                remaining_content = stripped_line[match.end() :].strip()
                 sections.append((header, [remaining_content] if remaining_content else []))
             else:
                 sections[-1][1].append(line)
--- a/tests/adapters/test_chat_adapter.py
+++ b/tests/adapters/test_chat_adapter.py
@@ -2648,6 +2648,27 @@ def test_chat_adapter_parses_float_with_underscores():
         assert result[0]["score"].score == 123456.789
 
 
+def test_chat_adapter_parse_handles_indented_field_header():
+    """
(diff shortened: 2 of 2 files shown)


Srivatsa03:
Gentle bump on this — small two-line fix for silent value corruption in `ChatAdapter.parse` when a field header is indented (markdown list / nested block), with a direct regression test. Greptile reviewed it 5/5 with no issues. Would appreciate a look when you have a moment @isaacbmiller, thanks!

Srivatsa03:
hey @isaacbmiller, mind approving the workflow run on this when you get a sec? CI hasn't kicked off yet since it's my first PR here. it's a small 2-line fix for silent value corruption in `ChatAdapter.parse` on indented field headers, with a regression test that fails before / passes after. happy to rebase if anything's stale 🙏

Srivatsa03:
@isaacbmiller could you approve the workflow run here? CI has never triggered on this first-time PR. Two-line fix, Greptile 5/5, merges clean.

PR #9946, opened 2026-06-25
Preserve input/label split when copying an Example

iamsharduld:
## Summary

`Example.copy()`, `Example.without()`, and `Example(base=existing_example)` all route through
`Example.__init__`, which copies `_store` but leaves `_input_keys = None`. As a result a copied
`Example` loses the input/label split set by `.with_inputs()` — `.inputs()` then raises and
`.labels()` breaks.

## Reproduce (no LM)

```python
import dspy
ex = dspy.Example(question="q", answer="a").with_inputs("question")
print(ex.inputs().keys())          # ['question']  ✓
cp = ex.copy(answer="b")
print(cp.inputs().keys())          # ValueError: Inputs have not been set for this example.
```

This matters because optimizers/evaluators copy examples (e.g. `BootstrapFewShot`, demo handling)
and rely on `.inputs()`/`.labels()`. `with_inputs()` only works because it *re-sets* `_input_keys`
after the internal `copy()`; a plain `.copy()` does not.

## Fix

In `Example.__init__`, when `base` is an `Example`, inherit its `_input_keys` (as a fresh `set` to
avoid aliasing). `with_inputs()` still overrides it explicitly, so its behavior is unchanged;
`inputs()`/`labels()` build from a **dict** base and are unaffected.

## Tests

`tests/primitives/test_example.py::test_example_copy_preserves_input_keys` — fail-before:
`copied._input_keys is None`; pass-after: `{"question"}`, and `.inputs()` / `.labels()` / `without()`
work on the copy. Full `test_example` / `test_module` / `signatures` (136) and `predict` /
`teleprompt` (270) pass. CPU-only, no GPU.

(Out of scope: the same `__init__` branch also doesn't carry `_demos` from an Example base; left
alone here to keep this focused on the `_input_keys` data-loss bug.)

### DIFF
--- a/dspy/primitives/example.py
+++ b/dspy/primitives/example.py
@@ -111,6 +111,12 @@ class Example:
         # Initialize from a base Example if provided
         if base and isinstance(base, type(self)):
             self._store = base._store.copy()
+            # Preserve the input/label split so copy()/without()/Example(base=...)
+            # don't silently break .inputs()/.labels() on the new instance.
+            # Use getattr: subclasses such as Prediction don't keep _input_keys.
+            base_input_keys = getattr(base, "_input_keys", None)
+            if base_input_keys is not None:
+                self._input_keys = set(base_input_keys)
 
         # Initialize from a dict if provided
         elif base and isinstance(base, dict):
--- a/tests/primitives/test_example.py
+++ b/tests/primitives/test_example.py
@@ -120,6 +120,36 @@ def test_example_copy_without():
         _ = without_a.a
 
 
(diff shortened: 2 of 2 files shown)


itxaiohanglover:
Nice catch! Resetting _input_keys to None on copy was a silent bug — .inputs()/.labels() would break on any copied Example. Test coverage for copy(), without(), and Example(base=...) is thorough.

isaacbmiller:
LGTm, thanks for the PR

PR #9948, opened 2026-06-25
Sample labeled demos independently per predictor in BootstrapFewShot

iamsharduld:
## Summary

`BootstrapFewShot._train` reassigns the shared `raw_demos` pool to the sampled subset **inside** the
per-predictor loop:

```python
raw_demos = self.validation
for name, predictor in self.student.named_predictors():
    augmented_demos = self.name2traces[name][: self.max_bootstrapped_demos]
    sample_size = max(0, min(self.max_labeled_demos - len(augmented_demos), len(raw_demos)))
    raw_demos = rng.sample(raw_demos, sample_size)   # <-- shrinks the shared pool
    predictor.demos = augmented_demos + raw_demos
```

So each subsequent predictor samples from the *previous* predictor's leftovers rather than the full
validation pool. Two consequences:

1. Predictors get an identical, monotonically shrinking demo set instead of independent samples.
2. When an early predictor consumes most of the pool via bootstrapped (augmented) demos, later
   predictors are **starved** of labeled demos.

## Reproduce

A 2-predictor program where the first predictor has 5 bootstrapped demos (`max_bootstrapped_demos=5`,
`max_labeled_demos=8`) and a 10-example validation pool: the first predictor takes `5 + 3` labeled, and
the second predictor — which should still get 8 labeled from the full pool — receives only **3** (the
first predictor's leftovers).

## Fix

Sample into a fresh local variable so the full pool is reused for every predictor:

```python
sampled_demos = rng.sample(raw_demos, sample_size)
predictor.demos = augmented_demos + sampled_demos
```

Behavior-preserving for the single-predictor case.

## Tests

`tests/teleprompt/test_bootstrap.py::test_bootstrap_labeled_demos_sampled_independently_per_predictor`
— fail-before: second predictor gets 3 demos; pass-after: 8. Full `test_bootstrap.py` suite passes.
Deterministic (`random.Random(0)`), no LM/GPU needed.

### DIFF
--- a/dspy/teleprompt/bootstrap.py
+++ b/dspy/teleprompt/bootstrap.py
@@ -266,7 +266,7 @@ class BootstrapFewShot(Teleprompter):
             sample_size = min(self.max_labeled_demos - len(augmented_demos), len(raw_demos))
             sample_size = max(0, sample_size)
 
-            raw_demos = rng.sample(raw_demos, sample_size)
-            predictor.demos = augmented_demos + raw_demos
+            sampled_demos = rng.sample(raw_demos, sample_size)
+            predictor.demos = augmented_demos + sampled_demos
 
         return self.student
--- a/tests/teleprompt/test_bootstrap.py
+++ b/tests/teleprompt/test_bootstrap.py
@@ -36,6 +36,42 @@ class SimpleModule(dspy.Module):
         return self.predictor(**kwargs)
 
 
+class TwoPredictorModule(dspy.Module):
+    def __init__(self):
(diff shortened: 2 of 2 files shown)


itxaiohanglover:
Good catch! The shared `raw_demos` reassignment was a subtle bug — later predictors were sampling from leftovers. Using a separate `sampled_demos` variable is the clean fix. The regression test with TwoPredictorModule nicely demonstrates the starvation scenario.

isaacbmiller:
LGTM

PR #9949, opened 2026-06-25
Match fields by name in Signature.load_state

iamsharduld:
## Summary

`Signature.dump_state` serializes each field's `prefix`/`description` **positionally**, with no field
name, and `Signature.load_state` restores them with a positional `zip`:

```python
# dump_state
state["fields"].append({"prefix": ..., "description": ...})   # no name
# load_state
for field, saved_field in zip(signature_copy.fields.values(), state["fields"], strict=False):
    field.json_schema_extra["prefix"] = saved_field["prefix"]
    field.json_schema_extra["desc"] = saved_field["description"]
```

So loading a saved state into a signature whose fields are declared in a **different order** silently
swaps prefixes/descriptions between fields, and a different field count drops or misaligns them.

## Reproduce

```python
import dspy

class Source(dspy.Signature):
    """Source instruction."""
    question = dspy.InputField(desc="the question")
    context  = dspy.InputField(desc="the context")
    answer   = dspy.OutputField(desc="the answer")

state = Source.dump_state()

class Target(dspy.Signature):
    """Target instruction."""
    context  = dspy.InputField(desc="orig context")
    question = dspy.InputField(desc="orig question")
    answer   = dspy.OutputField(desc="orig answer")

loaded = Target.load_state(state)
print(loaded.fields["question"].json_schema_extra["desc"])  # "the context"  <- wrong (swapped)
```

## Fix

Serialize the field `name` in `dump_state` and match by name in `load_state`, with a **positional
fallback** when the state lacks names so older saved state files still load:

```python
# dump_state: {"name": field, "prefix": ..., "description": ...}
# load_state: if every saved field has a "name", assign by name; else fall back to the positional zip.
```

## Tests

`tests/signatures/test_signature.py`:
- `test_load_state_matches_fields_by_name` — fail-before: `question` gets `"the context"`; pass-after:
  each field keeps its own prefix/desc across a reordered target signature.
- `test_load_state_backward_compatible_without_names` — legacy state (no `name`) still loads positionally.
- Existing `test_dump_and_load_state` updated for the added `name` key.

Full `tests/signatures` (49) and `tests/predict` (197) pass. No LM/GPU needed.

### DIFF
--- a/dspy/signatures/signature.py
+++ b/dspy/signatures/signature.py
@@ -526,6 +526,7 @@ class Signature(BaseModel, metaclass=SignatureMeta):
         for field in cls.fields:
             state["fields"].append(
                 {
+                    "name": field,
                     "prefix": cls.fields[field].json_schema_extra["prefix"],
                     "description": cls.fields[field].json_schema_extra["desc"],
                 }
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Nice fix! Matching by field name instead of position is much more robust. The backward compatibility fallback for state files saved before field names were serialized is a thoughtful touch — won't break existing serialized models. Test with reordered fields is exactly the right regression case.

PR #9964, opened 2026-07-02
feat: add tool call exact match metric

### LINKED ISSUE #9961: [Feature]  Add tool_call_exact_match for dspy.ToolCalls

DSPy now has first-class structured tool-call types (`dspy.ToolCalls` and
`dspy.ToolCallResults`), and ReActV2 records native tool calls in structured history. However,
`dspy.evaluate` does not currently provide a deterministic metric for comparing labeled and
predicted tool calls. Users with a `ToolCalls` output field must reimplement tool-name and argument
matching in each evaluation or optimization pipeline.

I would like to contribute `tool_call_exact_match`, a small standalone metric following DSPy's
existing `answer_exact_match(example, pred, trace=None)` convention.

For a narrow and predictable v1, I propose these semantics:

- Read `example.tool_calls` and `pred.tool_calls`, accepting `dspy.ToolCalls` or any shape already
  supported by `ToolCalls.model_validate`.
- Ignore provider-generated call IDs and `ToolCallResults`; compare only tool names and arguments.
- Compare multiple calls as an order-independent multiset so parallel-call ordering does not affect
  correctness while duplicate calls remain significant.
- Require the same tool names and exact argument structures, with no fuzzy string matching or type
  coercion.
- Return `False` for any missing or extra call, argument, or differing value; otherwise return
  `True`.
- Do not execute tools or use an LLM judge.

Example usage:

```python
import dspy
from dspy.evaluate import tool_call_exact_match

gold = dspy.Example(
    tool_calls=dspy.ToolCalls.from_dict_list(
        [{"name": "search", "args": {"query": "DSPy", "limit": 5}}]
    )
)
pred = dspy.Prediction(
    tool_calls=dspy.ToolCalls.from_dict_list(
        [{"name": "search", "args": {"query": "DSPy", "limit": 5}}]
    )
)

assert tool_call_exact_match(gold, pred)
```

Automatic extraction from `ReAct` trajectories or `ReActV2` history would remain out of scope for
this initial metric. Programs using those modules could explicitly surface the calls they want to
evaluate as a `tool_calls` output. Partial-credit metrics such as tool-selection accuracy and
argument/slot accuracy could be considered separately after the exact-match contract is established.

If this scope is welcome, I am happy to implement the metric, export it from `dspy.evaluate`, add
unit and `Evaluate` integration tests, and provide a short documentation example.

### Additional Context

- DSPy 3.3.0b1 introduced ReActV2 together with structured/native tool-call history based on
  `dspy.ToolCalls` and `dspy.ToolCallResults`, but the current evaluation package exports no
  corresponding deterministic tool-call metric:
  - https://github.com/stanfordnlp/dspy/releases/tag/3.3.0b1
  - https://github.com/stanfordnlp/dspy/blob/main/dspy/adapters/types/tool.py
  - https://github.com/stanfordnlp/dspy/blob/main/dspy/evaluate/metrics.py
- BFCL provides precedent for deterministic function-call evaluation using AST and execution-based
  checks. This proposal is intentionally smaller: exact comparison of DSPy's already-normalized
  `ToolCalls` objects without execution:
  https://gorilla.cs.berkeley.edu/blogs/8_berkeley_function_calling_leaderboard.html
- JTPRO evaluates tool-calling with Tool Selection Accuracy, Slot Filling Accuracy, and Overall
  Success Rate, demonstrating the usefulness of call-level correctness as an optimization signal:
  https://arxiv.org/abs/2604.19821
- A recent ICML study finds that tool-calling evaluation can vary substantially with implementation
  choices such as system prompts, history handling, and multi-turn templates, which motivates making
  the matching contract explicit and deterministic:
  https://arxiv.org/abs/2606.00135
- Tool-calling also remains outside the evaluated scope of several prompt-optimization frameworks.
  RPT lists tool-using agents among settings its experiments do not cover, while GTBP identifies
  deeper graphs, tool-calling, and dynamic routing as unsupported real-world workflow structures:
  - https://arxiv.org/abs/2605.21781
  - https://arxiv.org/abs/2606.14155

sjsjsjjs534:
## Summary
- Adds `tool_call_exact_match(example, pred, trace=None)` for deterministic comparison of `dspy.ToolCalls`.
- Accepts any shape supported by `ToolCalls.model_validate`, ignores provider-generated call IDs and `ToolCallResults`, and compares tool names plus argument structures exactly.
- Treats multiple calls as an order-independent multiset so parallel-call ordering does not affect correctness while duplicate calls remain significant.
- Exports the metric from `dspy.evaluate`.

Closes #9961.

## Test Plan
- `uv run --with pytest pytest tests/evaluate/test_metrics.py -q`
  - Before the implementation, collection failed because `tool_call_exact_match` was missing.
- `uv run --with ruff ruff check dspy/evaluate/metrics.py dspy/evaluate/__init__.py tests/evaluate/test_metrics.py`
- `uv run --with ruff ruff format --check dspy/evaluate/metrics.py dspy/evaluate/__init__.py tests/evaluate/test_metrics.py`

### DIFF
--- a/dspy/evaluate/__init__.py
+++ b/dspy/evaluate/__init__.py
@@ -1,12 +1,13 @@
 from dspy.evaluate.auto_evaluation import CompleteAndGrounded, SemanticF1
 from dspy.evaluate.evaluate import Evaluate, EvaluationResult
-from dspy.evaluate.metrics import EM, answer_exact_match, answer_passage_match, normalize_text
+from dspy.evaluate.metrics import EM, answer_exact_match, answer_passage_match, normalize_text, tool_call_exact_match
 
 __all__ = [
     "EM",
     "normalize_text",
     "answer_exact_match",
     "answer_passage_match",
+    "tool_call_exact_match",
     "Evaluate",
     "SemanticF1",
     "CompleteAndGrounded",
--- a/dspy/evaluate/metrics.py
+++ b/dspy/evaluate/metrics.py
@@ -323,16 +323,15 @@ def _freeze_tool_call_arg(value):
         return ("dict", tuple((key, _freeze_tool_call_arg(value[key])) for key in sorted(value)))
     if isinstance(value, list):
         return ("list", tuple(_freeze_tool_call_arg(item) for item in value))
-    return (type(value).__qualname__, value)
+    return (type(value), value)
 
 
 def _tool_call_counter(value):
     try:
         tool_calls = ToolCalls.model_validate(value).tool_calls
+        return Counter((tool_call.name, _freeze_tool_call_arg(tool_call.args)) for tool_call in tool_calls)
     except Exception:
         return None
-
-    return Counter((tool_call.name, _freeze_tool_call_arg(tool_call.args)) for tool_call in tool_calls)
 
 
 def tool_call_exact_match(example, pred, trace=None):
--- a/dspy/evaluate/metrics.py
+++ b/dspy/evaluate/metrics.py
@@ -5,6 +5,7 @@ import string
 import unicodedata
 from collections import Counter
 
+from dspy.adapters.types import ToolCalls
 from dspy.dsp.utils.utils import print_message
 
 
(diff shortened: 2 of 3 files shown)


YashDThapliyal:
Hi @sjsjsjjs534 — I’m surprised to see this PR. I spent substantial time researching and scoping #9961 against the existing DSPy APIs, and I explicitly stated that I wanted to implement it after maintainer confirmation. This PR appears to implement that specification without any prior coordination with me.

@maintainers, could you please advise how you’d like to proceed? I’m still interested in implementing the issue and had been waiting for confirmation in accordance with the contribution guidelines. I’m open to collaboration, but I would appreciate acknowledgment of the research and specification work already done and the opportunity to contribute to the implementation.

PR #9984, opened 2026-07-08
Flatten tools and tool_choice for the Responses API

### LINKED ISSUE #9943: [Bug]  model_type="responses" + native function calling fails with "Missing required parameter: 'tools[0].name'" — tools/tool_choice not flattened to the Responses shape
### What happened?

 With `dspy.LM(model_type="responses")` + native function calling
(`ChatAdapter(use_native_function_calling=True)`, e.g. via `ReActV2`), every request to the
OpenAI **Responses API** is rejected:

```
litellm.BadRequestError: OpenAIException - {
  "error": {
    "message": "Missing required parameter: 'tools[0].name'.",
    "type": "invalid_request_error",
    "param": "tools[0].name",
    "code": "missing_required_parameter"
  }
}
```

**Root cause.** `dspy/clients/lm.py::_convert_chat_request_to_responses_request` converts
`messages`, `reasoning_effort`, and `response_format` into Responses-API shapes, but
**passes `tools` (and Chat-object `tool_choice`) through untouched** (`lm.py:626-657`). The
native-FC adapter emits **Chat Completions**-format tools — `{"type":"function","function":
{"name":...}}` (`dspy/adapters/base.py:110-118` → `dspy/adapters/types/tool.py:152-164`) —
while the Responses API requires the **flattened** form `{"type":"function","name":...,
"parameters":...}` (top-level `name`; see `openai/types/responses/function_tool_param.py`).
So the function name lands at `tools[0].function.name`, and the Responses endpoint rejects
the request because `tools[0].name` is missing. LiteLLM doesn't reshape it on the native
OpenAI Responses path either (`litellm/llms/openai/responses/transformation.py` does no
param-mapping/request-transform), so the fix needs to live in DSPy's converter.

**Net effect:** native function calling is unusable on `model_type="responses"` — the exact
combination needed to run a tool-using agent (`ReActV2`) on a reasoning model via the
Responses API.

**Expected behavior.** When `model_type="responses"` and native tools are present, DSPy
should translate them to the Responses-API shape before the call:
- **tools** — flatten each `{"type":"function","function":{name, description, parameters,
  strict?}}` → `{"type":"function", name, description, parameters, strict?}`; pass
  already-flat/hosted tools through unchanged.
- **tool_choice** — string values (`"auto"`/`"required"`/`"none"`) are already valid on
  Responses and need no change; the **object** form must be flattened:
  `{"type":"function","function":{"name":"X"}}` → `{"type":"function","name":"X"}`.

**Suggested fix.** Add a `tools`/`tool_choice` translation branch to
`_convert_chat_request_to_responses_request`, alongside the existing
`messages`/`reasoning_effort`/`response_format` handling:

```python
if "tools" in request:
    request["tools"] = [_chat_tool_to_responses_tool(t) for t in request["tools"]]
if isinstance(request.get("tool_choice"), dict):
    request["tool_choice"] = _chat_tool_choice_to_responses(request["tool_choice"])
```

where `_chat_tool_to_responses_tool` hoists `t["function"]` to the top level for
`{"type":"function", ...}` tools (passing other tool types through). Happy to open a PR.

**Why it matters / context.** This blocks running `ReActV2` (native tool calling) on a
reasoning model via the Responses API. It compounds with the GPT-5-on-chat guard in
`dspy/adapters/types/reasoning.py:68-73` (`if "gpt-5" in lm.model and lm.model_type ==
"chat": return signature`, citing litellm [#14748](https://github.com/BerriAI/litellm/issues/14748)):
**that guard disables native reasoning for GPT-5-family models on chat**, so "GPT-5 + native
reasoning" pushes the setup toward `model_type="responses"` — where this tools bug then
blocks it. So today there is no path to "GPT-5 + native reasoning + native tools" in dspy;
fixing this converter is what unblocks it. (The guard is GPT-5-substring-only, so o-series
models are unaffected.)

**Workaround.** Use an o-series reasoning model (`o3` / `o4-mini`) on `model_type="chat"`.
o-series gets native reasoning on Chat Completions (the GPT-5 guard is GPT-5-specific;
`_is_openai_reasoning_model` at `lm.py:47-52` still classifies `o4-mini` as a reasoning
model), and native function calling works on chat because the tool format already matches
the chat endpoint. The only thing forgone vs. Responses is cross-tool-call reasoning
persistence — a quality/efficiency optimization, not a functional blocker.

### 1. Minimal — root cause, no network call

```python
from dspy.clients.lm import _convert_chat_request_to_responses_request

req = {
    "messages": [{"role": "user", "content": "hi"}],
    "reasoning_effort": "low",
    "tools": [{
        "type": "function",
        "function": {  # Chat Completions nesting, produced by the native-FC adapter
            "name": "search",
            "description": "Search a knowledge base.",
            "parameters": {"type": "object", "properties": {"query": {"type": "string"}}},
        },
    }],
}

out = _convert_chat_request_to_responses_request(req)
print(out["tools"])
# -> unchanged: {"type": "function", "function": {"name": "search", ...}}
#    Responses needs:  {"type": "function", "name": "search", "parameters": {...}}
#    so tools[0].name is missing and OpenAI rejects the request.
```

The converter handles `messages`→`input` (`lm.py:626-639`), `reasoning_effort`→`reasoning`
(`lm.py:640-643`), `response_format`→`text.format` (`lm.py:645-655`), then returns the
request otherwise unchanged (`lm.py:657`). The responses completion path forwards
`**request` to `litellm.responses`/`aresponses` (`lm.py:588-617`).

### 2. End-to-end (live call)

```python
import dspy
from dspy.predict.react_v2 import ReActV2  # native-tool-calling ReAct

def search(query: str) -> str:
    """Search a knowledge base for the query."""
    return "some result"

lm = dspy.LM("openai/gpt-5-mini", model_type="responses", temperature=1.0)
dspy.configure(lm=lm, adapter=dspy.ChatAdapter(use_native_function_calling=True))

ReActV2("question -> answer", tools=[search])(question="find something")
#  -> litellm.BadRequestError: ... Missing required parameter: 'tools[0].name'.
```

The same setup works on `model_type="chat"` (default). Only `model_type="responses"` fails,
and only because of the tool-definition shape.

> Note: for reasoning models, dspy maps `max_tokens` → `max_completion_tokens`
> (`lm.py:124-132`), which is not the Responses token param (`max_output_tokens`). If you
> pass a token cap and hit an unrelated param error first, drop it or pass
> `max_output_tokens=16000`. dspy also requires `temperature` ∈ {None, 1.0} for reasoning models.

### DSPy version

- `dspy == 3.3.0b1` - `litellm == 1.81.0` - Python 3.12, provider: OpenAI - Reproduced directly with `openai/gpt-5-mini` on `model_type="responses"`. The cause is   model-agnostic (Chat-format tools → Responses endpoint), so any model on   `model_type="responses"` with native tools hits it.

AnayGarodia:
Fixes #9943

Using `dspy.LM(model_type="responses")` with native function calling fails every request with `Missing required parameter: 'tools[0].name'`. `_convert_chat_request_to_responses_request` converts `messages`, `reasoning_effort`, and `response_format` to the Responses API shape, but has no branch for `tools` / `tool_choice`. They are emitted in Chat Completions shape (`{"type": "function", "function": {"name": ..., "parameters": ...}}`) by `Tool.format_as_litellm_function_call`, whereas the Responses API expects the function fields flattened onto the tool (`{"type": "function", "name": ..., "parameters": ...}`).

This flattens `tools` entries and a function-typed `tool_choice` into the Responses shape, leaving any already-flat or non-function entries untouched.

Added `test_responses_api_converts_tools_correctly`, mirroring the existing `test_responses_api_converts_images_correctly`, which calls the converter directly and asserts the tool is flattened (`name`/`parameters` present, no nested `function` key) and `tool_choice` is flattened too. It fails on current main and passes with the change.

### DIFF
--- a/dspy/clients/lm.py
+++ b/dspy/clients/lm.py
@@ -46,10 +46,13 @@ def _get_litellm():
 
 def _is_openai_reasoning_model(model: str) -> bool:
     model_family = model.split("/")[-1].lower() if "/" in model else model.lower()
-    return re.match(
-        r"^(?:o[1345](?:-(?:mini|nano|pro))?(?:-\d{4}-\d{2}-\d{2})?|gpt-5(?!-chat)(?:-.*)?)$",
-        model_family,
-    ) is not None
+    return (
+        re.match(
+            r"^(?:o[1345](?:-(?:mini|nano|pro))?(?:-\d{4}-\d{2}-\d{2})?|gpt-5(?!-chat)(?:-.*)?)$",
+            model_family,
+        )
+        is not None
+    )
 
 
 class LM(BaseLM):
@@ -205,12 +208,7 @@ class LM(BaseLM):
         exc_cls = _lm_error_class_from_litellm_exception(exc) or _lm_error_class_from_status(status)
         return exc_cls(message, **metadata)
 
-    def forward(
-        self,
-        prompt: str | None = None,
-        messages: list[dict[str, Any]] | None = None,
-        **kwargs
-    ):
+    def forward(self, prompt: str | None = None, messages: list[dict[str, Any]] | None = None, **kwargs):
         """Call the configured LM synchronously.
 
         LiteLLM/provider exceptions are wrapped in DSPy's structured LM error
@@ -654,6 +652,18 @@ def _convert_chat_request_to_responses_request(request: dict[str, Any]):
         text = request.pop("text", {})
         request["text"] = {**text, "format": response_format}
 
+    # Flatten `tools` / `tool_choice` from Chat Completions shape
+    # (`{"type": "function", "function": {"name": ...}}`) to the flat shape the
+    # Responses API requires (`{"type": "function", "name": ..., "parameters": ...}`).
+    if "tools" in request:
+        request["tools"] = [
+            {"type": "function", **tool["function"]} if tool.get("type") == "function" and "function" in tool else tool
+            for tool in request["tools"]
+        ]
+    tool_choice = request.get("tool_choice")
+    if isinstance(tool_choice, dict) and tool_choice.get("type") == "function" and "function" in tool_choice:
+        request["tool_choice"] = {"type": "function", **tool_choice["function"]}
+
     return request
 
 
@@ -704,9 +714,11 @@ def _add_dspy_identifier_to_headers(headers: dict[str, Any] | None = None):
         **headers,
     }
 
-#--------
+
+# --------
 # Errors
-#--------
+# --------
+
 
 def _safe_litellm_exception_class(name: str) -> type[Exception] | None:
     cls = getattr(_get_litellm(), name, None)
--- a/tests/clients/test_lm.py
+++ b/tests/clients/test_lm.py
@@ -282,7 +282,9 @@ def test_lm_wraps_litellm_errors_with_metadata():
     response.status_code = 429
     response.headers = {"x-request-id": "req-123", "retry-after": "2.5"}
 
-    error = litellm.RateLimitError(message="too many requests", llm_provider="openai", model="gpt-4o", response=response)
+    error = litellm.RateLimitError(
+        message="too many requests", llm_provider="openai", model="gpt-4o", response=response
+    )
     wrapped = lm._wrap_litellm_exception(error)
 
     assert isinstance(wrapped, dspy.LMRateLimitError)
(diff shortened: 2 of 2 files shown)


AnayGarodia:
Closing this: I see #9842 is already the designated fix for this issue and will land before 3.3.0. Didn't mean to add a duplicate; deferring to that one.

PR #9997, opened 2026-07-12
Fix: Evaluate straggler infinite loop and shutdown race condition

sYanXO:
# Fix: ParallelExecutor Infinite Straggler Loop & Shutdown Race Condition

Fixes [[#9574](https://github.com/stanfordnlp/dspy/issues/9574)]

## Problem

When an evaluation task hangs past the 120-second timeout, `ParallelExecutor` flags it as a straggler and submits a duplicate task to the thread pool to try to get it done. If that duplicate *also* takes longer than 120 seconds, it gets flagged as a straggler too, and a third task is spawned. This repeats indefinitely.

The result:

- The thread pool fills up with duplicate tasks.
- Requests stay hung, and no new tasks can be processed.
- If the evaluator eventually tries to shut down (either normally or via `Ctrl+C`), it calls `executor.shutdown(wait=False)`.
- The straggler logic never checked whether the executor was already shutting down before submitting the next retry, so `executor.submit()` throws immediately:

  ```
  ```

This crashes the process instead of shutting down gracefully.

## Root Cause

Two separate issues compound each other:

1. **No cap on straggler retries** — a task that keeps timing out spawns an unbounded number of duplicate submissions.
2. **No shutdown check before `submit()`** — once the executor starts shutting down, any in-flight straggler logic that tries to submit a new task crashes with an unhandled `RuntimeError`.

### 1. Expose the timeout
Added a `timeout` parameter to `dspy.Evaluate.__init__` and `dspy.Evaluate.__call__`, threaded down into `ParallelExecutor`. Users running complex, long-running evaluations can now raise the straggler timeout threshold instead of having it hardcoded, preventing false-positive straggler detection.

### 2. Cap straggler retries
Added a `resubmitted_count` dictionary to `ParallelExecutor` that tracks how many times each item has been resubmitted. Retries are now capped at **exactly 1** per item, eliminating the infinite resubmission loop and the resulting thread pool saturation.

### 3. Guard against the shutdown race condition
Wrapped the straggler `executor.submit()` call in a `try / except RuntimeError` block. If a submit is attempted while the executor is shutting down (or has already shut down), the exception is now caught gracefully and the loop breaks instead of crashing the program.

## Impact

- No more unbounded duplicate task submission during hung evaluations.
- Thread pools no longer saturate with straggler retries.
- Clean, non-crashing shutdown on `Ctrl+C` or normal evaluator exit.
- Users can tune the straggler timeout for legitimately long-running eval tasks.

## Testing

Reproduced the issue locally with a script that simulates hanging evaluation tasks exceeding the straggler timeout, confirmed the original `RuntimeError` on shutdown, then verified the fix resolves it under the same conditions.

> **Note:** Per DSPy's AI contribution guidelines, I paired with an AI coding assistant to investigate the root cause, write the reproduction script, and test this fix locally.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -78,6 +78,7 @@ class Evaluate:
         display_table: bool | int = False,
         max_errors: int | None = None,
         provide_traceback: bool | None = None,
+        timeout: float | None = None,
         failure_score: float = 0.0,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
@@ -106,6 +107,7 @@ class Evaluate:
         self.display_table = display_table
         self.max_errors = max_errors
         self.provide_traceback = provide_traceback
+        self.timeout = timeout
         self.failure_score = failure_score
         self.save_as_csv = save_as_csv
         self.save_as_json = save_as_json
@@ -125,6 +127,7 @@ class Evaluate:
         callback_metadata: dict[str, Any] | None = None,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: float | None = None,
     ) -> EvaluationResult:
         """
         Args:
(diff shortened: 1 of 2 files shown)


isaacbmiller:

PR #10006, opened 2026-07-13
fix(cache): include endpoint in request identity

isaacbmiller:
## 1. Issue / repro

DSPy treats identical prompts sent to different endpoints as the same cached request:

```python
lm("Query", api_base="https://endpoint-a.example/v1")
lm("Query", api_base="https://endpoint-b.example/v1")
```

If both endpoints expose the same model name, the second call can return endpoint A's cached response without contacting endpoint B. The same collision affects `base_url` and hosted embeddings.

`request_cache` hashes the request after removing ignored fields. DSPy explicitly removed both endpoint aliases:

```python
ignored_args_for_cache_key = ["api_key", "api_base", "base_url"]
```

That makes these requests identical from the cache's perspective:

```text
key(model, prompt, endpoint A) == key(model, prompt, endpoint B)
```

An endpoint is a semantic request input: model aliases are not globally unique across custom, local, Azure, or proxy deployments.

The regressions use fake backends whose output is determined solely by the endpoint. Each test verifies:

1. Endpoint A invokes the backend.
2. Endpoint A with a different API key is a cache hit.
3. Endpoint B invokes the backend and returns B's result.

This is covered for `api_base` and `base_url` at all four cache boundaries: generic `request_cache`, `dspy.LM`, synchronous embeddings, and asynchronous embeddings. The eight parameterized cases fail with the old ignore lists and pass with this change.

The production change is four one-line edits. It does not add endpoint normalization, a cache migration, another cache layer, or provider-specific routing. Every boundary follows one rule: ignore credentials, but hash semantic request inputs.

DSPy stores entries under a SHA-256 digest of the transformed request; the raw endpoint is not persisted as the cache key. LM sync/async calls share one cache wrapper, while embeddings have separate sync and async decorators.

Requests without an explicit endpoint keep their existing identity. Requests with an endpoint safely cold-miss older entries.

## 6. What the fix does

```diff
- ignored_args_for_cache_key = ["api_key", "api_base", "base_url"]
+ ignored_args_for_cache_key = ["api_key"]
```

After the change:

```text
key(model, prompt, endpoint A) != key(model, prompt, endpoint B)
```

## Verification

- 8 focused endpoint/credential regression cases passed.
- 38 broader cache/LM/Embedder tests passed.
- Pre-commit and `git diff --check` passed.

## Scope and limitations

- Explicit `api_base` and `base_url` fields are covered. Endpoints resolved only from LiteLLM globals or environment variables are outside this change.
- API keys remain intentionally excluded, preserving credential rotation and the current cache contract. Gateways that route different backends by API key should isolate caches per tenant.
- Equivalent URL spellings can produce harmless cache misses, never incorrect cross-endpoint hits.

### DIFF
--- a/dspy/clients/cache.py
+++ b/dspy/clients/cache.py
@@ -230,7 +230,7 @@ def request_cache(
         enable_memory_cache: Whether to enable in-memory cache at call time. If False, the memory cache will not be
             written to on new data.
     """
-    ignored_args_for_cache_key = ignored_args_for_cache_key or ["api_key", "api_base", "base_url"]
+    ignored_args_for_cache_key = ignored_args_for_cache_key or ["api_key"]
     # Deprecation notice
     if maxsize is not None:
         logger.warning(
--- a/dspy/clients/embedding.py
+++ b/dspy/clients/embedding.py
@@ -165,7 +165,7 @@ def _compute_embeddings(model, batch_inputs, caching=False, **kwargs):
         raise ValueError(f"`model` in `dspy.Embedder` must be a string or a callable, but got {type(model)}.")
 
 
-@request_cache(ignored_args_for_cache_key=["api_key", "api_base", "base_url"])
+@request_cache(ignored_args_for_cache_key=["api_key"])
 def _cached_compute_embeddings(model, batch_inputs, caching=True, **kwargs):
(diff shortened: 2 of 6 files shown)


isaacbmiller:
Actually closing this - Auto opened by one of my agents and we should cache by model; intended behavior

PR #10009, opened 2026-07-13
fix(adapter): preserve nested open mappings

isaacbmiller:
## 1. Issue / repro

`JSONAdapter` detects only a top-level `dict` as incompatible with Structured Outputs. Open mappings nested inside models, arrays, dataclasses, or `TypedDict`s are missed:

```python
class NestedOutput(pydantic.BaseModel):
    metadata: dict[str, Any]

class Signature(dspy.Signature):
    output: NestedOutput = dspy.OutputField()
```

Pydantic correctly emits an open object:

```json
{"type": "object", "additionalProperties": true}
```

DSPy's Structured Outputs normalization then silently rewrites it as an object that permits no keys:

```json
{"type": "object", "properties": {}, "required": [], "additionalProperties": false}
```

The request can succeed, but the response schema now contradicts the declared output type.

The old preflight looks only at each direct Python annotation:

```python
get_origin(annotation) is dict
```

It cannot see mappings represented inside Pydantic `$defs`, array `items`, unions/compositions, dataclasses, `TypedDict`s, nullable objects, or custom schema applicators.

The generated JSON Schema is the actual provider boundary. It must be checked before `enforce_required()` mutates every object into a closed object.

The tests exercise the public adapter call and verify intentional `json_object` mode for:

- `list[dict[str, Any]]`
- nested dataclass, Pydantic, and `TypedDict` fields
- Pydantic `extra="allow"`
- `propertyNames` and `unevaluatedProperties`
- implicit object applicators with no `type`
- nullable schemas with `type: ["object", "null"]`

Negative controls verify that Structured Outputs remains selected for:

- fixed-key `TypedDict`
- a default value containing `{"type": "object"}`
- an example containing `{"type": "object"}`

Those last two prove the scanner follows schema-bearing keywords instead of interpreting arbitrary annotation data as schemas.

The generated schema is the single source of truth instead of accumulating Python-type special cases. The change adds:

- one schema predicate,
- one internal exception for this known incompatibility, and
- symmetric sync/async selection of JSON mode.

There is no new dependency, provider retry, endpoint-specific rule, or parsing change.

Structured Outputs requires object keys to be explicitly declared. Plain `json_object` mode supports arbitrary keys and is already JSONAdapter's established mode for output shapes Structured Outputs cannot express.

The check runs during local response-model construction, before the provider call. `LMError` still propagates; this does not retry a failed provider request in a different mode. Native `ToolCalls` are still excluded before schema generation.

## 6. What the fix does

Before:

```python
if get_origin(annotation) is dict:
    raise ValueError(...)
```

After:

```python
schema = pydantic_model.model_json_schema()

if _schema_has_open_ended_mapping(schema):
    raise _OpenEndedMappingError(...)
```

The adapter catches only that internal signal:

```python
except _OpenEndedMappingError:
    return super().__call__(...)
```

The model therefore makes exactly one request, in the mode that can represent its declared output.

## Verification

```text
uv run --frozen pytest --deno -q tests/adapters/test_json_adapter.py
50 passed

uv run --frozen pytest --deno -q tests/adapters
242 passed
```

Pre-commit and `git diff --check` pass. Independent adversarial schema review: 5/5.

## Scope

Two files. Fixed closed schemas keep the existing Structured Outputs path; provider errors, parsing, type validation, and non-schema-capable LMs are unchanged.

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Any, get_origin
+from typing import Any
 
 import json_repair
 import pydantic
@@ -24,16 +24,48 @@ from dspy.utils.exceptions import AdapterParseError, LMError
 logger = logging.getLogger(__name__)
 
 
-def _has_open_ended_mapping(signature: SignatureMeta) -> bool:
-    """
-    Check whether any output field in the signature has an open-ended mapping type,
-    such as dict[str, Any]. Structured Outputs require explicit properties, so such fields
-    are incompatible.
-    """
-    for field in signature.output_fields.values():
-        annotation = field.annotation
-        if get_origin(annotation) is dict:
+class _OpenEndedMappingError(ValueError):
+    """Raised when a response schema permits arbitrary object keys."""
+
+
+def _schema_has_open_ended_mapping(schema: Any) -> bool:
+    if not isinstance(schema, dict):
+        return False
+
+    schema_type = schema.get("type")
+    has_object_type = schema_type == "object" or (isinstance(schema_type, list) and "object" in schema_type)
+    has_object_keywords = any(
+        keyword in schema
+        for keyword in ("properties", "patternProperties", "additionalProperties", "propertyNames", "unevaluatedProperties")
+    )
+    if has_object_type or (schema_type is None and has_object_keywords):
+        if "patternProperties" in schema:
             return True
+        if "propertyNames" in schema and schema.get("additionalProperties") is not False:
+            return True
+        if schema.get("additionalProperties") not in (None, False):
+            return True
+        if schema.get("unevaluatedProperties") not in (None, False):
+            return True
+        if has_object_type and "properties" not in schema and schema.get("additionalProperties") is not False:
+            return True
+
+    for keyword in ("properties", "$defs", "definitions", "dependentSchemas"):
+        subschemas = schema.get(keyword)
+        if isinstance(subschemas, dict) and any(
+            _schema_has_open_ended_mapping(subschema) for subschema in subschemas.values()
+        ):
+            return True
+
+    for keyword in ("items", "contains", "additionalProperties", "unevaluatedProperties", "not", "if", "then", "else"):
+        if _schema_has_open_ended_mapping(schema.get(keyword)):
+            return True
+
+    for keyword in ("prefixItems", "allOf", "anyOf", "oneOf"):
+        subschemas = schema.get(keyword)
+        if isinstance(subschemas, list) and any(_schema_has_open_ended_mapping(subschema) for subschema in subschemas):
+            return True
+
     return False
 
 
@@ -58,7 +90,7 @@ class JSONAdapter(ChatAdapter):
 
         has_tool_calls = any(field.annotation == ToolCalls for field in signature.output_fields.values())
 
-        if _has_open_ended_mapping(signature) or (not self.use_native_function_calling and has_tool_calls) or not lm.supports_response_schema:
+        if (not self.use_native_function_calling and has_tool_calls) or not lm.supports_response_schema:
             # We found that structured output mode doesn't work well with dspy.ToolCalls as output field.
             # So we fall back to json mode if native function calling is disabled and ToolCalls is present.
             lm_kwargs["response_format"] = {"type": "json_object"}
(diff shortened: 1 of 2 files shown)


percymcn:
I reimplemented `_schema_has_open_ended_mapping` verbatim from this diff and ran it against dspy 3.3.0 / pydantic 2.13.4 before writing any of this, so everything below is measured rather than read.

**The detector works, and it covers two cases the description doesn't claim.** Running it over the raw `model_json_schema()` for a batch of output types, it correctly returns `True` for the nested/`list[...]` cases in the body *and* for `Optional[dict[str, str]]` — which the old preflight missed for a second reason worth writing down: `get_origin(Optional[dict[str, str]])` is `Union`, not `dict`, so a *top-level* optional mapping walked past both the call-site guard and the "final guard" too. Today that one reaches the provider as a live open map and 400s; with this PR it falls back cleanly. Moving the check onto the generated schema is what makes both fall out of one predicate.

**What it leaves, and it is a different class rather than a gap in the walk.** `enforce_required` is not the only thing standing between a pydantic model and a strict-valid schema, and the other blockers are not mappings — so the detector correctly returns `False` and structured outputs still gets used. Verdicts are from `toStrictJsonSchema()` in `openai@7.4.0` (the JS SDK's strict transformer, i.e. the same subset the API enforces), run on the exact bytes dspy emits:

| output field | emits | `_schema_has_open_ended_mapping` | outcome after this PR |
|---|---|---|---|
| `dict[str, Any]` nested / in a list / under `Optional` | open map | `True` | ✅ json mode |
| `tuple[int, int, int, int]` | `prefixItems` | `False` | ❌ `unsupported keyword prefixItems` |
| `tuple[int, int]` inside a model (`$defs`) | `prefixItems` | `False` | ❌ same, at `$defs/…/properties/box` |
| `list[tuple[int, int]]` | `prefixItems` under `items` | `False` | ❌ same |
| `set[str]` | `uniqueItems` | `False` | ❌ `unsupported keyword uniqueItems` |

**Why this matters more after the PR than before it.** Before, both classes were broken-ish and the user got *something*. After, the two diverge: an open mapping is detected and degrades gracefully to json mode, while a `tuple` or `set` field goes to the provider and comes back a 400 — and that error is a provider failure, so it hits `except LMError: raise` rather than the json-mode fallback. Same adapter, same "this schema can't be expressed in strict mode" root cause, one path graceful and one path a hard error with no fallback.

**Two test-design notes, both measured, because I think they'd bite a reviewer here.**

1. Every fixture added in this PR is a `dict[str, Any]` shape (`_DataclassOutput`, `_PydanticOutput`, `_TypedDictOutput`, the `Annotated` map variants). All of them pass with the tuple/set gap wide open, so the suite can't distinguish "the preflight is complete" from "the preflight covers mappings." A discriminating fixture is one line: an output model with `box: tuple[int, int]`.
2. If you reach for the vendor's own Python SDK to assert compatibility, it will not help. `openai.lib._pydantic._ensure_strict_json_schema` (openai 2.53.0) **accepts all five schemas above verbatim** — including the three the API rejects — so a Python-side assertion built on it passes while the bug ships. The JS transformer is the one that actually throws; if you want this checkable in CI without an API key, that's the oracle.

**Shape of the remaining fix.** The architecture here is already right — walk the schema, raise, let the existing fallback fire — so it's one more predicate on the same walk. The walk already *visits* `prefixItems` (it's in the list of applicators it descends into, looking for maps underneath); it just never asks about the keyword itself. But I'd separate the two, because falling back is a heavier remedy than a tuple needs: a *homogeneous* `prefixItems` is losslessly expressible in strict mode as `items` + `minItems`/`maxItems` (both supported), so rewriting it keeps structured outputs *and* the fixed length. I verified the rewritten form is accepted by `toStrictJsonSchema()` for all four tuple rows above. A *heterogeneous* tuple has no lossless form and should raise rather than be widened to `items: {anyOf: [...]}`, which would let position 0 start accepting a string. `uniqueItems` has no strict form at all, so raising (and falling back) is the right answer there.

**One scoping question rather than a finding.** `_ExtraAllowedOutput` (`ConfigDict(extra="allow")`) is detected as an open mapping, so it now drops to json mode. Today it goes through structured outputs and is accepted — `enforce_required` closes it, extras are silently dropped, but the declared fields stay grammar-enforced. Both readings are defensible; it's just worth being deliberate about, since `extra="allow"` is common and this is a behaviour change for it rather than a fix.

PR #10010, opened 2026-07-13
fix(adapters): coerce non-string Literal members in parse_value

spjosyula:
---
parse_value rejects Literal[int] / Literal[bool] output fields coming from a text adapter (ChatAdapter, XMLAdapter). The value arrives as a string, so Literal[1, 2, 3] gets "2", which the Literal branch checks for membership and rejects before it reaches the coercion path (json_repair -> TypeAdapter) below. String literals work, which hid it.

parse_value("2", Literal[1, 2, 3])   # ValueError: '2' is not one of (1, 2, 3)

With XMLAdapter (or ChatAdapter without JSON fallback) it's a hard AdapterParseError for that field — e.g. rating scales like Literal[1, 2, 3, 4, 5]. Regression from #8143.

Fix: let non-matching literals fall through to the same coercion every other type uses; non-members still fail via pydantic. Added a test.

Used Claude Code to help find this: I reproduced it, traced it to #8143, and verified the fix myself.

---

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -170,7 +170,8 @@ def parse_value(value, annotation):
             if v in allowed:
                 return v
 
-        raise ValueError(f"{value!r} is not one of {allowed!r}")
+            # Coerce string forms of non-string members (e.g. "2" for Literal[1, 2, 3]) below.
+            value = v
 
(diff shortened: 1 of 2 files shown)


isaacbmiller:
lgtm - thanks for the PR

Issue #658, opened 2024-03-15
Shadowing signature fields causes weird errors


thomasnormal:
I was just defining a signature like this:
```python
class Signature(dspy.Signature):
    __doc__ = textwrap.dedent(
        """I gave my language model a task, but it failed. Figure out what went wrong,
        and write instructions to help it avoid the error next time.""",
    )

    task_description: str = dspy.InputField(desc="What I asked the model to do")
    model_output: str = dspy.InputField(desc="The output of the model")
    error: str = dspy.InputField(desc="The validation error trigged by the models output")
    explanation: str = dspy.OutputField(desc="Explain what the model did wrong")
    instructions: str = dspy.OutputField(desc="Instructions for the model to do better next time")
```
but this fails with 
```
    raise TypeError(
TypeError: Field 'instructions' in 'Signature' must be declared with InputField or OutputField. field.json_schema_extra=None
```
Why? Because `instructions` is a "reserved word" of `dspy.Signature`.

There are a number of other reserved words, like `fields`, `insert`, `prepend`, `append`, `equals`.
At least a few of those have already caused me trouble in the past.
On top of that there are "pydantic reserved words", like `schema` that I've also had issues with.

Pydantic 2.x is trying to clean this up, by moving all their special words to the prefix `model_`, such as `model_schema` instead of `schema`. Just now I made a field with name `model_output` and got this warning from pydantic: `"...pydantic/_internal/_fields.py:149: UserWarning: Field "model_output" has conflict with protected namespace "model_"`

I think we should similarly move all the dspy related words to some prefix, like `dspy_instructions`, `dspy_equals` etc.
This causes only a small inconvenience for dspy backend developers (e.g. you are making a new optimizer), but make life a lot easier for end users (that is, people just defining a casual signature.)

On top of that, we should make sure to throw informative errors in the cases where there's still a clash. Like if the user uses `model_schema` or `dspy_instructions` as a field name for some reason.

thomasnormal:
An alternative idea is to move everything out of the `Signature` namespace and into the module.
Then you would do `dspy.signature.instructions(sig)` instead of `sig.dspy_instructions`.

thomasnormal:
A related issue is "reserved keywords" for `dspy.Predict`.
The forward method assigns special meaning to words like `new_signature`, `signature`, `demos`, `config`, `lm`, and `_trace`.
So if a user uses any of those words as names in a Signature, they'll probably also run into trouble.

[...]

PR #10011, opened 2026-07-13
fix(signatures): raise clear error when fields shadow reserved DSPy attributes

TLXyloph:
## What

Declaring a signature field named `instructions` currently fails with a message that blames the user for something they didn't do:

```
TypeError: Field `instructions` in `Signature` must be declared with InputField or OutputField,
but field `instructions` has `field.json_schema_extra=None`
```

— even though the field *was* declared with `OutputField`. Worse, an input field named `demos` produces no error at all: its value is silently popped by `Predict._forward_preprocess` and never reaches the LM.

This PR makes both cases fail fast with an error that names the actual collision.

## Why the old error was misleading

`instructions`, `fields`, `input_fields`, `output_fields` and `signature` are **properties on `SignatureMeta`**. A property is a data descriptor, so it takes precedence over the class `__dict__` during attribute lookup — Pydantic's field collection does `getattr(cls, name)` and gets the property object back instead of the `FieldInfo`, hence `json_schema_extra=None`. The field could never have been read back from the class.

## Approach

- `signature.py` derives the reserved set from the metaclass at runtime — any non-dunder **data descriptor** on `SignatureMeta`. This is the mechanically correct criterion: it is exactly the set of names a field cannot survive, and it stays correct if a property is added later. It also explains why plain methods like `append`/`delete` keep working (non-data descriptors, shadowed by the class dict) — pinned by a test.
- `predict.py` rejects *input* fields colliding with `Predict`'s privileged kwargs (`signature`, `new_signature`, `demos`, `config`, `lm`), which would otherwise be silently swallowed.
- Docs updated to list the reserved names.

## Scope

This implements the minimal fix the thread converged on: *"we should at least make sure we document those reserved names well, and throw errors when the user unknowingly tries to use them."* It deliberately does **not** do the `dspy_`-prefix namespace migration discussed higher up — that is a breaking API change and a maintainer call.

### Known gap: `prediction`

`_forward_preprocess` (`predict.py:203`) also pops `prediction` into `config`, but **only** when its value is a `{"type": "content", "content": ...}` dict (the OpenAI predicted-outputs format). So an input field named `prediction` carrying that shape can still be silently swallowed — the exact failure mode this PR sets out to kill. I left it out of `RESERVED_PREDICT_KWARGS` because the pop is *conditional*: a `prediction: str` field passes through fine today, and unconditionally reserving the name would reject signatures that currently work.

**Proposed fix, if you want it in this PR:** the resolved `signature` is already in scope at that point, so the pop can simply defer to an explicit field declaration:

```python
if "prediction" in kwargs and "prediction" not in signature.input_fields:
    if (
        isinstance(kwargs["prediction"], dict)
        and kwargs["prediction"].get("type") == "content"
        and "content" in kwargs["prediction"]
    ):
        config["prediction"] = kwargs.pop("prediction")
```

This is backward compatible: signatures that do not declare a `prediction` field (i.e. everyone using predicted outputs today) behave identically, and `prediction: str` keeps working. It only changes the case that is currently broken — a declared `prediction` input field whose value happens to be dict-shaped now reaches the adapter instead of vanishing.

The tradeoff worth your call: it means a signature **cannot** both declare a `prediction` input field and use the predicted-outputs LM feature. Resolving in favor of the explicit signature declaration seems right (a declared field is more specific than an optional LM knob), but that is a product decision, so I have left it out pending your preference. Say the word and I will add it.

## Testing

Red→green verified against base: the reserved-name cases fail on `main` and pass with this change. `tests/signatures/` + `tests/predict/` → 214 passed, 2 xfailed, no regressions.

Addresses #658

### DIFF
--- a/docs/docs/diving-deeper/signatures-in-depth.md
+++ b/docs/docs/diving-deeper/signatures-in-depth.md
@@ -62,6 +62,14 @@ GEPA and the other instruction optimizers rewrite the Signature’s docstring. T
 
 Field names are part of the program’s public interface: caller code reads `result.haiku`, downstream modules look up `"haiku"` by name. An optimizer that renamed them would break the program around it. So name your fields carefully — the optimizer can’t fix `result` into `haiku` later. The same goes for field descriptions when you write them.
 
+### 10. A few field names are reserved
+
+`instructions`, `fields`, `input_fields`, `output_fields` and `signature` are properties on `SignatureMeta`, and a property is a data descriptor: it wins over the class `__dict__` during attribute lookup. So a field with one of those names could be declared but never read back — `MySignature.instructions` would return the docstring, not your field. Rather than let that fail confusingly downstream, `SignatureMeta.__new__` raises a `TypeError` naming the colliding field, for both the class form and the string form (`dspy.Signature("question -> instructions")`).
+
+`dspy.Predict` reserves a second, overlapping set: `lm`, `demos`, `config`, `signature` and `new_signature` are the privileged keyword arguments it pops off in `_forward_preprocess`, so an *input* field with one of those names would have its value swallowed instead of reaching the LM. `Predict` raises a `ValueError` when it is constructed with such a signature, and again if the signature is swapped out later (by an optimizer, or via a per-call `signature=` override). Output fields are unaffected — nothing pops them.
+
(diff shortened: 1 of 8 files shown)


TLXyloph:
Hi @isaacbmiller — flagging this in case it's useful now that your RLM namespace PR (#10020) has landed. 

Please do correct me if I'm wrong, but I thought:

PR #10020  hardened the RLM side, and that this PR covers the Signature/Predict reserved-name side of  Issue #658, which is still unaddressed on main: an instructions OutputField fails with the misleading json_schema_extra=None TypeError, and a demos input field is silently consumed by _forward_preprocess before it reaches the LM. I verified both against the current main.

Happy to rework the PR to match the style of #10020 if you'd prefer, or close it.

PR #10013, opened 2026-07-13
fix(docstring): Fix max_iters default value in docstring of ReAct

ellacroix:
Small fix in the docstring of ReAct to update the default value of max_iters from 10 to 20, following the update in the __init__() function in PR #9162.


isaacbmiller:
LGTM. Sneaky catch. Thanks for the PR!

PR #9801, opened 2026-05-22
fix(rlm): protect history from nested fences

he-yufeng:
> **Stack:** one contributor-authored commit on top of #10019. That parent owns exact source-character truncation; this PR owns only how the retained text is delimited in the RLM prompt.

Fixes #9573.

## 1. Issue / repro

RLM copies generated code, interpreter output, and input previews into later prompts using fixed triple-backtick fences. Those values can themselves contain Markdown code blocks:

```python
entry = REPLEntry(
    code='print("""```python\nx = 1\n```""")',
    output="Code:\n```python\nprint(1)\n```\nOutput: ok",
)
```

The inner ` ``` ` sequence can terminate the outer code block early. In #9573, a prior REPL output containing fenced code corrupted the next action prompt and the adapter could no longer parse the model's `reasoning`/`code` fields.

## 2. Why this is the root cause

The formatter currently assumes its payload never contains its delimiter:

```python
code_block = f"```python\n{self.code}\n```"
lines.append(f"Preview:\n```\n{self.preview}\n```")
```

`code`, `output`, and `preview` are derived from user/model/interpreter data, so that assumption is false. The adapter failure is downstream: once the transcript's Markdown structure is ambiguous, the next LM response can stop following the action signature.

## 3. How we know the fix addresses the root cause

The regression reproduces the reported connection directly: code and output contain their own ` ```python ` blocks, then the formatted history is inspected to prove the outer delimiter is four backticks and the inner block is unchanged.

A second exact regression combines truncation with a retained four-backtick run:

```text
raw output: ````abcdef````
limit:      9 source characters
outer fence: ````` (five backticks)
```

That proves the formatter chooses the delimiter from the text that actually reaches the prompt, after #10019's truncation step.

## 4. Why this is the concise fix

One helper finds the longest consecutive backtick run, and one helper chooses a fence exactly one character longer (with three as the Markdown minimum):

```python
fence = "`" * max(3, _longest_backtick_run(text) + 1)
return f"{fence}{language}\n{text}\n{fence}"
```

The same helper replaces the three delimiter-owning call sites. There is no escaping table, parser, retry, provider branch, or content mutation. The payload remains byte-for-byte intact inside an unambiguous fence.

## 5. Context needed to validate the change

Markdown permits a fenced block to use more than three backticks. A closing fence must be at least as long as the opening fence, so choosing `longest inner run + 1` guarantees that no retained payload sequence can close the outer block.

Truncation and fencing are separate contracts:

[...]

PR #10019, opened 2026-07-13
fix(rlm): honor truncation limits

isaacbmiller:
> **Review guide:** one commit, three files. The behavior change is two arithmetic corrections at the existing truncation sites plus explicit guards for the previously undefined non-positive limit case.

## 1. Issue / repro

Both RLM truncation paths promise a source-character limit but violate it when the limit is odd.

```python
REPLVariable.from_value("value", "abcdef", preview_chars=1).preview
REPLEntry.format_output("abcdef", max_output_chars=1)
```

On `main`, both retain all six source characters:

```text
...abcdef

Output (6 chars):

... (5 characters omitted) ...

abcdef
```

The formatted output is internally contradictory: it says five characters were omitted while displaying all six.

Non-positive limits expose the same undefined boundary. `0` currently leaks the complete value because Python treats `-0` as `0`; negative limits duplicate overlapping slices. An RLM “maximum” that retains more source text than requested is not enforcing a limit.

Both implementations floor-divide the limit and reuse that count for both slices:

```python
half = limit // 2
head = value[:half]
tail = value[-half:]
```

For a positive odd limit, this retains `limit - 1` source characters. At limit `1`, `half == 0`, and `value[-0:]` is the complete string. That exact slicing edge explains the full-value leak.

For `limit <= 0`, there is no coherent head/tail split. The public contract is therefore made explicit: truncation limits must be positive.

Regressions cover both truncation sites and both boundaries:

- limit `1`: retain exactly the final character, never the complete value;
- limit `5`: retain two head characters and three tail characters;
- limits `0` and `-1`: raise a precise `ValueError`;
- `RLM(max_output_chars=...)`: rejects the invalid configuration at construction, before an interpreter or LM call starts.

The positive-limit assertions count the retained source characters directly, so the test cannot pass by merely changing a marker or omission message.

The positive configured limit is divided once and its remainder goes to the tail:

```python
head_chars = limit // 2
tail_chars = limit - head_chars
```

Those counts always sum to the configured limit. The invalid boundary is one explicit guard at each callable API:

```python
if limit <= 0:
    raise ValueError("... must be greater than 0")
```

There is no helper abstraction, special-case slice, alternate marker, fallback output, or prompt rewrite.

The limit counts retained source characters; the existing `...` marker or omission message is additional formatting. RLM deliberately keeps both the beginning and end of long values. Assigning the odd remainder to the tail retains one extra trailing character, where final results and exception details commonly appear.

`max_output_chars` is constructor configuration, so invalid values fail there immediately. The primitive methods retain their own guards because they are public/tested formatting boundaries and can be called independently of RLM.

- Makes the retained source-character count equal every positive odd limit.
- Rejects zero and negative preview/output limits explicitly.
- Rejects invalid `RLM(max_output_chars=...)` configuration before execution.
- Adds exact limit-`1`, limit-`5`, zero, and negative regressions for both truncation paths.

- **Even positive limits are byte-for-byte unchanged**, including the defaults (`1,000` and `10,000`).
- **Positive odd limits intentionally retain one additional trailing character.** Snapshots for custom odd limits will change.
- **Limit `1` stops exposing the full value/output.** This is the security/correctness edge fixed by the PR.
- **Zero and negative limits now raise `ValueError`.** They previously returned nonsensical, over-limit output; callers using them must choose a positive bound.
- **The truncation marker is not included in the source-character budget.** That existing convention is unchanged.
- **No provider, adapter, interpreter lifecycle, tool, error-recovery, or fallback behavior changes.**
- **Adjacent open work:** #9801 edits `REPLEntry.format_output` for nested Markdown fences. The concerns are independent, but whichever lands second will need a small semantic rebase.

## Validation

- `uv run --frozen pytest --deno -q tests/predict/test_rlm.py` — `131 passed, 2 skipped`
- repository pre-commit hooks on all three changed files — passed
- `git diff --check` — passed
- branch is one commit over current `main`

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -164,6 +164,8 @@ class RLM(Module):
         """
         super().__init__()
         _validate_interpreter_factory(interpreter_factory)
+        if max_output_chars <= 0:
+            raise ValueError("max_output_chars must be greater than 0")
         self.signature = ensure_signature(signature)
         self.max_iters = max_iters
         self.max_llm_calls = max_llm_calls
--- a/dspy/primitives/repl_types.py
+++ b/dspy/primitives/repl_types.py
@@ -51,6 +51,9 @@ class REPLVariable(pydantic.BaseModel):
             field_info: Optional pydantic FieldInfo with desc/constraints metadata
             preview_chars: Max characters for preview
         """
+        if preview_chars <= 0:
+            raise ValueError("preview_chars must be greater than 0")
+
         jsonable = serialize_for_json(value)
         if isinstance(jsonable, (dict, list)):
             value_str = json.dumps(jsonable, indent=2)
@@ -58,8 +61,9 @@ class REPLVariable(pydantic.BaseModel):
             value_str = str(jsonable)
         is_truncated = len(value_str) > preview_chars
         if is_truncated:
-            half = preview_chars // 2
-            preview = value_str[:half] + "..." + value_str[-half:]
+            head_chars = preview_chars // 2
+            tail_chars = preview_chars - head_chars
+            preview = value_str[:head_chars] + "..." + value_str[-tail_chars:]
         else:
             preview = value_str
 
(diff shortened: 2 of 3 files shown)


isaacbmiller:
Superseded by #10340

PR #10020, opened 2026-07-13
fix(rlm): protect RLM-owned namespaces

isaacbmiller:
> **Scope:** one commit, two files. This PR owns only RLM namespace validation. Signature-default binding has been restored to a separate child PR.

## 1. Issue / repro

RLM combines signature inputs, user tools, built-in helpers, and returned metadata, but currently accepts names that it will later overwrite or silently discard.

```python
# The second tool silently replaces the first during dict construction.
RLM("context -> answer", tools=[
    Tool(first, name="lookup"),
    Tool(second, name="lookup"),
])

# Undeclared kwargs enter the sandbox and can shadow owned bindings.
rlm(context="x", lookup="not the lookup tool")
rlm(context="x", SUBMIT="not the submit helper")

# RLM later tries to add its own trajectory field to the Prediction.
RLM("context -> trajectory")
```

Call-time `tools=` has the same problem: it is currently treated as an ordinary undeclared input, not as an explicit tool-registration API.

Names are validated independently and after information has already been lost:

```python
# Duplicate names are destroyed here before validation can observe them.
return {tool.name: tool for tool in tool_list}

# Only missing inputs are checked; extra keys flow into the interpreter.
missing = set(self.signature.input_fields) - set(input_args)
```

These are not interpreter-precedence bugs. RLM creates the combined sandbox and result namespaces, so it must reject configurations with two meanings for one name before starting an interpreter.

## 3. How we know the fix targets the root cause

Tests exercise every ownership boundary directly:

- duplicate, keyword, and reserved tool names fail during construction;
- input/helper and input/tool collisions fail during construction;
- reserved result names fail during construction;
- undeclared sync and async call inputs fail before an interpreter is created;
- valid tools and declared inputs continue through the existing execution path.

The failure timing is part of the contract: invalid constructor configuration fails in `__init__`, while invalid call inputs fail before interpreter creation.

Tool normalization becomes an explicit loop so duplicates remain observable. One `_validate_namespace()` compares the sets owned at construction. `_validate_inputs()` adds one unexpected-key check at the call boundary.

There is no renaming scheme, precedence table, alias system, fallback path, or sandbox-specific repair. Ambiguous programs fail where the ambiguity is introduced.

| Boundary | Names that must remain unique |
|---|---|
| Sandbox bindings | declared inputs, user tools, `llm_query`, `llm_query_batched`, `SUBMIT`, `print` |
| Returned `Prediction` | declared outputs, `trajectory`, `final_reasoning` |

Tool names and output names do **not** share a binding namespace, so this remains valid:

```python
result = lookup(...)
SUBMIT(lookup=result)
```

The tool is a callable sandbox binding; the same-named output is only a keyword accepted by `SUBMIT`.

## 6. What the fix does

- Preserves duplicate tool names long enough to reject them.
- Rejects Python keywords and RLM-reserved tool names.
- Rejects declared input collisions with helpers or tools.
- Rejects undeclared sync and async call inputs before interpreter startup.
- Reserves `trajectory` and `final_reasoning` for RLM result metadata.

- **Undeclared kwargs now raise `ValueError`.** Callers using them as an undocumented REPL-variable injection path must declare them in the signature.
- **Call-time `tools=` is not tool registration.** It is rejected when undeclared; tools remain constructor configuration.
- **Duplicate tools no longer use last-write-wins.** Give them distinct names.
- **Conflicting signatures fail earlier.** Inputs that reuse helper/tool names and outputs named `trajectory` or `final_reasoning` must be renamed.
- **Tool/output overlap remains valid.** Outputs are `SUBMIT` keywords, not sandbox bindings.
- **No signature-default, provider, adapter, tool-execution, interpreter-lifecycle, output-coercion, or serialization behavior changes.**

## Validation

- `uv run --frozen pytest --deno -q tests/predict/test_rlm.py` — `141 passed, 2 skipped`
- repository pre-commit hooks on both changed files — passed
- `git diff --check` — passed
- one commit over current `main`

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -14,6 +14,7 @@ import base64
 import contextvars
 import functools
 import inspect
+import keyword
 import logging
 import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -172,7 +173,7 @@ class RLM(Module):
         self.sub_lm = sub_lm
         self._interpreter_factory = interpreter_factory
         self._user_tools = self._normalize_tools(tools)
-        self._validate_tools(self._user_tools)
+        self._validate_namespace(self._user_tools)
 
         # Build the action and extract signatures
         action_sig, extract_sig = self._build_signatures()
@@ -183,8 +184,9 @@ class RLM(Module):
     # Tool Creation and Validation
     # =========================================================================
 
-    # Reserved tool names that conflict with built-in sandbox functions
-    _RESERVED_TOOL_NAMES = frozenset({"llm_query", "llm_query_batched", "SUBMIT", "print"})
+    # Names owned by RLM rather than the user-provided signature or tools.
+    _RESERVED_SANDBOX_NAMES = frozenset({"llm_query", "llm_query_batched", "SUBMIT", "print"})
+    _RESERVED_RESULT_NAMES = frozenset({"trajectory", "final_reasoning"})
 
     def _normalize_tools(self, tools: list[Callable] | None) -> dict[str, Tool]:
         """Normalize tools list to a dict of Tool objects keyed by name."""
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Checked the 4/5 concern about tool/output overlap. No code change is needed: tool names are Python bindings in the sandbox, while output names are only keyword names accepted by SUBMIT. This is valid and unambiguous: result = lookup(...); SUBMIT(lookup=result). Rejecting it would block a coherent API use without preventing a runtime collision. I added this boundary explicitly to the compatibility section.

PR #10023, opened 2026-07-13
fix(rlm): normalize sub-LM responses

isaacbmiller:
> **Review guide:** one commit, two production decisions in the same `llm_query` boundary:
>
> 1. normalize DSPy's typed and legacy LM return contracts to text;
> 2. isolate only `dspy.LMError` failures in batched calls, so programming/contract errors stay visible.

## 1. Issue / repro

RLM's `llm_query()` tool changes meaning when DSPy's typed LM mode is active.

```python
lm = DummyLM([{"answer": "typed answer"}])
tools = RLM("context -> answer", sub_lm=lm)._make_llm_tools()

with dspy.context(experimental=True):
    print(tools["llm_query"]("test prompt"))
```

The same LM returns answer text in default mode. On `main`, typed mode instead returns the Pydantic representation of the entire response:

```text
model='dummy' outputs=[LMOutput(parts=[LMTextPart(... text='[[ ## answer ## ]]\ntyped answer')], ...)] ...
```

There is a second surprise at this boundary: `llm_query_batched()` converts every worker exception—including `TypeError` from a broken LM implementation—into an `[ERROR]` string.

`BaseLM` has two public return shapes during the typed migration:

- default calls return a non-empty legacy list of text outputs;
- explicit/experimental typed calls return `LMResponse`.

RLM handles only the first, then stringifies everything else:

```python
response = target_lm(prompt)
if isinstance(response, list) and response:
    ...
return str(response)
```

An `LMResponse` therefore reaches `str(response)` instead of `.text`. Separately, the batch path catches `Exception`, although DSPy already defines `LMError` as the provider/configuration/transport failure boundary.

The regression uses the real `DummyLM`/`BaseLM` path in both modes:

```python
assert legacy_tool("test prompt") == "[[ ## answer ## ]]\nlegacy answer"

with dspy.context(experimental=True):
    assert typed_tool("test prompt") == "[[ ## answer ## ]]\ntyped answer"
```

The typed assertion fails on `main` because the result begins with `model='dummy'`; it passes here because RLM reads `LMResponse.text`.

Batch tests prove the intended classification:

```python
LMTransportError(...)  -> "[ERROR] ..."  # one failed slot
LMNotConfiguredError   -> "[ERROR] ..."  # one failed slot
TypeError(...)         -> raises          # broken implementation/contract
```

A textless legacy response also proves that the diagnostic names the received type (`NoneType`) instead of failing with a generic message.

The scalar tool explicitly handles the two DSPy contracts and rejects anything else:

```python
if isinstance(response, dspy.LMResponse):
    text = response.text
elif isinstance(response, list) and response:
    first_output = response[0]
    text = first_output.get("text") if isinstance(first_output, dict) else first_output
else:
    raise TypeError(...)

if not isinstance(text, str):
    raise TypeError(f"... got {type(text).__name__}.")
```

The batch change is one type substitution: `except Exception` becomes `except dspy.LMError`. There is no attribute probing, provider-specific parsing, retry, or stringification fallback.

`llm_query()` is a text-returning semantic helper exposed inside the interpreter. Returning response metadata changes the generated program's input even when the underlying answer is identical.

Per-slot batch errors are still strategically useful: one expected provider failure should not discard successful siblings. DSPy's exception taxonomy makes that policy precise. Programming and response-contract errors mean the helper itself is broken and should remain visible.

- Returns `LMResponse.text` for typed DSPy calls.
- Preserves text extraction from the established legacy list shape.
- Requires either shape to contain a string.
- Reports the received type for a textless response.
- Classifies missing LM setup as `LMNotConfiguredError`.
- Keeps per-slot batch markers for `LMError` failures.
- Lets unexpected implementation errors escape the batch.

- **Raw callable return values are no longer stringified.** `sub_lm` is declared as a DSPy LM; custom implementations must return `LMResponse` or the documented legacy list shape.
- **Empty legacy lists and responses without string text now raise `TypeError`.** Actual empty-string text remains valid. Tool-call-only responses cannot satisfy a text-returning helper.
- **Unexpected batch exceptions now fail the batch.** This is intentionally less tolerant so programming and contract errors cannot masquerade as provider failures.
- **Expected LM failures remain isolated.** Configuration, transport, provider, rate-limit, and other `LMError` subclasses still become one `[ERROR]` entry.
- **Missing LM setup now raises `LMNotConfiguredError` instead of plain `RuntimeError`.** Scalar callers catching only that exact error type must update; batch behavior remains an `[ERROR]` result.
- **No provider-specific response parsing is added.** Both accepted branches are DSPy-level contracts.
- **Call limits, ordering, copied request context, threading, and valid scalar behavior are unchanged.**

## Validation

- `uv run --frozen pytest --deno -q tests/predict/test_rlm.py` — `127 passed, 2 skipped`
- focused legacy, typed, unsupported-shape, textless-shape, missing-configuration, LM-error, unexpected-error, and request-context tests — passed
- repository pre-commit hooks on both changed files — passed
- `git diff --check` — passed
- branch contains one commit over current `main`

### DIFF
--- a/dspy/predict/rlm.py
+++ b/dspy/predict/rlm.py
@@ -256,14 +256,24 @@ class RLM(Module):
         def _query_lm(prompt: str) -> str:
             target_lm = lm if lm is not None else dspy.settings.lm
             if target_lm is None:
-                raise RuntimeError("No LM configured. Use dspy.configure(lm=...) or pass sub_lm to RLM.")
+                raise dspy.LMNotConfiguredError(
+                    "No LM configured. Use dspy.configure(lm=...) or pass sub_lm to RLM."
+                )
             response = target_lm(prompt)
-            if isinstance(response, list) and response:
-                item = response[0]
-                if isinstance(item, dict) and "text" in item:
-                    return item["text"]
-                return item
-            return str(response)
+            if isinstance(response, dspy.LMResponse):
+                text = response.text
+            elif isinstance(response, list) and response:
+                first_output = response[0]
+                text = first_output.get("text") if isinstance(first_output, dict) else first_output
+            else:
+                raise TypeError(
+                    "Sub-LM must return dspy.LMResponse or a non-empty list of text outputs, "
+                    f"got {type(response).__name__}."
+                )
+
+            if not isinstance(text, str):
+                raise TypeError(f"Sub-LM response must contain text, got {type(text).__name__}.")
+            return text
 
         def llm_query(prompt: str) -> str:
             """Query the LLM with a prompt string."""
@@ -273,7 +283,7 @@ class RLM(Module):
             return _query_lm(prompt)
 
         def llm_query_batched(prompts: list[str]) -> list[str]:
-            """Query the LLM with multiple prompts concurrently."""
+            """Query prompts concurrently, isolating LM failures while propagating contract errors."""
             if not prompts:
                 return []
             _check_and_increment(len(prompts))
@@ -288,7 +298,7 @@ class RLM(Module):
                     idx = future_to_idx[future]
                     try:
                         results[idx] = future.result()
-                    except Exception as e:
+                    except dspy.LMError as e:
                         results[idx] = f"[ERROR] {e}"
             return [results[i] for i in range(len(prompts))]
 
--- a/tests/predict/test_rlm.py
+++ b/tests/predict/test_rlm.py
@@ -240,12 +240,51 @@ class TestRLMInitialization:
         with pytest.raises(TypeError, match="first positional argument"):
             rlm(query="test", interpreter=MockInterpreter())
 
+    def test_llm_query_returns_legacy_response_text(self):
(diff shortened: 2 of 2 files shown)


isaacbmiller:
Addressed the 4/5 boundary note in `eafab0b28`.

The intentional policy is now stated on `llm_query_batched()` itself: LM failures are isolated per prompt, while response-contract and programming errors propagate.

I also fixed the concrete configuration edge the review identified:

```python
if target_lm is None:
    raise dspy.LMNotConfiguredError(...)

...
except dspy.LMError as error:
    results[index] = f"[ERROR] {error}"
```

So a missing LM remains a per-slot `[ERROR]`, while raw exceptions from an implementation that does not use DSPy's LM error taxonomy remain visible intentionally. A regression pins the missing-configuration behavior, and the full RLM file now passes `115 passed, 2 skipped`.

PR #10026, opened 2026-07-14
fix(lm): send Responses-native tool and tool_choice shapes

isaacbmiller:
## Summary

Before, `model_type="responses"` reused several Chat Completions shapes and could reject, reorder, or corrupt otherwise valid Responses requests and replay history; after, DSPy emits the Responses wire dialect while preserving the existing public LM output contract and a deliberately narrow legacy-compatibility door.

This is primarily the request-side extraction from #9842. It also includes two narrow integration fixes required by the same flow: tool-only adapter outputs and cached strict provider response models.

### Change labels

- **BUG → FIX** — externally incorrect, rejected, lossy, or corrupting behavior that this PR corrects.
- **REFACTOR** — implementation moved through the shared typed request mapper; no separate public behavior change is intended.
- **COMPATIBILITY** — deliberate tolerance at the legacy `dspy.LM` boundary.
- **UNCHANGED** — an existing contract this PR preserves.
- **BOUNDARY** — where permissive legacy behavior stops and strict/provider validation begins.
- **OUT OF SCOPE** — related work intentionally not included in this PR.

### Request serialization

| Classification | Behavior | Before | Now |
| --- | --- | --- | --- |
| **REFACTOR** | Shared Responses mapper | The legacy LM door maintained a second, partial Chat-to-Responses converter. | The legacy door sanitizes compatibility inputs, builds `LMRequest`, and delegates wire emission to `to_openai_responses_request`; the rows below are the intended behavior changes. |
| **BUG → FIX** | Function tools | Responses requests emitted the Chat shape, with fields nested under `function`; OpenAI expects flat function-tool fields. | Chat-shaped and flat function tools normalize to flat Responses objects: `type`, `name`, `parameters`, optional `description`, and optional `strict`. |
| **BUG → FIX** | Forced function choice | Chat-shaped forced choices were sent as `{type, function: {name}}`, which is not the Responses shape. | They emit `{type: "function", name}`. Responses-native choices such as flat function, hosted-tool, and `allowed_tools` objects pass through unchanged. |
| **BUG → FIX** | `strict` and provider tool fields | `strict` could be lost, flat tool extras could fail typed validation, and provider echoes could overwrite canonical `type`, `name`, `parameters`, or `strict`. Chat function-scoped extras were also emitted at the wrong level. | `strict` is part of `LMToolSpec`; non-canonical provider fields are preserved at the dialect-correct function location, while canonical normalized fields always win. |
| **BUG → FIX / COMPATIBILITY** | Hosted tools and ordering | Hosted Responses tools could not pass through the normalized function-tool model, and separating them from function tools reordered mixed lists. | Hosted tools are preserved byte-for-byte and reconstructed in their original list positions alongside normalized function tools. |
| **BUG → FIX** | Assistant history text | All text was emitted as `input_text`; OpenAI rejects `input_text` on assistant history because it represents prior model output. | Assistant history emits `output_text`; system, developer, and user text remain `input_text`, while tool results remain `function_call_output`. String and list-form text history are both handled. |
| **REFACTOR / COMPATIBILITY** | Responses-native content history | `input_text`, `output_text`, `input_image`, and `input_file` blocks entering through the legacy LM door were not consistently normalized before typed conversion. | Known native blocks are normalized through the shared mapper and re-emitted with role-correct direction and lossless image/file fields. |
| **BUG → FIX / COMPATIBILITY** | File blocks | Raw base64 `file_data`, or a block containing both `file_data` and `file_id`, could be rewrapped with invented metadata or lose one of the original fields. | When the typed representation cannot be lossless, the original legacy block is retained and re-emitted verbatim. Normal data-URI and file-ID paths keep their existing typed behavior. |
| **BUG → FIX** | Pydantic response formats | `model_json_schema()` omitted `additionalProperties: false`, so otherwise valid structured-output requests could be rejected by the Responses API. | Object schemas are closed recursively at JSON Schema keyword positions. Explicit `additionalProperties` values and schema-shaped values inside defaults/examples are not rewritten. |
| **BUG → FIX / COMPATIBILITY** | Token and reasoning config | The compatibility converter did not consistently route `max_completion_tokens` through `max_output_tokens`; `reasoning_effort` could overwrite an explicit `reasoning` object. | `max_completion_tokens` maps to the shared `max_tokens` field and then to `max_output_tokens`; explicit `reasoning` wins, while the shorthand retains the existing `{effort, summary: "auto"}` behavior. |

### Tool-call replay and adapter behavior

| Classification | Behavior | Before | Now |
| --- | --- | --- | --- |
| **BUG → FIX** | Responses function-call arguments | Replay only read Chat-style `function.arguments`, so flat top-level Responses `arguments` became `{}`. | Replay accepts nested Chat arguments or top-level Responses arguments. |
| **BUG → FIX** | Tool-call identity | When a Responses item contained both `id="fc_*"` and `call_id="call_*"`, DSPy preferred the item ID, producing a tool result that OpenAI could not associate with the call. | `call_id` is canonical for replay and tool results; `id` remains the fallback for Chat-shaped calls. |
| **BUG → FIX** | Provider call metadata | Raw provider fields such as item ID, status, and parser diagnostics could leak back into a request or overwrite normalized call fields. | Internal provider fields are filtered; canonical name, arguments, type, and call identity are regenerated from the normalized part. |
| **BUG → FIX** | Tool-only LM outputs | Adapter postprocessing indexed `output["text"]`, so a valid tool-call-only result with no `text` key raised `KeyError`. | Missing and explicit `None` text are both accepted when the expected native tool-call field is present. Outputs with neither usable text nor expected tool calls still raise `AdapterParseError`. |
| **BUG → FIX / COMPATIBILITY** | SDK/output message dumps | Passing an OpenAI/LiteLLM message dump back through the legacy LM door could fail strict typed validation because output-only fields such as `id`, `type`, `status`, `annotations`, or `refusal` were retained. | The legacy door keeps only supported message input fields and strips provider output metadata before strict typed conversion. |

### Cache behavior

| Classification | Behavior | Before | Now |
| --- | --- | --- | --- |
| **BUG → FIX** | Cached strict provider models | Marking a cache hit with normal attribute assignment failed for Pydantic provider models configured with `extra="forbid"`. | DSPy deep-copies the cached response, clears usage, and marks the copy as a cache hit without mutating the original strict model. |

## Unchanged contracts

- **UNCHANGED** — The public `dspy.LM` return shape remains the existing legacy list/string-or-dict shape; this PR does not introduce a response-side API migration.
- **UNCHANGED** — Chat Completions tools and forced choices remain Chat-shaped. Only function-scoped `strict` and provider extras are placed correctly within that existing shape.
- **UNCHANGED** — Sync and async Responses calls use the same request converter and therefore have the same wire contract.
- **UNCHANGED** — Caller-supplied non-Pydantic response formats pass through unchanged.
- **UNCHANGED** — Existing explicit `additionalProperties` choices are preserved.
- **UNCHANGED** — Cache hits still report empty usage, fresh calls still report usage, and the original cached response is not mutated.
- **BOUNDARY** — The typed path remains strict (`extra="forbid"`); tolerance for SDK dumps and Responses-native legacy blocks exists only at the legacy `dspy.LM` compatibility boundary.
- **BOUNDARY** — Unknown or invalid request shapes are not handled with fallback cascades; they continue to raise through normal typed/provider validation.

## Purposefully out of scope

- **OUT OF SCOPE** — The broader response-side typed-LM redesign from #9842 is not included.
- **BOUNDARY** — Hosted tools remain provider-native passthrough objects; DSPy does not invent a provider-independent schema for web search, file search, code interpreter, MCP, or future hosted tools.
- **BOUNDARY** — Model-specific reasoning/temperature compatibility is left to the provider on the legacy Responses door, matching its prior behavior. The strict typed mapper can still enforce its existing validation.
- **OUT OF SCOPE** — This does not fix LiteLLM/OpenAI SDK parsing support for every newer hosted-tool response variant; the live hosted-tool probe uses the shape supported by the pinned SDK.
- **OUT OF SCOPE** — This does not add credentials to CI. OpenAI server-authority probes skip when `OPENAI_API_KEY` is unavailable.

## Review guide

1. **REFACTOR + FIX** — `dspy/clients/openai_format.py`: dialect-specific tools and choices, assistant text direction, response schemas, and provider-field precedence.
2. **REFACTOR + COMPATIBILITY** — `dspy/clients/lm.py`: the legacy compatibility door, native content/tool-choice tolerance, hosted-tool passthrough, ordering, and config precedence.
3. **FIX** — `dspy/core/types.py`: tool coercion, Responses call replay, and lossless legacy file blocks.
4. **FIX** — `dspy/adapters/base.py`: tool-only outputs plus top-level Responses arguments and `call_id` handling.
5. **FIX** — `dspy/clients/cache.py`: cache-hit marking for strict provider models.

The remaining diff is regression evidence at those boundaries plus seven consolidated live server-authority scenarios.

## Verification

- **TEST** — Focused affected suite: **195 passed, 27 skipped** with `uv run --frozen pytest --deno`.
- **LINT** — Ruff passed on every file changed by the final test-consolidation commit.
- **LIVE TEST** — The earlier 13-probe matrix passed locally against `gpt-5-nano`; it is now folded into seven equivalent server-authority scenarios without production-code changes.
- **CI BOUNDARY** — GitHub's real-LM job does not provide `OPENAI_API_KEY`, so the OpenAI Responses probes are expected to skip there; offline regression tests cover the mapped request, replay, adapter, and cache boundaries.

Includes the already-merged stacked cache fix #10027 and ports the assistant-history correction from #9652.


christianbundy:
@isaacbmiller would you consider fixing https://github.com/stanfordnlp/dspy/pull/9652 as part of this PR, or is it strictly out of scope? I noticed that the other PR will have a merge conflict once this lands. happy to help if you'd like me to rebase it on top of your changes or anything, but wanted to ask since this seems very relevant.

therewasaguy:
thanks @christianbundy ! Happy to update https://github.com/stanfordnlp/dspy/pull/9652 if this one lands first

AnnasMazhar:
Reviewed the tool-format changes — looks solid. The `tool_to_openai_responses()` and `tool_choice_to_openai_responses()` in `openai_format.py` correctly emit the flattened Responses shape (top-level `name`/`description`/`strict` instead of nested `function`), and the hosted-tool passthrough (`function not in tool and type not in (None, "function")`) ensures web_search/file_search/mcp tools arent coerced into a shape they dont fit. The `_convert_chat_request_to_responses_request` rewrite to route through the normalised mapper is the right call — this is cleaner than patching the hand-rolled converter.

One thing I verified: tools with `strict: true` on the Chat path produce `function.strict`, which was previously silently dropped. The new `LMToolSpec.strict` field hoists it from `provider_data` in the Chat path (`tool_to_openai`) and emits it as a top-level `strict` in the Responses path (`tool_to_openai_responses`) — correct per the Responses API behaviour where `strict` is indeed a top-level key. Good coverage.

MaximeRivest:
LGTM

PR #10030, opened 2026-07-14
fix(adapters): ChatAdapter.parse corrupts values on indented field headers

chuenchen309:
## What

`ChatAdapter.parse()` matches the field-header regex against `line.strip()`, but then slices the remaining content out of the *unstripped* `line` using the match offset computed from the *stripped* string:

```python
match = field_header_pattern.match(line.strip())
if match:
    header = match.group(1)
    remaining_content = line[match.end() :].strip()  # offset from stripped, sliced from unstripped
```

When a header line has leading whitespace, the offsets no longer line up, so the parsed value retains trailing characters of the header marker itself.

**Repro:**
```python
completion = "  [[ ## answer ## ]] Paris is the capital of France\n\n[[ ## completed ## ]]\n"
ChatAdapter().parse(MySignature, completion)
# => {'answer': ']] Paris is the capital of France'}   (should be {'answer': 'Paris is the capital of France'})
```

## Fix

Compute both the match and the slice against the same (stripped) string:

```python
stripped_line = line.strip()
match = field_header_pattern.match(stripped_line)
if match:
    header = match.group(1)
    remaining_content = stripped_line[match.end() :].strip()
```

## Test plan

- Added `test_parse_handles_indented_field_headers`, verified it fails against the current code with the exact corrupted output shown above, then passes with the fix.
- Ran the full `tests/adapters/` suite (231 tests) — all pass, no regressions.
- `pre-commit run` (ruff lint) — clean.

## AI-Generated Contribution disclosure

Found via an AI-assisted code review pass (Claude Code) across `dspy/adapters/`. I traced the root cause myself (confirmed the offset-mismatch mechanism by reading the surrounding parse logic), personally reproduced the bug with the exact snippet above, and wrote/verified the regression test before opening this PR.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -219,11 +219,12 @@ class ChatAdapter(Adapter):
         sections = [(None, [])]
 
         for line in completion.splitlines():
-            match = field_header_pattern.match(line.strip())
+            stripped_line = line.strip()
+            match = field_header_pattern.match(stripped_line)
             if match:
                 # If the header pattern is found, split the rest of the line as content
                 header = match.group(1)
-                remaining_content = line[match.end() :].strip()
+                remaining_content = stripped_line[match.end() :].strip()
                 sections.append((header, [remaining_content] if remaining_content else []))
             else:
                 sections[-1][1].append(line)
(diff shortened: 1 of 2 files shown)


chuenchen309:
While cross-checking my open PRs against this repo's file paths, I found #9942 (opened 2026-06-23) already fixes this exact same bug with an identical approach (using the stripped line's offset instead of the original line's), about three weeks before I found it independently. Closing this one in favor of the earlier PR rather than duplicate the review effort.

PR #10033, opened 2026-07-14
fix(teleprompt): validate restrict up front in BootstrapFewShotWithRandomSearch

chuenchen309:
## What

If `restrict` doesn't intersect any candidate seed in `range(-3, num_candidate_sets)`, every iteration of the seed loop hits `continue` at the restrict check, so the loop body — including the `best_program = program` assignment — never runs. The code after the loop unconditionally references `best_program`, raising a confusing `UnboundLocalError` instead of a clear validation error about the invalid `restrict` argument.

## Fix

Validate `restrict` up front and raise `ValueError` with a message naming the actual problem, before any candidate evaluation work starts.

## How I verified this

- Added `test_restrict_matching_no_candidate_seed_raises_clear_error`, confirmed it fails with `UnboundLocalError` before the fix and passes (raises the intended `ValueError`) after.
- Ran the full `tests/teleprompt/test_random_search.py` (2 passed) and the broader `tests/teleprompt/` suite (74 passed).
- pre-commit (ruff lint) clean.

## Disclosure

I used AI assistance (Claude) to help investigate and draft this fix, but I personally reviewed the root cause, wrote/ran the regression test, and reviewed the final diff before submitting.

### DIFF
--- a/dspy/teleprompt/random_search.py
+++ b/dspy/teleprompt/random_search.py
@@ -58,11 +58,13 @@ class BootstrapFewShotWithRandomSearch(Teleprompter):
         self.trainset = trainset
         self.valset = valset or trainset  # TODO: FIXME: Note this choice.
 
-        if restrict is not None and not any(seed in restrict for seed in range(-3, self.num_candidate_sets)):
-            raise ValueError(
-                f"`restrict` {restrict!r} does not match any candidate seed in "
-                f"range(-3, {self.num_candidate_sets}); no candidate programs would be evaluated."
-            )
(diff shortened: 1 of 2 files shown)


chuenchen309:
Good catch — pushed a fix that materializes `restrict` into a set once before validation, so a single-use iterable no longer gets consumed by the upfront check before the seed loop uses it. Added a regression test using `iter([-3])` (a valid seed) that reproduces the failure before this fix and passes after.

isaacbmiller:
lgtm

PR #10038, opened 2026-07-14
fix(utils): Unbatchify re-arms the full timeout on every queue.get(), violating max_wait_time

chuenchen309:
## What

The worker loop's outer `while` condition correctly checks elapsed time against `max_wait_time`, but each inner `queue.get(timeout=self.max_wait_time)` call re-arms with the *full* timeout again rather than the time remaining in the window. Under trickling input (items arriving spaced just under `max_wait_time` apart), each `get()` succeeds before its own fresh timeout expires, so the outer time check is only re-evaluated after another full wait — the batch can be flushed at a multiple of the configured `max_wait_time` instead of within it.

Measured: with `max_wait_time=0.1s` and two items arriving 0.08s apart, the batch flushed at **0.181s** — nearly double the configured bound.

```python
u = Unbatchify(batch_fn=..., max_batch_size=10, max_wait_time=0.1)
# item 1 submitted at t=0, item 2 submitted at t=0.08
# batch flushes at t≈0.18, not within t≈0.1
```

## Fix

Compute the time remaining in the window and pass that (floored at 0) to `queue.get()` instead of the full `max_wait_time`.

## Testing

Added `test_unbatchify_honors_max_wait_time_under_trickling_input` to `tests/utils/test_unbatchify.py`. Confirmed via `git stash` that it fails against the pre-fix code (measured 0.181s against a `wait_time*1.5 = 0.15s` bound) and passes after.

- `pytest tests/utils/test_unbatchify.py` — 3/3 passed.
- `ruff check`/`ruff format --check` clean on the changed lines (pre-existing unrelated formatting drift elsewhere in the file left untouched).
- pre-commit (ruff lint) passed on commit.

## AI disclosure

Developed with AI assistance (Claude Code), which located the timing bug while auditing `unbatchify.py`'s worker loop. I reviewed the diff, independently reproduced the max_wait_time violation and the fix by executing both against the actual class with real threads/timing, and ran the tests/linters myself before opening this PR.

### DIFF
--- a/dspy/utils/unbatchify.py
+++ b/dspy/utils/unbatchify.py
@@ -57,8 +57,9 @@ class Unbatchify:
             futures = []
             start_time = time.time()
             while len(batch) < self.max_batch_size and (time.time() - start_time) < self.max_wait_time:
+                remaining_time = self.max_wait_time - (time.time() - start_time)
                 try:
-                    input_item, future = self.input_queue.get(timeout=self.max_wait_time)
+                    input_item, future = self.input_queue.get(timeout=max(remaining_time, 0))
                     batch.append(input_item)
                     futures.append(future)
                 except queue.Empty:
--- a/tests/utils/test_unbatchify.py
+++ b/tests/utils/test_unbatchify.py
@@ -70,3 +70,28 @@ def test_unbatchify_timeout_trigger():
     assert results == [101, 201]
 
     unbatcher.close()
+
+
+def test_unbatchify_honors_max_wait_time_under_trickling_input():
+    """A batch must flush within max_wait_time of the FIRST item arriving, even if
+    later items keep trickling in just under max_wait_time apart from each other.
+    Each queue.get() must wait only the time remaining in the window, not the full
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Traced the timing bug: the outer `while` loop's elapsed-time check only gets re-evaluated after each `queue.get()` call returns, but the old code armed every `queue.get()` with the full `max_wait_time` regardless of how much of the window had already elapsed. So a batch that keeps receiving new items just under the timeout (each one resetting the wait to a fresh full window) can overrun the intended budget by close to 2x, not bounded to `max_wait_time` from the first item as documented. Computing `remaining_time` from `start_time` and using `max(remaining_time, 0)` as the per-call timeout fixes it, the `max(..., 0)` guards the case where time advances between the while-check and the `get()` call itself, avoiding a negative timeout to `queue.get`. The new test's trickling-input scenario is the right way to catch this, a test with items arriving all at once or all late wouldn't exercise the re-arming bug at all. Looks correct.

isaacbmiller:
lgtm, thanks for the PR

PR #10044, opened 2026-07-14
feat(callbacks): expose optimizer lifecycle hooks

isaacbmiller:
> **Optimizer tracking stack (1/2):** **#10044 (this PR)** → #10045 RandomSearch events. This PR is based on independently mergeable #10043 because both touch callback call-ID plumbing.

## Issue / reproduction

DSPy callbacks can observe modules, LMs, tools, adapters, and evaluations inside an optimizer, but they cannot observe the optimizer's `compile` boundary itself.

A callback therefore cannot answer:

```text
Which compile call caused this work?
What arguments started it?
Which program did it return?
Did it fail or get cancelled?
```

## Why this is the root cause

`Teleprompter.compile` is only an interface. Every concrete optimizer overrides it, so decorating the base method would never see those calls. Existing callback dispatch also treats an otherwise unknown wrapped method as a module call.

The missing boundary is where each `Teleprompter` subclass declares its concrete `compile` method.

## What the fix does

- Adds `BaseCallback.on_optimizer_start` and `on_optimizer_end`.
- Lets the existing callback wrapper carry an internal `optimizer` kind when selecting handlers.
- Wraps a `compile` method declared directly on a `Teleprompter` subclass when that subclass is created.
- Tracks whether a callback call scope is still live, so detached work cannot reuse a completed optimizer as its parent.
- Exports `Teleprompter` from `dspy.teleprompt`'s public surface.

The callback shape matches DSPy's existing components:

```python
class Tracker(BaseCallback):
    def on_optimizer_start(self, call_id, instance, inputs): ...
    def on_optimizer_end(self, call_id, outputs, exception): ...
```

## Why this fixes the root cause

The wrapper is installed at the one point shared by direct optimizer implementations: subclass definition. It preserves the visible signature with `functools.wraps`, assigns a call ID before entering the method body, restores the parent ID before the end callback, and reports optimizer cancellation or other `BaseException` exits before re-raising them.

`super().compile()` remains an ordinary nested method call. Its inner start callback sees the outer optimizer call ID as its parent, so the relationship uses the same callback tree DSPy already exposes elsewhere.

Python copies `ContextVar` values into child tasks. Restoring the parent task's string cannot invalidate that copy, so each wrapped call also carries a private mutable scope. Scope exit marks it complete; the next callback dispatch in a copied task or bound thread walks to the nearest still-live ancestor before invoking handlers.

## Why this is concise

The production change is one callback classification branch, an eight-line subclass hook, and one private linked call scope. The scope is the full detached-work fix: contexts copy the same tiny liveness object, so there is no global lookup or owner bookkeeping. It deliberately does not add:

- an optimizer invocation registry;
- task or thread ownership;
- recursion or `super()` deduplication;
- callback-wrapper discovery or suppression;
- opaque-decorator handling; or
- result/coroutine validation.

Those mechanisms accounted for most of the previous implementation and defended behavior no built-in optimizer currently exercises.

## Review map

- `dspy/teleprompt/teleprompt.py`: wraps direct `compile` declarations.
- `dspy/utils/callback.py`: defines and dispatches lifecycle handlers.
- `dspy/utils/callback_context.py`: drops completed call scopes copied into detached tasks or bound threads.
- `tests/callback/test_optimizer_lifecycle.py`: exercises signatures, inputs/output, nesting, inheritance, failures, and async cancellation.

## Compatibility and downsides

- This is a **compile method-call lifecycle**, not an inferred logical-run lifecycle. Real `super().compile()` paths such as `InferRules` produce nested pairs on the same instance. Trackers that only want the outer call can use the parent call ID.
- A subclass that only inherits `compile` gets no additional wrapper. A `compile` supplied solely by a mixin and absent from the subclass `__dict__` is not wrapped.
- `functools.wraps` preserves the visible signature, but function identity and the class `__dict__` entry change.
- An explicitly generic-`@with_callbacks`-decorated `compile` may retain its inner module callback pair inside the optimizer pair. DSPy does not silently suppress a decorator the author requested.
- Global callbacks now observe internal child optimizers, increasing callback volume.
- Async `compile` methods use the existing async callback path. Optimizer cancellation is passed to `on_optimizer_end` and re-raised; ordinary callback handler failures remain logged and isolated.
- Callback dispatch from a copied async/thread context skips completed wrapper-created scopes and resumes at the nearest live ancestor. This prevents a detached task from reporting work under an optimizer that has already ended.
- `ACTIVE_CALL_ID` remains the same public string-valued `ContextVar`. A raw `.get()` in detached code, before that code enters another callback-wrapped boundary, therefore retains normal Python snapshot semantics. Start/end handlers are normalized before dispatch; replacing the `ContextVar` with a proxy would be a wider compatibility break.
- Replacing `compile` after class creation does not install a wrapper automatically.

## Verification

Focused lifecycle and callback tests:

```bash
uv run --frozen --no-sync pytest --deno -q \
  tests/callback/test_callback.py \
  tests/callback/test_optimizer_lifecycle.py \
  tests/teleprompt/test_teleprompt.py
```

Direct result: `18 passed`.

The cumulative callback/evaluate/teleprompt integration sweep reports `125 passed, 25 skipped`. The sandboxed full suite reports `1148 passed, 251 skipped`; its five failures and five setup errors were all network lookups. Rerunning that exact network-dependent subset with network access reports `21 passed, 2 xfailed`.

Copy-paste demonstration:

```bash
DSPY_CACHEDIR=/tmp/dspy-pr-demo-cache uv run --frozen --no-sync python - <<'PY'
import warnings
warnings.filterwarnings("ignore", message="Core Pydantic V1 functionality")

import dspy
from dspy.utils.callback import BaseCallback


class Recorder(BaseCallback):
    def on_optimizer_start(self, call_id, instance, inputs):
        self.call_id = call_id
        print("start", type(instance).__name__, "examples", len(inputs["trainset"]))

    def on_optimizer_end(self, call_id, outputs, exception):
        print("end", "same_call", call_id == self.call_id, "output", type(outputs).__name__, "error", exception)


class IdentityOptimizer(dspy.Teleprompter):
    def compile(self, student, *, trainset):
        return student


student = dspy.Predict("question -> answer")
with dspy.context(callbacks=[Recorder()]):
    compiled = IdentityOptimizer().compile(student, trainset=["example"])

print("compiled_is_student", compiled is student)
PY
```

Direct output, unedited:

```text
start IdentityOptimizer examples 1
end same_call True output Predict error None
compiled_is_student True
```

The detached-task regression runs the child module only after optimizer end. Direct output, unedited:

```text
order ['optimizer_start', 'optimizer_end', 'module_start']
detached_parent None
```

### DIFF
--- a/dspy/teleprompt/__init__.py
+++ b/dspy/teleprompt/__init__.py
@@ -31,5 +31,6 @@ __all__ = [
     "LabeledFewShot",
     "InferRules",
     "SIMBA",
+    "Teleprompter",
     "bootstrap_trace_data",
 ]
--- a/dspy/teleprompt/teleprompt.py
+++ b/dspy/teleprompt/teleprompt.py
@@ -1,11 +1,19 @@
 from typing import Any
 
 from dspy.primitives import Example, Module
+from dspy.utils.callback import with_callbacks
 
 
 class Teleprompter:
     def __init__(self):
         pass
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        super().__init_subclass__(**kwargs)
+
+        compile_method = cls.__dict__.get("compile")
+        if compile_method is not None:
+            cls.compile = with_callbacks(compile_method, _callback_kind="optimizer")
 
     def compile(self, student: Module, *, trainset: list[Example], teacher: Module | None = None, valset: list[Example] | None = None, **kwargs) -> Module:
         """
--- a/dspy/utils/callback.py
+++ b/dspy/utils/callback.py
@@ -5,7 +5,8 @@ import uuid
 from typing import Any, Callable
 
 import dspy
-from dspy.utils.callback_context import ACTIVE_CALL_ID
+from dspy.utils.callback_context import ACTIVE_CALL_ID as ACTIVE_CALL_ID
+from dspy.utils.callback_context import _active_call_context, _normalize_active_call_context
 
 logger = logging.getLogger(__name__)
 
@@ -252,9 +253,40 @@ class BaseCallback:
         """
         pass
 
+    def on_optimizer_start(
+        self,
+        call_id: str,
+        instance: Any,
+        inputs: dict[str, Any],
+    ):
+        """A handler triggered when an optimizer starts compiling a program.
 
-def with_callbacks(fn):
+        Args:
+            call_id: A unique identifier for the optimizer call.
+            instance: The Teleprompter instance.
+            inputs: The arguments passed to the optimizer's ``compile`` method.
+        """
+        pass
+
+    def on_optimizer_end(
+        self,
+        call_id: str,
+        outputs: Any | None,
+        exception: BaseException | None = None,
+    ):
+        """A handler triggered after an optimizer finishes compiling a program.
+
+        Args:
+            call_id: The identifier shared with the corresponding start handler.
+            outputs: The compiled program, or None if compilation failed.
+            exception: The exception raised by compilation, if any.
+        """
+        pass
+
+
+def with_callbacks(fn, *, _callback_kind: str | None = None):
     """Decorator to add callback functionality to instance methods."""
+    is_optimizer_callback = _callback_kind == "optimizer"
 
     def _execute_start_callbacks(instance, fn, call_id, callbacks, args, kwargs):
         """Execute all start callbacks for a function call."""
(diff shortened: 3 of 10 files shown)


isaacbmiller:
@greptileai

isaacbmiller:
superseded by #10120.

PR #10046, opened 2026-07-14
Make Audio URL download opt-in; add SSRF guard + timeout to Audio & Image (fixes #9993)

### LINKED ISSUE #9993: dspy.Audio auto-downloads from any http(s) string with no timeout and no SSRF guard
### Description

`dspy.Audio` auto-downloads from any URL-shaped string it's given, with no timeout and no restriction on what the URL points to. This happens unconditionally, not behind an opt-in flag like `dspy.Image` has.

`encode_audio()` (`dspy/adapters/types/audio.py`) is invoked from `Audio`'s pydantic `model_validator(mode="before")`, so it runs on every `Audio(...)` construction, including whenever a plain string is coerced into an `Audio`-typed field (a tool return value, a field parsed out of retrieved document content, a model output that an adapter is coercing into a typed field):

```python
elif isinstance(audio, str) and audio.startswith("http"):
    a = Audio.from_url(audio)
    return {"data": a.data, "audio_format": a.audio_format}
```

```python
@classmethod
def from_url(cls, url: str) -> "Audio":
    response = requests.get(url)
    response.raise_for_status()
    ...
```

Two issues with this specific call:

1. **No timeout.** `requests.get(url)` has no `timeout=`, so a slow or intentionally-hanging endpoint blocks the request indefinitely.
2. **No URL/host validation.** Nothing stops the URL from pointing at `http://169.254.169.254/latest/meta-data/` (cloud instance metadata), a private-network service, or `localhost`. Compare `dspy.Image`, where `_encode_image_from_url` has the same missing-timeout issue, but at least the download only happens when the caller explicitly passes `download=True` — `Audio` has no equivalent gate at all, so any code path that builds an `Audio` from a plain string (e.g. an adapter coercing a field, or application code doing `dspy.Audio(some_value)` on data that ultimately traces back to a retrieved document or tool output) triggers the fetch with no way to opt out short of pre-filtering the string before it reaches the constructor.

In an agent/RAG context where documents, tool results, or model output can influence what string ends up being passed into an `Audio`-typed field, this is a real internal-network-probing/DoS surface, not just a hardening nicety.

### Suggested fix

- Add `timeout=` to both `requests.get` calls (`Audio.from_url` here and `_encode_image_from_url` in `image.py`, which has the same gap).
- For `Audio`, either gate the auto-download behind an explicit flag the same way `Image` does with `download`, or at minimum validate the resolved URL/IP isn't pointing at a private/link-local/loopback address before fetching (the same class of check `Predict.load_state`'s SSRF fix in #9549 is doing for a different code path).

### Version

Checked against current `main`.

immu4989:
Fixes #9993. Implements option 1 from the issue thread (opt in via constructor, mirroring `Image`), plus the destination validation @ErenAta16 asked for on both types.

## The problem

`Audio` downloaded from any `http(s)` string unconditionally, and `encode_audio` runs from `Audio`'s `model_validator(mode="before")`. So any string coerced into an `Audio`-typed field (a tool return, text parsed from a retrieved document, a model output an adapter is casting) triggered a live outbound `requests.get` with no timeout and no SSL control. That is an SSRF surface and a hanging-host DoS, and it fired without any explicit `Audio(...)` call.

## The fix

**Trigger gating (Audio).** Downloading from a URL is now opt-in, mirroring `Image`. A bare `Audio(url)` or a string coerced into an `Audio` field raises a clear error pointing at `Audio.from_url(url)`; `Audio(url, download=True)` opts in.

**Destination validation (Audio + Image).** A shared `assert_public_url` guard (`dspy/adapters/types/_url_safety.py`) rejects loopback / private / link-local / reserved addresses (including the `169.254.169.254` cloud metadata endpoint) and non-`http(s)` schemes. Applied in `Audio.from_url` **and** `Image._encode_image_from_url`, so an explicit `download=True` still cannot be pointed at an internal address when the URL itself came from untrusted input. Gating the trigger and validating the destination are complementary, per @ErenAta16's note.

**Timeout + verify.** Added a request timeout to both download paths (`Audio` had neither a timeout nor SSL-verify control; `Image` lacked a timeout).

## Breaking change

`Audio("http://...")` no longer auto fetches. The raised error gives the one line migration (`Audio.from_url(url)`, or `download=True`). This is the intended tradeoff for a real SSRF/DoS fix, as discussed in the issue.

## Scope note

The guard validates at DNS resolution time and does not defend against DNS rebinding (a host that resolves public here but private at request time). That is a deeper mitigation than a media loader should own; it is called out in the code. This PR closes the direct SSRF surface and the unconditional-download/no-timeout surface.

## Tests

`tests/adapters/test_url_safety.py` (new, fully offline: the guard and the opt in gate both raise before any network call). Covers: guard rejects loopback/private/link-local/metadata/ipv6-loopback/non-http schemes and allows public IPs; `Audio` refuses to download by default via both the constructor and the coercion path; `Audio.from_url` and `Image._encode_image_from_url` apply the guard; non-URL Audio paths still work. Existing `tests/adapters/test_audio.py` and `tests/signatures/test_adapter_image.py` pass.

### DIFF
--- a/dspy/adapters/types/_url_safety.py
+++ b/dspy/adapters/types/_url_safety.py
@@ -1,42 +1,76 @@
 """Shared URL-safety helpers for media types that download from URLs.
 
-`Image` and `Audio` can fetch media from a URL. Two things must be true for that
-to be safe against SSRF and hanging-host DoS:
+`Image` and `Audio` can fetch media from a URL. Three things must be true for
+that to be safe against SSRF and hanging-host DoS:
 
 1. The fetch must be gated behind an explicit opt-in, so untrusted content
    (tool output, a retrieved document, a model output being cast to a media
    type) can never *trigger* an outbound request on its own. That gating lives
    in the media types themselves.
 
-2. Even an opted-in fetch must validate its destination, because the URL itself
-   may have come from untrusted input. `assert_public_url` rejects URLs that
-   resolve to loopback, private, link-local, or otherwise non-public addresses
-   (e.g. the cloud metadata endpoint ``169.254.169.254``).
+2. Every request the fetch actually makes must go to a public destination, at
+   every hop. `safe_get` validates the initial URL *and every redirect target*
+   with `assert_public_url` before connecting, so a public URL that redirects
+   to an internal address (e.g. the cloud metadata endpoint) is refused.
 
-These are complementary: (1) controls the trigger, (2) validates the target.
+3. Neither the DNS lookup nor the HTTP request may hang indefinitely. The HTTP
+   request is bounded by `timeout`; name resolution is bounded separately (see
+   `_resolve`), since `socket.getaddrinfo` does not honor the socket timeout on
+   most platforms.
+
+Known limitation: `assert_public_url` validates at resolution time and does not
+defend against DNS rebinding (a host that resolves public here and private when
+the socket actually connects). Closing that requires resolving once and
+connecting to the pinned IP with the original host in the SNI/Host header, which
+is a deeper change than a media loader should own.
 """
 from __future__ import annotations
 
 import ipaddress
 import socket
-from urllib.parse import urlparse
+import threading
+from urllib.parse import urljoin, urlparse
+
+import requests
 
 DEFAULT_DOWNLOAD_TIMEOUT = 30.0
+MAX_REDIRECTS = 5
 
 
-def assert_public_url(url: str) -> None:
+def _resolve(host: str, port: int, timeout: float):
+    """`socket.getaddrinfo` with a hard wall-clock timeout.
+
+    `getaddrinfo` ignores `socket.setdefaulttimeout` on most platforms, so a
+    host pointed at a dead DNS server can hang the caller indefinitely. Run it
+    on a daemon thread and give up after `timeout` seconds; the caller never
+    blocks longer than that even if the lookup thread is still stuck.
+    """
+    box: dict = {}
+
+    def work():
+        try:
+            box["ok"] = socket.getaddrinfo(host, port)
+        except Exception as e:  # noqa: BLE001 - surfaced below as ValueError
+            box["err"] = e
+
+    th = threading.Thread(target=work, daemon=True)
+    th.start()
+    th.join(timeout)
+    if th.is_alive():
+        raise ValueError(f"DNS resolution for {host!r} timed out after {timeout}s (SSRF check).")
+    if "err" in box:
+        raise ValueError(f"Could not resolve host {host!r}: {box['err']}")
+    return box["ok"]
+
+
+def assert_public_url(url: str, *, dns_timeout: float = DEFAULT_DOWNLOAD_TIMEOUT) -> None:
     """Raise ``ValueError`` unless every address ``url``'s host resolves to is
     a public, routable IP address.
 
     Guards against the common SSRF targets: loopback (``127.0.0.1``,
     ``localhost``), private ranges (``10/8``, ``192.168/16``, ...), link-local
     (``169.254/16``, including the cloud metadata endpoint), and other reserved
-    or non-global ranges.
-
-    Note: this validates at resolution time and does not defend against DNS
-    rebinding (a host that resolves to a public IP here but a private one when
-    the request is actually made). That is a deeper mitigation than a media
-    loader should own; this closes the direct SSRF surface.
+    or non-global ranges. Name resolution is bounded by ``dns_timeout``.
     """
     parsed = urlparse(url)
     if parsed.scheme not in ("http", "https"):
@@ -48,11 +82,7 @@ def assert_public_url(url: str) -> None:
     if not host:
         raise ValueError(f"Could not parse a host from URL: {url!r}")
 
-    try:
-        addrinfos = socket.getaddrinfo(host, parsed.port or (443 if parsed.scheme == "https" else 80))
-    except socket.gaierror as e:
-        raise ValueError(f"Could not resolve host {host!r}: {e}") from e
-
+    addrinfos = _resolve(host, parsed.port or (443 if parsed.scheme == "https" else 80), dns_timeout)
     for info in addrinfos:
         ip = ipaddress.ip_address(info[4][0])
         if not ip.is_global or ip.is_multicast or ip.is_reserved:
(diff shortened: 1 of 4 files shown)


ErenAta16 (commented):
This matches what we discussed on #9993: the opt-in gate closes the trigger (untrusted content can no longer cause an implicit fetch by being coerced into an `Audio` field), and `assert_public_url` closes the destination side for both `Audio.from_url` and `Image._encode_image_from_url`. Checked `assert_public_url` against the test matrix: `is_global`/`is_multicast`/`is_reserved` correctly reject loopback, RFC1918 private ranges, link-local (including `169.254.169.254`), IPv6 loopback, and `0.0.0.0`, and the scheme check rejects `file://`/`ftp://`/`gopher://` before any resolution happens. The `model_validate` coercion test is the important one, that's the exact path the original issue was about, and it correctly raises instead of fetching.

Two notes, neither blocking:

1. The DNS-rebinding gap you flagged in the docstring is accurate and worth having called out explicitly rather than silently, a host that resolves to a public IP during `assert_public_url` and a private one by the time `requests.get()` actually connects would still get through. Agreed this is a deeper mitigation than a media loader should own on its own; resolving once and connecting to the resolved IP directly (with the original host sent via SNI/Host header) is the usual fix if this ever needs closing, but that is a bigger change than this PR's scope.

2. `assert_public_url`'s own `socket.getaddrinfo` call has no timeout, only the subsequent `requests.get()` does. So the "hanging-host DoS" protection is not fully complete yet, a URL whose host points at a DNS server that never responds can still hang the caller indefinitely during the SSRF check itself, before `timeout` ever comes into play. Might be worth `socket.setdefaulttimeout()` scoped around the `getaddrinfo` call (or `signal.alarm` on platforms that support it) if that path matters for your threat model.

Good fix for the actual reported issue either way.

immu4989:
Thanks both, pushed a fix for the redirect gap and the resolution timeout.

**Redirect-based SSRF bypass (@greptile-apps P1).** Correct and a real hole, the guard validated the initial URL but `requests.get` followed redirects to an unvalidated destination. Fixed by routing both download paths through a new `safe_get()` that follows redirects manually and runs `assert_public_url` on every hop (the initial URL and each `Location`), so a public URL that 3xx-redirects to `169.254.169.254` or any private address is refused instead of followed. Redirect depth is capped too. Added tests: redirect-to-private is refused, a redirect to another public address is still followed, and a redirect loop terminates.

**DNS-resolution timeout (@ErenAta16 note 2).** Also correct, `socket.getaddrinfo` ignores the socket timeout on most platforms, so the SSRF check itself could hang on a dead resolver before the request `timeout` ever applied. `assert_public_url` now bounds resolution with a daemon-thread wall-clock timeout (`dns_timeout`), so the caller never blocks longer than that even if the lookup thread is still stuck. Test covers it with a slow-resolver mock.

**DNS rebinding (@ErenAta16 note 1).** Agreed this stays out of scope, resolve-and-pin (connect to the resolved IP with the original host in SNI/Host) is the right fix but a larger change than this loader should carry. Kept the limitation documented in the module so it is explicit rather than silent.

52 tests pass. Happy to adjust anything else.

immu4989:
Thanks, and good catch on the CGNAT range. Since DSPy supports Python 3.10 and the `100.64.0.0/10` `is_global` mislabel (bpo-38655) isn't fixed until 3.11, that's a real bypass on a supported interpreter rather than a theoretical one, so I closed it rather than only documenting it: `_EXTRA_BLOCKED_NETS` now rejects the range explicitly, version-independently, with a test (`http://100.64.0.1/...` is refused). Left a pointer to it in the module's known-limitation note next to the DNS-rebinding caveat.

Agreed on the multicast/reserved checks being intentional defence-in-depth (they also cover the IPv6 multicast case where `is_global` can read True), and on the daemon-thread trade-off. 53 tests pass.

ErenAta16:
Both notes from my earlier pass are resolved: the redirect-based bypass is closed by `safe_get()` validating every hop (initial URL + each `Location`, with a capped depth), and the DNS-resolution hang is closed by the daemon-thread wall-clock timeout on `getaddrinfo`. The remaining DNS-rebinding gap is correctly scoped out and documented rather than silently left. 52 passing tests, no regressions on the existing Image path. Good to merge from my side.

immu4989:
Checking in since this has been open a couple of weeks. @ErenAta16 confirmed above that both review points are resolved, the branch still merges cleanly, and there's nothing outstanding on my side.

Is there anything you'd like changed before this can be looked at? Happy to rebase or split it up if the three-file diff is the awkward part.

ErenAta16 (commented):
Nothing outstanding from me on the code. The concrete thing standing between this and a review is mechanical, so here is where it actually sits.

The branch is 39 commits behind `main` (`0312f0da`), and the combined status on `fc8dd65f` is `pending` with **zero** reported contexts — no check ever posted, so there is nothing for a maintainer to look at as green. That is what `mergeable_state: unstable` is reflecting, not a test failure.

I rebased it locally onto current `main` to see whether those 39 commits broke anything:

```
Rebasing (1/3) (2/3) (3/3)
Successfully rebased and updated refs/heads/rebase-probe.
```

No conflicts, all three commits replay. After the rebase:

| suite | result |
|---|---|
| `tests/adapters/test_url_safety.py` | 24 passed |
| `tests/adapters/test_audio.py` | 9 passed |

So a rebase-and-push is enough to re-trigger CI, and worth doing since the current head has no status at all. I would not split the diff — the three files are one change (`_url_safety.py` is the guard, `audio.py`/`image.py` are its two call sites), and separating them would land a guard with no callers.

One thing I could not check: the wider `tests/adapters/` run doesn't finish for me locally, it hangs on network-dependent tests. Of the failures I did see, `test_baml_adapter.py::test_baml_adapter_async_functionality` fails identically on a clean `main` checkout, so it isn't from this branch; the rest are in files this PR doesn't touch and I stopped the run before it labelled them, so I can't say more than that.

I'm not a maintainer here, so I can't move this along beyond that.

immu4989:
Closing this one. #10111 landed in 3.3.0 and covers the implicit-fetch problem more thoroughly than this PR did, and it draws the SSRF boundary differently on purpose (trust policy stays with the caller invoking `from_url()`), so most of what's here is either superseded or declined by design.

The one piece that didn't make it into 3.3.0 is the missing `timeout` on the two `requests.get` calls, which is the remaining half of #9993. I've split that into a small follow-up PR.

Issue #8782, opened 2025-09-09
Update docs for GEPA train/val split advise


LakshyAAAgrawalP:
https://github.com/stanfordnlp/dspy/blob/main/docs/docs/learn/optimization/overview.md

PR #10048, opened 2026-07-14
docs(gepa): explain trainset/valset roles and split guidance

ymxlx:
## What

Adds a "Choosing the training and validation sets" subsection to the GEPA getting-started walkthrough (`docs/docs/getting-started/gepa-optimization.md`). The walkthrough previously passed `trainset` and `valset` into `compile()` without explaining their distinct roles or the risks of how you split the data.

The new subsection covers:

- **Roles:** `trainset` drives reflective prompt updates, while `valset` tracks Pareto scores and selects the program returned by `compile`.
- **Split guidance:** keep `trainset` as large as possible and make `valset` the smallest sample that still represents the downstream distribution — an oversized `valset` also consumes metric-call budget that would otherwise let GEPA explore more candidate prompts.
- **Omitting `valset`:** GEPA reuses `trainset` for both, which is useful as an inference-time search strategy but deliberately overfits; pass a separate validation set when you care about performance on unseen data.

Addresses #8782.

## Verification

Every claim is grounded in the current implementation, `dspy/teleprompt/gepa/gepa.py`:

- Roles of the two sets — docstring, lines 485–486.
- `valset = valset or trainset` fallback plus the overfitting / inference-time-search warning — lines 515–519.
- The "a smaller `valset` lets GEPA explore more within the same budget" tradeoff — the budget computation at lines 504–512 and the warning at line 525.

Docs-only change; no code or tests are affected.

## AI assistance disclosure

Per `CONTRIBUTING.md`: this documentation was written with AI assistance (Anthropic's Claude, via Claude Code). I used it to draft the prose and to cross-check every statement against the GEPA source cited above. I have read and verified the wording and the technical claims myself, I can explain them without the tool, and I own the correctness of this change.

### DIFF
--- a/docs/docs/getting-started/gepa-optimization.md
+++ b/docs/docs/getting-started/gepa-optimization.md
@@ -86,6 +86,24 @@ Finally, we compile our optimized program:
 optimized_haiku_bot = optimizer.compile(haiku_bot, trainset=train, valset=val)
 ```
 
+### Choosing the training and validation sets
+
+GEPA uses these two splits for different jobs. The `trainset` supplies examples
+for reflective prompt updates, while the `valset` tracks Pareto scores and
+selects the program returned by `compile`. Keep a final test set separate so you
+can evaluate the optimized program on examples that influenced neither step.
+
+There is no universal split ratio. For generalization, keep as much data as
+possible in `trainset` and make `valset` the smallest sample that still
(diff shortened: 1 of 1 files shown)


isaacbmiller:
lgtm. Thanks for the PR!

PR #10051, opened 2026-07-15
fix(adapters): unwrap markdown code fence in parse_value so None survives (#8181)

### LINKED ISSUE #8181: [Bug] chatadapter parse bug
### What happened?

# Input
"[[ ## reasoning ## ]]
用户希望记录网络连接不稳定的问题以及他的配置信息。根据对话历史，用户提到网络连接不稳定可能是由于路由器问题，并且他已经尝试过重启路由器但没有效果。用户没有提供任何网页链接信息，因此 `memory_url_info` 字段应设置为 `None`。`memory_text` 字段应包含用户描述的问题和配置信息。

[[ ## tool_parameters ## ]]
```python
{
  "memory_text": "网络连接不稳定，可能是由于路由器问题。已尝试重启路由器但没有效果。",
  "memory_url_info": None
}
```

[[ ## completed ## ]]"

# Output

{
reasoning: "用户希望记录网络连接不稳定的问题以及他的配置信息。根据对话历史，用户提到网络连接不稳定可能是由于路由器问题，并且他已经尝试过重启路由器但没有效果。用户没有提供任何网页链接信息，因此 `memory_url_info` 字段应设置为 `None`。`memory_text` 字段应包含用户描述的问题和配置信息。"
tool_parameters: {
      memory_text: "网络连接不稳定，可能是由于路由器问题。已尝试重启路由器但没有效果。"
      memory_url_info: "None}```" }
}

As we can see, the value of  memory_url_info is  **"None}```"** 

It should be  memory_url_info: None.

### Steps to reproduce

# Input
"[[ ## reasoning ## ]]
用户希望记录网络连接不稳定的问题以及他的配置信息。根据对话历史，用户提到网络连接不稳定可能是由于路由器问题，并且他已经尝试过重启路由器但没有效果。用户没有提供任何网页链接信息，因此 `memory_url_info` 字段应设置为 `None`。`memory_text` 字段应包含用户描述的问题和配置信息。

[[ ## tool_parameters ## ]]
```python
{
  "memory_text": "网络连接不稳定，可能是由于路由器问题。已尝试重启路由器但没有效果。",
  "memory_url_info": None
}
```

[[ ## completed ## ]]"

### DSPy version

2.6.23

marcodelpin:
Fixes #8181.

When an LM wraps a dict/list output value in a markdown code fence (```python ... ```), `json_repair.loads` coerces the Python literal `None` into the string `"None"` — json_repair only understands lowercase JSON `null`. Because `parse_value` tries json_repair first and only falls back to `ast.literal_eval` when json_repair returns `""`, the fenced value never reaches `ast.literal_eval` and the `None` is silently turned into the string `"None"`.

## What changed

`dspy/adapters/utils.py`:

- New `_strip_markdown_code_fence(text)` helper: returns the inner body of a single ```` ```lang ... ``` ```` block, or `None` when `text` is not a single fenced block (non-fenced input is left untouched).
- In `parse_value`, when the value is a single fenced block, strip the fence and `ast.literal_eval` the unwrapped body **before** the json_repair fallback, so `None`/`True`/`False` literals survive. Non-fenced input keeps the existing json_repair-then-ast path unchanged.

## Reproduction

On stock `main`:

```python
from dspy.adapters.utils import parse_value

fenced = "```\n[1, None, 3]\n```"
parse_value(fenced, list)
# -> [1, 'None', 3]   # the None became the string "None"
```

With this patch it returns `[1, None, 3]`.

## Tests

Two offline unit regressions added to `tests/adapters/test_adapter_utils.py` (fenced dict with `None`, and a plain fence without a language tag). Full `tests/adapters/test_adapter_utils.py` is green (8/8). Verified A/B: the two new tests **fail on stock `main`** and **pass with the patch**. `ruff check` clean on the touched files.

## Disclosure

AI-assisted (Claude Code): the change and tests were drafted with AI assistance, then reviewed, adversarially re-checked, and verified (stock-vs-patched A/B) by me before submission. This contribution is my own and offered under the project's MIT license.

### DIFF
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -2,6 +2,7 @@ import ast
 import enum
 import inspect
 import json
+import re
 import types
 from collections.abc import Mapping
 from typing import Any, Literal, Union, get_args, get_origin
@@ -146,6 +147,19 @@ def find_enum_member(enum, identifier):
     raise ValueError(f"{identifier} is not a valid name or value for the enum {enum.__name__}")
 
 
+_MARKDOWN_CODE_FENCE = re.compile(r"\s*```[^\n]*\n(?P<body>.*?)\n?```\s*\Z", re.DOTALL)
+
+
+def _strip_markdown_code_fence(text):
+    """Return the inner body of a single ```lang ... ``` markdown code fence, or None.
+
+    An LM sometimes wraps a dict/list value in a fenced code block. Returns None when
+    `text` is not a single fenced block, so callers can leave non-fenced input untouched.
+    """
+    match = _MARKDOWN_CODE_FENCE.fullmatch(text)
+    return match.group("body") if match else None
+
+
 def parse_value(value, annotation):
     if annotation is str:
         return str(value)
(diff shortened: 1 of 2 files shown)


marcodelpin:
Closing this as superseded: the behaviour it fixes already works on current `main`.

While rebasing the branch onto `main` I re-checked the fix against upstream:

- the two regression tests this PR adds pass on plain `main` with the unchanged upstream `dspy/adapters/utils.py` (`pytest tests/adapters/test_adapter_utils.py -k fence` -> 2 passed);
- replacing the helper this PR introduces with a no-op leaves all 9 adapter-utils tests green, so the change is no longer distinguishable from its absence;
- calling upstream `parse_value` directly on the three fenced shapes from #8181 (` ```python {...} ``` `, ` ``` [1, None, 3] ``` `, ` ```json {"a": null} ``` `) returns the parsed value with `None` preserved in every case.

I have not traced which of the intervening ~97 commits repaired it, so I am not attributing it. If the two regression tests would still be useful on their own - they pin behaviour `main` now has but does not test under that name - I am happy to open them as a separate, tests-only PR.

Analysis and rebase verification were AI-assisted (Claude Code), same as disclosed in the PR description.

PR #10054, opened 2026-07-15
fix(react): ReAct.truncate_trajectory silently empties trajectory on single tool call

chuenchen309:
## Symptom

`ReAct.truncate_trajectory` documents (in its own comment and error message) that a trajectory holding exactly one tool call — 4 keys: `thought_i`, `tool_name_i`, `tool_args_i`, `observation_i` — cannot be truncated further and should raise `ValueError`. The guard is:

```python
keys = list(trajectory.keys())
if len(keys) < 4:
    # Every tool call has 4 keys: thought, tool_name, tool_args, and observation.
    raise ValueError(
        "The trajectory is too long so your prompt exceeded the context window, but the trajectory cannot be "
        "truncated because it only has one tool call."
    )

for key in keys[:4]:
    trajectory.pop(key)
```

`len(keys) < 4` is `False` when `len(keys) == 4` (the single-tool-call case), so instead of raising, the code falls through and pops all 4 keys, silently returning an **empty trajectory**.

This is reached from `_call_with_potential_trajectory_truncation`'s retry loop: after exactly one tool call has been recorded, if the LM keeps exceeding the context window on retries, the trajectory is silently wiped rather than the loop raising as documented, and the extraction step then runs against an empty trajectory instead of surfacing a clear error.

## Fix

Change the guard to `len(keys) <= 4`, matching the documented invariant.

## Verification

- Added `test_truncate_trajectory_raises_on_single_tool_call` in `tests/predict/test_react.py`, constructing a 4-key trajectory directly and asserting `truncate_trajectory` raises `ValueError`.
- Confirmed **red** against the unfixed guard (test failed with "DID NOT RAISE ValueError" — trajectory became `{}`).
- Confirmed **green** after the one-line fix.
- Full `tests/predict/` suite: 236 passed, 62 skipped (pre-existing, unrelated — no LM API key in this environment).
- `ruff check` clean on changed files.

## AI-Generated Contributions

Per CONTRIBUTING.md's AI-disclosure guidance: I used Claude (AI-assisted, human-in-the-loop) to help find and fix this bug. Claude located the off-by-one by reading `truncate_trajectory` against its own docstring/comment, wrote the failing test first, confirmed it red against the unfixed code, applied the minimal fix, and confirmed it green plus the full `tests/predict/` suite. I reviewed and verified the diagnosis and fix before submitting — this was not an unsupervised autonomous agent opening the PR.

### DIFF
--- a/dspy/predict/react.py
+++ b/dspy/predict/react.py
@@ -172,7 +172,7 @@ class ReAct(Module):
         Users can override this method to implement their own truncation logic.
         """
         keys = list(trajectory.keys())
-        if len(keys) < 4:
+        if len(keys) <= 4:
             # Every tool call has 4 keys: thought, tool_name, tool_args, and observation.
             raise ValueError(
                 "The trajectory is too long so your prompt exceeded the context window, but the trajectory cannot be "
--- a/tests/predict/test_react.py
+++ b/tests/predict/test_react.py
@@ -252,6 +252,23 @@ def test_trajectory_truncation():
     assert result.output_text == "Final output"
 
 
+def test_truncate_trajectory_raises_on_single_tool_call():
+    # A trajectory with exactly one tool call has 4 keys (thought, tool_name, tool_args,
+    # observation). Per truncate_trajectory's own docstring/error message, this is the
+    # smallest trajectory that cannot be truncated further, so it must raise instead of
+    # silently popping every key and returning an empty trajectory.
+    react = dspy.ReAct("input_text -> output_text", tools=[])
+    trajectory = {
+        "thought_0": "Thought 0",
+        "tool_name_0": "finish",
+        "tool_args_0": {},
+        "observation_0": "Completed.",
+    }
+
+    with pytest.raises(ValueError):
+        react.truncate_trajectory(trajectory)
+
+
 @pytest.mark.asyncio
 async def test_context_window_exceeded_after_retries():
     def echo(text: str) -> str:


ErenAta16 (commented):
Off-by-one is real, and it is worse than "returns an empty trajectory". I ran it rather than eyeballing the diff.

The part the description does not spell out is that `truncate_trajectory` mutates the caller's dict in place (`trajectory.pop(key)`), and `forward` hands that same object back as `dspy.Prediction(trajectory=trajectory, ...)`. So on `main` the user's returned trajectory is wiped, not just the copy used for the retry:

```
## main ##
returned   : {}  | caller's dict now: {}
## this branch ##
RAISED     : ValueError: The trajectory is too long so your prompt exceeded the context window ...
```

End to end is the more convincing demo. One completed tool call, then the LM starts raising `ContextWindowExceededError`, with an `extract` that succeeds once the prompt is small enough (which is exactly the situation truncation is supposed to rescue):

```
## main ##
RESULT     : 'Fallback output'
trajectory : {}
extract saw: ['']

## this branch ##
RAISED     : ValueError: The trajectory is too long so your prompt exceeded the context window,
             but the trajectory cannot be truncated because it only has one tool call.
```

On `main` the user gets a confident-looking answer that the extractor produced from a completely empty trajectory, and `result.trajectory` comes back `{}` even though `echo` actually ran. That is the silent-wrong-answer case, which is the one you least want. This branch turns it into the documented error.

Same result through `aforward`, so sync and async do not diverge:

```
## main (async) ##      RESULT: 'Fallback output' | trajectory: {}
## branch (async) ##    RAISED: ValueError: The trajectory is too long ...
```

The new test is not vacuous. Reverting only `dspy/predict/react.py` to `main` while keeping the branch's test file:

```
FAILED tests/predict/test_react.py::test_truncate_trajectory_raises_on_single_tool_call
E   Failed: DID NOT RAISE ValueError
1 failed, 8 passed, 1 skipped
```

And the whole file is green on the branch (`9 passed, 1 skipped`). I specifically checked `test_context_window_exceeded_after_retries`, since it is the one that could plausibly have been pinned to the buggy behavior. It starts from an empty trajectory, so it takes the `0 <= 4` path either way and is unaffected.

I also checked that a partially-built trajectory cannot exist at truncation time. In both `forward` and `aforward`, `observation_{idx}` is always assigned (the tool call is wrapped in try/except that writes the error string into the observation), so key counts at the truncation point are always multiples of 4 and `<= 4` never eats a partial call. `react_v2.py` has no truncation path at all, it just breaks on `ContextWindowExceededError`, so there is no sibling to fix.

One thing I would mention in passing, not a blocker and arguably pre-existing. When this fires from the react loop rather than from `extract`, `forward` catches it here:

```python
except ValueError as err:
    logger.warning(f"Ending the trajectory: Agent failed to select a valid tool: {_fmt_exc(err)}")
    break
```

so a context-window failure gets logged as a tool-selection failure. The user still ends up with the right exception from the `extract` call afterwards, so behavior is correct, but the intervening log line points at the wrong cause. The same broad `except ValueError` is what would swallow this if `extract` ever moved inside the loop. Also worth noting the error string says "it only has one tool call" for every trajectory with 4 or fewer keys, including the empty one, which was already true before this change.

Good catch, and the one-character diff is the right size for it.

Issue #10053, opened 2026-07-15
AvatarOptimizer worker threads drop DSPy context and callback lineage


isaacbmiller:
## Reproduction

`AvatarOptimizer.thread_safe_evaluator()` changes DSPy semantics when work crosses its private thread pool:

```python
import dspy
from dspy.teleprompt.avatar_optimizer import AvatarOptimizer
from dspy.utils.callback import ACTIVE_CALL_ID, BaseCallback

GLOBAL_LM = object()
LOCAL_LM = object()


class Actor(dspy.Module):
    def forward(self, value):
        return dspy.Prediction(saw_local_lm=dspy.settings.lm is LOCAL_LM)


class Recorder(BaseCallback):
    def __init__(self):
        self.parents = []

    def on_module_start(self, call_id, instance, inputs):
        if isinstance(instance, Actor):
            self.parents.append(ACTIVE_CALL_ID.get())


# Bypass the separately broken constructor; see #7931 and the note below.
optimizer = object.__new__(AvatarOptimizer)
optimizer.metric = lambda example, prediction: float(prediction.saw_local_lm)

example = dspy.Example(value=1).with_inputs("value")
recorder = Recorder()
dspy.configure(lm=GLOBAL_LM, callbacks=[recorder])

token = ACTIVE_CALL_ID.set("parent")
try:
    with dspy.context(lm=LOCAL_LM):
        score = optimizer.thread_safe_evaluator([example], Actor(), num_threads=2)
finally:
    ACTIVE_CALL_ID.reset(token)

print("score", score)
print("actor_parent_ids", recorder.parents)
```

Output on current `main`:

```text
score 0.0
actor_parent_ids [None]
```

Expected:

```text
score 1.0
actor_parent_ids ['parent']
```

The worker therefore sees neither the request-scoped LM nor the causal callback parent. Callbacks installed through `dspy.context(callbacks=[...])` are also absent in the worker entirely.

## Root cause

`dspy/teleprompt/avatar_optimizer.py` creates a raw `ThreadPoolExecutor` and submits `process_example` directly:

```python
with ThreadPoolExecutor(max_workers=num_threads) as executor:
    futures = [
        executor.submit(self.process_example, actor, example, return_outputs)
        for example in devset
    ]
```

[...]

PR #10060, opened 2026-07-16
fix(teleprompt): preserve DSPy context and callbacks in AvatarOptimizer workers

TLXyloph:
## Intent

Fix stanfordnlp/dspy issue #10053: AvatarOptimizer's thread_safe_evaluator uses a raw ThreadPoolExecutor, so dspy.context(lm=...) overrides and context-installed callbacks never reach worker threads. Route per-example execution through the context-aware ParallelExecutor so scoped LMs and callbacks work in workers, and open a PR upstream against stanfordnlp/dspy.

## What Changed

- `AvatarOptimizer.thread_safe_evaluator` now routes per-example execution through the context-aware `ParallelExecutor` instead of a raw `ThreadPoolExecutor`, so `dspy.context(lm=...)` overrides and context-installed callbacks propagate into worker threads (issue #10053). The executor is configured with `max_errors=len(devset) + 1` to preserve the legacy behavior of never aborting on a single bad example, and `straggler_limit=0` to keep Avatar's at-most-once contract (no duplicate actor/metric LM calls).
- The public `process_example` method is replaced by a closure that no longer swallows exceptions itself; failures now surface as `None` outcomes and are scored 0, with results zipped back against `devset` (`strict=True`) to guarantee ordering.
- Adds `tests/teleprompt/test_avatar_optimizer.py` covering context propagation into workers, devset result ordering, failing-example scoring, `max_errors` tolerance, callback firing in worker threads, and `ACTIVE_CALL_ID` ancestry.

## Risk Assessment

✅ Low: Well-bounded change swapping a raw ThreadPoolExecutor for the context-aware ParallelExecutor in a single method, with executor options chosen to preserve the prior swallow-errors and at-most-once semantics, and covered by new tests for context propagation, ordering, failure scoring, and callback transport.

## Testing

I validated the fix end-to-end rather than relying on unit tests alone: I wrote a user-level repro that supplies an LM only through `dspy.context(lm=...)` and ran it against both the base commit and the fixed code, producing a clean before/after — base scores 0.0 with every worker thread raising "no LM configured in this thread" and zero callback invocations, while the fixed code scores 1.0 with all 8 predictions from the scoped LM and the callback firing 8 times across 4 non-main worker threads, proving the override genuinely transits into workers. The 6 new tests pass, and the neighboring parallelizer, settings, and callback suites (40 tests total) show no regressions from the ParallelExecutor routing. This is a library-internal concurrency change with no rendered UI surface, so the reviewer-visible evidence is a CLI transcript rather than a screenshot. I restored the `uv.lock` churn that `uv run` introduced and confirmed the worktree is clean.

<details>
<summary>Evidence: Before/after repro transcript for issue #10053</summary>

<code>=== BEFORE (base 96bae53d — raw ThreadPoolExecutor) ===&#10;dspy.settings.lm before context: None&#10;no LM configured in this thread (x5)&#10;&#10;average score over 8 examples: 0.0&#10;per-example: 0, 0, 0, 0, 0, 0, 0, 0&#10;scoped-LM predictions: 0/8&#10;context callback fired: 0 times across 0 thread(s)&#10;RESULT: FAIL - issue #10053 reproduced (context lost in worker threads)&#10;&#10;=== AFTER (target f6b4db4e — ParallelExecutor) ===&#10;dspy.settings.lm before context: None&#10;Average Metric: 8.00 / 8 (100.0%): 100%|##########| 8/8 [00:00&lt;00:00, 8.69it/s]&#10;&#10;average score over 8 examples: 1.0&#10;per-example: 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0&#10;scoped-LM predictions: 8/8&#10;context callback fired: 8 times across 4 thread(s) (main thread among them: False)&#10;RESULT: PASS - scoped LM + callbacks reached workers</code>

```text
Repro: dspy issue #10053 — AvatarOptimizer loses dspy.context(lm=...) + callbacks in worker threads
Script: repro_10053.py  (8-example devset, num_threads=4, LM supplied ONLY via dspy.context)

=================== BEFORE (base 96bae53d — raw ThreadPoolExecutor) ===================
dspy.settings.lm before context: None
no LM configured in this thread
no LM configured in this thread
no LM configured in this thread
no LM configured in this thread
no LM configured in this thread

average score over 8 examples: 0.0
per-example: 0, 0, 0, 0, 0, 0, 0, 0
scoped-LM predictions: 0/8
context callback fired: 0 times across 0 thread(s)

RESULT: FAIL - issue #10053 reproduced (context lost in worker threads)

=================== AFTER (target f6b4db4e — ParallelExecutor) ===================
dspy.settings.lm before context: None
Average Metric: 8.00 / 8 (100.0%): 100%|##########| 8/8 [00:00<00:00,  8.69it/s]

average score over 8 examples: 1.0
per-example: 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0
scoped-LM predictions: 8/8
context callback fired: 8 times across 4 thread(s) (main thread among them: False)

RESULT: PASS - scoped LM + callbacks reached workers
```
</details>
<details>
<summary>Evidence: Repro script (user-level reproduction of issue #10053)</summary>

```text
"""End-user repro for dspy issue #10053.

A user scopes an LM with dspy.context(lm=...) -- never calling dspy.configure --
and evaluates an Avatar actor over a devset. The actor uses dspy.settings.lm, so
if the context override does not reach AvatarOptimizer's worker threads the actor
fails and every example scores 0.
"""
import threading
import dspy
from dspy.teleprompt.avatar_optimizer import AvatarOptimizer
from dspy.utils.callback import BaseCallback


class MyActor(dspy.Module):
    def forward(self, **kwargs):
        lm = dspy.settings.lm
        if lm is None:
            raise ValueError("no LM configured in this thread")
        return dspy.Prediction(answer="Paris")


class TraceCallback(BaseCallback):
    def __init__(self):
        self.threads = set()
        self.calls = 0
        self.lock = threading.Lock()

    def on_module_start(self, call_id, instance, inputs):
        with self.lock:
            self.calls += 1
            self.threads.add(threading.get_ident())


def metric(example, prediction):
    return 1.0 if getattr(prediction, "answer", None) == example.answer else 0.0


devset = [dspy.Example(question=f"Capital of France? #{i}", answer="Paris").with_inputs("question") for i in range(8)]

optimizer = AvatarOptimizer.__new__(AvatarOptimizer)   # skip __init__ (unrelated TypedPredictor issue #7931)
optimizer.metric = metric

cb = TraceCallback()
print("dspy.settings.lm before context:", dspy.settings.lm)
with dspy.context(lm=DummyLM([{"answer": "Paris"}]), callbacks=[cb]):
    avg, results = optimizer.thread_safe_evaluator(devset, MyActor(), return_outputs=True, num_threads=4)

print(f"\naverage score over {len(devset)} examples: {avg}")
print("per-example: " + ", ".join(f"{s}" for _, _, s in results))
print(f"scoped-LM predictions: {sum(1 for _, p, _ in results if p is not None)}/{len(devset)}")
print(f"context callback fired: {cb.calls} times across {len(cb.threads)} thread(s) "
      f"(main thread among them: {threading.main_thread().ident in cb.threads})")
print("\nRESULT:", "PASS - scoped LM + callbacks reached workers" if avg == 1.0 and cb.calls == len(devset)
      else "FAIL - issue #10053 reproduced (context lost in worker threads)")
```
</details>

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>🔧 **Review** - 4 issues found → auto-fixed (2) ✅</summary>

- ⚠️ `tests/teleprompt/test_avatar_optimizer.py:167` - The xfail reason is factually stale: it claims &#34;ParallelExecutor uses a bare executor.submit without contextvars.copy_context&#34;, but the base commit 96bae53d (#10043) already added `_bind_active_call_id`, which `ParallelExecutor.execute` calls (dspy/utils/parallelizer.py:49). ACTIVE_CALL_ID ancestry therefore does propagate, so this test most likely XPASSes. Because `strict=False`, an XPASS is silently reported as a pass and the assertion is never enforced — the test guards nothing. Either drop the xfail marker and let it be a real regression test for the ancestry transport, or delete it.
- ⚠️ `dspy/teleprompt/avatar_optimizer.py:92` - `process_example` was a public method on AvatarOptimizer and has been deleted (inlined as a closure). Any downstream code that subclassed AvatarOptimizer to override per-example handling, or called `optimizer.process_example(...)` directly, breaks silently with no deprecation path. Consider keeping a thin public `process_example(self, actor, example)` and having the closure delegate to it, so subclass overrides still take effect.
- ℹ️ `dspy/teleprompt/avatar_optimizer.py:99` - `max_errors=len(devset) + 1` hard-overrides `dspy.settings.max_errors` for this code path, so a user who deliberately sets a low max_errors to fail fast on a broken actor gets no abort — the run silently completes with a near-zero average instead. This does preserve the legacy swallow-everything behavior (and the comment says so), but it&#39;s now the one evaluator in the codebase that ignores the setting. Worth confirming this is the intended contract rather than honoring the setting.
- ℹ️ `dspy/teleprompt/avatar_optimizer.py:104` - `straggler_limit=0` disables resubmission via the `len(not_done) &lt;= self.straggler_limit` guard, which is a somewhat indirect way to express intent — the branch is only skipped because `not_done` is non-empty whenever the check runs. Passing `timeout=0` would disable the straggler path at its explicit guard (`if 0 &lt; self.timeout`) and read more directly. Current form is correct; this is purely clarity.

🔧 Fix: test(avatar): drop stale xfail from call-ancestry test
2 infos still open:
- ℹ️ `uv.lock:764` - The uv.lock change (dspy 3.2.1 → 3.3.0b1) is an incidental re-sync against pyproject.toml&#39;s version and is unrelated to the AvatarOptimizer fix. It&#39;s harmless locally but adds noise to an upstream PR against stanfordnlp/dspy — consider dropping it from the commit unless maintainers expect the lock to be kept in sync.
- ℹ️ `dspy/teleprompt/avatar_optimizer.py:108` - `process_example` returns `(example, prediction, score)` but the caller discards the first element (`_, prediction, score = outcome`) since it already has `example` from zipping over `devset`. Returning just `(prediction, score)` would be simpler; note ParallelExecutor&#39;s `compare_results` progress bar reads `r[-1]`, so score must remain last either way.

🔧 Fix: chore: drop unrelated uv.lock version churn from branch
✅ Re-checked - no issues remain.

</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- <code>`uv run pytest tests/teleprompt/test_avatar_optimizer.py -v` — 6 new tests pass (context propagation, devset ordering, failing-example scoring, max_errors tolerance, callback firing in workers, ACTIVE_CALL_ID ancestry)</code>
- <code>`uv run --frozen pytest tests/teleprompt/test_avatar_optimizer.py tests/utils/test_parallelizer.py tests/utils/test_settings.py tests/callback -q` — 40 passed, no regressions in the ParallelExecutor/settings/callback paths the fix depends on</code>
- <code>Manual end-to-end repro: authored `repro_10053.py` (Avatar actor reading `dspy.settings.lm`, 8-example devset, `num_threads=4`, LM supplied only via `dspy.context`) and ran it against the fixed code — avg 1.0, 8/8 scoped-LM predictions, callback fired 8x across 4 non-main threads</code>
- <code>Baseline comparison: temporarily restored `git show 96bae53d:dspy/teleprompt/avatar_optimizer.py` and re-ran the same repro — reproduced the bug (avg 0.0, 0/8 predictions, 0 callback invocations, workers raising &#39;no LM configured in this thread&#39;), then restored the fixed file</code>
- <code>`git status --porcelain` — confirmed clean worktree; restored `uv.lock` churn introduced by `uv run` and used `--frozen` for subsequent runs</code>

</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>

### DIFF
--- a/dspy/teleprompt/avatar_optimizer.py
+++ b/dspy/teleprompt/avatar_optimizer.py
@@ -1,14 +1,13 @@
-from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from random import sample
 from typing import Callable
 
 from pydantic import BaseModel
-from tqdm import tqdm
 
 import dspy
 from dspy.predict.avatar import ActionOutput
 from dspy.teleprompt.teleprompt import Teleprompter
+from dspy.utils.parallelizer import ParallelExecutor
 
 DEFAULT_MAX_EXAMPLES = 10
 
@@ -90,26 +89,18 @@ class AvatarOptimizer(Teleprompter):
         self.comparator = dspy.TypedPredictor(Comparator)
         self.feedback_instruction = dspy.Predict(FeedbackBasedInstruction)
 
-    def process_example(self, actor, example, return_outputs):
-        actor = deepcopy(actor)
+    def process_example(self, actor, example):
+        """Run a single example through a copy of ``actor`` and score it.
 
-        try:
-            prediction = actor(**example.inputs().toDict())
-            score = self.metric(example, prediction)
-
-            if return_outputs:
-                return example, prediction, score
-            else:
-                return score
-
-        except Exception as e:
-            print(e)
-
-            if return_outputs:
-                return example, None, 0
-            else:
-                return 0
-
+        This is the per-example extension point: ``thread_safe_evaluator`` calls it once per
+        example from inside a worker, so overriding it here changes how every example is
+        handled. Exceptions are intentionally left uncaught -- the caller's ParallelExecutor
+        records the failure and scores the example 0.
+        """
+        actor_clone = deepcopy(actor)
+        prediction = actor_clone(**example.inputs().toDict())
+        score = self.metric(example, prediction)
+        return example, prediction, score
 
     def thread_safe_evaluator(self, devset, actor, return_outputs=False, num_threads=None):
         total_score = 0
@@ -117,17 +108,26 @@ class AvatarOptimizer(Teleprompter):
         results = []
         num_threads = num_threads or dspy.settings.num_threads
 
-        with ThreadPoolExecutor(max_workers=num_threads) as executor:
-            futures = [executor.submit(self.process_example, actor, example, return_outputs) for example in devset]
+        executor = ParallelExecutor(
+            num_threads=num_threads,
+            # Never abort on per-example failures; failing examples are scored 0 (issue #10053).
+            max_errors=len(devset) + 1,
+            compare_results=True,
+            # Disable straggler resubmission to preserve Avatar's at-most-once contract:
+            # each actor/metric runs exactly once per example, with no duplicate LM calls.
+            straggler_limit=0,
+        )
 
-            for future in tqdm(futures, total=total_examples, desc="Processing examples"):
-                result = future.result()
-                if return_outputs:
-                    example, prediction, score = result
-                    total_score += score
-                    results.append((example, prediction, score))
-                else:
-                    total_score += result
+        outcomes = executor.execute(lambda example: self.process_example(actor, example), devset)
+
+        for example, outcome in zip(devset, outcomes, strict=True):
+            if outcome is None:
+                # The example failed or was cancelled; count it as a zero score.
+                results.append((example, None, 0))
+            else:
+                _, prediction, score = outcome
+                total_score += score
+                results.append((example, prediction, score))
 
         avg_metric = total_score / total_examples
 
--- a/tests/teleprompt/test_avatar_optimizer.py
+++ b/tests/teleprompt/test_avatar_optimizer.py
@@ -1,0 +1,229 @@
+"""Tests for AvatarOptimizer.thread_safe_evaluator.
+
+These tests isolate the parallel-evaluation path from AvatarOptimizer.__init__ (which
+currently constructs a dspy.TypedPredictor and is affected by the unrelated #7931). We
+build the optimizer via ``__new__`` and set ``self.metric`` directly, since
+``thread_safe_evaluator`` only depends on ``self.metric`` plus dspy settings.
+"""
+
+import threading
+
+import pytest
+
+import dspy
+from dspy.teleprompt.avatar_optimizer import AvatarOptimizer
+from dspy.utils.callback import ACTIVE_CALL_ID, BaseCallback
+from dspy.utils.dummies import DummyLM
+
+
+def _make_optimizer(metric):
+    optimizer = AvatarOptimizer.__new__(AvatarOptimizer)
+    optimizer.metric = metric
+    return optimizer
+
+
+def _devset(n):
+    return [dspy.Example(question=f"q{i}", answer="ok").with_inputs("question") for i in range(n)]
+
+
+class ContextReadingActor(dspy.Module):
+    """Actor that requires an LM to be visible in the thread that runs it.
+
+    This mirrors the real failure in issue #10053: the actor reads ``dspy.settings.lm``,
+    which is only supplied via a ``dspy.context`` override. If that override does not reach
+    the worker thread, the actor raises and the example scores 0.
+    """
+
+    def forward(self, **kwargs):
+        if dspy.settings.lm is None:
+            raise ValueError("no LM configured in this thread")
+        return dspy.Prediction(answer="ok")
+
+
+class FailingActor(dspy.Module):
+    """Actor that raises for a specific input and succeeds otherwise."""
+
+    def forward(self, question=None, **kwargs):
+        if question == "boom":
+            raise RuntimeError("boom")
+        return dspy.Prediction(answer="ok")
+
+
+class AlwaysFailingActor(dspy.Module):
+    """Actor that raises for every input."""
+
+    def forward(self, **kwargs):
+        raise RuntimeError("boom")
+
+
+def _match_metric(example, prediction):
+    return 1.0 if getattr(prediction, "answer", None) == "ok" else 0.0
+
+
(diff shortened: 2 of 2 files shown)


TLXyloph:
I believe I fixed Greptile's P1 in d861242d. It's a public method again and the executor calls it per example, so overrides drive every example and the context propagation still holds. Covered by test_process_example_is_the_per_example_override_point.

Two deliberate deviations from the original signature: no return_outputs, and it no longer swallows the exceptions — ParallelExecutor records the failure and the example still scores 0, as before. Happy to restore the exact arity if a researcher/maintainer wants.

TLXyloph:
Closing this — @isaacbmiller mentioned that Avatar/COPRO aren't a current focus and may be deprecated soon, so a fix here isn't worth maintainer review time. Thank you!

PR #10063, opened 2026-07-18
feat(retrievers): add ChromaDB retriever

katherineahn:
## What
Adds `ChromadbRM`, a `Retrieve` module that returns the top-k passages from a ChromaDB collection, mirroring `dspy/retrievers/weaviate_rm.py`.

## Details
`forward()` accepts a query or list of queries (empty strings are filtered out), runs `collection.query(query_texts=[query], n_results=k)` for each, and wraps every returned document as a `dotdict({"long_text": ...})` so results stay interchangeable with other DSPy retrievers. `chromadb` is an optional dependency, guarded by an import check with an install hint.

## Tests
`tests/retrievers/test_chromadb_rm.py` — mocked tests that run `forward()` against a fake ChromaDB client (no live instance): documents are wrapped as `long_text`, `k` is passed through as `n_results`, and empty queries are skipped.

### DIFF
--- a/dspy/retrievers/chromadb_rm.py
+++ b/dspy/retrievers/chromadb_rm.py
@@ -1,6 +1,5 @@
 import dspy
 from dspy.dsp.utils import dotdict
-from dspy.primitives.prediction import Prediction
 
 try:
     import chromadb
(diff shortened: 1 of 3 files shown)


katherineahn:
Addressed the missing-extra note: added a `chromadb` optional-dependency to `pyproject.toml` (mirroring the existing `weaviate` extra), and corrected the install hint to `pip install dspy[chromadb]` to match the current package name. Rebased onto latest `main`; tests pass and lint is clean.

PR #10066, opened 2026-07-19
fix(predict): stop majority() from dropping article-only answers (e.g. multiple-choice "A")

lntutor:
### What's broken

`dspy.majority` normalizes each candidate answer with `default_normalize`, which runs `normalize_text` (SQuAD-style normalization that strips the English articles `a`/`an`/`the`). `majority` treats a `None`/empty normalization as "ignore this completion", so any answer that is **entirely an article** normalizes to `""` and is silently dropped from the vote:

```python
from dspy import majority
majority([{"answer":"A"}, {"answer":"A"}, {"answer":"A"}, {"answer":"B"}]).completions[0]["answer"]
# 'B'   -- 'A' won 3-to-1 but all three 'A' votes were discarded
```

Multiple-choice QA with options `A`/`B`/`C`/`D` is an extremely common self-consistency use case, so this silently corrupts the majority winner. (`normalize_text("A") == ""`, so `default_normalize("A") is None`.)

### Fix

Fall back to the lowercased/stripped answer when full normalization yields `""`:

```python
return normalize_text(s) or (s.strip().lower() or None)
```

Article-only answers now count and still group correctly (`"A"` and `"a"` together), while genuinely empty/whitespace answers still map to `None` and stay ignored. Any answer that normalizes to non-empty is **unchanged** (`"The Eiffel Tower"` → `"eiffel tower"` as before).

### Tests

Adds `test_majority_does_not_drop_article_answers` and `test_majority_the_answer`. Existing `tests/predict/test_aggregation.py` (6) and `tests/evaluate/test_metrics.py` (3) still pass.

### DIFF
--- a/dspy/evaluate/metrics.py
+++ b/dspy/evaluate/metrics.py
@@ -84,18 +84,19 @@ def HotPotF1(prediction, answers_list):  # noqa: N802
     return max(hotpot_f1_score(prediction, ans) for ans in answers_list)
 
 
-def normalize_text(s):
+def normalize_text(s, remove_articles=True):
     """Normalize text for string and token comparisons.
 
     Steps:
         1) Unicode NFD normalization
         2) lowercasing
         3) punctuation removal
-        4) English article removal ("a", "an", "the")
+        4) Optional English article removal ("a", "an", "the")
         5) whitespace collapse
 
     Args:
         s (str): Input string.
+        remove_articles (bool): Whether to remove English articles.
 
     Returns:
         str: Normalized string.
@@ -107,7 +108,7 @@ def normalize_text(s):
     """
     s = unicodedata.normalize("NFD", s)
 
-    def remove_articles(text):
+    def _remove_articles(text):
         return re.sub(r"\b(a|an|the)\b", " ", text)
 
     def white_space_fix(text):
@@ -120,7 +121,10 @@ def normalize_text(s):
     def lower(text):
         return text.lower()
 
-    return white_space_fix(remove_articles(remove_punc(lower(s))))
+    normalized = remove_punc(lower(s))
+    if remove_articles:
+        normalized = _remove_articles(normalized)
+    return white_space_fix(normalized)
 
 
 def em_score(prediction, ground_truth):
--- a/dspy/predict/aggregation.py
+++ b/dspy/predict/aggregation.py
@@ -3,7 +3,11 @@ from dspy.primitives.prediction import Completions, Prediction
 
 
 def default_normalize(s):
-    return normalize_text(s) or None
+    # `normalize_text` strips English articles ("a", "an", "the"), so answers that consist
+    # solely of an article (e.g. the multiple-choice option "A", or "The") would normalize to
+    # an empty string and be discarded from the vote. Fall back to the lowercased, stripped
+    # answer so such answers still count, while genuinely empty answers still map to None.
+    return normalize_text(s) or (s.strip().lower() or None)
 
 
 def majority(prediction_or_completions, normalize=default_normalize, field=None):
--- a/dspy/predict/aggregation.py
+++ b/dspy/predict/aggregation.py
@@ -5,9 +5,9 @@ from dspy.primitives.prediction import Completions, Prediction
 def default_normalize(s):
     # `normalize_text` strips English articles ("a", "an", "the"), so answers that consist
     # solely of an article (e.g. the multiple-choice option "A", or "The") would normalize to
-    # an empty string and be discarded from the vote. Fall back to the lowercased, stripped
-    # answer so such answers still count, while genuinely empty answers still map to None.
-    return normalize_text(s) or (s.strip().lower() or None)
+    # an empty string and be discarded from the vote. Retain the same normalization pipeline
+    # without article removal so punctuation-only answers remain ignored.
+    return normalize_text(s) or normalize_text(s, remove_articles=False) or None
 
 
 def majority(prediction_or_completions, normalize=default_normalize, field=None):
--- a/tests/predict/test_aggregation.py
+++ b/tests/predict/test_aggregation.py
@@ -41,3 +41,17 @@ def test_majority_with_no_majority():
     completions = [{"answer": "2"}, {"answer": "3"}, {"answer": "4"}]
     result = majority(completions)
     assert result.completions[0]["answer"] == "2"  # The first completion is returned in case of a tie
+
+
+def test_majority_does_not_drop_article_answers():
+    # "A" normalizes to "" under SQuAD-style normalization (articles stripped);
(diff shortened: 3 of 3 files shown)


ErenAta16 (commented):
The underlying bug is real and I reproduced it, but I think the fallback as written trades one wrong answer for another. Details, with output.

First, confirming the diagnosis. `normalize_text` is `white_space_fix(remove_articles(remove_punc(lower(s))))`, so an article-only answer collapses to `""` and `majority` drops it:

```
3x 'A' vs 1x 'B'   main -> 'B'   this branch -> 'A'
```

And the two new tests are not vacuous. Reverting only `dspy/predict/aggregation.py` to `main`:

```
FAILED tests/predict/test_aggregation.py::test_majority_does_not_drop_article_answers
FAILED tests/predict/test_aggregation.py::test_majority_the_answer
2 failed, 6 passed
```

The problem is that `s.strip().lower()` is not the same normalizer minus article removal, it is *no* normalizer. It skips `remove_punc` and `white_space_fix` too, and that has two consequences.

**1. Junk answers now win votes they used to lose.** Anything that is only punctuation used to normalize to `""` and be ignored. Now it survives the fallback and counts:

```
case                              main        this branch
junk '...' x2 vs 'Paris'          'Paris'     '...'
junk '-'   x2 vs 'Paris'          'Paris'     '-'
```

Two refusals or dangling dashes out of three samples now beat the one real answer. Whitespace-only is still fine because `"  ".strip()` is empty, but `"..."`, `"-"`, `"?"` and friends are not.

**2. It does not actually fix multiple choice once formatting varies.** This is the use case the PR is aimed at, and models are not consistent about how they emit the letter:

```
normalize_text     this branch
'A'    -> ''       -> 'a'
'A.'   -> ''       -> 'a.'
'(A)'  -> ''       -> '(a)'
'A)'   -> ''       -> 'a)'
```

Four buckets for one answer. So:

```
case                                       main     this branch
mixed MC formatting: A, A., (A) vs B, B    'B'      'B'
```

A got three of five votes and still loses, because the three A votes are split across three distinct keys. Self-consistency sampling is exactly where you would expect that spread.

**What I would do instead.** Fall back to the same pipeline with only `remove_articles` skipped, rather than to raw `.strip().lower()`:

```python
def _normalize_keep_articles(s):
    s = unicodedata.normalize("NFD", s)
    s = "".join(ch for ch in s.lower() if ch not in set(string.punctuation))
    return " ".join(s.split())

def default_normalize(s):
    return normalize_text(s) or (_normalize_keep_articles(s) or None)
```

(Cleaner still would be a `normalize_text(s, remove_articles=True)` keyword in `dspy/evaluate/metrics.py` so the two paths cannot drift, but either shape works.)

I applied that on top of your branch and re-ran the same table:

```
case                                       as-shipped   punct-stripped fallback
3x 'A' vs 1x 'B'                           'A'          'A'
mixed MC formatting: A, A., (A) vs B, B    'B'          'A'
junk '...' x2 vs 'Paris'                   '...'        'Paris'
junk '-'   x2 vs 'Paris'                   '-'          'Paris'
blank x2 vs 'Paris'                        'Paris'      'Paris'
'The Eiffel Tower' x2 vs 'x'               ok           ok
```

All four of your intended cases still behave, the junk regressions go away, and the MC formatting variants collapse into one bucket. Your two new tests pass unchanged:

```
tests/predict/test_aggregation.py tests/evaluate/test_metrics.py
11 passed
```

(identical to the 11 passed on your branch as-is, so this is a strict improvement, not a trade).

One small thing in your favor that is worth stating: `default_normalize` is only referenced as the default argument of `majority` and nowhere else in the tree, so the blast radius is contained to self-consistency voting. Answers that normalize to something non-empty are untouched by either version.

If you take the alternative, `test_majority_does_not_drop_article_answers` is worth extending with the `A` / `A.` / `(A)` mix, since that is the case the current test cannot distinguish.

lntutor:
Thanks for the detailed reproduction. Addressed in 6c023fa:

- `normalize_text` now supports `remove_articles=False`, so the majority fallback reuses the same Unicode, lowercase, punctuation-removal, and whitespace-collapse pipeline while preserving article-only answers.
- Added regressions for mixed multiple-choice formatting (`A`, `A.`, `(A)` versus `B`) and punctuation-only answers (`...` versus `Paris`).

Verified locally with:

```text
uv run pytest tests/predict/test_aggregation.py tests/evaluate/test_metrics.py -q
13 passed
uv run ruff check dspy/predict/aggregation.py dspy/evaluate/metrics.py tests/predict/test_aggregation.py
uv run ruff format --check dspy/predict/aggregation.py dspy/evaluate/metrics.py tests/predict/test_aggregation.py
```

lntutor:
Correction: the fix commit is `9f88737e` (`fix(predict): preserve normalization for article answers`).

lntutor:
Rebased onto current `main`; the reviewer-requested normalization change remains intact. Reverified `tests/predict/test_aggregation.py` and `tests/evaluate/test_metrics.py` (13 passed), plus Ruff check and format verification. Fresh CI should now run against the updated branch.

ErenAta16 (commented):
Rebase looks clean and both points from my earlier review are properly closed. I re-ran the two counterexamples against the current diff rather than going off the description.

The `remove_articles` parameter is the right shape for this. Keeping the default `True` means `normalize_text` is unchanged for every existing caller (`em_score`, `hotpot_f1_score`, the metrics), and renaming the inner helper to `_remove_articles` avoids the parameter shadowing it, which would have been an easy thing to get wrong.

The fallback chain in `default_normalize` is what makes it work:

```python
return normalize_text(s) or normalize_text(s, remove_articles=False) or None
```

Full normalization first, then retry without article stripping only when the first pass emptied the string, then `None`. That keeps punctuation-only answers out of the vote while letting article-only answers back in, which is exactly the distinction that was missing.

Verified against both cases I raised:

```
multiple-choice formats splitting the vote:  ["A", "A.", "(A)", "B", "B"]
   main : ('b', 2)      <- A/A./(A) counted as three separate keys
   PR   : ('a', 3)      <- grouped, A wins

punctuation-only answers winning:            ["...", "...", "Paris"]
   main : ('paris', 1)
   PR   : ('paris', 1)  <- still correctly ignored

original reported bug:                       ["A", "A", "A", "B"]
   main : ('b', 1)      <- all three A votes discarded
   PR   : ('a', 3)
```

`default_normalize("A") == default_normalize("A.") == default_normalize("(A)")` confirms the grouping is real and not incidental.

Test coverage matches the behavior one-to-one, including `test_majority_normalizes_multiple_choice_punctuation` and `test_majority_ignores_punctuation_only_answers`, which are the two I'd have asked for. Having both directions asserted matters here, since a fix that just skipped normalization entirely would pass the article test and quietly fail the punctuation one.

Nothing further from me on this one.

PR #10070, opened 2026-07-20
fix: roll up nested track_usage() into enclosing tracker

### LINKED ISSUE #10064: [Bug] Nested `dspy.track_usage()` silently under-counts at the outer scope
### What happened?

### Steps to reproduce

```python
import dspy


MODEL = "gpt"


        )

```

(Assuming `dspy` installed)
```bash
python cost_bug.py
```

Output:
```bash
orchestrator billed:  105
actual workflow cost: 1605
```

### DSPy version

3.2.1

PiedPiper911:
## Summary

Fixes #10064.

When two `dspy.track_usage()` blocks are nested, the outer tracker previously did not include LM usage that occurred inside the inner block. Each LM call was recorded only by the innermost active tracker, so the outer scope silently under-counted.

This is a real-world scenario in multi-agent setups (coordinator + sub-agents both tracking cost) and training loops (top-level + per-epoch tracking).

## Root Cause

`track_usage()` sets a new `UsageTracker` via `settings.context(usage_tracker=tracker)`. When nested, the inner tracker replaces the outer one in settings. LM calls inside the inner block go only to the inner tracker. On exit, the outer tracker is restored but never received the inner block's usage.

## Fix

On exit of the inner `track_usage()` context, roll up all recorded usage entries into the parent tracker (if one exists). This ensures outer scopes observe the full workflow cost while inner scopes still report their own isolated usage.

## Changes

- `dspy/utils/usage_tracker.py`: Save reference to enclosing tracker before entering context; on exit, iterate inner tracker's `usage_data` and call `parent_tracker.add_usage()` for each entry.

## Testing

The fix was verified against the reproduction script from the issue:

```python
with track_usage() as orchestrator_cost:
    record_lm_call(100, 10)
    a = subagent(1000, 100)  # nested track_usage
    b = subagent(500, 50)
    record_lm_call(5, 1)

# Before: orchestrator billed 105
# After:  orchestrator billed 1605 (correct)
```

Backward compatible: non-nested usage of `track_usage()` is unaffected (parent_tracker is None).

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -67,8 +67,20 @@ class UsageTracker:
 
 @contextmanager
 def track_usage() -> Generator[UsageTracker, None, None]:
-    """Context manager for tracking LM usage."""
+    """Context manager for tracking LM usage.
+
+    When nested, the inner tracker rolls its recorded usage up into the
+    enclosing tracker on exit so that outer scopes observe the full cost
+    of the workflow (fixes #10064).
+    """
     tracker = UsageTracker()
+    parent_tracker = settings.usage_tracker  # may be None if not nested
 
     with settings.context(usage_tracker=tracker):
         yield tracker
+
+    # Roll up into the enclosing tracker so outer scopes see nested usage.
+    if parent_tracker is not None:
+        for lm, entries in tracker.usage_data.items():
+            for entry in entries:
+                parent_tracker.add_usage(lm, entry)
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -76,11 +76,13 @@ def track_usage() -> Generator[UsageTracker, None, None]:
     tracker = UsageTracker()
     parent_tracker = settings.usage_tracker  # may be None if not nested
 
-    with settings.context(usage_tracker=tracker):
-        yield tracker
-
-    # Roll up into the enclosing tracker so outer scopes see nested usage.
-    if parent_tracker is not None:
-        for lm, entries in tracker.usage_data.items():
-            for entry in entries:
-                parent_tracker.add_usage(lm, entry)
+    try:
+        with settings.context(usage_tracker=tracker):
+            yield tracker
+    finally:
+        # Roll up into the enclosing tracker so outer scopes see nested usage.
+        # This runs even if the nested block raises an exception.
+        if parent_tracker is not None:
+            for lm, entries in tracker.usage_data.items():
+                for entry in entries:
+                    parent_tracker.add_usage(lm, entry)


PiedPiper911:
Friendly ping - this fixes nested track_usage() not rolling up into the enclosing tracker. Happy to adjust if needed. Thanks!

ErenAta16 (commented):
The mechanism is right and this is the earliest of the four PRs on #10064, so it has priority. One gap worth closing before it lands: the roll-up isn't exception-safe.

```python
    yield tracker
if parent_tracker is not None:      # skipped when the body raises
    ...
```

In a `@contextmanager`, an exception from inside the `with track_usage()` body propagates through the `yield`, so anything after it never runs. Usage recorded before the failure is dropped from the parent's total. Simulated on an inner block that spends 500 tokens and then raises, with the exception caught outside:

```
this shape  : outer billed 105   <- the 500 is lost
try/finally : outer billed 605
```

Wrapping it fixes it without changing anything else:

```python
try:
        yield tracker
finally:
        ...
```

This is worth closing because a failed LM call is precisely when you want the cost recorded: the tokens were spent and the provider bills for them, so a retry loop that swallows the exception would under-count on every failed attempt, which is the same class of silent under-counting the issue is about.

The other thing reviewers will likely ask for is a regression test, since this is the only one of the four without one. #10085 and #10096 both have a nested-scope test you could adapt, and an exception-path test would pin the fix above.

Disclosure: I have a competing PR (#10086) on this issue, opened without checking for existing ones first, which was my mistake given yours predates it by five days. I've put a full four-way comparison on #10064 and I'd rather this one land than mine.

PiedPiper911:
Hi! Just checking in on this PR — it fixes nested track_usage() calls to properly roll up into the enclosing tracker. Let me know if any changes are needed. Thanks!

ErenAta16 (commented):
Nothing further needed from me. The gap I raised is closed by `0eb46c52`, and I re-ran the case rather than reading the diff.

Inner block spends 500 tokens and then raises, outer has already spent 100:

```
before (no try/finally)    outer total = 100    (expected 600)
this PR (try/finally)      outer total = 600    (expected 600)

clean path, unchanged:
before                     outer total = 600
this PR                    outer total = 600
```

So the roll-up now survives the exception and the non-failing path is untouched, which is the pair that matters.

Two details that are right and worth having on the record rather than left implicit:

**`parent_tracker` is read before the context is entered.** `settings.usage_tracker` is captured while it still refers to the enclosing tracker, so the roll-up targets the parent rather than the tracker this call is about to install. Reading it one line later would make the whole thing a no-op that still looks correct.

**Usage from a failed block counts toward the parent.** That is a decision, not a side effect of using `finally`, and I think it is the right one: the tokens were spent whether or not the block completed, and a cost accounting that only counts successful work will understate every run that retries. Worth keeping the comment that says the roll-up runs on the exception path, because someone optimising later will read `finally` as defensive rather than deliberate.

One small thing, non-blocking: the roll-up is now on the exception path, so anything that raises inside it would replace the in-flight exception and hide the real failure. `add_usage` is a `setdefault` plus an append, so it will not, but it does mean that block should stay simple. If it ever grows a serialisation step or a callback, it wants its own guard.

@PiedPiper911 this reads as ready to me. It is also the earliest of the PRs on #10064, which I mentioned in my first pass and still think should count for something when picking which one lands.

PR #10075, opened 2026-07-23
fix(teleprompt): propagate DSPy context into AvatarOptimizer worker threads

### LINKED ISSUE #10053: AvatarOptimizer worker threads drop DSPy context and callback lineage

Goutham-Annem:
## Summary

Fixes context propagation in `AvatarOptimizer` when using `ThreadPoolExecutor`. Python's `concurrent.futures` does not automatically copy `contextvars` state to worker threads, so DSPy context (active LM, settings) was silently unavailable to workers.

Fixes #10053

## Changes

- Use `contextvars.copy_context()` to capture context before submitting to the thread pool
- Each worker runs inside `ctx.run(...)` so it sees the same DSPy context as the calling thread

## Testing

Existing test suite covers optimizer correctness. This is a thread-safety fix with no behavioral change when running single-threaded.

### DIFF
--- a/dspy/teleprompt/avatar_optimizer.py
+++ b/dspy/teleprompt/avatar_optimizer.py
@@ -1,3 +1,4 @@
+import contextvars
 from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from random import sample
@@ -118,7 +119,8 @@ class AvatarOptimizer(Teleprompter):
         num_threads = num_threads or dspy.settings.num_threads
 
         with ThreadPoolExecutor(max_workers=num_threads) as executor:
-            futures = [executor.submit(self.process_example, actor, example, return_outputs) for example in devset]
+            ctx = contextvars.copy_context()
+            futures = [executor.submit(ctx.run, self.process_example, actor, example, return_outputs) for example in devset]
 
             for future in tqdm(futures, total=total_examples, desc="Processing examples"):
                 result = future.result()
--- a/dspy/teleprompt/avatar_optimizer.py
+++ b/dspy/teleprompt/avatar_optimizer.py
@@ -119,8 +119,7 @@ class AvatarOptimizer(Teleprompter):
         num_threads = num_threads or dspy.settings.num_threads
 
         with ThreadPoolExecutor(max_workers=num_threads) as executor:
-            ctx = contextvars.copy_context()
-            futures = [executor.submit(ctx.run, self.process_example, actor, example, return_outputs) for example in devset]
+            futures = [executor.submit(contextvars.copy_context().run, self.process_example, actor, example, return_outputs) for example in devset]
 
             for future in tqdm(futures, total=total_examples, desc="Processing examples"):
                 result = future.result()


ErenAta16 (commented):
One `Context` object can't be entered twice at the same time, so sharing `ctx` across the futures raises as soon as two of them overlap.

```python
ctx = contextvars.copy_context()
futures = [executor.submit(ctx.run, self.process_example, actor, example, return_outputs) for example in devset]
```

Three tasks, three worker threads, each task sleeping 20 ms so they actually overlap:

| | results |
|---|---|
| shared `ctx` (this PR) | `['OUTER', 'RuntimeError: cannot enter context: <Context object at 0x...> is already entered', 'RuntimeError: ...']` |
| `copy_context()` per task | `['OUTER', 'OUTER', 'OUTER']` |

The first task enters the context and the rest hit `RuntimeError`. It only looks fine with a single task or when tasks are short enough not to overlap — my first attempt at this passed for exactly that reason, which is also why a test that submits two trivial callables wouldn't catch it.

`dspy/predict/rlm.py:313` already has the correct form in this repo:

```python
executor.submit(contextvars.copy_context().run, _query_lm, prompt)
```

so moving the `copy_context()` inside the comprehension matches existing usage and fixes it:

```python
futures = [
    executor.submit(contextvars.copy_context().run, self.process_example, actor, example, return_outputs)
    for example in devset
]
```

Worth confirming `thread_safe_evaluator` is reachable with `num_threads > 1` in the test suite — `num_threads = num_threads or dspy.settings.num_threads`, so a default of 1 in the test environment would hide this.

Reproduced standalone with `contextvars` + `ThreadPoolExecutor` on Python 3.10.12; the `RuntimeError` comes from CPython's `Context.run`, not from anything dspy-specific.

Goutham-Annem:
Friendly ping — CI is passing and this PR has been open for 18 days. Would appreciate a review when you get a chance!

PR #10065, opened 2026-07-18
Proposed fix of Issue #10064, rolling up nested cost trackers to the parent on closing

asparagus:
# Description
- Rolling up nested cost trackers to the parent on closing.
- Prevent string values on the usage tracker from being concatenated. e.g:
  ```python
  {
      ...
      "service_tier": "standardstandardstandardstandardstandardstandardstandard",
      "inference_geo": "ususususususususususususus",
  }
  ```
- Fix bug in `_merge_usage_entries` where `dict(usage_entry2)` could be returned without performing a `None` check

# Context
Resolves https://github.com/stanfordnlp/dspy/issues/10064

# Caveats
- Are any users relying on the current behavior, or already accounting for it?
- Handling of non-numeric values: the proposed implementation will keep the first value that `is not None` and drop others

# Testing
Added 4 tests all of which fail under current `main` code:
- `tests/utils/test_usage_tracker.py::test_non_numeric_usage_values_are_not_accumulated`
- `tests/utils/test_usage_tracker.py::test_non_numeric_usage_value_present_in_only_some_entries`
- `tests/utils/test_usage_tracker.py::test_nested_usage_trackers_roll_up`
- `tests/utils/test_usage_tracker.py::test_nested_usage_trackers_roll_up_despite_errors`

Added an additional test to document intended `asyncio` usage:
- `tests/utils/test_usage_tracker.py::test_nested_usage_trackers_support_async_work`

See errors on main below:
```
=================================================================== short test summary info ====================================================================
FAILED tests/utils/test_usage_tracker.py::test_non_numeric_usage_values_are_not_accumulated - AssertionError: assert {'prompt_tokens': 30, 'completion_tokens': 6, 'service_tier': 'standardstandardstandard', 'inference_geo': 'ususus', 'iterations': None, 'is_byok': 0} == {'prompt_tokens': 30, 'completion_tokens': 6, 'service_tier': 'standard', 'inference_geo': 'us', 'iterations': None, 'is_byok': False}

  Common items:
  {'completion_tokens': 6, 'is_byok': 0, 'iterations': None, 'prompt_tokens': 30}
  Differing items:
  {'inference_geo': 'ususus'} != {'inference_geo': 'us'}
  {'service_tier': 'standardstandardstandard'} != {'service_tier': 'standard'}

  Full diff:
    {
        'prompt_tokens': 30,
        'completion_tokens': 6,
  -     'service_tier': 'standard',
  +     'service_tier': 'standardstandardstandard',
[...]

PR #10086, opened 2026-07-25
fix(usage): roll nested track_usage() up into the enclosing tracker

### LINKED ISSUE #10064: [Bug] Nested `dspy.track_usage()` silently under-counts at the outer scope

ErenAta16:
## What

Fixes #10064.

`track_usage()` records each LM call against the innermost active tracker only. When two blocks are nested, the inner `settings.context(usage_tracker=...)` shadows the outer tracker for the duration of the inner scope, so LM calls made inside the inner block never reach the outer one. The outer block silently under-counts.

This shows up whenever cost tracking exists at two levels, which the issue lays out well: a coordinator tracking its own usage while spawning sub-agents that track theirs, or a top-level training-cost tracker with per-epoch tracking underneath.

## Reproduction

From the issue, recording usage the same way `dspy/clients/lm.py` does on each call (no live LM needed):

```python
with track_usage() as orchestrator:
    record_lm_call(100, 10)
    a = subagent(1000, 100)   # each subagent opens its own track_usage() block
    b = subagent(500, 50)
    record_lm_call(5, 1)
```

```
before:  orchestrator billed:  105     # only the two direct calls
         actual workflow cost: 1605
after:   orchestrator billed:  1605
```

## Fix

Capture the tracker already active in an enclosing block before swapping in the new one, then on exit roll this tracker's usage up into it:

```python
parent_tracker = settings.usage_tracker
try:
        yield tracker
finally:
        for lm, usage_entries in tracker.usage_data.items():
            for usage_entry in usage_entries:
                parent_tracker.add_usage(lm, usage_entry)
```

The roll-up runs in `finally`, so usage incurred before an exception inside the inner scope still counts (those tokens were spent regardless of how the block exited). A top-level block has no parent, so nothing rolls up and its behavior is unchanged. Each inner block still reports exactly its own usage while it is open, so there is no double counting within a scope.

This matches the roll-up-on-exit approach the issue proposed.

## Tests

Added three regression tests to `tests/utils/test_usage_tracker.py`:

- `test_nested_track_usage_rolls_up_into_outer_tracker` (the main case, checks both prompt and completion tokens)
- `test_nested_track_usage_rolls_up_on_exception`
- `test_top_level_track_usage_has_no_parent_to_roll_up_into`

The first two fail on current `main` and pass with the fix; the third is the no-op case and passes either way (I confirmed the first two go red when only the source change is reverted). Existing usage-tracker tests still pass: 7 passed, 1 skipped.

I also checked three-level nesting by hand (each level correctly sees the innermost cost) and that a solo top-level block still records its own usage once without erroring.

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -70,5 +70,19 @@ def track_usage() -> Generator[UsageTracker, None, None]:
     """Context manager for tracking LM usage."""
     tracker = UsageTracker()
 
-    with settings.context(usage_tracker=tracker):
-        yield tracker
+    # Capture any tracker already active in an enclosing `track_usage()` block.
+    # While this block is open it shadows the enclosing one in settings, so each
+    # LM call is recorded only by the innermost active tracker. Rolling this
+    # tracker's usage up into the parent on exit keeps an outer block reflecting
+    # everything that happened inside it, including work done in nested blocks
+    # (e.g. sub-agents or per-epoch tracking under a top-level cost tracker).
+    parent_tracker = settings.usage_tracker
+
+    try:
+        with settings.context(usage_tracker=tracker):
+            yield tracker
+    finally:
+        if parent_tracker is not None:
+            for lm, usage_entries in tracker.usage_data.items():
+                for usage_entry in usage_entries:
+                    parent_tracker.add_usage(lm, usage_entry)
--- a/tests/utils/test_usage_tracker.py
+++ b/tests/utils/test_usage_tracker.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel
 
 import dspy
 from dspy.utils.usage_tracker import UsageTracker, track_usage
+from dspy.dsp.utils.settings import settings
 
 
 def test_add_usage_entry():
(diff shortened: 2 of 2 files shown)


ErenAta16:
Closing as a duplicate.

#10065 predates this by a week and is the better fix. Beyond the nested roll-up, it also handles a case none of the rest of us covered: `_merge_usage_entries` concatenates non-numeric values, so `service_tier` accumulates as `'standardstandardstandard'` and a `False` flag tallies to `0`. I verified that on `main`.

That matters here specifically rather than being an unrelated bonus. Rolling nested entries up into the parent means the parent holds more entries per LM, so `get_total_tokens()` folds them together more times and the concatenation gets *more* visible. Landing a roll-up without the `_is_summable` guard would make that symptom worse, so #10065 is closer to a prerequisite than an alternative.

Full comparison and evidence on #10064. Apologies to @asparagus for missing it when I opened this.

PR #10087, opened 2026-07-26
fix(teleprompt): make COPRO.compile eval_kwargs optional

### LINKED ISSUE #9080: [Bug] COPRO implementation does not have optional `eval_kwargs`
### What happened?

# Issue: COPRO `eval_kwargs` Parameter Not Actually Optional

## Problem Description
The COPRO optimizer documentation at https://dspy.ai/api/optimizers/COPRO/ indicates that `eval_kwargs` is an optional parameter, but in practice it is required. This creates a discrepancy between the documented API and the actual implementation.

## Current Behavior
- `eval_kwargs` must be provided or the code will fail
- **Workaround**: Pass an empty dictionary (`{}`)

### Option 1: Update Implementation (Recommended)
Make `eval_kwargs` truly optional by setting a default value:

```python
def __init__(self, ..., eval_kwargs=None):
    self.eval_kwargs = eval_kwargs or {}
```

### Steps to reproduce

A simple example with prompt auto-tuning, using COPRO should be enough.

### DSPy version

3.0.4

katherineahn:
Fixes #9080.

`COPRO.compile` documents `eval_kwargs` as optional, but the signature marked it as a required keyword-only argument, so `compile(student, trainset=...)` raised `TypeError: missing 1 required keyword-only argument: 'eval_kwargs'`.

This defaults `eval_kwargs` to `None` and normalizes it with `eval_kwargs = eval_kwargs or {}` at the top of `compile()`.

`eval_kwargs` is splatted in **two** places, not one: `Evaluate(devset=trainset, metric=self.metric, **eval_kwargs)` at `copro_optimizer.py:138`, and `evaluate(module_clone, devset=trainset, **eval_kwargs)` at `:227` inside the depth/candidate loop. The normalization rebinds the local before either is reached, so both call sites are covered and an empty dict is a no-op for both.

The docstring now states the default: `eval_kwargs: optional, dict, defaults to None (no extra kwargs)`.

Adds `test_compile_without_eval_kwargs`, which calls `compile()` without `eval_kwargs` (fails with TypeError on current main, passes with this change).

Thanks @TomeHirata for confirming this was welcome in the issue thread.

### DIFF
--- a/dspy/teleprompt/copro_optimizer.py
+++ b/dspy/teleprompt/copro_optimizer.py
@@ -121,7 +121,7 @@ class COPRO(Teleprompter):
         assert hasattr(predictor, "signature")
         predictor.signature = updated_signature
 
-    def compile(self, student, *, trainset, eval_kwargs):
+    def compile(self, student, *, trainset, eval_kwargs=None):
         """
         optimizes `signature` of `student` program - note that it may be zero-shot or already pre-optimized (demos already chosen - `demos != []`)
 
@@ -133,6 +133,7 @@ class COPRO(Teleprompter):
 
         Returns optimized version of `student`.
         """
+        eval_kwargs = eval_kwargs or {}
         module = student.deepcopy()
         evaluate = Evaluate(devset=trainset, metric=self.metric, **eval_kwargs)
         total_calls = 0
--- a/dspy/teleprompt/copro_optimizer.py
+++ b/dspy/teleprompt/copro_optimizer.py
@@ -128,7 +128,7 @@ class COPRO(Teleprompter):
         parameters:
         student: program to optimize and left modified.
         trainset: iterable of `Example`s
-        eval_kwargs: optional, dict
+        eval_kwargs: optional, dict, defaults to None (no extra kwargs)
            Additional keywords to go into `Evaluate` for the metric.
 
         Returns optimized version of `student`.
--- a/tests/teleprompt/test_copro_optimizer.py
+++ b/tests/teleprompt/test_copro_optimizer.py
@@ -65,6 +65,28 @@ def test_signature_optimizer_optimization_process():
     # such as checking the instructions of the optimized student's predictors.
 
 
+def test_compile_without_eval_kwargs():
+    # `eval_kwargs` is documented as optional, so compile() must run when it is omitted.
+    # Regression test for #9080 (previously raised TypeError: missing keyword-only argument).
+    optimizer = COPRO(metric=simple_metric, breadth=2, depth=1, init_temperature=1.4)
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
The docstring already says `eval_kwargs: optional, dict`, and `COPRO` is the only optimizer in `dspy.teleprompt` that requires it:

| optimizer | required `compile` args |
|---|---|
| `BootstrapFewShot` | `student`, `trainset` |
| `MIPROv2` | `student`, `trainset` |
| `BootstrapFinetune` | `student`, `trainset` |
| **`COPRO`** | `student`, `trainset`, **`eval_kwargs`** |

Before: `COPRO.compile() missing 1 required keyword-only argument: 'eval_kwargs'`. After: signature is `(self, student, *, trainset, eval_kwargs=None)` and `tests/teleprompt/test_copro_optimizer.py` is 6 passed.

`eval_kwargs` is splatted in two places, not one — `evaluate = Evaluate(..., **eval_kwargs)` at line 137 and `evaluate(module_clone, devset=trainset, **eval_kwargs)` at line 226 (inside the depth/candidate loop). The `eval_kwargs = eval_kwargs or {}` at the top of `compile` covers both since it rebinds the local before either is reached, so the fix is complete — worth stating in the PR body, because "one-line default" reads like it only guards the first use.

Two notes on the shape:

`eval_kwargs or {}` also swallows an explicitly-passed `{}`, which is harmless here (both are empty) but differs from `if eval_kwargs is None`. Fine either way; mentioning it only because the same idiom bites when the falsy value is meaningful.

The docstring says optional but doesn't say what the default is. Since the signature now carries `None`, a `defaults to None (no extra kwargs)` on that line would close the loop.

Measured on `dspy` @ 67bf5eb with the patch applied, Python 3.10.12.

katherineahn:
Thanks for the detailed review, and for running it rather than eyeballing it.

Took the docstring suggestion. The parameter line now reads `eval_kwargs: optional, dict, defaults to None (no extra kwargs)`, so both the default and its effect are stated. 6 copro tests pass, ruff clean.

On the two splat sites, agreed that's the part worth being explicit about. I've updated the PR body with both line references (`:138` for the `Evaluate` construction, `:227` inside the candidate loop) so it's clear the single normalization covers both.

On `or {}` vs `is None`: you're right that they diverge when the falsy value carries meaning. Here an explicitly-passed `{}` and the substituted `{}` are indistinguishable to `**` unpacking, so I kept `or {}` to hold the diff to one line. Happy to switch to the stricter form if a maintainer would rather have it.

isaacbmiller:
lgtm. Thanks for the PR!

PR #10088, opened 2026-07-27
fix(usage_tracker): roll nested track_usage() up into the outer tracker

### LINKED ISSUE #10064: [Bug] Nested `dspy.track_usage()` silently under-counts at the outer scope

cardoner1993:
## 📝 Changes Description

Fixes #10064.

Each LM call is recorded only by the innermost active tracker: `track_usage()` replaces
`settings.usage_tracker` for the duration of its block and discards the child on exit. Nested blocks
therefore make the outer scope silently under-count — the repro in the issue bills 105 tokens for a
workflow that actually cost 1605.

This implements the roll-up the issue proposes:

- **`UsageTracker.merge(other)`** — absorbs every usage entry from another tracker, per LM key.
- **`track_usage()`** captures the enclosing tracker before entering its own context and merges the
  child into it on exit.

Design notes, since these are the likely review questions:

- **The merge is entry-level, not total-level.** `get_total_tokens()` aggregation and the per-call
  granularity of `usage_data` are unchanged, so the pydantic/nested-dict flattening logic keeps a single
  code path.
- **It runs in a `finally`.** Tokens spent before an exception were still spent.
- **It is skipped when there is no parent.** The paths `BaseLM` deliberately runs under
  `settings.context(usage_tracker=None)` stay unbilled.
- **`ParallelExecutor` is unaffected.** It deep-copies the tracker into each worker
  (`dspy/utils/parallelizer.py`) instead of going through `track_usage()`, so workers remain independent
  of the parent — the existing `test_parallel_executor_with_usage_tracker` assertion that the parent stays
  empty still holds.
- **`Module.forward` is unaffected.** It only opens a tracker when none is active, so it never nests.

### Tests

Four tests added to `tests/utils/test_usage_tracker.py`:

| Test | Asserts |
|---|---|
| `test_nested_track_usage_rolls_up_to_outer_tracker` | Outer sees all three calls; the inner block still reports only its own usage |
| `test_nested_track_usage_rolls_up_multiple_models_and_levels` | Roll-up is per-model and survives three levels of nesting |
| `test_nested_track_usage_rolls_up_on_exception` | Tokens spent before a raise still bill the outer scope |
| `test_track_usage_restores_outer_tracker_as_active` | Leaving a nested block restores the enclosing tracker in `settings` |

The first three fail on `main` and pass with this change; the fourth guards the context stack against
regression.

## ✅ Contributor Checklist

## ⚠️ Warnings

Nothing behaviour-changing beyond the roll-up itself. Anyone who nests `track_usage()` and relies on the
outer tracker excluding inner blocks would see higher outer totals — but that is the bug being fixed.

## 🤖 AI assistance disclosure

Added after the fact, per `CONTRIBUTING.md` — apologies for the omission on the original submission.

Tool: Claude Code (Opus). Used to sweep the repo's open issues, to draft the reproduction and the patch,
and for test scaffolding. I reviewed and ran everything myself and can explain every line.

Prompt trail: a request to find a contributable issue in this repo, then to reproduce #10064 offline and
work out where the outer-scope total is lost. The diagnosis — that `UsageTracker` never rolls inner-scope
usage back up to the outer tracker — I confirmed by reading `dspy/utils/usage_tracker.py` and by watching
the assertion fail on `main` before writing the fix.

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -54,6 +54,15 @@ class UsageTracker:
         if len(usage_entry) > 0:
             self.usage_data[lm].append(self._flatten_usage_entry(usage_entry))
 
+    def merge(self, other: "UsageTracker") -> None:
+        """Absorb every usage entry recorded by another tracker.
+
+        Used to roll a nested `track_usage()` block's usage up into the enclosing
+        tracker, so that an outer block reflects everything that happened inside it.
+        """
+        for lm, usage_entries in other.usage_data.items():
+            self.usage_data[lm].extend(usage_entries)
+
     def get_total_tokens(self) -> dict[str, dict[str, Any]]:
         """Calculate total tokens from all tracked usage."""
         total_usage_by_lm = {}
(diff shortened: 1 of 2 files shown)


cardoner1993:
Already to many people with solutions for the issue.

PR #10094, opened 2026-07-28
fix(tool): exclude args with default values from required in native function-calling schema

### LINKED ISSUE #9882: [Bug] Tool args with defaults are emitted as `required` in native function-calling schema

SpiliosDimakopoulos:
Fixes #9882

## Problem
Tool.format_as_litellm_function_call() listed every argument in the
schema's "required" list, even arguments that already had a "default"
key in their own schema entry:

    "required": list(self.args.keys())

For example, `def search(query: str, k: int = 3)` produced a schema
where `k` had `{"type": "integer", "default": 3}` in "properties", but
was still listed in "required" -- contradicting the schema itself and
never actually telling the LM it could omit `k`.

## Fix
"required" now only includes args that don't have a "default" key,
matching standard JSON Schema / OpenAI function-calling conventions:

    required = [name for name, schema in self.args.items() if "default" not in schema]

## Testing
Added a regression test with a defaulted arg. It fails against
unmodified dspy (produces ["x", "y"] instead of ["x"]) and passes
after this fix (verified against a separate, untouched checkout).

Found 2 existing tests (test_chat_adapter.py, test_json_adapter.py)
that had hardcoded the old, incorrect behavior ("required": ["query",
"k"] where k has a default) as their expected output -- updated both
to reflect the corrected behavior.

Ran the full tests/adapters/ suite (231 tests, all pass) plus
tests/predict/ and tests/primitives/ (296 tests, all pass) for extra
confidence, since Tool is used broadly (ReAct, native tool calling,
etc.). ruff check on the changed source file: clean.

### DIFF
--- a/dspy/adapters/types/tool.py
+++ b/dspy/adapters/types/tool.py
@@ -150,6 +150,9 @@ class Tool(Type):
         return str(self)
 
     def format_as_litellm_function_call(self):
+        # Args with a default value are optional: they shouldn't be listed as "required",
+        # even though they're still advertised (with their default) in "properties".
+        required = [name for name, schema in self.args.items() if "default" not in schema]
         return {
             "type": "function",
             "function": {
@@ -158,7 +161,7 @@ class Tool(Type):
                 "parameters": {
                     "type": "object",
                     "properties": self.args,
-                    "required": list(self.args.keys()),
+                    "required": required,
                 },
             },
         }
--- a/tests/adapters/test_chat_adapter.py
+++ b/tests/adapters/test_chat_adapter.py
@@ -1165,7 +1165,7 @@ def test_chat_adapter_format_exact_messages_and_lm_kwargs_with_native_tool_calli
                              "parameters": {"type": "object",
                                             "properties": {"query": {"type": "string"},
                                                            "k": {"type": "integer", "default": 3}},
-                                            "required": ["query", "k"]}}}]}
+                                            "required": ["query"]}}}]}
     assert lm_kwargs == expected_lm_kwargs
 
 
--- a/tests/adapters/test_json_adapter.py
+++ b/tests/adapters/test_json_adapter.py
@@ -604,7 +604,7 @@ def test_json_adapter_format_exact_messages_and_lm_kwargs_with_native_tool_calli
                              "parameters": {"type": "object",
                                             "properties": {"query": {"type": "string"},
                                                            "k": {"type": "integer", "default": 3}},
-                                            "required": ["query", "k"]}}}]}
+                                            "required": ["query"]}}}]}
     assert lm_kwargs == expected_lm_kwargs
 
 def test_json_adapter_format_exact_messages_with_tool_calls_output_demo():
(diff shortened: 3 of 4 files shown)


ErenAta16 (commented):
Confirmed the self-contradiction on current `main`:

```python
def search(query: str, k: int = 3): ...

properties: {'query': {'type': 'string'}, 'k': {'type': 'integer', 'default': 3}}
required  : ['query', 'k']
```

`k` carries a `default` in `properties` while simultaneously being listed as `required`, which is the schema telling the model two opposite things about the same argument. Providers generally treat `required` as authoritative, so the default was advertised but unusable, and a tool author writing `k: int = 3` got no optionality at all.

Keying off `"default" not in schema` rather than re-deriving from the function signature is the right call: `self.args` is already the built schema, so the two views stay consistent by construction. If a `default` key is present the argument is optional by definition, so I can't construct a case where this drops something that should have stayed required.

Updating the two adapter tests is correct rather than a compromise. Those assertions were encoding the buggy schema (`"required": ["query", "k"]`), so the fix couldn't land without changing them. Worth flagging for reviewers that those diffs are the point, not collateral damage.

The new test asserts both directions (`required == ["x"]` and `"y" not in required`) while still checking `properties` keeps both args, which is exactly the right distinction here: the fix should change what's required, not what's advertised.

Looks correct to me.

PR #10095, opened 2026-07-28
fix(streaming): propagate exceptions from apply_sync_streaming instead of swallowing them

SpiliosDimakopoulos:
Fixes #9142

## Problem
When using `dspy.streamify(..., async_streaming=False)`,
`apply_sync_streaming` runs the async generator in a background thread
and bridges it to a sync generator via a Queue. Its `finally` block
always sent the `stop_sentinel`, regardless of whether the generator
raised. This meant any exception was completely invisible to the sync
consumer -- iterating over the stream just looked like it ended
normally. The actual traceback only appeared on stderr via Python's
default `threading.excepthook`, in a different thread, which is easy
to miss or not connect to the calling code at all.

## Fix
Wrap exceptions raised in the producer thread in a distinct
`_StreamingException` marker before putting them on the queue (this
avoids any ambiguity with a legitimately-yielded stream item that
happens to be an `Exception` instance itself), and re-raise on the
consumer side when that marker is seen.

## Testing
Added a minimal, isolated regression test using a plain async
generator that raises partway through (no LM/program needed to
reproduce this). It fails against unmodified dspy -- the exception
never reaches `pytest.raises`, only a
`PytestUnhandledThreadExceptionWarning` shows up -- and passes after
this fix (verified against a separate, untouched checkout). Added a
companion test confirming normal (non-erroring) streams still complete
correctly, with no regression.

Ran the full `tests/streaming/` suite, plus `tests/predict/`,
`tests/primitives/`, and `tests/adapters/` (560+ tests total): all
pass, except 2 pre-existing errors unrelated to this change (missing
`litellm` CLI binary needed by a local-proxy-server test fixture in my
sandbox; confirmed identical on unmodified dspy). `ruff check` on the
changed source file: clean.

### DIFF
--- a/dspy/streaming/streamify.py
+++ b/dspy/streaming/streamify.py
@@ -228,6 +228,18 @@ def streamify(
         return sync_streamer
 
 
+class _StreamingException:
+    """Wraps an exception raised inside apply_sync_streaming's producer thread.
+
+    A distinct wrapper (rather than checking isinstance(item, Exception) directly)
+    avoids any ambiguity with a legitimately-yielded stream item that happens to be
+    an exception instance itself.
+    """
+
+    def __init__(self, exc: Exception):
+        self.exc = exc
+
+
 def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
     """Convert the async streaming generator to a sync generator."""
     queue = Queue()  # Queue to hold items from the async generator
@@ -243,6 +255,10 @@ def apply_sync_streaming(async_generator: AsyncGenerator) -> Generator:
             try:
                 async for item in async_generator:
                     queue.put(item)
+            except Exception as e:
+                # Propagate the exception to the consumer instead of letting it
+                # vanish into the background thread once the stream ends.
+                queue.put(_StreamingException(e))
             finally:
                 # Signal completion
                 queue.put(stop_sentinel)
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
`except Exception` leaves `asyncio.CancelledError` on the old path, and that's the one that matters most for a stream.

Four generators through `apply_sync_streaming` with this PR applied:

| generator raises | items delivered | consumer sees |
|---|---|---|
| nothing | `c1`, `c2` | clean end |
| `ValueError` | `c1`, `c2` | `ValueError: boom` |
| `asyncio.CancelledError` | `c1` | **clean end** |
| `KeyboardInterrupt` | `c1` | **clean end** |

Rows 3 and 4 aren't silent, though — they're worse than silent. `Exception in thread Thread-3 (producer)` and a full traceback go to stderr from the worker thread, while the `for chunk in stream:` loop on the calling side exits normally. So the log says something blew up and the caller's control flow says the response finished.

`CancelledError` has been a `BaseException` since 3.8, and on a streaming path it's a routine event rather than an edge case: a client disconnecting or an upstream `wait_for` timing out cancels the async generator. The consumer then gets a truncated LLM response that looks complete, which is the same failure #9142 is about, just arriving through a different door.

`except BaseException as e:` in the `runner()` handler covers all four rows. The `finally` already queues `stop_sentinel` either way, so the shutdown path doesn't change.

Measured on `dspy` @ 67bf5eb with the PR applied, Python 3.10.12.

PR #10096, opened 2026-07-28
fix(usage_tracker): roll nested track_usage() usage up into the enclosing tracker

### LINKED ISSUE #10064: [Bug] Nested `dspy.track_usage()` silently under-counts at the outer scope

SpiliosDimakopoulos:
Fixes #10064

## Problem
dspy.settings.usage_tracker is a single slot. When track_usage() blocks
are nested (e.g. an orchestrator with its own tracking, calling
sub-agents that also track their own usage), each LM call is only
recorded by the innermost active tracker. The outer tracker never
learns about anything that happened inside the nested block, silently
under-counting.

Using the issue's own repro (orchestrator + 2 subagents):
    orchestrator billed:  105
    actual workflow cost: 1605

## Fix
track_usage() now captures the enclosing tracker (if any) before
overriding settings.usage_tracker, and on exit rolls its own recorded
usage up into that parent tracker.

## Why this doesn't break dspy.Parallel's usage isolation
There's an existing test (test_parallel_executor_with_usage_tracker)
that explicitly checks that dspy.Parallel() workers do NOT pollute a
shared parent tracker. I checked this carefully: Parallel's
ThreadPoolExecutor runs workers in separate OS threads without copying
contextvars, so settings.usage_tracker correctly evaluates to None
inside each worker thread regardless of this change -- the isolation
comes from Python's threading/contextvars model itself, not from
track_usage() avoiding roll-up. Confirmed empirically: this test still
passes.

## Testing
Ran the exact repro script from the issue -- billed now matches actual
cost (1605 instead of 105).

Added a regression test for the nested case. It fails against
unmodified dspy (105 != 1605) and passes after this fix (verified
against a separate, untouched checkout). Added a companion test
confirming top-level (non-nested) tracking is unaffected.

Ran the full tests/utils/test_usage_tracker.py suite (all pass) plus a
broader sweep of every other file that references track_usage
(test_gepa.py, test_base_module.py, test_predict.py, test_lm.py,
test_settings.py): identical pass/fail/error counts to unmodified
dspy -- no new failures introduced. ruff check on the changed source
file: clean.

### DIFF
--- a/dspy/utils/usage_tracker.py
+++ b/dspy/utils/usage_tracker.py
@@ -67,8 +67,20 @@ class UsageTracker:
 
 @contextmanager
 def track_usage() -> Generator[UsageTracker, None, None]:
-    """Context manager for tracking LM usage."""
+    """Context manager for tracking LM usage.
+
+    If this is used inside another `track_usage()` block, the usage recorded here is
+    also rolled up into the enclosing tracker on exit, so the outer block's totals
+    reflect everything that happened within it, not just what happened outside the
+    nested block.
+    """
+    parent_tracker = settings.usage_tracker
     tracker = UsageTracker()
 
     with settings.context(usage_tracker=tracker):
         yield tracker
+
+    if parent_tracker is not None:
+        for lm, usage_entries in tracker.usage_data.items():
+            for usage_entry in usage_entries:
+                parent_tracker.add_usage(lm, usage_entry)
--- a/tests/utils/test_usage_tracker.py
+++ b/tests/utils/test_usage_tracker.py
@@ -155,6 +155,45 @@ def test_track_usage_context_manager(lm_for_test):
     assert isinstance(total_usage[lm_for_test], dict)
 
 
+def test_nested_track_usage_rolls_up_into_outer_tracker():
+    """Regression test for #10064: usage recorded inside a nested
+    track_usage() block must also be reflected in the enclosing tracker,
+    not just the innermost one.
+    """
+    model = "openai/gpt-4o-mini"
+
+    def record(prompt_tokens, completion_tokens):
+        dspy.settings.usage_tracker.add_usage(
+            model, {"prompt_tokens": prompt_tokens, "completion_tokens": completion_tokens}
+        )
+
+    def subagent(prompt_tokens, completion_tokens):
+        with track_usage() as usage:
+            record(prompt_tokens, completion_tokens)
+        return usage.get_total_tokens()[model]["prompt_tokens"]
+
+    with track_usage() as orchestrator_cost:
+        record(100, 10)
+        a = subagent(1000, 100)
+        b = subagent(500, 50)
+        record(5, 1)
+
+    billed = orchestrator_cost.get_total_tokens()[model]["prompt_tokens"]
+    assert billed == 100 + a + b + 5  # 1605, not 105
+
+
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Same note I left on #10070, since this shares its structure: the roll-up isn't exception-safe.

```python
    yield tracker
if parent_tracker is not None:      # skipped when the body raises
    ...
```

In a `@contextmanager`, an exception from inside the `with track_usage()` body propagates through the `yield`, so anything after it never runs. Usage recorded before the failure is dropped from the parent's total. Simulated on an inner block that spends 500 tokens and then raises, with the exception caught outside:

```
this shape  : outer billed 105   <- the 500 is lost
try/finally : outer billed 605
```

Wrapping it fixes it without changing anything else:

```python
try:
        yield tracker
finally:
        ...
```

This is worth closing because a failed LM call is precisely when you want the cost recorded: the tokens were spent and the provider bills for them, so a retry loop that swallows the exception would under-count on every failed attempt, which is the same class of silent under-counting the issue is about.

Your two tests are good and `test_top_level_track_usage_unaffected_by_rollup` is a case not everyone covered, worth keeping whichever PR lands. An exception-path test would catch the above.

Heads up that there are four open PRs for #10064 now (#10070 Jul 20, #10085 and #10086 Jul 25, this one). #10070 has priority by date. Full comparison on the issue. Disclosure: #10086 is mine.

ErenAta16 (commented):
The roll-up sits after the `with settings.context(...)` block with no `try`/`finally`, so it's skipped whenever the nested block raises — and a sub-agent whose tool call blows up is the case people hit.

Same script against `main`, this PR, and #10065:

```python
with track_usage() as outer:
    rec(100, 10)
    try:
        with track_usage():
            rec(1000, 100)
            raise RuntimeError("tool failed")
    except RuntimeError:
        pass
    rec(5, 1)
```

| | nested block returns | nested block raises | non-numeric usage fields |
|---|---|---|---|
| `main` @ 67bf5eb | 105 | 105 | `is_byok: 0`, `provider: 'openaiopenai'` |
| this PR | **1105** | **105** | `is_byok: 0`, `provider: 'openaiopenai'` |
| #10065 | **1105** | **1105** | `is_byok: False`, `provider: 'openai'` |

Expected is 1105 in both of the first two columns. Wrapping the `yield` in `try`/`finally` closes the second column.

Third column is a separate bug in `_merge_usage_entries` that #10065 also fixes. `elif current_v is not None or v is not None: result[k] = (current_v or 0) + (v or 0)` runs on every non-dict value, so `False + False` becomes `0` and two `'openai'` strings concatenate into `'openaiopenai'`. Both litellm's `Usage` and dspy's `LMUsage` let providers pass extra fields through, so this fires whenever one of them reports something that isn't a count — OpenRouter's `is_byok` is the one I could name.

Worth flagging that #10065 predates this by a week and is still open with no review, and #10086 / #10088 were the same fix and got closed by their authors. Four PRs against #10064 now.

Measured on `dspy` @ 67bf5eb, Python 3.10.12.

Issue #10093, opened 2026-07-27
Align dspy.Image validation with Audio/File (add model_validator; make I/O boundary explicit)


isaacbmiller:
## Summary

`dspy.Image` normalizes/validates its input only through a custom `__init__`, while `dspy.Audio` and `dspy.File` use a custom `__init__` **plus** a `@pydantic.model_validator(mode="before")` (`validate_input`) that delegates to a shared `encode_*` function. This asymmetry — introduced/left by #10069 — has an observable consequence and leaves `Image`'s I/O boundary implicit.

## Observable inconsistency

A bare data-URI string validates cleanly for `Audio`/`File` through the pydantic validation path (nested models, `TypeAdapter`, output-parse fallback) but **fails for `Image`**, even though direct construction of the same value works:

```python
from pydantic import TypeAdapter, BaseModel
import dspy

img_uri  = "data:image/png;base64,iVBORw0KGgo...=="
aud_uri  = "data:audio/wav;base64,AA=="
file_uri = "data:text/plain;base64,aGk="

TypeAdapter(dspy.Image).validate_python(img_uri)   # ValidationError: Input should be a valid dictionary...
TypeAdapter(dspy.Audio).validate_python(aud_uri)   # OK
TypeAdapter(dspy.File).validate_python(file_uri)   # OK

class Wrap(BaseModel):
    x: dspy.Image
Wrap.model_validate({"x": img_uri})                # ValidationError (Audio/File equivalents pass)

dspy.Image(img_uri)                                # OK (direct positional construction)
```

## Why it happens

Pydantic v2 detects `Image`'s custom `__init__` (`__pydantic_custom_init__`) and routes **dict** validation through it, which is why `validate_python({"url": ...})` reaches `encode_image` and the resource-loading boundary holds. But a **bare string** never gets there: with no `mode="before"` validator, pydantic's core schema rejects it ("Input should be a valid dictionary or instance of Image") before any normalization runs. `Audio`/`File`'s `validate_input` catches the bare string and normalizes it.

## Why it's worth fixing

1. **Consistency** — `Image` cannot be validated from a bare string anywhere pydantic drives validation (nested models, `TypeAdapter`, parse fallbacks), even for a legitimate data URI; the other two types can.
[...]

PR #10098, opened 2026-07-29
refactor(adapters): name I/O factories open()/download() (ergonomics follow-up)

isaacbmiller:
## Summary

Follow-up to #10069 that addresses reviewer feedback on API ergonomics. Renames the two **I/O** factories to verbs and keeps the in-memory `from_*` constructors, so the resource types read like the rest of the Python ecosystem:

- `Image.open(path)` / `Image.download(url)` (was `from_path` / `from_url`)
- `Audio.open(path)` / `Audio.download(url)`
- `File.open(path)` (File has no remote-fetch content block, so no `download`)
- in-memory constructors unchanged: `from_bytes`, `from_array`, `from_file_id`

The old names — `from_path`, `from_url`, `from_file` — remain as **deprecated aliases** (warn, delegate, removed in 3.4), so this is non-breaking on top of #10069.

## Why (the ecosystem convention)

pandas and the stdlib draw a line this PR follows: `from_*` classmethods construct from **in-memory Python data** (`DataFrame.from_records`, `int.from_bytes`, `datetime.fromisoformat`), while **I/O** — reading a file or a URL — lives behind a differently-named callable (`open`, `read_csv`, `load`). pandas even *removed* `DataFrame.from_csv` in favor of `pd.read_csv` for exactly this reason. Under that convention, `from_path`/`from_url` were the misfits: they do disk/network I/O but wore the in-memory `from_*` prefix.

Their own reference libraries agree on the other half too — a *constructor* never reads a file: `np.array("x.npy")` returns a string array, `pd.DataFrame("x.csv")` raises, and `PIL.Image` reads via the named `Image.open()`. So keeping I/O out of `__init__` (from #10069) is idiomatic; this PR only fixes the *naming* of the explicit loaders.

Net surface:

| | Image | Audio | File |
|---|---|---|---|
| local file (I/O) | `open` | `open` | `open` |
| remote (I/O) | `download` | `download` | — |
| in-memory | `Image(bytes/PIL/data-URI)` | `from_array`, `Audio(bytes,…)` | `from_bytes`, `from_file_id`, `File(bytes)` |

## Migration

- `Image.from_path` → `Image.open`; `Image.from_url` → `Image.download` (same for `Audio`)
- `File.from_path` → `File.open`
- `from_file` (already deprecated in #10069) now points at `open`
- all deprecated aliases warn and are scheduled for removal in **3.4**

## Validation

- `uv run --frozen pytest --deno -q tests/adapters tests/signatures` (`370 passed, 2 xfailed`)
- deprecation of every alias is covered (`test_local_read_aliases_are_deprecated`, `test_deprecated_factory_aliases_warn`)
- Ruff lint/format; pre-commit hooks

Stacked on #10069 (merge that first). Also advances the convergence tracked in #10093.

### DIFF
--- a/docs/docs/api/primitives/Audio.md
+++ b/docs/docs/api/primitives/Audio.md
@@ -10,8 +10,8 @@
             - extract_custom_type_from_annotation
             - format
             - from_array
-            - from_path
-            - from_url
+            - open
+            - download
             - is_streamable
             - parse_lm_response
             - parse_stream_chunk
--- a/docs/docs/api/primitives/Image.md
+++ b/docs/docs/api/primitives/Image.md
@@ -10,8 +10,8 @@
             - extract_custom_type_from_annotation
             - format
             - from_PIL
-            - from_path
-            - from_url
+            - open
+            - download
             - is_streamable
             - parse_lm_response
             - parse_stream_chunk
--- a/docs/docs/diving-deeper/adapters.md
+++ b/docs/docs/diving-deeper/adapters.md
@@ -125,15 +125,15 @@ The base class. Subclass it (it’s a `pydantic.BaseModel`) and implement `forma
 
 **`dspy.Image(source)`**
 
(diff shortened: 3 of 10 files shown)


isaacbmiller:
Shelving this for now. Given the underlying issue is low-severity hardening (not a real vulnerability), the naming churn — a second deprecation cycle on `from_path`/`from_url` right after introducing them — isn't worth it near-term. Leaving it as a reference for the ideal `open()`/`download()` surface if we revisit resource-type ergonomics later (see #10093).

PR #10103, opened 2026-07-29
fix(clients): scrub LM secrets and call history from save_program pickles

MaximeRivest:
> **Stacked PR 1/3** — #10104 and #10116 build on this.

## Problem

`module.save(path, save_program=True)` embeds live secrets and private data into `program.pkl` in cleartext:

- explicit `api_key` strings (and sibling credentials such as `azure_ad_token`, `aws_secret_access_key`)
- credential **callables**, closure included — serializing the very object whose purpose is keeping the secret out of the process image
- `Authorization`-style headers passed via `extra_headers` / `headers`
- the entire `lm.history` — every prompt and completion the LM made before saving

`allow_pickle=True` gates the *loader's* trust in the file, but nothing gated what the file carries to whoever receives it — and full-program artifacts exist precisely to be shipped to other teams and services. The JSON state path has always excluded `api_key` (`LM.dump_state`); the pickle path silently didn't.

Repro:

```python
lm = dspy.LM("openai/gpt-4o-mini", api_key="sk-live-SECRET")
predict = dspy.Predict("q -> a")
predict.lm = lm
predict.save("prog/", save_program=True)
assert b"sk-live-SECRET" in open("prog/program.pkl", "rb").read()  # passes before this fix
```

## Fix

`BaseLM` gains `__getstate__` / `__setstate__`. When pickling a program artifact, they scrub credentials (a named `SENSITIVE_LM_KWARG_NAMES` set), sensitive headers (`SENSITIVE_HEADER_NAMES`), `callbacks`, and `history`. Loaded programs work, but resolve credentials fresh on the loading side — the same contract the JSON state path already documents.

The subtle part: `__getstate__` also fires during `copy.copy` / `copy.deepcopy`, which optimizers and the transactional `load_state` trial run rely on — an unconditional scrub would strip working credentials mid-optimization. Scrubbing is therefore scoped by a contextvar that only `BaseModule.save`'s cloudpickle dump sets. In-process copies keep credentials and history; only saved artifacts are scrubbed. Both behaviors are pinned by tests.

## Tests

- `test_program_pickle_scrubs_lm_secrets_and_history` — no secret/header/history bytes in `program.pkl`; loaded program has no `api_key` kwarg, scrubbed headers, empty history.
- `test_copy_and_deepcopy_keep_lm_credentials_and_history` — `deepcopy` and `lm.copy()` keep credentials and history.

Note: `tests/evaluate/test_evaluate.py::test_multi_thread_evaluate_call_cancelled` is flaky/environment-sensitive on current `main` in my test setup, independent of this change.

### DIFF
--- a/dspy/clients/base_lm.py
+++ b/dspy/clients/base_lm.py
@@ -1,3 +1,5 @@
+import contextlib
+import contextvars
 import copy as copy_module
 import datetime
 import importlib
@@ -24,6 +26,41 @@ GLOBAL_HISTORY = []
 LM_CLASS_STATE_KEY = "_dspy_lm_class"
 _BUILTIN_LM_CLASS_PATH = "dspy.clients.lm.LM"
 ForwardContract = Literal["legacy", "typed_lm"]
+
+# Kwarg names that hold credentials and must never reach a saved artifact.
+SENSITIVE_LM_KWARG_NAMES = frozenset(
+    {"api_key", "azure_ad_token", "aws_access_key_id", "aws_secret_access_key", "aws_session_token"}
+)
+# Header names (lowercase) that hold credentials and must never reach a saved artifact.
+SENSITIVE_HEADER_NAMES = frozenset(
+    {"authorization", "proxy-authorization", "api-key", "x-api-key", "cookie", "set-cookie"}
+)
+
+# Scrubbing is scoped to program-artifact pickling only: `copy.copy`,
+# `copy.deepcopy`, and in-process pickling (optimizer copies, the
+# transactional-load trial run) must keep working credentials and history.
+_scrub_lm_state_on_pickle = contextvars.ContextVar("_scrub_lm_state_on_pickle", default=False)
+
+
+@contextlib.contextmanager
+def scrub_lm_state_on_pickle():
+    """Scrub credentials and history from LMs pickled inside this context.
+
+    `BaseModule.save(save_program=True)` wraps its cloudpickle dump in this
+    context so program artifacts never embed API keys, credential callables,
+    sensitive headers, or call history.
+    """
+    token = _scrub_lm_state_on_pickle.set(True)
+    try:
+        yield
+    finally:
+        _scrub_lm_state_on_pickle.reset(token)
+
+
+def _scrub_sensitive_headers(headers):
+    if not isinstance(headers, dict):
+        return headers
+    return {key: value for key, value in headers.items() if str(key).lower() not in SENSITIVE_HEADER_NAMES}
 
 
 def _import_lm_class(class_path: str) -> type:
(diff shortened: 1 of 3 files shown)


ErenAta16 (commented):
Confirmed the leak and the fix, then went looking for what the scrub set misses. Five things still reach `program.pkl` on this branch.

First, the baseline. On `origin/main` (`0312f0da`) with one LM carrying five different secrets:

```
LEAK api_key
LEAK Authorization header
LEAK history prompt
LEAK callable closure token      (azure_ad_token_provider=lambda: ...)
LEAK vertex_credentials
```

On `4e5642d5`, the first three are gone. The last two are not:

| carrier | on `main` | on this branch |
|---|---|---|
| `api_key="sk-live-..."` | leak | scrubbed |
| `api_key=lambda: "..."` | leak | scrubbed |
| `extra_headers={"Authorization": ...}` | leak | scrubbed |
| `lm.history` | leak | scrubbed |
| `azure_ad_token_provider=lambda: "..."` | leak | **leak** |
| `vertex_credentials='{"private_key": ...}'` | leak | **leak** |
| `default_headers={"Authorization": ...}` | leak | **leak** |
| `extra_headers=[("Authorization", ...)]` | leak | **leak** |
| subclass setting `self.api_key = ...` | leak | **leak** |

Details on each:

`azure_ad_token_provider` and `vertex_credentials` are both real litellm kwargs — 23 and 95 files in the installed litellm reference them — and `**kwargs` lands them in `self.kwargs` verbatim, so adding them to `SENSITIVE_LM_KWARG_NAMES` covers both. `azure_ad_token_provider` is the one I'd flag hardest, because the description says credential callables are handled "closure included": that holds for `api_key=lambda: ...` (it's under the `api_key` key) but not for this one, which is the canonical Azure credential callable. `aws_web_identity_token` and `aws_role_name` are in litellm too and fall in the same bucket.

`default_headers` is a third header field alongside `extra_headers` / `headers`, and only the latter two are looped over.

`_scrub_sensitive_headers` returns non-dicts unchanged, so `extra_headers=[("Authorization", "Bearer ...")]` passes straight through. httpx accepts a sequence of pairs, so this isn't only theoretical.

The subclass case: `__getstate__` filters `state["kwargs"]` but `state` itself is `self.__dict__.copy()`, so a `BaseLM` subclass that stores a credential as an attribute keeps it. Applying `SENSITIVE_LM_KWARG_NAMES` to the top level of `state` as well would close it with the set you already have.

Two things that check out and are worth recording so nobody re-tests them. The contextvar scoping does what it claims — after `copy.deepcopy` the copy still has `kwargs["api_key"]` and its one history entry, so optimizer copies aren't affected. And `GLOBAL_HISTORY` does not end up in the artifact: it's module-level, so cloudpickle stores it by reference, and a private entry I put there before saving is absent from the bytes.

`test_program_pickle_scrubs_lm_secrets_and_history` and its sibling both pass. I couldn't run the whole of `tests/clients/test_lm.py` — `pytest-asyncio` isn't in this env, so the `@pytest.mark.asyncio` tests error at collection, and the run then hangs on something network-bound.

PR #10104, opened 2026-07-29
feat(primitives): refuse untrusted LM endpoint state; add lm= override on load

MaximeRivest:
> **Stacked PR 2/3** — builds on #10103; #10116 builds on this. This diff includes #10103's commit until it merges.

## Problem

Loading saved state whose LM config carries endpoint routing (`api_base`, `base_url`, `model_list`) strips those keys with a warning and continues. For anyone using a local or self-hosted server — vLLM, SGLang, Ollama, a gateway — the endpoint lives in exactly those keys, so this is the default experience for that entire population, and it produces the worst possible failure mode:

```python
lm = dspy.LM("openai/my-local-model", api_base="http://localhost:8000/v1")
predict = dspy.Predict("q -> a")
predict.lm = lm
predict.save("prog.json")

loaded = dspy.Predict("q -> a")
loaded.load("prog.json")   # warning, then "success"
loaded(q="hello")          # → api.openai.com, under ambient OPENAI_API_KEY
```

The reloaded program silently sends requests — and the prompts inside them — to the provider's default endpoint under whatever ambient key resolves. If the model name happens to exist there (a real model name behind a corporate proxy, say), it *works*, rerouted, bypassing the gateway and billed to the ambient key with no signal beyond one load-time warning. If not, it fails later with a misleading model-not-found error far from the cause.

## Change

**Untrusted loads refuse, typed.** State whose LM carries endpoint configuration now raises `dspy.LMStateError` (new, under `LMConfigurationError`) naming the offending keys and both exits. A program loads exactly as it was saved, or fails naming the gap — it never loads "successfully" onto a different endpoint than it was saved with.

**`lm=` is the safe exit.** `load()` / `load_state()` accept:

- a single `BaseLM` — applied to every predictor, saved LM state ignored;
- a dict mapping predictor names (as returned by `named_predictors()`) to LMs — for multi-LM programs. Predictors absent from the dict load their saved LM state normally, so callers supply routes only for the predictors that carried endpoint configuration. Unknown names fail fast, listing the module's actual predictor names.

```python
program.load("prog.json", lm={
    "draft.predict": dspy.LM("openai/local-model", api_base="http://localhost:8000/v1"),
    "critique": dspy.LM("anthropic/claude-sonnet-5"),
})
```

Routes come from the caller's code, never from the file.

**`allow_unsafe_lm_state=True` keeps its meaning** for trusted files: every saved LM — however many the program has — reconstructs exactly as saved.

## Behavior change

The previous strip-and-warn tests (`test_load_ignores_serialized_endpoint_override_by_default` and friends) asserted the silent-degrade behavior; they are rewritten to assert the typed refusal, plus new coverage for the `lm=` single/mapping paths and the unknown-name error. The `saving-and-loading` docs now present the two exits side by side.

Note: `tests/primitives/test_base_module.py::test_usage_tracker_in_parallel` and `tests/evaluate/test_evaluate.py::test_multi_thread_evaluate_call_cancelled` fail on current `main` in my environment independent of this change.

Related: #10103 (artifact-side hygiene for the pickle path).

### DIFF
--- a/docs/docs/diving-deeper/saving-and-loading.md
+++ b/docs/docs/diving-deeper/saving-and-loading.md
@@ -6,6 +6,23 @@ Saving a DSPy program preserves what an optimizer produced: the rewritten signat
 
 Read this when you’re shipping an optimized program to another team, version-controlling optimizer output, or trying to figure out why a `.json` save round-trips but a `.pkl` save needs an extra flag to load.
 
+The examples below use this two-predictor program:
+
+```python
+import dspy
+
+
+class HaikuEnsemble(dspy.Module):
+    def __init__(self):
+        super().__init__()
+        self.draft = dspy.ChainOfThought("topic -> haiku")
+        self.critique = dspy.Predict("haiku -> improved_haiku")
+
+    def forward(self, topic):
+        draft = self.draft(topic=topic)
+        return self.critique(haiku=draft.haiku)
+```
+
 ## Design decisions
 
 ### 1. Two paths: state-only and full-program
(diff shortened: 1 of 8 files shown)


docs/docs/diving-deeper/saving-and-loading.md
```diff
@@ -6,6 +6,23 @@ Saving a DSPy program preserves what an optimizer produced: the rewritten signat
 
 Read this when you’re shipping an optimized program to another team, version-controlling optimizer output, or trying to figure out why a `.json` save round-trips but a `.pkl` save needs an extra flag to load.
 
+The examples below use this two-predictor program:
+
+```python
+import dspy
+
+
+class HaikuEnsemble(dspy.Module):
```
isaacbmiller, line 15: Nit: this is not an ensemble

isaacbmiller (commented):
Seems strange to do things other than load state from a file in a function to load from a file.

Would prefer one of the solutions we discussed a while ago along the lines of:
register custom `{custom_lm_identifier: CustomLMClass}` in some registry and then load_state will check against that registry rather than pass a per predictor dict or instance at runtime.

Totally fine with raising instead of warning when we strip state automatically tho - happy to include that separately from the load fn change.

Instead of:
```python
program.load("prog.json", lm={
    "draft.predict": dspy.LM("openai/local-model", api_base="http://localhost:8000/v1"),
    "critique": dspy.LM("anthropic/claude-sonnet-5"),
})
```
I prefer
```python
program.load("prog.json")
# if we need to redefine these at runtime
program.set_predictor_lms(lm={
    "draft.predict": dspy.LM("openai/local-model", api_base="http://localhost:8000/v1"),
    "critique": dspy.LM("anthropic/claude-sonnet-5")
}) # we would add this method
```

or if we want to have the per predictor LM serialized:

prog.json:
```json
{"<module>.<predictor>": {"lm": {"kind": "dspy/OpenAICompatLM", "base_url": "..."}}}
```
```python
program.load("prog.json", allow_unsafe_lm_state=True)
```

And instead of
```python
program.load("prog.json", lm=dspy.LM("anthropic/claude-sonnet-5")})
```
I prefer:
```python
program.load("prog.json")
program.set_lm(dspy.LM("anthropic/claude-sonnet-5"))
```

MaximeRivest:
> Seems strange to do things other than load state from a file in a function to load from a file.

this is not my perspective. program.load()  loads _state_ not _state_from_file_? Is there any reason to restrict it to _state_from_file_?

isaacbmiller:
hmmm yeah you're right.

wdyt of splitting the parameters into `lm` and something like `lm_overrides`?

Issue #8481, opened 2025-07-02
[Feature] XMLAdapter fails on nested models & repeated tags due to one-level regex parser


Bhuvanesh09:
## TL;DR

* XMLAdapter only handles **flat** key/value XML.  
* Nested models, lists or any repeated tags raise `AdapterParseError`.  
* Root cause: a greedy one–level regex (`<(\w+)>(.*?)</\1>`) with no recursion.  
* Effect: Structured-data extraction use-cases are broken on every DSPy version tested (0.1.7 – 0.2.0).  
* I am preparing a PR that rewrites the parser to perform proper XML traversal; details & full analysis are linked below.

---

## Minimal Reproduction

```python
from typing import List
from pydantic import BaseModel
import dspy

class Person(BaseModel):
    name: str
    age: int

class PersonExtraction(dspy.Signature):
    text: str = dspy.InputField()
    person: Person = dspy.OutputField()

sig = PersonExtraction()
lm_output = """
<person>
  <name>John Doe</name>
  <age>30</age>
</person>
"""

dspy.XMLAdapter(sig).parse(lm_output)   # Raises AdapterParseError
```

Expected result

`{"person": Person(name="John Doe", age=30)}`

Actual result

`AdapterParseError: Failed to parse LM response …`

```
Environment
DSPy            : 3.0.0b1
Python          : 3.11.
OS              : macOS (Apple Silicon)
```
## Impact
- Any signature with nested Pydantic models or List[...] cannot be parsed.
- Few-shot examples generated by format_field_with_value teach the LM an incorrect JSON-in-XML format, compounding errors.
- Users are pushed to JSONAdapter even when XML would be preferable.

## Proposed Fix
Replace the one–level regex with an XML traversal (e.g., xml.etree.ElementTree) that:
- Recursively walks the tree,
- Builds Python containers (dict, list) matching the signature,
- Delegates scalar conversion to parse_value.
- Update format_field_with_value to emit canonical XML rather than JSON-in-XML.

Add unit tests covering:
- Nested models,
- Repeated tags → lists,
- Mixed scalar & container fields.

I have an implementation that passes these tests and will open a draft PR once this issue is acknowledged.

### Would you like to contribute?

- [x] Yes, I'd like to help implement this.
- [ ] No, I just want to request it.

PR #10106, opened 2026-07-30
fix(adapters): stop XMLAdapter from silently truncating values at a nested closing tag

### LINKED ISSUE #10102: [Bug] XMLAdapter.parse silently truncates a field value that contains its own closing tag
### What happened?

`XMLAdapter.parse` silently truncates a field's value at the first occurrence of that field's own closing tag inside the content, so anything after it is dropped without an error.

```python
field_pattern = re.compile(r"<(?P<name>\w+)>((?P<content>.*?))</\1>", re.DOTALL)
```

`.*?` is non-greedy, so the match ends at the *first* `</name>`, not the last one. `parse` then finds a value for the field, `fields.keys() == signature.output_fields.keys()` holds, and no `AdapterParseError` is raised — the caller gets a short string back.

`ChatAdapter` and `JSONAdapter` handle the same content fine, so this is specific to the XML path.

### Steps to reproduce

```python
import dspy
from dspy.adapters import XMLAdapter, ChatAdapter, JSONAdapter

class CodeSig(dspy.Signature):
    task: str = dspy.InputField()
    code: str = dspy.OutputField()

lm_output = "<code>\nif x:\n    print('</code>')\n</code>\n"
XMLAdapter().parse(CodeSig, lm_output)
```

| adapter | parsed `code` |
|---|---|
| `XMLAdapter` | `"if x:\n    print('"` |
| `ChatAdapter` | `"if x:\n    print('</code>')"` |
| `JSONAdapter` | `"if x:\n    print('</code>')"` |

With two output fields it takes the second one's value too:

```python
class Two(dspy.Signature):
    q: str = dspy.InputField()
    reasoning: str = dspy.OutputField()

XMLAdapter().parse(Two, "<reasoning>\nthink\n</reasoning>\n<answer>\nsee </answer> above\n</answer>\n")
# {'reasoning': 'think', 'answer': 'see'}
```

Round-tripping through the adapter's own formatter shows the same thing — `format_assistant_message_content` doesn't escape the value, so a demo whose output contains the closing tag produces a message that this parser reads back short:

```python
a = XMLAdapter()
a.parse(CodeSig, a.format_assistant_message_content(CodeSig, {"code": "before </code> after"}))
# {'code': 'before'}
```

### Why it matters

An output field holding code, HTML, XML, or a quoted log line is the normal case for a field named `code` or `html`, and the field names are the user's own — a signature with an `answer` field asking the model to critique XML output hits it too. The failure is silent: no exception, no warning, just a shorter string, so it surfaces downstream as a wrong answer rather than a parse failure.

### What a fix has to decide

Making `.*` greedy fixes the single-field case and breaks the multi-field one — a greedy `<reasoning>...</reasoning>` would swallow through to the last `</reasoning>` in the completion, and with a repeated tag that's the wrong span.

Since `signature.output_fields` is known up front, one option is per-field extraction against the last `</name>` that precedes the next known field's opening tag, rather than one pattern scanned across the whole completion. Another is to raise instead of truncating when a field's content contains its own closing tag — noisier, but at least not silent.

Noting #8482 is open and rewrites `XMLAdapter` for nested data; I haven't checked whether that rewrite changes this behaviour, so it may already be covered there.

### DSPy version

`main` @ 67bf5eb, Python 3.10.12

TLXyloph:
Fixes #10102: `XMLAdapter.parse` silently truncates a field value that contains its own closing tag, returning a short string with no error. The goal here is to recover the values that are unambiguous, report the ones that are not, and leave every value that parses correctly today untouched.

```
completion:  <code>\n<code>x = 1</code>\n</code>

before   {'code': '<code>x = 1'}          silently truncated
after    {'code': '<code>x = 1</code>'}   recovered in full
```

## What changed

- **Escaping on the wire.** `format` now escapes output values (`&` as `&amp;`, `<` as `&lt;`), the prompt instructs the model to do the same, and `parse` decodes — so the format → parse round trip is lossless even for a value containing its own closing tag. A compliant completion never hits any of the recovery machinery below.
- **A linear tag scan instead of the lazy regex.** For raw, non-compliant completions, `_scan_tags` / `_balanced_partners` / `_extract_field_blocks` replace `field_pattern`. A value that is itself a same-named element (`<code><code>x</code></code>`) widens to its depth-balanced closing tag instead of being cut at the inner one. A widened span is accepted only when it's self-contained and swallows no other output field, decided from prefix data in constant time.
- **Ambiguity is reported, not resolved by guessing.** When a field's closing tag is left over in open text with real text before it, and when a nested span is structurally sound but holds a sibling output field (so widening would swallow the sibling and the lazy reading would drop the tail of the value), `parse` raises `AdapterParseError` naming the field. Whitespace-only duplicated closing tags stay tolerated as the benign model slip they are.
- **Streaming agrees with parse.** The stream listener asks the same block scan where a value ends rather than matching the closing tag directly, so nested values reach the consumer exactly as `parse` returns them, buffered until the boundary is known. This also fixed two chunk-boundary cases where the cache-hit shortcut ended the stream on a tag that was really part of the value and emitted nothing at all.
- **Docs.** The parse contract, the escape scheme, and the remaining limitations are documented in the `XMLAdapter` docstrings and `docs/docs/diving-deeper/adapters.md`, with the streaming behavior noted in the streaming guide.

## Behavior changes that deserve sign-off

- Completions that used to parse can now raise. Under `__call__` the inherited `use_json_adapter_fallback` retries through JSONAdapter, so programs keep working at the cost of a second LM call; pass `use_json_adapter_fallback=False` for a hard error instead.
- `parse` decodes `&lt;`/`&amp;` unconditionally, so a legacy completion that spells those literally now reads back decoded. That's the price of a reversible wire format; documented as a tradeoff.
- `XMLAdapter.field_pattern` is gone. Nothing in the repo used it, but it was a public-looking attribute, so a downstream override now breaks — probably worth a release note.
- The cache-hit guard sits in the adapter-agnostic streaming branch, so ChatAdapter is affected too: a listener whose field arrives in a later chunk now streams the value where it previously streamed nothing (details in the first review comment below).
- Two shapes of raw non-compliant output remain silently truncated by design and are documented: a value that *ends* with its own closing tag (only whitespace separates the readings), and a structurally-rejected nested span whose surplus tag a later block covers or separates. Both need a raw `<` in the value, which a compliant completion never produces.

## Testing

- Acceptance tests went in red-first against the pre-fix parser: the issue's exact completions fail on base and pass here, asserting full values against hand-computed oracles.
- Full suites on the branch: `tests/adapters` 358 passed; the XML adapter + streaming test files 160 passed, 3 skipped, with 2 pre-existing fixture errors from a litellm proxy binary missing in the venv (predates this branch).
- A 20,000-case differential fuzz against the pre-fix parser: 9,895 values identical, 699 recovered longer, 5,033 previously-silent truncations now reported, zero values changed incompatibly.
- Reviewed over five automated review passes (see the comments); every finding was either fixed with a pinned regression test or refuted with a runtime reproduction, and the latest pass generated no comments.

For triage: #8482 rewrites XMLAdapter on ElementTree for nested Pydantic models (#8481). It doesn't fix this bug, and it would reject things the adapter accepts today (bare `&`, stray tags, duplicate closing tags), since ElementTree wants well-formed XML.


TLXyloph:
Greptile's first pass flagged three findings and its second flagged two more. All five are now fixed. It last reviewed `e7174c9a`; head is now `4f1f4d03`.

## The two chunk-boundary P1s — fixed

Both reproduced, and both were real. They also reproduce on the base commit, so they weren't regressions from this branch, but they did contradict what it claims, so they're fixed here.

They turned out to be one bug rather than two, and not in the place the review pointed at. `_xml_nested_stream_chunk` was never reached in either case — both died earlier, in the pre-existing cache-hit shortcut that ends the stream when a chunk holds the start identifier followed by the end one. Two problems with it:

- It matched the end identifier directly, so for a value nesting its own tag the *inner* `</answer>` looked like the end. That's P1a.
- It ran on every chunk, with no guard for a field that had already started or was still being detected, so a later chunk carrying the inner `<answer>` read as a second start. That's P1b.

It now runs only before the field starts and before a start identifier is part-way through arriving, and for XMLAdapter it asks `_field_value_end` where the value stops instead of matching the tag. A value that genuinely starts and ends inside one chunk still takes the shortcut, unchanged.

```
                              before          after
P1a  streamed                 ''              '<answer>inner</answer>'
     parse                    same            same
P1b  streamed                 ''              '<answer>inner'
     parse                    same            same
```

**Worth flagging: this changes ChatAdapter too.** The guard sits in the adapter-agnostic branch. With a listener on `answer`, chunk 1 `[[ ## reasoning ## ]]\ntext\n\n` and chunk 2 `[[ ## answer ## ]]\n42\n\n[[ ## completed ## ]]`:

```
before:  streamed ''    predictions 1
after:   streamed '42'  predictions 0
```

I think the new behaviour is right — a listener on `answer` should stream `answer` — but it changes what `streamify(..., include_final_prediction_in_output_stream=False)` yields, so it shouldn't slip by unnoticed.

## P1 from the first pass: streaming vs parse on nested values

Fixed, though the original framing was a bit off. "Streaming truncates nested values" sounds like this PR broke streaming, but `streaming_listener.py` was byte-identical to main at `d0db8c05`, and so were the streamed bytes. What happened is I fixed `parse` and left streaming alone, which created a disagreement that didn't exist before:

```
completion:  <code>\n<code>x</code>\n</code>

main         streamed '<code>x'          parse '<code>x'          both truncated, so they agreed
d0db8c05     streamed '<code>x'          parse '<code>x</code>'   disagreed  <- the P1
4f1f4d03     streamed '<code>x</code>'   parse '<code>x</code>'   both correct
```

The listener asks the adapter where the value ends, and that helper runs `_extract_field_blocks`, the same scan `parse` uses. I originally wrote a second implementation of the rule and it drifted twice, once on an unpaired `<div>` and once on an unpaired `</html>`, both widening spans that `parse` rejects. Sharing the code seemed better than keeping two copies honest.

On why nested values are buffered rather than streamed incrementally: you can't tell which closing tag ends the value until the balancing one arrives, and once a chunk goes out you can't take it back. An eager guess either truncates or runs into the next field.

## P2: the non-field block case

Greptile was right. This pair makes it obvious:

```
<code>print(1)</code> tail text <note></code></note> more ...   parses, no error
<code>print(1)</code> tail text </code> more ...                raises
```

Same surplus tag, same lost text, and the only difference is a balanced `<note>` block. Docs now say any later balanced block, and spell out both remaining shapes inline instead of pointing at a private function.

## P2: prose heuristic

Right again — my explanation didn't match what the code does:

```
<answer>x<answer> prose</answer> trailing</answer>            stays lazy
<answer> <answer> prose mention</answer> trailing</answer>    widens
```

It comes down to whether the mention opens the value, not whether it looks like prose. Reworded.

## Other tradeoffs worth knowing about

- A value that *ends* with its closing tag (`<code>x</code></code>`) still truncates silently. Nothing is left over to detect, and expressing it would need escaping this format doesn't have. There's a test covering it.
- A nested-shaped value on the *last* output field, whose span never balances, is held until `finalize()` — nothing settles the boundary earlier when there's no later sibling. The value is correct and matches `parse`, just delivered in one piece at the end.
- Completions that used to parse can now raise. Under `__call__` the inherited `use_json_adapter_fallback` retries through JSONAdapter, so things keep working but you pay a second LM call, and on degenerate input that fallback can fail too. If you'd rather this be opt-in, happy to put it behind a flag.
- `AdapterParseError.parsed_result` now has uncast values on the missing-field path (`{"score": "5"}` instead of `{"score": 5}`) because the ambiguity check needs the raw text. `bootstrap_trace.py` is the only reader and it just uses `.keys()`.
- `XMLAdapter.field_pattern` is gone. Nothing in the repo used it, but it was a public-looking attribute on an exported class, so anyone overriding it downstream gets an `AttributeError`. Probably worth a release note.
- Single-chunk responses don't emit a StreamResponse, but that's the existing `cache_hit` path and it does the same for plain values on main.

## One question

`STREAM_NESTED_XML_VALUES` in `dspy/streaming/streaming_listener.py` flips this back in one line. I left it undocumented and out of `__all__` since I didn't want to commit the project to supporting it. Happy to wire it into `dspy.settings`, or just delete it and let `git revert` be the escape hatch.

Unrelated, for triage: #8482 rewrites XMLAdapter on ElementTree for nested Pydantic models (#8481). It doesn't fix this bug, and it'd reject things the adapter accepts today (bare `&`, stray tags, duplicate closing tags) since ElementTree wants well-formed XML.

TLXyloph:
Another round since `4f1f4d03`, which is the last commit the review saw. Head is now `1420a6ca`. What changed:

- Rebased onto current main.
- Two follow-ups from the earlier review rounds: the span acceptor builds its prefix data lazily on the first candidate span now, so the no-nesting path (nearly every completion) stops paying for it, and the stream listener holds a chunk whose nesting is still undecided instead of guessing. Also dropped a duplicate stream buffer and made the tests count parse work.
- Actually new: `format` now escapes `<` and `&` in output values as `&lt;` / `&amp;`, so a value containing its own closing tag survives a format→parse round trip losslessly instead of leaning on the recovery scan. The tradeoff is that `parse` decodes those entities unconditionally, so a legacy completion that literally spells `&lt;` reads back decoded now. The adapters guide covers it.
- A ruff format pass on top, no behavior in it.

I re-ran the exact repro shapes from both review passes against the new head. The split sibling (`<answer><answer>inner</answer>` then `<ot` + `her>sibling</other>`) streams `<answer>inner` and matches parse, and the token-split nested value streams `<answer>x = 1</answer>`, byte-identical to parse. The chunk-boundary tests pin both shapes.

That leaves the one-chunk case the summary still flags: `<answer><answer>inner</answer></answer><other>sibling</other>` arriving whole in a single chunk. Greptile found that the stream side emits nothing there, and that's true on this head too. But the same harness on main with a plain value in one chunk — `<answer>plain</answer><other>sibling</other>` — also emits nothing. A completion that arrives in one chunk never yields a StreamResponse; that's the cache-hit path, it's adapter-agnostic, and it predates this branch. What this PR changes is what the consumer actually gets at the end:

```
one chunk: <answer><answer>inner</answer></answer><other>sibling</other>

           streamed   final prediction
main       ''         <answer>inner              (truncated)
this PR    ''         <answer>inner</answer>

main, plain one-chunk <answer>plain</answer>:   streamed '' as well
```

So the value isn't dropped, it arrives the way every one-chunk response arrives, and it's no longer truncated on the way. Making the cache-hit path emit a final chunk would be a real behavior change for ChatAdapter's one-chunk responses too, which felt out of scope for this fix. If that contract is worth changing I'd rather do it as a follow-up; in the meantime I can add a small test pinning "one-chunk nested value lands whole in the prediction" if that helps the next pass.

TLXyloph:
The sibling-masked case from the last pass is fixed. Head is now `ae39d1fc`.

Greptile found that `<answer><answer>nested</answer><other>sibling</other></answer>` parsed to a truncated `<answer>nested` with no error: the widened span gets rejected for containing `other`, the lazy reading stands, and the sibling's masked block hides the surplus `</answer>` from the ambiguity check. That reproduced exactly as described.

The suggested fix — reject a declared field's leftover closing tag even when everything in between is masked — would have gone too far, though. `<answer>v</answer><other>s</other></answer>` has the same leftover tag with everything in between masked, and there the value is complete and the tag is just a duplicated-closer slip, which the adapter deliberately tolerates (there are four tests pinning that). What separates the two shapes is *why* the widening was rejected. So the span acceptor now reports its reason, and only the case where a structurally sound span holds a declared sibling raises — that's the one where both readings genuinely lose text the model wrote. A span rejected for borrowing structure from another element keeps the silent lazy fallback (the wide reading was never believable, so there's no second reading to be ambiguous against), and duplicate closers keep parsing.

Tests went in red-first: the exact completion from the review now raises in `parse`, the streaming run for it ends in the parse error instead of a plausible truncated prediction (the stream still emits the lazy reading as tokens arrive — a sent chunk can't be recalled), and the stitched and duplicate-closer counter-shapes are asserted in the same test so the narrow scope is pinned too. One earlier test asserted the silent truncation as correct behavior; it now asserts the raise. Also added the one-chunk regression promised last round: a nested value arriving whole in a single chunk lands untruncated in the prediction.

Docs and docstrings updated to match: the silently-truncated list is down to two shapes (a value ending with its own closing tag, and a stitched-rejected span whose surplus tag a later block covers or separates), and the streaming guide notes the ambiguous shape ends in the parse error. Full suites: 358 adapter tests, streaming green apart from the two pre-existing litellm-proxy fixture errors.

---

Update: the re-review of `ae39d1fc` generated no comments — the sibling case is closed, and the runtime checks re-verified the escape paths end to end, including chasing a suspected `&amp;lt;` double-decode and disproving it (the literal `&lt;` comes back intact). The merge-blocker note is gone and the CI check is green. One thing I can't explain: the confidence score reads 3/5 with an empty rationale, down from 4/5 last pass despite zero findings. If there's something behind that number I'm happy to dig, but as of this pass there's nothing actionable left in the review.

Issue #10067, opened 2026-07-19
dspy.Image/dspy.Audio output-field parsing reads arbitrary local files into the next prompt (file exfiltration via untrusted LM output)


geo-chen:
### What happened?

reported on 13 June 2026 - https://github.com/stanfordnlp/dspy/security/advisories/GHSA-frf5-486g-243f

### Summary

When a DSPy signature has an output field typed `dspy.Image` (or `dspy.Audio`), the adapter parses the language model's returned text into that type, and the type constructor treats a string that happens to be a filesystem path as a local file to read and base64-encode. Because the LM completion is an untrusted channel in agentic/RAG/ReAct loops (prompt injection, or a malicious/compromised model endpoint), a model can be steered to emit `{"out_field": {"url": "/etc/passwd"}}`, causing DSPy to read that file on the victim host and, when the result is round-tripped into a subsequent prompt, send its contents to the attacker-influenced LM endpoint. Confirmed end-to-end against the `JSONAdapter.parse`: a secret file was read into the parsed object and embedded in the outgoing messages.

### Details

Sink: `dspy/adapters/types/image.py:150-152` `encode_image`:

```python
elif os.path.isfile(image):
    return _encode_image_from_file(image)   # open(path,"rb").read() + base64, ~179-190
```

The same pattern is in `dspy/adapters/types/audio.py:146` (`encode_audio` -> `encode_audio_from_file`, with audio constrained to a valid audio MIME).

Source/flow: the LM completion string. `JSONAdapter.parse` (`json_adapter.py:190`) / `ChatAdapter.parse` (`chat_adapter.py:237`) call `parse_value(v, annotation)` (`adapters/utils.py:190`) -> `TypeAdapter(Image).validate_python({"url": "<path>"})` -> `Image.__init__` -> `encode_image(..., download_images=False)`. The local-file branch fires regardless of `download` flags. When the parsed `Image` is later serialized into a prompt (`Type.serialize_model` -> `Image.format()` -> `encode_image`, `image.py:76`), the file contents are embedded in the outgoing user message and sent to the LM endpoint, i.e. exfiltration.

[...]

PR #10111, opened 2026-07-30
fix(adapters): prevent implicit resource loading during validation

isaacbmiller:
## Before / after

Before, constructing or parsing a DSPy resource type could interpret a string as a locator and silently read a host file or fetch a remote URL; after, construction and validation do not dereference locator-shaped inputs, while filesystem and network I/O require an explicit `from_path()` / `from_url()` call or the narrowly retained deprecated direct call `Image(url, download=True)`.

## Summary

- **[SECURITY HARDENING]** Make `Image`, `Audio`, and `File` construction and LM-output validation free of implicit DSPy-initiated filesystem and network access.
- **[FIX]** Prevent Pydantic dictionary validation from activating the deprecated `Image(..., download=True)` compatibility shim.
- **[API]** Make `from_path()` the canonical local-read factory and `from_url()` the explicit remote-download factory.
- **[MIGRATION]** Preserve safe positional/in-memory construction and existing helper signatures instead of requiring all values to move to named factories.
- **[BREAKING / DELIBERATE]** Reject implicit local paths and locator-shaped embedded audio/file values without touching the host.

This is the lower-migration-churn alternative to #10069.

## Why this boundary

DSPy adapters parse LM output through Pydantic. Previously, that validation path reused convenience coercion that could call `open()` or `requests.get()`. A value being validated as data could therefore cause host I/O merely because it resembled a path or URL—or, in the image case, because a dictionary contained `download=True`.

This is defense-in-depth hardening, not a claim of a high-severity practical exploit. The reported PoCs require an unusual signature with `Image` / `Audio` / `File` output fields. The useful invariant is: **validation does not dereference locators; host I/O must be requested through an explicit API.**

The deprecated positional `Image(url, download=True)` shim remains for one release because it represents a direct developer-authored request, not coercion of untrusted dictionary data. Pydantic invokes the custom constructor with keyword data, so `download` and `verify` now require a positional source and are rejected from `TypeAdapter`, nested-model, and deserialization paths.

### Python and ecosystem rationale

Python does not prohibit side effects in `__init__`; this PR draws a narrower capability boundary:

- [PEP 20](https://peps.python.org/pep-0020/) says “Explicit is better than implicit.” Calls named `from_path()` and `from_url()` expose host I/O at the call site; bytes, PIL images, arrays, and embedded data remain in-memory transformations.
- Python's [`pathlib` design](https://docs.python.org/3/library/pathlib.html#pure-paths) separates path representation from filesystem operations such as [`Path.open()` and `Path.read_bytes()`](https://docs.python.org/3/library/pathlib.html#reading-and-writing-files). DSPy now similarly separates locator-shaped data from explicit dereferencing.
- Pydantic documents [`TypeAdapter`](https://docs.pydantic.dev/latest/concepts/type_adapter/) as a validation boundary for arbitrary types. DSPy uses this path for LM output, nested models, and deserialization, so it must not grant implicit access to the host.
- Pydantic [discourages custom `BaseModel.__init__` implementations](https://docs.pydantic.dev/latest/concepts/models/#defining-a-custom-init). This alternative retains the existing constructors as a migration constraint, but isolates and tests the security-sensitive compatibility branch rather than treating it as general validation behavior.
- The [OWASP SSRF Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html) treats application-initiated requests to caller-controlled URLs as a security boundary. `from_url()` remains explicit and caller-owned; callers must validate untrusted destinations for their environment.

## Security and I/O contract

| Input | Construction / validation | Explicit operation |
|---|---|---|
| Local image path | rejected without touching the filesystem | `Image.from_path(path)` |
| Local audio path | rejected without touching the filesystem | `Audio.from_path(path)` |
| Local file path | rejected without touching the filesystem | `File.from_path(path)` |
| HTTP(S) image URL | retained as a provider-fetched reference | `Image.from_url(url)` downloads into a data URI |
| HTTP(S) audio URL | rejected as embedded audio data | `Audio.from_url(url)` downloads and embeds it |
| URL-shaped file data | rejected | no `File.from_url()` in this PR |
| Raw bytes / arrays / PIL objects | normalized in memory; no host I/O | existing constructors and factories remain available |
| Validation payload containing `download` / `verify` | rejected without fetching | no validation-time migration by design |
| Direct positional `Image(url, download=True)` | downloads with a deprecation warning | migrate to `Image.from_url(url)` |

`from_url()` is a synchronous, caller-initiated network request. It follows redirects and provides no SSRF protection beyond requiring an HTTP(S) URL. If a URL is derived from untrusted input, the caller owns destination validation or allowlisting.

## Public API after this PR

| Capability | Image | Audio | File |
|---|---|---|---|
| In-memory constructor | `Image(URL_or_data_URI)`, `Image(bytes)`, `Image(PIL_image)` | `Audio(data_URI_or_dict)`, `Audio(bytes, audio_format=...)`, array input with `sampling_rate=` | `File(data_URI_or_dict)`, `File(bytes)` |
| Canonical keyword fields | `Image(url=...)` | `Audio(data=<base64>, audio_format=...)` | `File(file_data=..., file_id=..., filename=...)` |
| Non-fetching remote reference | positional source or keyword `url=` | — | provider `file_id=` only |
| Read local resource | `from_path()` | `from_path()` | `from_path()` |
| Download remote resource | `from_url()` | `from_url()` | — |
| Existing specialized factory | deprecated `from_PIL()` | `from_array()` | `from_bytes()`, `from_file_id()` |

Deprecated APIs retained through 3.3 and scheduled for removal in 3.4:

- `Image.from_file()` → `Image.from_path()`
- `Image.from_PIL()` → `Image(PIL_image)`
- `Audio.from_file()` → `Audio.from_path()`
- `Image(url, download=True)` → `Image.from_url(url)`
- Constructor `download` / `verify` arguments generally → explicit reference construction or `Image.from_url(url, verify=...)`

### Constructors and validation

| Previous call / behavior | New behavior | Migration |
|---|---|---|
| `Image(local_path)` implicitly read a file | rejects without reading | `Image.from_path(path)` |
| `Image(url=local_path)` could be interpreted as a file | rejects without reading | `Image.from_path(path)` |
| Pydantic validation of `{"url": ..., "download": true}` could fetch | rejects without fetching | explicit application code may call `Image.from_url()` |
| `Image(url=..., download=...)` keyword-only construction | rejects because the shim is direct-positional only | use `Image(url=...)` for a reference or `Image.from_url()` to download |
| Direct `Image(url, download=True)` | still downloads, but warns | migrate to `Image.from_url(url)` before 3.4 |
| `Audio(path_or_url)` could be treated as embedded data or loaded | rejects without reading/fetching | `Audio.from_path()` / `Audio.from_url()` |
| `File(path)` could read the filesystem | rejects without reading | `File.from_path()` |
| LM-output parsing of locator-shaped resources | rejects or retains a non-fetching image reference; never dereferences | no implicit-I/O migration by design |

### Factories and helpers

| Previous call / behavior | New behavior | Migration |
|---|---|---|
| `Image.from_url(url)` retained a reference by default | downloads and returns a data URI | use `Image(url)` / `Image(url=url)` for a reference |
| `Image.from_url(url, download=...)` | `download` removed | choose reference construction or `from_url()` explicitly |
| `encode_image(path)` | rejects local paths | `Image.from_path(path)` |
| `encode_image(..., download_images=..., verify=...)` | options removed | `Image.from_url(..., verify=...)` for a fetch |
| `encode_audio(path_or_url)` | rejects locator strings | `Audio.from_path()` / `Audio.from_url()` |
| `encode_file_to_dict(path)` | rejects local paths | `File.from_path(path)` |
| Image/Audio `from_file()` | deprecated alias | `from_path()` |

## Unchanged / intentionally out of scope

- Safe in-memory positional construction remains supported for bytes, PIL images, arrays, data URIs, dictionaries, and existing resource instances where previously accepted.
- Existing `encode_audio(..., format=...)`, `Audio.from_array(..., format=...)`, and positional optional arguments are not renamed or made keyword-only.
- Embedded audio/file payload validation is not tightened beyond the locator/no-host-I/O boundary in this alternative.
- `Image(url="https://...")` remains a non-fetching reference; the downstream model provider may fetch it.
- The custom Pydantic constructors remain temporarily; this PR secures their validation boundary rather than redesigning the entire resource API.
- No async resource-loading API or `File.from_url()` is added.
- DSPy does not implement a universal SSRF allowlist; trust policy remains with the caller explicitly invoking `from_url()`.

## Review guide

1. `dspy/adapters/types/image.py`: non-fetching construction, explicit image loading, and the direct-positional compatibility guard.
2. `dspy/adapters/types/audio.py`: in-memory normalization plus explicit path/URL factories.
3. `dspy/adapters/types/file.py`: in-memory normalization plus explicit path factory.
4. `tests/adapters/test_resource_loading.py`: constructor and `TypeAdapter.validate_python()` host-I/O boundary, exact metadata-endpoint regression, compatibility-shim coverage, and explicit factories.
5. Resource documentation under `docs/docs/api/primitives/` and `docs/docs/diving-deeper/adapters.md`.

## Validation

- `uv run --frozen pytest --deno -q tests/adapters/test_resource_loading.py`
- 30 tests passed locally.
- `test_construction_performs_no_host_io` and `test_validation_performs_no_host_io` forbid `open()` and `requests.get()` for locator-shaped inputs.
- `test_image_validation_rejects_download_without_host_io` covers the exact `download=True` metadata-endpoint payload.
- `test_image_positional_download_compatibility_shim` proves the intentionally retained direct-call migration path still works and warns.
- Full GitHub CI is pending on this draft.

Addresses #10067.

### DIFF
--- a/docs/docs/api/primitives/Audio.md
+++ b/docs/docs/api/primitives/Audio.md
@@ -10,7 +10,7 @@
             - extract_custom_type_from_annotation
             - format
             - from_array
-            - from_file
+            - from_path
             - from_url
             - is_streamable
             - parse_lm_response
--- a/docs/docs/api/primitives/Audio.md
+++ b/docs/docs/api/primitives/Audio.md
@@ -10,6 +10,7 @@
             - extract_custom_type_from_annotation
             - format
             - from_array
+            - from_file
             - from_path
             - from_url
             - is_streamable
--- a/docs/docs/api/primitives/Image.md
+++ b/docs/docs/api/primitives/Image.md
@@ -10,6 +10,7 @@
             - extract_custom_type_from_annotation
             - format
             - from_PIL
+            - from_file
             - from_path
             - from_url
             - is_streamable
--- a/docs/docs/api/primitives/Image.md
+++ b/docs/docs/api/primitives/Image.md
@@ -10,7 +10,7 @@
             - extract_custom_type_from_annotation
             - format
             - from_PIL
-            - from_file
+            - from_path
             - from_url
             - is_streamable
             - parse_lm_response
--- a/docs/docs/diving-deeper/adapters.md
+++ b/docs/docs/diving-deeper/adapters.md
@@ -127,9 +127,9 @@ The base class. Subclass it (it’s a `pydantic.BaseModel`) and implement `forma
 
 URL reference, data URI, bytes, or PIL image. `format()` returns the provider’s image content block (`{"type": "image_url", "image_url": {"url": ...}}`). Constructors and adapter parsing never access the filesystem or network. Use `Image.from_file(path)` to read a local file or `Image.from_url(url)` to download and base64-encode a remote resource.
 
-**`dspy.Audio(data, audio_format)`**
+**`dspy.Audio(source)`**
 
-Base64 audio data plus format string (`"wav"`, `"mp3"`), or in-memory bytes/array data. Renders as the provider’s audio content block. Use `Audio.from_file(path)` or `Audio.from_url(url)` for resource loading.
+A data URI, in-memory bytes, or array data; raw base64 must be passed as `Audio(data=..., audio_format=...)`. Renders as the provider’s audio content block. Use `Audio.from_file(path)` or `Audio.from_url(url)` for resource loading.
 
 **`dspy.File(file_data=None, file_id=None, filename=None)`**
 
--- a/docs/docs/diving-deeper/adapters.md
+++ b/docs/docs/diving-deeper/adapters.md
@@ -123,14 +123,17 @@ Types adapters know how to render and parse beyond Python’s standard ones. Eac
 **`dspy.adapters.types.Type`**  
 The base class. Subclass it (it’s a `pydantic.BaseModel`) and implement `format()` to plug in a new type. Adapters wrap the output of `format()` with `<<CUSTOM-TYPE-START-IDENTIFIER>>...<<END-IDENTIFIER>>` so multi-modal content can be inserted into a single message stream and later split out.
 
-**`dspy.Image(url, download=False, verify=True)`**  
-URL, local path, bytes, or PIL image. `format()` returns the provider’s image content block (`{"type": "image_url", "image_url": {"url": ...}}`). `download=True` fetches the image and base64-encodes it, useful when the LM provider can’t reach the URL.
+**`dspy.Image(url)`**
 
-**`dspy.Audio(data, audio_format)`**  
-Base64 audio data plus format string (`"wav"`, `"mp3"`). Renders as the provider’s audio content block.
+URL reference, data URI, bytes, or PIL image. `format()` returns the provider’s image content block (`{"type": "image_url", "image_url": {"url": ...}}`). Constructors and adapter parsing never access the filesystem or network. Use `Image.from_file(path)` to read a local file or `Image.from_url(url)` to download and base64-encode a remote resource.
 
-**`dspy.File(file_data=None, file_id=None, filename=None)`**  
-Either a data URI or a file ID (some providers preupload files and reference them by ID).
+**`dspy.Audio(data, audio_format)`**
+
+Base64 audio data plus format string (`"wav"`, `"mp3"`), or in-memory bytes/array data. Renders as the provider’s audio content block. Use `Audio.from_file(path)` or `Audio.from_url(url)` for resource loading.
+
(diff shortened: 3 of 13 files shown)


isaacbmiller:
discussed offline

PR #10112, opened 2026-07-31
fix: deduplicate signatures by content in get_dspy_source_code (fixes silent drop of dynamic signatures)

### LINKED ISSUE #10100: [Bug] get_dspy_source_code drops every dynamic signature after the first from program_code_string
### What happened?

`get_dspy_source_code` (`dspy/propose/utils.py`) de-duplicates signatures with

```python
item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig"
```

For anything built through `make_signature` — `dspy.Signature("q -> a")`, `ChainOfThought`, `.prepend`, `.with_instructions` — that key is the literal string `"Signature"`, the same for every one of them. So the first dynamic signature in a program lands in `program_code_string` and every later one is silently skipped.

`grounded_proposer` passes that string to the instruction-proposing LM as `program_code_string`, so from MIPROv2's point of view the second predictor's signature does not exist.

Where `"Signature"` comes from: pydantic captures the calling frame's locals into `__pydantic_parent_namespace__`, and the caller here is dspy's own `SignatureMeta.__new__`, whose first parameter is named `signature_name` (`dspy/signatures/signature.py:138`). The value read is the metaclass's local variable, not a property of the signature. `__name__` doesn't disambiguate them either — it is `StringSignature` for all of them.

### Steps to reproduce

```python
import dspy
from dspy.propose.utils import get_dspy_source_code

    def __init__(self):
        super().__init__()
        self.qa = dspy.Predict(QA)

code = get_dspy_source_code(TwoDynamicSignatures())
header = code.split("class TwoDynamic")[0]
print("-> answer  in header:", "-> answer" in header)   # True
print("-> summary in header:", "-> summary" in header)  # False
```

| | key | in `program_code_string` |
|---|---|---|
| `dspy.Signature("question -> answer")` | `Signature_sig` | yes |
| `dspy.Signature("text -> summary")` | `Signature_sig` | **no** |
| `ChainOfThought(QA).predict.signature` | `Signature_sig` | **no** |

Class-defined signatures are unaffected — `signature_name` is the real class name there, so `QA_sig` and `Summarize_sig` differ.

Filing separately from #9937 because that one is the `TypeError` when the attribute is `None` (Python 3.14, where `SignatureMeta` sets `__pydantic_reset_parent_namespace__ = False`). This is the ≤3.13 path where the attribute is populated and the value collides. Guarding the `None` alone doesn't reach it — I checked #9941 and the `__name__` fallback collides the same way, since it is `StringSignature` for all of them.

The `dspy.Module` branch four lines below already keys on the emitted source, which sidesteps both.

### DSPy version

`main` @ 67bf5eb, pydantic 2.12.5, Python 3.10.12

AbdullahRasheed45:
## Bug

`get_dspy_source_code` (`dspy/propose/utils.py`) only includes the **first** dynamic signature when a program has multiple — every subsequent one is silently dropped from `program_code_string`. This means MIPROv2's `grounded_proposer` is passed a truncated view of the program and any predictor after the first is invisible to the optimizer.

Fixes #10100.

## Root cause

The deduplication key is:

```python
item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig"
```

For any signature built dynamically (`dspy.Signature("q -> a")`, `ChainOfThought`, `.prepend`, `.with_instructions`, …), Pydantic captures `SignatureMeta.__new__`'s own local variable `signature_name`, whose value is always the string literal `"Signature"`. Every dynamic signature therefore maps to the same key `"Signature_sig"`, so only the first one passes the `not in completed_set` guard.

```python
import dspy
from dspy.propose.utils import get_dspy_source_code

class TwoDynamic(dspy.Module):
    def __init__(self):
        super().__init__()
        self.qa = dspy.Predict(QA)

code = get_dspy_source_code(TwoDynamic())
header = code.split("class TwoDynamic")[0]
print("-> answer  in header:", "-> answer"  in header)   # True
print("-> summary in header:", "-> summary" in header)   # False  ← BUG
```

## Fix

Use the signature's actual source/string representation as the deduplication key — already the approach used four lines below for `dspy.Module` items:

```python
# Before
if hasattr(item, "signature") and item.signature is not None and item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig" not in completed_set:
    try:
        header.append(inspect.getsource(item.signature))
        print(inspect.getsource(item.signature))          # leftover debug print
    except (TypeError, OSError):
        header.append(str(item.signature))
    completed_set.add(item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig")

# After
if hasattr(item, "signature") and item.signature is not None:
    try:
        sig_source = inspect.getsource(item.signature)
    except (TypeError, OSError):
        sig_source = str(item.signature)
    if sig_source not in completed_set:
        header.append(sig_source)
        completed_set.add(sig_source)
```

Also removes the debug `print()` that was logging signature source to stdout on every `get_dspy_source_code` call.

### DIFF
--- a/dspy/propose/utils.py
+++ b/dspy/propose/utils.py
@@ -173,6 +173,9 @@ def get_dspy_source_code(module):
                 continue
             if isinstance(item, Parameter):
                 if hasattr(item, "signature") and item.signature is not None:
+                    # Which branch runs is determined by the signature's origin (source availability),
+                    # not by call order, so a given signature always yields the same sig_source here
+                    # and can't be emitted twice under two different representations.
                     try:
                         sig_source = inspect.getsource(item.signature)
                     except (TypeError, OSError):
--- a/dspy/propose/utils.py
+++ b/dspy/propose/utils.py
@@ -172,13 +172,14 @@ def get_dspy_source_code(module):
             except TypeError:
                 continue
             if isinstance(item, Parameter):
-                if hasattr(item, "signature") and item.signature is not None and item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig" not in completed_set:
+                if hasattr(item, "signature") and item.signature is not None:
                     try:
-                        header.append(inspect.getsource(item.signature))
-                        print(inspect.getsource(item.signature))
+                        sig_source = inspect.getsource(item.signature)
                     except (TypeError, OSError):
-                        header.append(str(item.signature))
-                    completed_set.add(item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig")
+                        sig_source = str(item.signature)
+                    if sig_source not in completed_set:
+                        header.append(sig_source)
+                        completed_set.add(sig_source)
             if isinstance(item, dspy.Module):
                 code = get_dspy_source_code(item).strip()
                 if code not in completed_set:


ErenAta16 (commented):
Carrying the analysis over from #10113, which @saket3395 has just closed in favour of this one, so the reasoning is not lost with that thread.

The bug both PRs fix is that `__pydantic_parent_namespace__["signature_name"]` is not derived from the signature's content. It comes from `SignatureMeta.__new__`'s parameter name, so every `dspy.Signature("a -> b")` built the same way collides on one key and only the first survives. MIPROv2 then proposes instructions having seen a single signature out of however many the program defines, with nothing logged.

**Deduping on the emitted text is the right key, and it is the reason this approach is better than the alternative.** #10113 keyed on `id(item.signature)`, which fixes the drop but stops collapsing genuinely identical signatures, so the proposer can receive the same source twice. This PR keys on `sig_source`:

```python
if sig_source not in completed_set:
    header.append(sig_source)
    completed_set.add(sig_source)
```

which is exactly the string being appended. Since the stated purpose of `completed_set` is to avoid redundant signature text in what goes to the instruction-proposing LM, keying on the text itself makes the guard match its purpose precisely: distinct signatures all survive, byte-identical ones appear once, and there is no proxy that can drift from what is actually emitted.

Removing the stray `print(inspect.getsource(item.signature))` is worth calling out in the description. It writes signature source to stdout on every call today, which is noise in any program that uses MIPROv2.

One thing worth confirming since the key now depends on which branch runs: the `inspect.getsource` path and the `str(item.signature)` path produce different text for the same signature. That is fine as long as a given signature always takes the same branch, which I would expect since the branch is a property of how the signature was defined rather than of call order. If a signature could take different branches across two visits it would be emitted twice, so a one-line comment noting the invariant would be cheap insurance.

Nothing blocking. This is the one to merge.

PR #10113, opened 2026-07-31
Fix get_dspy_source_code dropping dynamic signatures after the first

### LINKED ISSUE #10100: [Bug] get_dspy_source_code drops every dynamic signature after the first from program_code_string

saket3395:
## Summary
Fixes #10100 — `get_dspy_source_code` (`dspy/propose/utils.py`) silently drops every dynamically-built signature in a program after the first, so `grounded_proposer`/MIPROv2 never sees them when proposing instructions.

**Root cause (as diagnosed in the issue):** the function dedupes signatures using

```python
item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig"
```

For any signature built through `make_signature` (`dspy.Signature("q -> a")`, `.prepend`, `.with_instructions`, `ChainOfThought`, ...), `__pydantic_parent_namespace__` is pydantic capturing the *calling frame's locals* — and the caller here is always `SignatureMeta.__new__`, whose first parameter happens to be named `signature_name`. So this key evaluates to the literal string `"Signature"` for every dynamically-built signature, regardless of its actual content. The first one populates `program_code_string`; every later one looks like a duplicate of the first and gets dropped.

## Fix
Dedupe on `id(item.signature)` instead of the shared metaclass-parameter string:

```python
if hasattr(item, "signature") and item.signature is not None and id(item.signature) not in completed_set:
    ...
    completed_set.add(id(item.signature))
```

`completed_set` is local to a single call of `get_dspy_source_code`, so using object identity is exactly the right granularity here — it prevents re-adding the *same* signature object twice (the original intent), without conflating distinct signature instances that happen to share a class name.

## Verification
Traced the reproduction from the issue by hand against the patched logic:
- `QA = dspy.Signature("question -> answer")`, `SUM = dspy.Signature("text -> summary")` — two distinct objects, previously both keyed to `"Signature_sig"` (second one dropped); with `id()`, each has its own key, so both are retained.
- Class-defined signatures were already unaffected by the bug and remain unaffected by this change (they don't go through `make_signature`, so `id()` dedup applies to them the same way the old key did).

Wasn't able to run this against dspy's actual test suite locally — dspy pulls in a fairly heavy dependency chain (litellm, pydantic, etc.) and I didn't want to install a new stack just to verify a two-line fix. Flagging for CI/maintainer verification, happy to iterate.

## Test plan
- [x] Traced the exact repro from #10100 by hand against the patched code.
- [ ] Not run through dspy's test suite locally — CI/maintainer verification appreciated.

### DIFF
--- a/dspy/propose/utils.py
+++ b/dspy/propose/utils.py
@@ -172,13 +172,13 @@ def get_dspy_source_code(module):
             except TypeError:
                 continue
             if isinstance(item, Parameter):
-                if hasattr(item, "signature") and item.signature is not None and item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig" not in completed_set:
+                if hasattr(item, "signature") and item.signature is not None and id(item.signature) not in completed_set:
                     try:
                         header.append(inspect.getsource(item.signature))
                         print(inspect.getsource(item.signature))
                     except (TypeError, OSError):
                         header.append(str(item.signature))
-                    completed_set.add(item.signature.__pydantic_parent_namespace__["signature_name"] + "_sig")
+                    completed_set.add(id(item.signature))
             if isinstance(item, dspy.Module):
                 code = get_dspy_source_code(item).strip()
                 if code not in completed_set:
--- a/tests/propose/test_utils.py
+++ b/tests/propose/test_utils.py
@@ -1,0 +1,39 @@
+import dspy
+from dspy.propose.utils import get_dspy_source_code
+
+
+class TwoDynamicSignatures(dspy.Module):
+    def __init__(self):
+        super().__init__()
+        self.qa = dspy.Predict(dspy.Signature("question -> answer"))
+        self.summarize = dspy.Predict(dspy.Signature("text -> summary"))
+
+
+def test_get_dspy_source_code_keeps_distinct_dynamic_signatures():
+    # Two separately constructed dynamic signatures share the same
+    # __pydantic_parent_namespace__["signature_name"] ("Signature"), since that
+    # value comes from SignatureMeta.__new__'s parameter name, not the
+    # signature's own content. Deduping on that string alone would drop the
+    # second signature; deduping on object identity must not.
+    code = get_dspy_source_code(TwoDynamicSignatures())
+    header = code.split("class TwoDynamicSignatures")[0]
+
+    assert "-> answer" in header
+    assert "-> summary" in header
+
+
+class RepeatedSignatureReference(dspy.Module):
+    def __init__(self):
+        super().__init__()
+        shared = dspy.Signature("question -> answer")
+        self.first = dspy.Predict(shared)
+        self.second = dspy.Predict(shared)
+
+
+def test_get_dspy_source_code_dedupes_repeated_signature_reference():
+    # Two Predictors referencing the *same* signature object should still
+    # only contribute one copy of that signature's source to the header.
+    code = get_dspy_source_code(RepeatedSignatureReference())
+    header = code.split("class RepeatedSignatureReference")[0]
+
+    assert header.count("-> answer") == 1


saket3395:
Added regression coverage in `tests/propose/test_utils.py` addressing the review feedback: one test asserting two distinct dynamic signatures are both retained, and one asserting a signature referenced by two Predictors is only emitted once. Confirmed `ensure_signature` doesn't clone/copy an already-constructed Signature, so the second test is a meaningful check of the identity-based dedup rather than a tautology.

ErenAta16 (commented):
#10112 is open against the same function for the same bug, by a different author. Both fix it; they differ on one case, so flagging it rather than letting two PRs get reviewed separately.

The root cause checks out. Every dynamically built signature carries the same parent-namespace name:

```
sig0: signature_name='Signature'  fields=['a0']
sig1: signature_name='Signature'  fields=['a1']
sig2: signature_name='Signature'  fields=['a2']
```

so `"Signature_sig"` lands in `completed_set` after the first one and the rest are skipped.

Ran main and both dedupe keys over the same two modules:

| | 3 distinct dynamic signatures | 2 signatures with identical content |
|---|---|---|
| `main` | **1 emitted** (two dropped) | 1 |
| this PR — `id(signature)` | 3 | **2** |
| #10112 — source text | 3 | 1 |

Both recover the dropped signatures, which is the reported bug. The difference is the second column: keying on `id()` makes two distinct-but-identical signature objects each emit their source, so the proposer prompt gets the same block twice. `main` doesn't do that, so it's a new case rather than a pre-existing one. Whether that matters depends on whether duplicate blocks in the generated prompt are considered harmless.

Two smaller things in favour of #10112's shape:

`id()` is only unique among live objects, so it's a fragile identity key in general. Here the signatures are held by the module for the duration of the call, so I couldn't construct a case where it actually collides — noting it as a property of the approach, not a bug I can demonstrate.

`print(inspect.getsource(item.signature))` on the line below writes to stdout on every call. #10112 drops it; this PR keeps it. That looks like leftover debugging rather than intended output.

Nothing else in either diff changes behaviour, and both leave the `dspy.Module` branch below untouched.

ErenAta16 (commented):
The diagnosis is right: `__pydantic_parent_namespace__["signature_name"]` is not derived from the signature's content, so every `dspy.Signature("a -> b")` built the same way lands on the same key and only the first survives the dedupe. For MIPROv2 that means the proposer silently sees one signature out of however many the program defines, which is a bad failure because nothing errors and the proposals just come out underinformed.

**`id()` is safe in this loop, and it is worth saying why**, because `id()` as a dict key normally deserves suspicion: CPython reuses addresses after collection, so a stale `id` can collide with a freshly allocated object. That cannot happen here, since every `item.signature` is reachable from the module being walked for the whole duration of the walk, so nothing is collected mid-iteration and no address is recycled. A one-line comment saying that would stop a future reader from "fixing" it.

**The question is what the dedupe is for.** The old key, broken as it was, would have collapsed two separately constructed but identical signatures into one. `id()` will not: two `dspy.Signature("question -> answer")` objects created independently now both emit their source, so the returned code can contain the same signature twice.

If the set exists to avoid redundant text in what gets handed to the proposer, that is a regression in the other direction, and keying on the signature's content, its field names and instructions, would satisfy both goals: distinct signatures all survive, identical ones appear once. If the set exists only to avoid infinite recursion or reprocessing on shared references, then `id()` is exactly right and there is nothing to change.

Worth answering in the description either way, because "we dedupe by object identity" and "we dedupe by content" are different contracts and the next person to touch this will need to know which one was intended.

The test comment is doing real work. Spelling out that `signature_name` comes from `SignatureMeta.__new__`'s parameter name rather than from the signature itself is the non-obvious fact the whole bug rests on, and it is the sentence that stops someone reverting to a string key later.

saket3395:
@ErenAta16 Thanks for the thorough side-by-side — you're right, and it settles the open question cleanly.

The purpose of `completed_set` here is to avoid redundant signature text in what gets handed to the instruction-proposing LM, not to guard against reprocessing a live object. Given that goal, content-based dedup is the correct contract: it recovers all the dropped signatures (fixing the reported bug) *and* collapses genuinely duplicate content, which `id()` can't do — two independently-built signatures with identical content would each emit their source under this PR's approach.

#10112 implements content-based dedup and also drops the leftover `print(inspect.getsource(item.signature))` debug statement that I left in, which this PR doesn't. Rather than ask the maintainers to arbitrate between two PRs fixing the same bug, I'm closing this one in favor of #10112.

PR #10114, opened 2026-07-31
fix(dspy): enforce output field constraints when parsing LM responses

### LINKED ISSUE #7925: [Bug] The parse function in dspy's Adapters (ChatAdapter, JSONAdapter) can only partially validate output fields.
### What happened?

The current implementation is as follows.
```
import pydantic

class MyModel(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(validate_assignment=True)
    value:str = pydantic.Field(max_length=3)

fi = MyModel.model_fields["value"]

pydantic.TypeAdapter(fi.annotation).validate_python("1234")
```
With this approach, validating "1234" will not raise an error.

Image

The following approach can be used to validate with complete constraints.

```
getattr(MyModel.__pydantic_validator__.validate_assignment(MyModel.model_construct(), "value", "1234"),"value")
```

Image

The above approach is based on a Pydantic issue, where you can find more detailed discussions.

https://github.com/pydantic/pydantic/discussions/7367

### Steps to reproduce

import pydantic

class MyModel(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(validate_assignment=True)
    value:str = pydantic.Field(max_length=3)

fi = MyModel.model_fields["value"]

pydantic.TypeAdapter(fi.annotation).validate_python("1234")

### DSPy version

2.6.10

cardoner1993:
## Changes Description

Fixes #7925

A signature can declare constraints on an output field, and DSPy already renders them into the prompt; a field declared as `dspy.OutputField(max_length=3)` carries `'constraints': 'maximum length: 3'` and the model is told about it. The response was never checked against them, so DSPy stated a rule and then accepted any violation of it silently

The reason is that constraints are stored in `FieldInfo.metadata` rather than in the annotation, and every adapter validated with `parse_value(value, field.annotation)`. A `TypeAdapter` built from the annotation alone enforces the type and nothing else. On top of that, `parse_value` short-circuits on `if annotation is str: return str(value)`, so a constrained `str` field, which is the case in the issue, skipped validation entirely

`parse_value` now takes the field it is parsing for and re-attaches that field's constraint metadata through `Annotated` before validating. The four call sites (`chat_adapter.py`, `json_adapter.py`, `xml_adapter.py`, `predict/rlm.py`) pass their `FieldInfo`. Fields that declare no constraints are unaffected, and the two-argument form of `parse_value` keeps its previous behaviour for any caller that does not have a `FieldInfo` to hand

```python
class Sig(dspy.Signature):
    answer: str = dspy.OutputField(max_length=3)

field = Sig.output_fields["answer"]
parse_value("toolong", field.annotation, field)
# before: 'toolong'
# after:  ValidationError: String should have at most 3 characters
```

The `Annotated` route was chosen over the `__pydantic_validator__.validate_assignment` approach in the issue because it needs no model construction per parse and works on bare annotations. It was checked against `str`, `Optional[str]`, `str | None`, `list[str]`, `dict[str, int]` and `Annotated[Any, Ge(1)]`; all enforce correctly, including the union shapes

Branches that resolve a value themselves rather than handing it to pydantic (the `str` fast path, the enum member, and both literal returns) go through the same enforcement helper, so a `Literal` or str `Enum` field cannot slip a value past a constraint the way it could in the first revision of this PR

Tests cover string length, numeric bounds, an optional annotation, a container annotation, a constrained literal, a constrained str enum, fields with no constraints, the two-argument call, and the adapter call sites in ChatAdapter and JSONAdapter. Ten of them fail without the change to `dspy/adapters/utils.py`

## Contributor Checklist

## Warnings

This changes behaviour for existing programs. A program that declares a constraint and receives output violating it used to get that output back; it now raises. That is the point of the fix, and it only affects fields that declare constraints, but it is not backwards compatible for anyone who declared one decoratively and relied on it being ignored. Happy to put it behind an opt-in flag if you would rather it not be on by default

Separately, ChatAdapter falls back to JSONAdapter whenever parsing raises, so a constraint violation costs an extra LM call and the user sees the fallback's error rather than the real one:

```
AdapterParseError: LM response cannot be serialized to a JSON object.
Adapter JSONAdapter failed to parse the LM response.
LM Response: [[ ## answer ## ]]
toolong
```

The real cause, `String should have at most 3 characters`, is swallowed. That fallback masks every ChatAdapter parse failure the same way, so it is not introduced here, but this change does create a new class of failure that runs into it. Whether a validation failure should be retried through another adapter at all felt like a separate question, so it is left alone and the adapter tests call `adapter.parse` directly. Glad to follow up if you want it addressed

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -234,7 +234,7 @@ class ChatAdapter(Adapter):
         for k, v in sections:
             if (k not in fields) and (k in signature.output_fields):
                 try:
-                    fields[k] = parse_value(v, signature.output_fields[k].annotation)
+                    fields[k] = parse_value(v, signature.output_fields[k].annotation, signature.output_fields[k])
                 except Exception as e:
                     raise AdapterParseError(
                         adapter_name="ChatAdapter",
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -187,7 +187,7 @@ class JSONAdapter(ChatAdapter):
         # Attempt to cast each value to type signature.output_fields[k].annotation.
         for k, v in fields.items():
             if k in signature.output_fields:
-                fields[k] = parse_value(v, signature.output_fields[k].annotation)
+                fields[k] = parse_value(v, signature.output_fields[k].annotation, signature.output_fields[k])
 
         if fields.keys() != signature.output_fields.keys():
             raise AdapterParseError(
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -4,7 +4,7 @@ import inspect
 import json
 import types
 from collections.abc import Mapping
-from typing import Any, Literal, Union, get_args, get_origin
+from typing import Annotated, Any, Literal, Union, get_args, get_origin
 
 import json_repair
 import pydantic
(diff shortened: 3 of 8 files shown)


cardoner1993:
Both real. The enum and literal branches resolve their own value and return it, so they never touched the constrained target I built at the top of the function; a Literal["ok", "okay"] or a str Enum declaring max_length=3 happily returned "okay" while pydantic rejects it.

Pushed 36b87ef. Every branch that resolves a value itself now goes through one small enforce() helper: the str fast path, the enum member coming out of find_enum_member, and both literal returns, including the quoted and Literal[...] bracketed forms that get stripped before the membership check. Three tests added, one for a constrained literal, one for a constrained str enum, one pinning that unconstrained literal and enum fields parse exactly as before. All three fail against a1d42a0 and pass on the new commit.

Two notes from checking the rest of the parse surface while I was in there. BAMLAdapter inherits JSONAdapter.parse and TwoStepAdapter rebuilds its extractor signature with the whole FieldInfo, so both pick this up without changes. The one route where constraints still do not apply is base.py's native parse_lm_response path for dspy.Type fields such as Image and Audio, which resolve themselves and never reach parse_value; a length or range constraint on those would not mean much, so I left it as is rather than widen the diff.

Also worth knowing for anyone reading the tests: pydantic phrases literal and enum violations as Value should have at most 3 items after validation, not 4 rather than the String should have at most 3 characters wording you get for a plain str field, so those tests match on the former.

PR #10115, opened 2026-07-31
fix(persist): scrub known LM config from program pickles

isaacbmiller:
## Summary

A deliberately restricted alternative to #10103:

- scrub known DSPy-owned LM configuration (`api_key`, `api_base`, `base_url`, and `model_list`) from whole-program pickle artifacts
- remove known authentication headers from built-in `dspy.LM` and `dspy.Predict` configuration
- clear built-in `dspy.LM` history from program artifacts
- leave custom LM/Predict subclasses and ordinary copy/pickle behavior unchanged

## Design

`save_program=True` uses a per-save cloudpickle dispatch table for exact built-in `dspy.LM` and `dspy.Predict` types. This avoids adding context-sensitive `__getstate__` behavior to `BaseLM` and intentionally does not claim to sanitize arbitrary custom authentication schemes or arbitrary object graphs.

Program pickles remain trusted, potentially sensitive artifacts and must only be loaded from trusted sources.

## Testing

- `uv run --frozen pytest --deno tests/clients/test_lm.py -k 'program_pickle or copy_and_deepcopy_keep' -q`
- `uv run --frozen pytest --deno tests/primitives/test_base_module.py -q`
- `uv run --frozen ruff check dspy/clients/base_lm.py dspy/primitives/base_module.py dspy/utils/saving.py tests/clients/test_lm.py`
- `git diff --check`

### DIFF
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import cloudpickle
 import orjson
 
-from dspy.utils.saving import get_dependency_versions
+from dspy.utils.saving import dump_program, get_dependency_versions
 
 # NOTE: Note: It's important (temporary decision) to maintain named_parameters that's different in behavior from
 # named_sub_modules for the time being.
@@ -218,7 +218,7 @@ class BaseModule:
                     cloudpickle.register_pickle_by_value(module)
 
                 with open(path / "program.pkl", "wb") as f:
-                    cloudpickle.dump(self, f)
+                    dump_program(self, f)
             except Exception as e:
                 raise RuntimeError(
                     f"Saving failed with error: {e}. Please remove the non-picklable attributes from your DSPy program, "
--- a/dspy/utils/saving.py
+++ b/dspy/utils/saving.py
@@ -1,7 +1,8 @@
+import copyreg
 import logging
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, BinaryIO
 
 import cloudpickle
 import orjson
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
Ran the same probe against this branch and #10103. They're complementary — each catches something the other misses, and five carriers survive both.

One LM carrying every secret at once, saved with `save_program=True`, then grepping `program.pkl` for each value:

| carrier | `main` | #10103 | this PR |
|---|---|---|---|
| `api_key="sk-live-..."` | leak | scrubbed | scrubbed |
| `api_key=lambda: "..."` | leak | scrubbed | scrubbed |
| `extra_headers={"Authorization": ...}` | leak | scrubbed | scrubbed |
| `lm.history` | leak | scrubbed | scrubbed |
| `azure_ad_token` | leak | scrubbed | **leak** |
| `aws_secret_access_key` | leak | scrubbed | **leak** |
| `aws_session_token` | leak | scrubbed | **leak** |
| `Predict.config["api_key"]` | leak | **leak** | scrubbed |
| `azure_ad_token_provider=lambda: ...` | leak | leak | leak |
| `vertex_credentials='{"private_key":...}'` | leak | leak | leak |
| `default_headers={"Authorization": ...}` | leak | leak | leak |
| `extra_headers=[("Authorization", ...)]` | leak | leak | leak |
| subclass setting `self.api_key` | leak | leak | leak |

`UNSAFE_PROGRAM_LM_CONFIG_KEYS` here is `{"api_key", "api_base", "base_url", "model_list"}`; #10103's set adds `azure_ad_token`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`. Going the other way, `_reduce_program_predict` scrubbing `Predict.config` is coverage #10103 doesn't have at all — that path carries an `api_key` straight through there.

On the five neither catches:

`azure_ad_token_provider` and `vertex_credentials` are both real litellm kwargs — 23 and 95 files in the installed litellm reference them — and `**kwargs` puts them in `self.kwargs` verbatim. The first is the canonical Azure credential callable, so its closure pickles with the token inside.

`default_headers` is a third header field next to `headers` / `extra_headers`, and only the latter two are looped over. `_sanitize_program_lm_config` also guards with `isinstance(headers, dict)`, so `extra_headers=[("Authorization", "Bearer ...")]` passes through — httpx accepts a sequence of pairs.

The subclass case is structural: both approaches filter `state["kwargs"]`, but `state` is `lm.__dict__`, so a `BaseLM` subclass storing a credential as a plain attribute keeps it.

Two things that hold on this branch and are worth not re-testing: `copy.deepcopy` still returns an LM with `kwargs["api_key"]` and its history intact, so optimizer copies aren't affected; and `GLOBAL_HISTORY` doesn't reach the artifact, since it's module-level and cloudpickle stores it by reference.

The branch is currently `dirty` against `main`.

PR #10124, opened 2026-08-03
fix(callback): keep with_callbacks wrappers picklable

### LINKED ISSUE #10123: [Bug] settings.save(modules_to_serialize=...) can never serialize a custom Adapter subclass: with_callbacks wrapper closes over a ContextVar
## Summary

`dspy.settings.save(path, modules_to_serialize=[my_module])` — the documented way to serialize settings including custom components by value so the target machine doesn't need the source code — **cannot ever work for a custom `Adapter` subclass**. It always raises:

```
RuntimeError: Saving failed with error: cannot pickle '_contextvars.ContextVar' object.
Please remove the non-picklable attributes from the values in the `dspy.settings`.
```

The error message is misleading: the user's settings values are all picklable. The real culprit is dspy's own callback decorator.

## Root cause

`Adapter.__init_subclass__` re-wraps `format` and `parse` with `with_callbacks` on **every** subclass ([dspy/adapters/base.py#L76-L81](https://github.com/stanfordnlp/dspy/blob/main/dspy/adapters/base.py#L76-L81)), installing the wrapper functions into the subclass's own `__dict__`. The `with_callbacks` wrapper references a `ContextVar` (`ACTIVE_CALL_ID` in `dspy/utils/callback.py`). When the subclass's module is registered with `cloudpickle.register_pickle_by_value`, cloudpickle serializes the class — including those wrapper functions — by value, and `ContextVar` objects are not picklable.

Minimal proof that the decorator alone is the problem:

```python
import cloudpickle
from dspy.utils.callback import with_callbacks

cloudpickle.dumps(with_callbacks(lambda self: None))
# TypeError: cannot pickle '_contextvars.ContextVar' object
```

## Reproduction

```python
import os, sys, tempfile
tmp = tempfile.mkdtemp()
open(os.path.join(tmp, "my_adapter.py"), "w").write(
    "from dspy.adapters.chat_adapter import ChatAdapter\n"
    "class MyAdapter(ChatAdapter):\n"
    "    pass\n"
)
sys.path.insert(0, tmp)

import dspy, my_adapter
dspy.configure(adapter=my_adapter.MyAdapter())
dspy.settings.save(os.path.join(tmp, "settings.pkl"), modules_to_serialize=[my_adapter])
# RuntimeError: Saving failed with error: cannot pickle '_contextvars.ContextVar' object. ...
```

Note the subclass doesn't even need to override anything — `class MyAdapter(ChatAdapter): pass` is enough, because `__init_subclass__` installs the unpicklable wrappers regardless.

Verified on 3.3.0b1 (Python 3.12.12); the relevant code is unchanged on `main`.

By-reference saving (no `modules_to_serialize`, adapter's `.py` shipped to the target and on `sys.path`) works fine end to end — the loaded adapter is genuinely used by programs. So this specifically breaks the "no code shipped" path that `modules_to_serialize` exists for.

## Suggested directions

- Make the `with_callbacks` wrapper picklable (e.g. look up the ContextVar at call time via module attribute access instead of closing over it), or
- Have `Adapter.__init_subclass__` skip re-wrapping / store the undecorated functions so cloudpickle can serialize the class, or
- At minimum, improve the `Settings.save` error message — "remove the non-picklable attributes from the values in `dspy.settings`" sends users hunting through their own objects when the unpicklable object is injected by dspy itself.

aiedwardyi:
Fixes #10123

The `with_callbacks` wrappers referenced `ACTIVE_CALL_ID` as a module global, so
cloudpickle pulled the ContextVar in when serializing a decorated class by value.
Importing it inside the wrappers makes it a local, so it is never captured. The
module-level import stays as a re-export.

Adds a unit test that the wrappers pickle, and an end-to-end test that
`settings.save(..., modules_to_serialize=[...])` round-trips a custom Adapter
subclass with the source module removed.

AI-assisted: Claude Code was used to reproduce the bug, write the failing tests,
and draft the fix. I reviewed and verified every line, ran the affected suites
against a clean-main baseline, and confirmed the issue's reproduction end to end.

### DIFF
--- a/dspy/utils/callback.py
+++ b/dspy/utils/callback.py
@@ -5,7 +5,9 @@ import uuid
 from typing import Any, Callable
 
 import dspy
-from dspy.utils.callback_context import ACTIVE_CALL_ID
+
+# Re-exported for backwards compatibility; the wrappers below import it lazily instead.
+from dspy.utils.callback_context import ACTIVE_CALL_ID  # noqa: F401
 
 logger = logging.getLogger(__name__)
 
@@ -297,6 +299,10 @@ def with_callbacks(fn):
 
             _execute_start_callbacks(instance, fn, call_id, callbacks, args, kwargs)
 
+            # Imported here, not at module level, so cloudpickle does not capture the ContextVar
+            # by value when serializing a decorated class.
+            from dspy.utils.callback_context import ACTIVE_CALL_ID
+
             # Active ID must be set right before the function is called, not before calling the callbacks.
             parent_call_id = ACTIVE_CALL_ID.get()
             ACTIVE_CALL_ID.set(call_id)
@@ -326,6 +332,10 @@ def with_callbacks(fn):
             call_id = uuid.uuid4().hex
 
             _execute_start_callbacks(instance, fn, call_id, callbacks, args, kwargs)
+
+            # Imported here, not at module level, so cloudpickle does not capture the ContextVar
+            # by value when serializing a decorated class.
+            from dspy.utils.callback_context import ACTIVE_CALL_ID
 
             # Active ID must be set right before the function is called, not before calling the callbacks.
             parent_call_id = ACTIVE_CALL_ID.get()
--- a/tests/callback/test_callback.py
+++ b/tests/callback/test_callback.py
@@ -1,5 +1,6 @@
 import time
 
+import cloudpickle
 import pytest
 
 import dspy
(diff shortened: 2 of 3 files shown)


aiedwardyi:
Closing this - #10119 fixed the same root cause by accessing the ContextVar through the module,
so the wrappers already pickle cleanly on main. Thanks!

PR #10144, opened 2026-08-04
fix(interpreter): pass SUBMIT payloads through as Python-serialized json

### LINKED ISSUE #10143: [Bug] SUBMIT() silently rounds ints beyond 2**53 and crashes on nan/inf payloads
### What happened?

The `FinalOutput`/`SUBMIT` path in `runner.js` parses the payload with JS `JSON.parse` before re-serializing it for the host. Two consequences:

```python
interp = PythonInterpreter()

interp.execute("SUBMIT(2**60 + 1)")
# FinalOutput(output={'output': 1152921504606847000})  — silently off by 23

interp.execute("SUBMIT(float('nan'))")
# CodeExecutionError: Unexpected token 'N', "[{"output": NaN}]" is not valid JSON
```

The rounding happens because `JSON.parse` represents all numbers as float64, so ints beyond 2**53 lose precision silently. The crash happens because `last_exception_args()` serializes the payload with Python `json.dumps`, which emits bare `NaN`/`Infinity` literals — and JS `JSON.parse` rejects those.

Same family as #10140, which covered the expression-result leg of the sandbox boundary; this is the SUBMIT leg. It is reachable through `ProgramOfThought`, where SUBMIT carries the final answer — a correct big-integer answer comes back silently wrong, and a nan/inf answer errors the whole run.

PR incoming: serialize the payload on the Python side and pass the JSON text through the RPC channel untouched, mirroring #10141.

### Steps to reproduce

Script above; reproduced on macOS (arm64), deno 2.9.4.

### DSPy version

3.3.0 (main @ 8ca62a490)

vari39:
Fixes #10143.

## Problem

`SUBMIT()` payloads cross the sandbox boundary through JS `JSON.parse` in the `FinalOutput` handler in `runner.js`, then get re-serialized for the host. Failure modes:

| `execute(...)` | Before | After |
|---|---|---|
| `"SUBMIT(2**60 + 1)"` | `{'output': 1152921504606847000}` — **silently off by 23** | exact |
| `"SUBMIT(float('nan'))"` | **CodeExecutionError (crash)** | `{'output': nan}` |
| `"SUBMIT({'value': float('inf')})"` | **CodeExecutionError (crash)** | `{'output': {'value': inf}}` |
| `"SUBMIT({3, 1, 2})"` | **CodeExecutionError (crash)** | `{'output': [1, 2, 3]}` |
| `"SUBMIT(x)"` where `x` is circular | **unhandled rejection, Deno session terminated** | structured `CodeExecutionError`, session survives |

Rounding: `JSON.parse` represents all numbers as float64. Crashes: the payload is serialized by `last_exception_args()` with a plain `json.dumps`, which emits bare `NaN`/`Infinity` literals that JS `JSON.parse` rejects, and raises `TypeError` outright on sets.

This is the SUBMIT leg of the boundary issue fixed for expression results in #10141 — reachable through `ProgramOfThought`, where SUBMIT carries the final answer.

## Fix

Same pattern as #10141: a `final_output_json()` helper serializes the payload inside Pyodide with Python `json`, the resulting string passes through JSON-RPC untouched, and the host parses it with Python `json.loads` (exact big ints, accepts `NaN`/`Infinity`). Payloads Python `json` cannot serialize (`TypeError` for unsupported types, `ValueError` for circular references) fall back to the previous `JSON.parse` path; if that path also fails to serialize — it runs the same Python `json.dumps` inside `last_exception_args()` — the failure is emitted as a structured JSON-RPC `ValueError` instead of escaping as an unhandled rejection that kills the Deno session. Previously a circular `SUBMIT` payload terminated the session outright (on main as well); it now raises a clean `CodeExecutionError` and the session remains usable.

## Type conventions

Deliberately identical to #10141 and to the host-side `_serialize_value` injection convention: sets serialize as sorted lists (unsorted list fallback for mixed types), tuples as lists — covered by `test_submit_set_round_trip`. The serializer's `_default` handler is intentionally duplicated from #10141 rather than shared, so the two PRs stay independent (based on main, either can land first). Happy to unify the two helpers in whichever branch lands second — it will need a rebase over the other at that point anyway.

## Testing

- 4 new regression tests (big-int payload exactness; nan and nested inf payloads returning instead of crashing; set payloads following the injection convention; circular payloads raising a structured error with the session surviving).
- Full deno-marked suite green: `test_python_interpreter.py`, `test_code_interpreter.py`, `test_program_of_thought.py`, `test_rlm.py` — 233 passed, 2 pre-existing "requires actual LM" skips.

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -663,7 +663,11 @@ class PythonInterpreter:
                 if not isinstance(result, dict):
                     self._raise_terminal_error(f"Malformed execution result: {msg}")
                 self._sync_files()
-                # Check for SUBMIT (encoded as success with "final" field)
+                # Check for SUBMIT (encoded as success with "final_json" or "final")
+                if "final_json" in result:
+                    # Payload serialized with Python json inside the sandbox, so
+                    # big ints and nan/inf survive the JS boundary.
+                    return FinalOutput(json.loads(result["final_json"]))
                 if "final" in result:
                     return FinalOutput(result["final"])
                 return result.get("output", None)
--- a/dspy/primitives/runner.js
+++ b/dspy/primitives/runner.js
@@ -19,6 +19,27 @@ sys.stdout, sys.stderr = buf_stdout, buf_stderr
 def last_exception_args():
     return json.dumps(sys.last_exc.args) if sys.last_exc else None
 
+def final_output_json():
+    # Serialize the FinalOutput payload with Python json and hand the string
+    # through JSON-RPC untouched, so values that are valid Python but not
+    # JS-JSON-safe survive: ints beyond 2**53 would be silently rounded by
+    # JSON.parse in JS, and nan/inf (emitted as bare NaN/Infinity by Python
+    # json) would make JSON.parse throw. The host parses with Python
+    # json.loads, which accepts both.
+    if not (sys.last_exc and sys.last_exc.args):
+        return None
+    def _default(o):
+        if isinstance(o, (set, frozenset)):
+            try:
+                return sorted(o)
+            except TypeError:
+                return list(o)
+        raise TypeError(f"Object of type {type(o).__name__} is not JSON serializable")
+    try:
+        return json.dumps(sys.last_exc.args[0], default=_default)
+    except TypeError:
+        return None
+
 class FinalOutput(BaseException):
     # Control-flow exception to signal completion (like StopIteration)
     pass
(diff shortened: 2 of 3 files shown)


vari39:
Good catch on the circular-reference path — confirmed and fixed in 53a83e3.

`json.dumps` raises `ValueError` (not `TypeError`) on circular references, so the payload escaped `final_output_json`, and the fallback couldn't save it either: the legacy path serializes with the same Python `json.dumps` inside `last_exception_args()`, so it re-raised, and the rejection escaped as an unhandled async error that terminated the Deno session (this is also the behavior on main today).

Now: `final_output_json` catches `(TypeError, ValueError)`, and the legacy branch is wrapped so a serialization failure emits a structured JSON-RPC `ValueError`. The host raises a clean `CodeExecutionError("... SUBMIT payload is not JSON-serializable ...")` and the session remains usable afterward — covered by `test_submit_circular_payload_raises_cleanly`, which asserts both the error and that the interpreter still executes code after. Full deno suite: 233 passed.

PR #10148, opened 2026-08-04
feat(adapter): Fix behavior for optional/default output fields

michaelisaac-dev:
Output fields with a default, a default_factory, or a None-allowing annotation no longer fail parsing when the LM omits them. The parsed result falls back to the default (or None). Required fields still raise AdapterParseError.                                                                                                                                   
 
- New shared helpers in dspy/adapters/utils.py (apply_output_field_defaults, annotation_allows_none), applied in ChatAdapter, JSONAdapter, and XMLAdapter parsing; BAML and TwoStep inherit the behavior. bridge.py changed to use the common helper method.                       
- predict.py and flex/bridge.py now use the shared annotation_allows_none instead of private copies, so the input-warning, adapter, and sandbox contracts can't drift.                                                                 
- Docs: new "Optional Output Fields" section in learn/programming/signatures.md.

### DIFF
--- a/docs/docs/learn/programming/signatures.md
+++ b/docs/docs/learn/programming/signatures.md
@@ -235,6 +235,22 @@ predictor = dspy.Predict("number: int -> result: str")
 predictor(number="42")  # No warning
 ```
 
+### Optional Output Fields
+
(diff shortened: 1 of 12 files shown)


isaacbmiller (approved):
Can we add a test using the Optional[Type] syntax? Otherwise lgtm

PR #10149, opened 2026-08-04
fix(adapters): add a default timeout to Image.from_url / Audio.from_url

immu4989:
The explicit download factories that shipped in 3.3.0 call `requests.get` without a `timeout`, which waits forever by default. A slow or unresponsive host hangs the caller indefinitely. This is the remaining half of #9993; the implicit-fetch half was resolved by #10111.

This adds `timeout: float | None = 30.0` to `Image.from_url`, `Audio.from_url`, and the internal `_encode_image_from_url` helper (which also serves the deprecated `Image(url, download=True)` shim), threaded straight into `requests.get`.

Behavior change: a download that previously hung forever now raises `requests.exceptions.Timeout` after 30 seconds. Callers who want the old behavior can pass `timeout=None`.

No change to the I/O boundary from #10111. Construction and validation still perform no host I/O; this only bounds the explicit fetches.

Tests: `test_from_url_applies_a_timeout_by_default` covers the default, an explicit value, and `None`, for both factories. `tests/adapters/test_resource_loading.py` passes 33/33 locally.

Addresses #9993.

### DIFF
--- a/dspy/adapters/types/audio.py
+++ b/dspy/adapters/types/audio.py
@@ -80,7 +80,7 @@ class Audio(Type):
         return encode_audio(values)
 
     @classmethod
-    def from_url(cls, url: str, verify: bool = True) -> "Audio":
+    def from_url(cls, url: str, verify: bool = True, timeout: float | None = 30.0) -> "Audio":
         """
         Download an audio file from URL and encode it as base64.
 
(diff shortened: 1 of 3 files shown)


isaacbmiller:
lgtm, thanks for the PR

PR #10152, opened 2026-08-06
fix(dspy): discover predictors nested inside containers in named_parameters

anxkhn:
## What

`BaseModule.named_parameters` only traversed one level of `list`, `tuple`, and `dict` attributes. Containers nested inside another container were never entered, so any predictor stored below the first level was invisible to the rest of the framework.

## Reproduction

```python
import dspy

class Program(dspy.Module):
    def __init__(self):
        super().__init__()
        self.groups = [[dspy.Predict("question -> answer")]]

program = Program()
print(program.named_predictors())  # [] -- expected [("groups[0][0]", Predict(...))]
print(program.dump_state())        # {} -- expected the nested predictor's state
```

The program runs fine when the predictor is called directly, but optimizers see no predictors to optimize, `dump_state`/`load_state` silently drop it, `reset_copy` does not reset it, and `set_lm` does not configure it. `named_sub_modules` already walks arbitrarily nested containers, so the two traversals disagreed.

## Change

`named_parameters` now uses a single recursive walker over `Parameter`, `dspy.Module`, `list`, `tuple`, and `dict` values instead of a one-level loop plus a two-case helper. Behaviour that was already there is preserved:

- shared references are still deduplicated by object identity, first path wins
- pre-compiled sub-modules are still skipped and kept frozen
- path formatting is unchanged (`name.sub`, `name[0]`, `name['key']`) and now matches `named_sub_modules` for nested cases

The walker carries one `visited` set seeded with the module itself, so reference cycles (a module reachable from its own attributes, or a self-referencing dict) terminate instead of recursing forever. Sub-modules are traversed through their `__dict__` rather than by re-entering `named_parameters`, which is what makes the shared `visited` set and the cycle guard work across module boundaries.

## Tests

Added a regression in `tests/primitives/test_module.py` covering a predictor buried in `[[{...}]]` with a duplicated reference and a self-referencing dict, and asserting the discovered path plus `dump_state`, `load_state`, `reset_copy`, and `set_lm` all reach it.

```
pytest tests/primitives/test_module.py tests/primitives/test_base_module.py
```

## AI disclosure

An AI coding assistant was used to draft this patch and its test, under human review; the change and the test output were checked by hand before submitting. Prompt used:

> Fix `dspy/primitives/base_module.py` so `named_parameters` recursively discovers predictors nested inside lists, tuples, and dicts (for example `self.groups = [[dspy.Predict("q -> a")]]`). Keep deduplication of shared references, keep skipping pre-compiled sub-modules, make the traversal cycle-safe, and keep the path format consistent with `named_sub_modules`. Add a regression test in `tests/primitives/test_module.py` covering nested containers, shared references, cycles, state round-trip, and `set_lm`.

### DIFF
--- a/dspy/primitives/base_module.py
+++ b/dspy/primitives/base_module.py
@@ -22,47 +22,39 @@ class BaseModule:
 
     def named_parameters(self):
         """
-        Unlike PyTorch, handles (non-recursive) lists of parameters too.
+        Unlike PyTorch, handles nested lists, tuples, and dictionaries of parameters too.
         """
 
         import dspy
         from dspy.predict.parameter import Parameter
 
-        visited = set()
+        visited = {id(self)}
         named_parameters = []
 
         def add_parameter(param_name, param_value):
-            if isinstance(param_value, Parameter):
-                if id(param_value) not in visited:
-                    visited.add(id(param_value))
-                    named_parameters.append((param_name, param_value))
+            if id(param_value) in visited:
+                return
+            visited.add(id(param_value))
 
+            if isinstance(param_value, Parameter):
+                named_parameters.append((param_name, param_value))
             elif isinstance(param_value, dspy.Module):
                 # When a sub-module is pre-compiled, keep it frozen.
                 if not getattr(param_value, "_compiled", False):
-                    for sub_name, param in param_value.named_parameters():
-                        add_parameter(f"{param_name}.{sub_name}", param)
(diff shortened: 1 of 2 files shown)


anxkhn:
fixed dictionary key path serialization so integer and quote-containing keys round-trip through map_named_predictors, with regression coverage for both cases. rebased onto the latest main; the focused tests and ruff pass. a maintainer review would be appreciated.

PR #10154, opened 2026-08-06
test(dspy): reuse LiteLLM server per pytest worker

isaacbmiller:
## 📝 Changes Description

Start one LiteLLM proxy per pytest worker instead of one process per test. A function-scoped wrapper truncates the request log before each test, preserving per-test request-log isolation.

The readiness loop also polls at 50ms instead of 500ms and fails immediately if the child process exits before opening its port.

### Isolated benchmark

The benchmark includes all 10 fixture consumers:

```bash
python -m pytest -q --durations=15 \
  tests/clients/test_lm.py::test_chat_lms_can_be_queried \
  tests/clients/test_lm.py::test_dspy_cache \
  tests/clients/test_lm.py::test_text_lms_can_be_queried \
  tests/clients/test_lm.py::test_lm_calls_support_callables \
  tests/clients/test_lm.py::test_lm_calls_support_pydantic_models \
  tests/clients/test_lm.py::test_responses_api_tool_calls \
  tests/streaming/test_streaming.py::test_streamify_yields_expected_response_chunks \
  tests/streaming/test_streaming.py::test_streaming_response_yields_expected_response_chunks
```

|---|---:|---:|
| Runtime | 49.47s | 7.96s / 7.24s |
| Proxy startups | 10 | 1 per worker |
| Result | 10 passed | 10 passed |

**41.5–42.2 seconds removed: 83.9–85.4% faster.**

Additional xdist validation:

```bash
python -m pytest -q -n 4 --dist worksteal <same 10 tests>
# 10 passed in 11.02s
```

Each xdist worker owns a separate random port and temporary log path. Tests execute sequentially within each worker, so truncating the worker-local log at function setup preserves isolation.

## ✅ Contributor Checklist

- [x] All fixture consumers pass serially and under xdist

## ⚠️ Warnings

The session fixture assumes consumers finish their requests before test teardown. All current consumers do so; none leave background requests running.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,11 @@ from typing import Any
 
 import pytest
 
-from tests.test_utils.server import litellm_test_server, read_litellm_test_server_request_logs  # noqa: F401
+from tests.test_utils.server import (  # noqa: F401
+    _litellm_test_server,
+    litellm_test_server,
+    read_litellm_test_server_request_logs,
+)
 
 SKIP_DEFAULT_FLAGS = ["reliability", "extra", "llm_call", "deno"]
 
--- a/tests/test_utils/server/__init__.py
+++ b/tests/test_utils/server/__init__.py
@@ -5,6 +5,7 @@ import subprocess
 import sys
 import tempfile
 import time
+from collections.abc import Iterator
 from typing import Any
 
 import pytest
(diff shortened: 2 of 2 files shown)


isaacbmiller:

|---|---:|---:|---:|
| Base → PR | 92s | 81s | 11s |
| PR → base | 83s | 74s | 9s |
| **Mean** | **87.5s** | **77.5s** | **10s / 11.4%** |

Both directions improved on the same runner with unchanged outcomes. The full-CI percentage is smaller than the targeted 10-test percentage because four xdist workers overlap server startups and the full suite contains unrelated work.

PR #10155, opened 2026-08-06
test(dspy): make asyncify concurrency test deterministic

isaacbmiller:
## 📝 Changes Description

Replace three waves of 500ms worker sleeps and elapsed-time bounds with explicit synchronization.

The test now blocks worker bodies on an event and directly asserts:

- Exactly the configured number of worker functions have started.
- The AnyIO limiter has exactly the configured number of borrowed tokens.
- All queued calls complete with the expected results after release.

Failure-only two-second guards prevent regressions from hanging the suite.

### Isolated benchmark

```bash
python -m pytest -q --durations=20 tests/utils/test_asyncify.py::test_asyncify
```

|---|---:|---:|
| Call time | 4.51s / 4.52s | 0.02s / 0.16s |
| Result | passed | passed |

**96.5–99.6% lower call time, removing at least 4.35 seconds.**

The new assertions are stricter than the previous broad elapsed-time window: a serialized executor cannot pass because every expected worker body must enter before the release event is set.

## ✅ Contributor Checklist

- [x] Focused test passes in repeated runs

## ⚠️ Warnings

None. The timeouts are failure guards and add no successful-path delay.

### DIFF
--- a/tests/utils/test_asyncify.py
+++ b/tests/utils/test_asyncify.py
@@ -1,6 +1,5 @@
 import asyncio
-import math
-from time import sleep, time
+import threading
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


isaacbmiller:

| Round | Baseline | This PR | Difference |
|---|---:|---:|---:|
| Base → PR | 144s | 111s | 33s faster |
| PR → base | 94s | 107s | 13s slower |

The runner changed speed materially during this job, so the full-suite CI effect is **inconclusive**; I do not claim the 10s mean difference as a proven CI saving. The changed asyncify test itself still has strong isolated evidence (4.51–4.52s → 0.02–0.16s), direct synchronization assertions, 5/5 Greptile review, and successful full CI. This PR should be assessed as determinism/flakiness improvement rather than a measurable matrix critical-path win.

isaacbmiller:
Closing under the CI-first acceptance criterion. The targeted test is substantially faster and more deterministic, but paired same-runner full-suite measurements disagreed by order: 33s faster in one pair and 13s slower in the reverse pair. That does not establish a reproducible CI critical-path improvement. The branch and benchmark evidence remain available if the determinism change is reconsidered separately.

PR #10156, opened 2026-08-06
test(dspy): use memory-only cache by default

isaacbmiller:
## 📝 Changes Description

Use a fresh memory-only DSPy cache in the global per-test isolation fixture instead of creating and closing a temporary 16-shard `FanoutCache` for every test.

This preserves per-test cache isolation and within-test cache hits. Tests that exercise disk persistence already explicitly configure their own `tmp_path`-backed disk caches.

### Isolated benchmark

Same checkout, Python environment, and serial command on both branches:

```bash
python -m pytest -q --durations=20
```

|---|---:|---:|
| Pytest runtime | 266.35s | 112.74s |
| Result | 1,192 passed / 417 skipped / 2 xfailed | 1,192 passed / 417 skipped / 2 xfailed |

**153.61 seconds removed: 57.7% faster, or 2.36x throughput.** An earlier untouched baseline measured 242.89s, so even against that lower baseline this change removes 130.15s (53.6%).

### Why behavior is preserved

- A new cache object is still created for every test.
- Memory caching remains enabled, so repeated requests within a test still hit cache.
- The original global cache is still restored during teardown.
- DSPy settings are still reset from `DEFAULT_CONFIG`.
- Dedicated cache, LM cache, embedding cache, and disk-serialization tests explicitly enable isolated disk caches when disk behavior is under test.

## ✅ Contributor Checklist

- [x] Complete serial suite passes with unchanged outcomes

## ⚠️ Warnings

No production code changes. This intentionally changes only the default storage tier used by the test isolation fixture; disk behavior remains covered by explicit disk-cache tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,6 @@
 import copy
 import os
 from collections.abc import Iterator
-from pathlib import Path
 from typing import Any
 
 import pytest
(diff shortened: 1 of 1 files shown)


isaacbmiller:

|---|---:|---:|---:|
| Base → PR | 83s | 41s | 42s |
| PR → base | 77s | 33s | 44s |
| **Mean** | **80s** | **37s** | **43s / 53.8%** |

Both runs preserved the same outcomes. This same-runner CI result is close to the isolated offline result (57.7%) and removes cross-runner variance from the comparison.

PR #10158, opened 2026-08-06
test(dspy): reduce unbatchify batch trigger timeout

isaacbmiller:
## 📝 Changes Description

Reduce `test_unbatchify_batch_size_trigger`'s `max_wait_time` from 5 seconds to 500ms.

The test only observes the single-item state for 20ms before submitting the second item. A 500ms timeout retains a 25x scheduling margin while avoiding a five-second `queue.get()` wait when `close()` joins the worker after the second batch.

### Isolated benchmark

```bash
python -m pytest -q --durations=20 tests/utils/test_unbatchify.py::test_unbatchify_batch_size_trigger
```

|---|---:|---:|
| Call time | 5.03s / 5.03s | 0.53s / 0.53s |
| Result | passed | passed |

**4.50 seconds removed: 89.5% faster, or 9.5x throughput.**

The assertions, inputs, batch size, expected batches, and results are unchanged. The batch is still triggered by reaching `max_batch_size=2`, not by the timeout.

## ✅ Contributor Checklist

- [x] Focused test passes in repeated runs

## ⚠️ Warnings

No production behavior changes. The 500ms value was chosen instead of the minimum possible timeout to retain substantial headroom on loaded xdist workers.

### DIFF
--- a/tests/utils/test_unbatchify.py
+++ b/tests/utils/test_unbatchify.py
@@ -23,7 +23,7 @@ Unbatchify.submit = submit
 def test_unbatchify_batch_size_trigger():
     """Test that the batch processes exactly when max_batch_size is reached."""
     batch_fn_mock = MagicMock(wraps=simple_batch_processor)
-    unbatcher = Unbatchify(batch_fn=batch_fn_mock, max_batch_size=2, max_wait_time=5.0)
+    unbatcher = Unbatchify(batch_fn=batch_fn_mock, max_batch_size=2, max_wait_time=0.5)
 
     futures = []
     futures.append(unbatcher.submit(10))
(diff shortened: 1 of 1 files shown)


isaacbmiller:

|---|---:|---:|---:|
| Base → PR | 88s | 85s | 3s |
| PR → base | 78s | 77s | 1s |
| **Mean** | **83s** | **81s** | **2s / 2.4%** |

The direction is consistent but the whole-suite impact is intentionally small: this removes 4.5s from one test, while four xdist workers overlap it with the rest of the suite. Outcomes remained unchanged.

isaacbmiller:
Closing under the CI-first impact threshold. Paired same-runner full-suite measurements were consistently favorable but small: 3s and 1s, averaging 2s / 2.4%. The targeted test remains 4.5s faster and deterministic, but the real four-worker CI benefit is not large enough to justify carrying this change. The branch and benchmark evidence remain available.

PR #10160, opened 2026-08-06
fix(evaluate): prevent futures-after-closure on long evals (fixes #9574)

Aryan-Pardeshi:
## Summary
Fixes #9574 — `ParallelEvaluate` raises `RuntimeError: cannot schedule new futures after executor close` on long evals.

Root cause: `ParallelExecutor._execute_parallel` resubmits stragglers via `executor.submit()` inside the wait loop. Three failure modes:
1. The submit can fire while the executor is mid-teardown (SIGINT / error-cancel), raising `cannot schedule new futures after closure` and crashing the whole eval.
2. Only the original future was marked `resubmitted`, so a slow resubmission could spawn a 3rd copy -> duplicate explosion saturating the pool.
3. `dspy.Evaluate` never exposed the straggler timeout, so users of legitimately long evals couldn't opt out.

Changes (`dspy/utils/parallelizer.py`, `dspy/evaluate/evaluate.py`):
- Wrap the straggler `executor.submit()` in `try/except RuntimeError` -> log + skip, never crash.
- Mark **both** original and resubmission as `resubmitted` (resubmit at most once per item).
- Expose `timeout` / `straggler_limit` on `Evaluate` (`timeout=0` disables straggler resubmission — the documented escape hatch for long evals). Defaults unchanged.

## Tests
- `tests/utils/test_parallelizer.py` -> 14 passed.
- Behavioral harness verified: pool raising `RuntimeError` on resubmit -> eval completes without crash; `timeout=0` -> each item runs exactly once.

## AI assistance disclosure
Per CONTRIBUTING.md ("Do not submit PRs from fully autonomous AI agents"): this change was authored with AI tooling assistance, but I (the human author) have reviewed every line, understand the fix, and the PR includes tests and passes CI. Not an autonomous-agent submission.

## Maintainer direction
Matches the approach referenced in the issue (guard the submit + expose opt-out), consistent with PRs #9912/#9997 discussion.

### DIFF
--- a/dspy/evaluate/evaluate.py
+++ b/dspy/evaluate/evaluate.py
@@ -132,11 +132,11 @@ class Evaluate:
         num_threads: int | None = None,
         display_progress: bool | None = None,
         display_table: bool | int | None = None,
-        timeout: int | None = None,
-        straggler_limit: int | None = None,
         callback_metadata: dict[str, Any] | None = None,
         save_as_csv: str | None = None,
         save_as_json: str | None = None,
+        timeout: int | None = None,
+        straggler_limit: int | None = None,
     ) -> EvaluationResult:
         """
         Args:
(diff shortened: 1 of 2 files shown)


Aryan-Pardeshi:
Good catch from the review — that was a real break, not a false positive.

`timeout` and `straggler_limit` were inserted **before** `callback_metadata`, `save_as_csv` and `save_as_json`, so any caller passing those three positionally would have bound them to the wrong parameters and the eval would fail before writing the files it was asked for.

`a6bd041` appends the two new parameters instead, restoring the established prefix:

```
self, program, metric, devset, num_threads, display_progress, display_table,
callback_metadata, save_as_csv, save_as_json, timeout, straggler_limit
```

`Evaluate.__init__` is keyword-only (`*`), so its ordering was never affected. Also added the missing `save_as_csv`/`save_as_json` docstring entries while reordering that block to match.

isaacbmiller:

PR #10163, opened 2026-08-07
ci: benchmark isolated pytest runtime changes

isaacbmiller:
## Purpose

Temporary CI instrumentation for paired, same-runner performance measurements of the four unmerged pytest optimizations.

Each matrix job checks out the exact common parent (`485ce44`) and one isolated candidate, then runs the full four-worker CI suite twice on each revision. Round two reverses execution order to balance runner warmup and filesystem-cache effects.

This is measurement infrastructure only and is not intended to merge. The resulting paired GitHub-runner times will be copied to the individual PRs and the investigation portal; this PR will then be closed.

### DIFF
--- a/.github/workflows/paired_ci_benchmarks.yml
+++ b/.github/workflows/paired_ci_benchmarks.yml
@@ -1,0 +1,72 @@
+name: Paired CI Performance Benchmarks
+
+on:
+  pull_request:
+
+permissions: {}
+
+jobs:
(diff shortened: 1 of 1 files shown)


isaacbmiller:
Paired CI measurements completed successfully and were copied to #10154, #10155, #10156, and #10158. This PR was temporary measurement infrastructure only and is not intended to merge.

PR #10167, opened 2026-08-07
fix(dspy): keep interpreter synced after base exceptions

### LINKED ISSUE #10165: [Bug] `dspy.Flex`/`dspy.RLM` sandbox dies with `CodeInterpreterError: Deno process exited (code 0)`
## Environment

- dspy 3.3.0 (Flex and RLM are `@experimental`)
- Python 3.12.4 (asdf), macOS 15 (Darwin 24.6.0, arm64)
- Deno 2.6.9
- Worker LMs that hit it: `accounts/fireworks/models/glm-5p2`, `accounts/fireworks/models/kimi-k2p7-code` (the bug is model-independent; see Root cause)

## Symptom

An unoptimized `dspy.Flex` with tools intermittently fails mid-run:

```
CodeExecutionError: RuntimeError: ['CodeInterpreterError: Deno process exited (code 0);
interpreter state was lost. Create a new interpreter for a fresh session.']
```

raised from `dspy/predict/flex/bridge.py:280` → `dspy/primitives/python_interpreter.py:692`.
The same call sometimes succeeds. Under GEPA the crash shows up as:

```
WARNING dspy.teleprompt.bootstrap_trace: Program raised on an example; capturing it as a
failed prediction: CodeExecutionError: RuntimeError: ['CodeInterpreterError: Deno process
exited (code 0); ...']
```

## Root cause

1. **A fresh `dspy.Flex(sig, tools=[...])` baselines to `dspy.RLM`**, not `dspy.Predict`
   (`dspy/predict/flex/flex.py:139`). So one Flex call runs **two** Deno/Pyodide sandboxes:
   the outer Flex bridge, plus a nested interpreter where RLM has the worker LM write and
   execute REPL code step by step.

2. **The nested sandbox's JSON-RPC session is fragile.** Tool calls cross the bridge via
   Pyodide's `run_sync` (JSPI stack switching) in `runner.js`. When LM-written REPL code
   raises `SystemExit` or `KeyboardInterrupt` (e.g. the model writes `exit()`, `sys.exit()`,
   or `quit()` to "end" its REPL session), the resulting `PythonError` escapes
   `runPythonAsync` as an **unhandled promise rejection** instead of hitting the
   `try/catch` in the `execute` handler.

3. **The unhandled-rejection handler double-responds.** `runner.js` catches the rejection
   and emits a JSON-RPC *error with `id: null`*. The host (`python_interpreter.py:672-678`)
   deliberately treats `id: null` errors as the response to the in-flight request — but the
   real response for that request also arrives later. Every subsequent request/response is
   now off by one. The session is unrecoverable.

4. **The next REPL step finds a corpse.** Depending on timing, the next `execute` either
   polls a Deno process that has since drained its event loop and exited cleanly
   (`_ensure_deno_process` → `"Deno process exited (code 0)"`) or reads the stale response
   (`"Response ID mismatch"`). RLM raises `CodeInterpreterError`; the Flex bridge relays it
   into the outer sandbox as a tool-call error; the sandbox shim re-raises it as
   `RuntimeError`; the outer `execute` surfaces it as the `CodeExecutionError` above.

**Why it's intermittent:** it depends entirely on what code the worker LM writes and whether
a tool call fails mid-run (the cached trajectory of a failing run shows the model reporting
*"The read_webpage call failed ... with an async error"* — i.e. `Unhandled async error:
PythonError` — one step before death). Identical requests succeed when the model takes a
different path.

**Threads are not required.** The minimal repro below is single-threaded. Four concurrent
threads each with their own interpreter (the `num_threads=4` GEPA shape) run cleanly, and a
thread-*shared* interpreter fails fast with a different, explicit error ("PythonInterpreter
is not thread-safe..."). A 4-thread GEPA run simply executes ~4× as many RLM REPL sessions
per unit time, so it multiplies the odds of hitting the fatal path — which is plausibly why
it was first seen there.

## Minimal deterministic repro (no LM, no tools, no network)

```python
# repro_deno_desync.py
# Requires: dspy 3.3.0, deno on PATH.
# If $HOME has a package.json/node_modules, also: export DENO_NO_PACKAGE_JSON=1

interp = PythonInterpreter()

# Step 1: something an LM plausibly writes to "end" a REPL session.
# Raises: CodeExecutionError: Error: Unhandled async error: PythonError
try:
    interp.execute("import sys\nsys.exit(0)")
except Exception as e:
    print(f"step 1: {type(e).__name__}: {e}")

print("deno poll after step 1:", interp.deno_process.poll())

# Step 2: the session is permanently desynced. This raises
#   CodeInterpreterError: Response ID mismatch: expected 4, got 3
# or, if Deno has drained and exited by now (typical with LM latency between steps):
#   CodeInterpreterError: Deno process exited (code 0); interpreter state was lost. ...
try:
    interp.execute("print('still alive?')")
except Exception as e:
    print(f"step 2: {type(e).__name__}: {e}")
```

Observed output:

```
step 1: CodeExecutionError: Error: Unhandled async error: PythonError
deno poll after step 1: None
step 2: CodeInterpreterError: Response ID mismatch: expected 4, got 3
```

`raise KeyboardInterrupt()` in place of `sys.exit(0)` produces the same corruption.
An ordinary exception (`raise ValueError(...)`), a raising host tool, an 8-second tool, and
a 3 MB tool result are all handled cleanly — the desync is specific to the
`SystemExit`/`KeyboardInterrupt` async-escape path.

## Full-context repro (how it actually bites)

```python
# Intermittent: fails when the worker LM's REPL code hits the path above.
# In one observed failure the fatal step was a plain `read_webpage(...)` call whose
# "async error" the model reported one step before the session died.
import os
import dspy
from tavily import TavilyClient

os.environ["DENO_NO_PACKAGE_JSON"] = "1"  # see footnote
tavily_client = TavilyClient(api_key=os.environ["TAVILY_API_KEY"])

def internet_search(query: str, max_results: int = 5, include_raw_content: bool = False):
    """Run a web search"""
    return tavily_client.search(query, max_results=max_results,
                                include_raw_content=include_raw_content, topic="general")

def read_webpage(url: str) -> dict:
    """Read the content of a URL."""
    return tavily_client.extract(url)

lm = dspy.LM("openai/accounts/fireworks/models/glm-5p2",
             api_key=os.environ["FIREWORKS_API_KEY"],
             api_base="https://api.fireworks.ai/inference/v1")
dspy.configure(lm=lm)

# Baselines to dspy.RLM because tools are present -> nested REPL sandbox per call.
researcher = dspy.Flex("research_request: str -> report: str",
                       tools=[internet_search, read_webpage])

# Run repeatedly (or under GEPA with num_threads=4): some runs complete, others die with
# CodeExecutionError: RuntimeError: ['CodeInterpreterError: Deno process exited (code 0); ...']
result = researcher(research_request=(
    "Write a history of Coyote Hills, a park in the East Bay Regional Parks "
    "District in California."))
print(result.report)
```

## Suggested fixes (upstream, `dspy/primitives`)

1. `runner.js`: catch `SystemExit`/`KeyboardInterrupt` (Pyodide `PythonError`s) from the
   rejection path and report them as ordinary app errors **for the current request id**,
   not as `id: null`.
2. `python_interpreter.py` `execute()`: don't consume `id: null` errors as the in-flight
   request's response — log and skip them so the real response keeps the session in sync.
3. Alternatively/additionally: have RLM treat a terminal `CodeInterpreterError` as a
   restartable event (new interpreter, replay variable state or fail just that step) rather
   than aborting the whole forward.

## Workarounds

- **Hardened interpreter factory (keeps the RLM baseline).** Fix (2) can be applied from
  user code without touching site-packages: subclass `PythonInterpreter` to skip
  unsolicited `id: null` errors, and pass it as `interpreter_factory` — Flex hands the same
  factory to the nested RLM, so both sandboxes are covered. Verified against the minimal
  repro: the trigger surfaces as a clean `CodeExecutionError: SystemExit: [0]` step error,
  and the session survives (subsequent executes and tool calls succeed).

  ```python

  class HardenedInterpreter(PythonInterpreter):
      """Skip unsolicited id:null sandbox errors instead of consuming them as the
      in-flight request's response, so one async hiccup doesn't desync the session."""
      def _parse_response_line(self, line, context):
          msg = super()._parse_response_line(line, context)
          if (
              isinstance(msg, dict)
              and msg.get("id") is None
              and "error" in msg
              and "method" not in msg
          ):
              return None  # treated as a skipped line; the real response follows
          return msg

  researcher = dspy.Flex("research_request: str -> report: str",
                         tools=[internet_search, read_webpage],
                         interpreter_factory=HardenedInterpreter)
  ```

  Caveat: if the sandbox ever emitted an `id: null` error *without* a real response
  following, `execute()` would block on the next read instead of raising. Not observed in
  testing (the `try/catch` response always arrived), but worth noting upstream.

- **Bind a non-RLM module source** before calling/optimizing. A ReAct+Predict `module_src`
  (e.g. a saved GEPA-optimized program: `researcher.load("flex_optimized_researcher.json")`)
  runs its predictors host-side — no nested REPL sandbox — and is robust in testing.
- GEPA already scores these crashes as failed predictions (score 0 with an
  empty-report-guarding metric), so optimization survives them; they just waste rollouts.

### Footnote: `DENO_NO_PACKAGE_JSON=1`

A separate startup failure: if `$HOME` (or any cwd ancestor) contains `package.json` +
`node_modules`, Deno resolves `npm:pyodide` there and the sandbox dies immediately on a
read-permission error (`NotCapable: Requires read access to ".../node_modules/pyodide/..."`,
then `Top-level await promise never resolved`, exit 1). Set `DENO_NO_PACKAGE_JSON=1`
*before* the first interpreter is created.

### Steps to reproduce

See above.

### DSPy version

3.3.0

isaacbmiller:
## 📝 Changes Description

This PR prevents `SystemExit`, `KeyboardInterrupt`, and their subclasses from desynchronizing a `PythonInterpreter` session.

Pyodide reports these exceptions through the global `unhandledrejection` handler before `runPythonAsync` reaches the request-local `catch`. The global handler consequently emitted an `id: null` JSON-RPC error followed by the correctly identified response, leaving the second response queued for the next execution. The runner now asks Python whether the current exception inherits from either control-flow exception, suppresses that duplicate global event, and lets the request-local handler produce the sole response.

A regression test verifies that the built-in exceptions and custom subclasses surface as `CodeExecutionError` and that subsequent execution succeeds in the same interpreter.

Closes #10165.

## 🧪 Validation

- Reported deterministic repro and subclass regression pass on Deno 2.6.9 and 2.7.7.
- `python -m pytest tests/primitives/test_python_interpreter.py --deno -q` on Deno 2.7.7: 73 passed.
- `pre-commit run --files dspy/primitives/runner.js tests/primitives/test_python_interpreter.py`: passed.

## ✅ Contributor Checklist

- [x] PR title follows the required format
- [x] Commit messages follow the required format `{label}(dspy): {message}`

## ⚠️ Warnings

None.

### DIFF
--- a/dspy/primitives/runner.js
+++ b/dspy/primitives/runner.js
@@ -131,10 +131,12 @@ const jsonrpcError = (code, message, id, data = null) => {
   return JSON.stringify({ jsonrpc: "2.0", error: err, id });
 };
 
-// Global handler to prevent uncaught promise rejections from crashing Deno
-// These can occur during async Python <-> JS interop
+// Prevent promise rejections from async Python <-> JS interop from crashing Deno.
 globalThis.addEventListener("unhandledrejection", (event) => {
   event.preventDefault();
+  // Pyodide reports these as unhandled before runPythonAsync rejects into the execute
+  // handler, which will respond with the request ID. Do not emit a duplicate id:null error.
+  if (event.reason?.type === "SystemExit" || event.reason?.type === "KeyboardInterrupt") return;
   console.log(jsonrpcError(JSONRPC_APP_ERRORS.RuntimeError, `Unhandled async error: ${event.reason?.message || event.reason}`, null));
 });
 
--- a/dspy/primitives/runner.js
+++ b/dspy/primitives/runner.js
@@ -19,6 +19,9 @@ sys.stdout, sys.stderr = buf_stdout, buf_stderr
 def last_exception_args():
     return json.dumps(sys.last_exc.args) if sys.last_exc else None
 
+def last_exception_ends_execution():
+    return isinstance(sys.last_exc, (SystemExit, KeyboardInterrupt))
+
 class FinalOutput(BaseException):
     # Control-flow exception to signal completion (like StopIteration)
     pass
@@ -136,7 +139,7 @@ globalThis.addEventListener("unhandledrejection", (event) => {
   event.preventDefault();
   // Pyodide reports these as unhandled before runPythonAsync rejects into the execute
   // handler, which will respond with the request ID. Do not emit a duplicate id:null error.
-  if (event.reason?.type === "SystemExit" || event.reason?.type === "KeyboardInterrupt") return;
+  if (event.reason?.name === "PythonError" && pyodide.runPython("last_exception_ends_execution()")) return;
   console.log(jsonrpcError(JSONRPC_APP_ERRORS.RuntimeError, `Unhandled async error: ${event.reason?.message || event.reason}`, null));
 });
 
--- a/tests/primitives/test_python_interpreter.py
+++ b/tests/primitives/test_python_interpreter.py
@@ -115,6 +115,8 @@ def test_base_exceptions_do_not_desync_interpreter():
         for code, error_type in [
             ("import sys\nsys.exit(0)", "SystemExit"),
             ("raise KeyboardInterrupt()", "KeyboardInterrupt"),
+            ("class ExitSignal(SystemExit): pass\nraise ExitSignal(0)", "ExitSignal"),
+            ("class InterruptSignal(KeyboardInterrupt): pass\nraise InterruptSignal()", "InterruptSignal"),
         ]:
             with pytest.raises(CodeExecutionError, match=error_type):
                 interpreter.execute(code)
--- a/tests/primitives/test_python_interpreter.py
+++ b/tests/primitives/test_python_interpreter.py
@@ -110,6 +110,17 @@ def test_exception_args(pooled_interpreter):
         interpreter.execute(code)
 
 
+def test_base_exceptions_do_not_desync_interpreter():
+    with PythonInterpreter() as interpreter:
+        for code, error_type in [
+            ("import sys\nsys.exit(0)", "SystemExit"),
+            ("raise KeyboardInterrupt()", "KeyboardInterrupt"),
+        ]:
+            with pytest.raises(CodeExecutionError, match=error_type):
+                interpreter.execute(code)
+            assert interpreter.execute("print('still alive')") == "still alive\n"
+
+
 def test_generated_exception_name_cannot_spoof_interpreter_failure(pooled_interpreter):
     interpreter = pooled_interpreter
     with pytest.raises(CodeExecutionError, match="CodeInterpreterError"):


isaacbmiller:
closing in favor of #10190

PR #10169, opened 2026-08-07
ci: prefetch Ollama image during test setup

isaacbmiller:
## Summary

Start pulling the pinned Ollama Docker image while the independent Python/uv setup and model-cache restore run. The real-LM tests, model, image, warm-up, and assertions are unchanged.

The service step waits for the pull to finish and propagates a failed/timed-out pull before starting Ollama.

## Why

In a clean pre-change Actions run, the real-LM job spent 39 seconds in `Start Ollama service`, after dependency setup and a 25-second model-cache restore had already completed. These operations are independent. Prefetching should hide most or all of the image-download latency behind setup without reducing coverage.

Pre-change evidence:
- [Run 31188418077](https://github.com/stanfordnlp/dspy/actions/runs/31188418077): real-LM job 170s
- Setup/cache before service: setup-uv + dependency install + model-cache restore
- Start Ollama: 39s
- Real-LM pytest: 86s; all 5 tests passed

I will add exact-head Actions timing and a repeated run before claiming a speedup.

## Validation

- `pre-commit run --files .github/workflows/run_tests.yml`
- `git diff --check`
- Normal GitHub Actions pending

PR #10173, opened 2026-08-07
ci: shard real-LM tests

isaacbmiller:
## Summary

Run the five Ollama-backed tests as two parallel jobs on top of #10169's image-prefetch optimization. This is an isolated stacked PR; once #10169 merges, it can be retargeted to `main` with only this matrix change remaining.

The two jobs use the same Python version, pinned Ollama image/model, model cache, dependencies, warm-up, environment, pytest markers, and assertions. Only the complementary `-k` selections differ.

## Coverage proof

The existing `llm_call` selection contains 24 nodes: five Ollama tests that execute in this job and 19 provider-specific tests that skip without credentials.

| Set | Selected nodes | Ollama tests expected to execute |
|---|---:|---:|
| First LM shard | 3 | 3 |
| Remaining LM shard | 21 | 2 |
| overlap | **0** | **0** |
| omitted | **0** | **0** |

Based on the clean unsharded Actions durations, generated-call time is balanced at roughly 41s versus 28s. With #10169 reducing setup/service overhead, the hypothesis is a real-LM critical path around 85–100s instead of 136–138s.

No timing claim is made until both normal Actions jobs pass and repeat.

## Tradeoff

Adds one runner job and duplicates model-cache restore/Ollama setup to reduce wall-clock latency. The existing `Run Tests with Real LM` context is retained for the first shard; the new `Run Remaining Tests with Real LM` context must also become required before merge.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -139,13 +139,8 @@ jobs:
               echo "$?" > "$RUNNER_TEMP/ollama-pull.status"
             fi
           ) &
-      - name: Install uv with caching
+      - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
-        with:
-          enable-cache: true
-          cache-dependency-glob: |
-            **/pyproject.toml
-            **/uv.lock
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -111,10 +111,18 @@ jobs:
         run: uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n auto --dist worksteal
   
   llm_call_test:
-    name: Run Tests with Real LM
+    name: ${{ matrix.job_name }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - job_name: Run Tests with Real LM
+            test_expression: stream_listener_chat_adapter or stream_listener_json_adapter or single_module_call_with_usage_tracker
+          - job_name: Run Remaining Tests with Real LM
+            test_expression: not (stream_listener_chat_adapter or stream_listener_json_adapter or single_module_call_with_usage_tracker)
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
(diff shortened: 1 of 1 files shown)


isaacbmiller:
### First Actions run: tests pass; one setup stall requires repeat

[Run 31192206458](https://github.com/stanfordnlp/dspy/actions/runs/31192206458) passed both real-LM shards:

- 3-test shard: 104s job, 49s pytest
- 2-test remainder: 161s job, 38s pytest

The 2-test job is not a valid clean timing sample: its unchanged `setup-uv` action stalled for 75s, versus 7s in the sibling job. Excluding that external stall, its measured stages total about 93s. The first shard's clean 104s result is 32–34s below #10169's repeated 136–138s unsharded jobs.

Coverage is intact (3 executing + 2 executing tests; all provider-specific skips retained). Triggering an unchanged repeat and withholding the speed claim until both shards get clean setup timings.

isaacbmiller:
Rebased the benchmark stack logically onto isolated #10174 (disable the 154KB uv cache for real-LM jobs). #10173's own diff remains only the two-way LM sharding change. The new exact head will test both shards without the repeatedly observed 75s cache-restore stall.

isaacbmiller:
### Corrected-parent paired Actions result

[Exact-head run 31193946419](https://github.com/stanfordnlp/dspy/actions/runs/31193946419) passed every normal CI job after updating this stack to #10174's explicit `enable-cache: false` correction.

This gives a particularly clean comparison because the parent [#10174 run 31193870063](https://github.com/stanfordnlp/dspy/actions/runs/31193870063) started only 56 seconds earlier under the same repository load:

| real-LM shape | jobs | critical lane |
|---|---:|---:|
| unsharded corrected parent | 1 | **139s** |
| this PR | 2 | **106s / 97s** |

**Paired critical-lane saving: 33s / 23.7%.**

The test calls themselves were balanced at 50s and 43s. Runtime collection confirms exactly the same five Ollama-backed tests passed; the same 19 provider-credential tests skipped. This is also the second green sharded result (the prior run was 107s / 98s), so the result is repeatable rather than a single lucky runner.

isaacbmiller:
Closing after validation against current `main`. The shard partition is correct: all 24 selected real-LM nodes remain covered as disjoint 3 + 21 sets, and the historical paired Actions result genuinely reduced the isolated lane from 139s to 106s / 97s.

The current cost/benefit no longer supports the extra job. Each shard duplicates Ollama/Docker setup and restores the 1,982,906,397-byte model cache. Across three recent main runs, reducing real-LM to 106s would improve whole-workflow completion by only 8s, 14s, and 0s because optional tests become the critical path. That does not justify another heavy runner and required check context.

PR #10174, opened 2026-08-07
ci: skip uv cache for real-LM tests

isaacbmiller:
## Summary

Disable setup-uv's dependency cache only in the real-LM job. This isolated PR stacks on #10169.

No dependencies, versions, tests, model setup, or assertions change. `uv sync` still installs the exact lockfile environment; only the action cache around uv's download cache is removed.

## Why

The cache is restoring only ~154KB and dependency installation takes 3–5s, but concurrent real-LM measurements repeatedly showed one identical-key restore stall for 75s. Another primary job saw a 185s restore stall. In the real-LM shard runs:

| Job | setup-uv cache | dependency install |
|---|---:|---:|
| sibling A, run 1 | 7s | 4s |
| sibling B, run 1 | **75s** | 4s |
| sibling A, run 2 | normal | normal |
| sibling B, run 2 | **~75s** | normal |

Caching 154KB cannot repay a 75s failure mode. Normal Actions will measure the uncached setup and exact dependency install before this is considered ready.

## Validation

- Workflow YAML/pre-commit passes
- Exact normal Actions pending

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -133,6 +133,8 @@ jobs:
           ) &
       - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: false
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -131,13 +131,8 @@ jobs:
               echo "$?" > "$RUNNER_TEMP/ollama-pull.status"
             fi
           ) &
-      - name: Install uv with caching
+      - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
-        with:
-          enable-cache: true
-          cache-dependency-glob: |
-            **/pyproject.toml
-            **/uv.lock
       - name: Create and activate virtual environment
         run: |
           uv venv .venv


isaacbmiller:
### First exact-head Actions result

[Run 31193168275](https://github.com/stanfordnlp/dspy/actions/runs/31193168275) passed every normal check.

Real-LM job: **130s** versus #10169's 136–138s repeats. The uncached setup was deterministic and cheap:

- setup-uv install: 8s
- exact locked dependency install: 3s
- no 75s cache stall

This is primarily a latency-tail/reliability change; the clean-path observed saving is 6–8s / 4.4–5.8%. Triggering an unchanged repeat before finalizing.

isaacbmiller:
Correction after Greptile's exact review: omitting `enable-cache` leaves setup-uv's automatic cache enabled. Commit 00430b4 now sets `enable-cache: false` explicitly. Therefore I am **discarding the earlier 130s run as evidence for cache disablement** and will only report timings from the corrected head. Thanks to the review gate for catching this.

isaacbmiller:
### Corrected exact-head Actions result

[Run 31193870063](https://github.com/stanfordnlp/dspy/actions/runs/31193870063) passed every normal CI job with `enable-cache: false` explicitly active.

Real-LM job: **139s**. Install uv took 7s and the exact locked dependency install took 6s. This confirms disabling this tiny uv cache does not create a material setup regression, but it is **not a clean-path speedup** versus #10169's 136–138s runs.

The narrow value is tail-latency protection: identical-key setup-uv cache restores previously stalled 75s twice in real-LM shard runs and 185s in another test job, while the uncached install is 13s total and does not contact that restore path. I am not counting the pathological avoided stalls as a guaranteed per-run saving.

isaacbmiller:
Closing after independently tracing the reported setup-uv stall. In the cited 75s job, “Downloading uv” began at 15:21:56, while “Trying to restore cache” did not begin until 15:23:10. The actual 154,577-byte cache restore then completed in about 0.5s. The delay therefore occurred before cache restoration, so `enable-cache: false` would not avoid the observed failure mode.

Normal current restores remain under a second, locked dependency installation takes roughly 3–6s, and the corrected uncached run took 139s versus cached parent runs of 136–138s. The corrected configuration works, but the claimed tail-latency benefit is not supported and there is no clean-path improvement.

PR #10172, opened 2026-08-07
ci: benchmark optional test shards on split suite

isaacbmiller:
## Temporary CI benchmark — draft until #10164 merges

This measures optional-suite sharding on top of #10164's still-open split-suite workflow and current optimized `main`. Once #10164 lands, its overlapping diff drops out and this becomes an isolated optional-sharding PR; the combined workflow must then be rerun before readiness.

## Candidate

The 166-test optional lane has one severe xdist straggler: seven Deno tests account for about 60s of 132s aggregate test time, including a single 22s 100MB-list roundtrip. Run those seven tests in their own parallel job while the other 159 optional tests run unchanged in the existing job.

Tests, markers, dependencies, Deno version, Python versions, worker count, distribution strategy, payload sizes, and assertions are unchanged. The two `-k` expressions are exact complements; future tests default to the 159-test remainder side.

## Coverage proof

| Set | Tests |
|---|---:|
| Optional unsharded | 166 |
| Heavy Deno shard | 7 |
| Optional remainder | 159 |
| overlap | **0** |
| omitted | **0** |

Both exact shard commands passed locally with Deno 2.7.7:

- heavy shard: 7 passed, 43.07s wall
- remainder: 159 passed, 29.52s wall
- unsharded: 166 passed, 59.51s wall

Local critical-path improvement: **16.44s / 27.6%**. Normal Actions is the deciding measurement.

## Tradeoff

Adds five runner jobs to lower wall-clock latency. If retained, branch protection must require the five new `Run Heavy Deno Tests` contexts in addition to the existing optional checks.

PR #10171, opened 2026-08-07
ci: shard primary tests

isaacbmiller:
## Summary

Split the 1,446-node primary test suite into two deterministic path shards on every supported Python version:

- `Primary Tests: Clients, Adapters, and Utils`: `tests/clients tests/adapters tests/utils`
- `Primary Tests: All Other Directories`: all other tests

Tests, markers, dependencies, Python versions, worker count, distribution strategy, and assertions are unchanged. New tests default to the remainder shard.

## Coverage proof

| Set | Tests |
|---|---:|
| Primary unsharded | 1,446 |
| clients/adapters/utils | 538 |
| remainder | 908 |
| overlap | **0** |
| omitted | **0** |

Both exact shard commands passed. Greptile independently reproduced the partition and execution proof at 5/5 confidence.

## GitHub Actions evidence

Across repeated green runs, the primary critical path fell from a 134.6s unsharded mean to approximately 44.6s after the final Deno cleanup: about **90s / 66.9% faster**.

This does not burn additional aggregate runner time despite adding five checks:

- unsharded primary jobs: 673s aggregate
- sharded primary jobs: 411–429s aggregate
- reduction: approximately **36–39%**

The tradeoff is five additional visible `Primary Tests: All Other Directories (3.x)` checks. Branch protection must include them, or use the stacked aggregate scheduling/gate policy in #10184.

## Validation

- Exact-head normal `Lint, Test, and Build` workflow: green
- Greptile: 5/5
- No production code changes
- No test behavior or coverage changes

PR #10164, opened 2026-08-07
ci: run extra tests in parallel

isaacbmiller:
## Summary

Run the existing extra/Deno test matrix in parallel with the primary pytest matrix instead of serially at the end of every primary test job.

No test command, marker, Python version, assertion, dependency extra, or Deno version changes. The existing required `Run Tests (3.x)` checks retain their names; this only introduces `Run Extra and Deno Tests (3.x)` checks for the work that was previously embedded at the end of those jobs.

## Measured CI impact

Two complete five-version GitHub Actions runs passed:

| Run | End-to-end workflow | Slowest test job |
|---|---:|---:|
| [First](https://github.com/stanfordnlp/dspy/actions/runs/31136302554) | 232s | 228s |
| [Warm-cache repeat](https://github.com/stanfordnlp/dspy/actions/runs/31136607226) | 234s | 231s |

The fastest nearby successful workflow using the old serial structure took [267s](https://github.com/stanfordnlp/dspy/actions/runs/31132102278), with a 265s slowest test job. That comparison is conservative because the baseline already contains #10154's separate LiteLLM speedup.

Observed saving:

- **End-to-end workflow:** 33–35s / 12.4–13.1%
- **Test critical path:** 34–37s / 12.8–14.0%

The exact-parent main run took 287s, but the faster 267s comparison is used to avoid overstating the result.

## Tradeoff

This intentionally trades runner consumption for feedback latency. The five test jobs consumed 21.2 runner-minutes in the serial baseline; the primary plus extra matrices consumed 25.5 and 26.2 runner-minutes in the two split runs, an increase of roughly **20–24%**. Public-repository hosted-runner billing is not the constraint here, but the extra concurrent capacity is a real operational cost and should be weighed against the repeatable 12–13% wall-clock improvement.

Total test coverage and behavior are unchanged.

## Validation

- `pre-commit run --files .github/workflows/run_tests.yml`
- Two complete normal Actions runs
- All primary, extra/Deno, real-LM, build, lint, and security jobs passed in both runs

PR #10175, opened 2026-08-07
ci: benchmark combined pytest runtime optimizations

isaacbmiller:
## Benchmark only — do not merge

Temporary integration benchmark for measuring the **actual end-to-end GitHub Actions critical path** after applying the independently reviewed pytest runtime changes:

- #10164: run optional/Deno tests in their own Python matrix
- #10169: overlap Ollama image prefetch with environment setup
- #10171: split the 1,446 primary nodes into complete 538/908 path shards
- #10172: split the 166 optional nodes into complete 159/7 shards
- #10174: explicitly disable pathological setup-uv caching only for real-LM jobs
- #10173: split the 24-node real-LM collection into complementary matrix selections

Every shard uses complementary selectors previously collection-tested for zero overlaps and zero omissions. This PR exists solely to obtain a normal Actions run containing all lanes simultaneously. Each production change remains isolated in its own PR.

I will report exact job timing, workflow wall time, coverage/selection evidence, and an unchanged repeat. **Do not merge this integration branch.**

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -133,6 +133,8 @@ jobs:
           ) &
       - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: false
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -139,13 +139,8 @@ jobs:
               echo "$?" > "$RUNNER_TEMP/ollama-pull.status"
             fi
           ) &
-      - name: Install uv with caching
+      - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
-        with:
-          enable-cache: true
-          cache-dependency-glob: |
-            **/pyproject.toml
-            **/uv.lock
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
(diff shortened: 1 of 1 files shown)


isaacbmiller:
### Integration run 1: correctness pass, timing rejected

[Run 31194084418](https://github.com/stanfordnlp/dspy/actions/runs/31194084418) passed all **29 jobs**, proving the merged workflow shape is valid. All five Python versions passed in all four pytest partitions, and both real-LM shards passed.

I am rejecting the 238s workflow wall time as performance evidence: this run overlapped #10173 and #10174's benchmark workflows, so many jobs waited for runner slots. The executed job critical path was 107s, but start delay contaminated end-to-end latency.

Now triggering one unchanged repeat with the other benchmark workflows complete. That quiet repeat is the timing result.

isaacbmiller:
### Integration run 2: fully green, but 20-slot cap makes all-shards composition slower

[Run 31194509197](https://github.com/stanfordnlp/dspy/actions/runs/31194509197) passed all 29 jobs again. No other workflow was active.

- Workflow wall time: **160s**
- Peak concurrent Actions jobs, reconstructed from exact job timestamps: **20**
- Slowest executed job: **112s**
- One real-LM shard did not start until **50s** after the workflow began

Therefore the extra 48s was internal runner-slot queueing, not pytest. This all-shards composition is **11s slower** than the previous 149s combined state and should not ship as a bundle.

This result proves isolated lane savings do not add linearly. The slot-aware follow-up #10177 retains primary and real-LM sharding, drops the lower-yield optional sub-shard, and defers five short builds so all 17 long test jobs can start under the 20-job cap.

isaacbmiller:
Closing completed benchmark: all 29 jobs were green, but the quiet 160s run proved unthrottled all-shards composition exceeds the 20-job cap and regresses wall time.

PR #10178, opened 2026-08-07
ci: run optional pytest partitions concurrently per job

isaacbmiller:
## Summary

Run the existing complete optional-suite partition inside each Python-version job as two concurrent pytest processes:

- 159 remaining optional/Deno nodes with 2 xdist workers
- 7 heavy Deno nodes with 2 xdist workers

This retains four total xdist workers and does not add GitHub Actions jobs, which matters because the combined benchmark measured a hard 20-job concurrency ceiling.

## Behavior and coverage proof

Collection against the production marker/flags:

| set | nodes |
|---|---:|
| unsplit `extra or deno` | 166 |
| remaining selection | 159 |
| heavy selection | 7 |
| overlap | 0 |
| omitted | 0 |
| extra | 0 |

Both subprocesses are always waited. Any non-zero process status fails the Actions step. No test, marker, assertion, dependency, Python version, or Deno version changes.

## Why this shape

#10172 proved the same 159/7 split reduces the optional critical lane by 24.1–29.2% when represented as ten Actions jobs. #10175 then proved those five additional jobs overflow the repository's 20-runner cap and make the complete workflow slower. This PR tests whether preserving that partition *inside* the existing five jobs captures the load-balancing gain without consuming runner slots.

Normal Actions timing and an unchanged repeat will determine whether this is worth merging; no offline speed claim is used.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -141,7 +141,18 @@ jobs:
       - name: Install dependencies
         run: uv sync -p .venv --extra dev --extra test_extras
       - name: Run extra and deno tests
-        run: uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n auto --dist worksteal
+        run: |
+          uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n 2 --dist worksteal \
+            -k 'not (large_list_variable or multiple_large_variables or enable_write_flag or interpreter_security_filesystem_access or read_file_access_control or enable_env_vars_flag or enable_net_flag)' &
+          remaining_pid=$!
+          uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n 2 --dist worksteal \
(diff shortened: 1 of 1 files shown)


isaacbmiller:
### Actions result: reject — no critical-path improvement

The optional jobs all passed, including both concurrent complementary selections in every Python version.

| metric | parent #10164 | this PR |
|---|---:|---:|
| range across five versions | 103–119s | 102–119s |
| mean | 113.6s | 110.4s |
| critical path | **119s** | **119s** |

The 159/7 partition remains complete (166 union, zero overlap, zero omissions), but statically assigning two workers to each process did not shorten the slowest version. The 3.2s mean movement does not affect workflow wall time and is below the bar for another layer of shell orchestration.

Closing as a disproven candidate. No merge recommendation.

isaacbmiller:
Closing: real Actions preserved coverage but showed zero critical-path improvement (119s → 119s).

PR #10180, opened 2026-08-07
ci: benchmark six optional test workers

isaacbmiller:
## Benchmark candidate

Use six xdist workers only for the optional/Deno matrix. The 166-node selection, `worksteal` scheduler, tests, assertions, flags, dependencies, Python versions, and Deno version are unchanged.

The default Actions runner exposes four logical CPUs. Eight-worker oversubscription was already rejected as 22% slower on Actions; this tests whether a smaller 1.5× oversubscription can hide Deno subprocess waits without the eight-worker contention penalty.

Acceptance bar: reduce the slowest optional job (parent 119s exact-head reference), then repeat unchanged. Otherwise close. No offline speed claim.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -141,7 +141,7 @@ jobs:
       - name: Install dependencies
         run: uv sync -p .venv --extra dev --extra test_extras
       - name: Run extra and deno tests
-        run: uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n auto --dist worksteal
+        run: uv run -p .venv pytest tests/ -m 'extra or deno' --extra --deno -n 6 --dist worksteal
 
   llm_call_test:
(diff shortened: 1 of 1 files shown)


isaacbmiller:
### Actions result: reject — 31% critical-path regression

All five optional jobs passed, but six workers oversubscribed the four-vCPU runner and materially regressed runtime:

| metric | parent 4 workers | this PR 6 workers |
|---|---:|---:|
| range | 103–119s | 123–156s |
| mean | 113.6s | 132.0s |
| critical path | **119s** | **156s** |

Critical-path regression: **37s / 31.1%**. Together with the prior eight-worker rejection, this confirms the existing four-worker `worksteal` configuration is the right per-job scheduler.

isaacbmiller:
Closing: real Actions showed a 37s / 31.1% critical-path regression.

PR #10177, opened 2026-08-07
ci: benchmark slot-aware pytest workflow

isaacbmiller:
## Benchmark only — do not merge

This temporary integration benchmark accounts for the **measured 20-job Actions concurrency cap**. The all-shards benchmark #10175 proved that 29 jobs queue internally and completed in 160s despite every executed job being at most 112s.

Composition:

- #10164: optional/Deno tests in their own five-version matrix
- #10169: prefetch Ollama image
- #10171: shard the primary lane (largest isolated gain)
- #10173: shard the real-LM lane
- #10174: avoid pathological uv cache restores in real-LM jobs
- deliberately **exclude #10172's optional sub-shard**, because its five extra jobs exceed the cap for only ~28s of isolated gain
- defer the five 14–24s build jobs until the two primary shards release runners; `if: ${{ !cancelled() }}` preserves build execution even when a prerequisite test fails

At startup this leaves exactly 19 runnable jobs: 10 primary shards, 5 optional jobs, 2 real-LM shards, lint, and Ruff. All critical work can start without queueing. Builds should finish after the ~35–60s primary lanes and before the expected ~100–114s critical lane.

The goal is an exact normal-Actions end-to-end measurement. Every production change remains isolated. **Do not merge this integration branch.**

PR #10181, opened 2026-08-07
ci: benchmark all shards within runner capacity

isaacbmiller:
## Benchmark only — do not merge

This follow-up reconciles the successful isolated optional shards (#10172) with the measured 20-job Actions concurrency cap:

- primary shard matrices: `max-parallel: 3` each = 6 runnable primary jobs
- optional shard matrices: 10 runnable jobs
- real-LM shards: 2 runnable jobs
- workflow lint + Ruff: 2 runnable jobs
- **startup total: exactly 20**
- builds wait for both primary matrices, still run after failures via `if: ${{ !cancelled() }}`

The four throttled primary entries should enter as 35–45s jobs release slots. Prior runs put the slowest primary shard at 61s, so the second wave should finish near or below the ~100–110s optional/real-LM lanes. Builds then take 15–24s.

Compared with:

- #10175 all shards unthrottled: 160s because jobs queued unpredictably
- #10177 no optional sub-shard: repeated 131–139s / 135s mean

This run determines whether explicit capacity scheduling can retain all isolated lane gains. Every test partition is unchanged from the twice-green #10175 composition. **Do not merge this integration branch.**

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -133,6 +133,8 @@ jobs:
           ) &
       - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: false
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -139,13 +139,8 @@ jobs:
               echo "$?" > "$RUNNER_TEMP/ollama-pull.status"
             fi
           ) &
-      - name: Install uv with caching
+      - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
-        with:
-          enable-cache: true
-          cache-dependency-glob: |
-            **/pyproject.toml
-            **/uv.lock
       - name: Create and activate virtual environment
         run: |
           uv venv .venv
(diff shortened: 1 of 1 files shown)


isaacbmiller:
### Run 1: all 29 jobs green; 135s upper bound with one Docker pull outlier

[Run 31196384779](https://github.com/stanfordnlp/dspy/actions/runs/31196384779) passed everything.

The capacity schedule behaved as intended:

- exactly 20 startup jobs launched immediately
- second-wave primary jobs finished by ~98s
- deferred builds finished by ~114s
- optional critical job finished by ~110s

Workflow wall time was **135s** because one real-LM shard's background Docker pull made `Start Ollama` wait 39s. Its sibling shard in the same run took 8s at that stage. The affected shard's pytest step itself was a normal 42s.

Treating this as correctness proof and a conservative upper bound. Triggering an unchanged repeat to measure without the single registry outlier.

isaacbmiller:
### Final result: reject in favor of #10177

Both capacity-scheduled all-shards runs passed all 29 jobs:

| run | wall time | critical job |
|---|---:|---:|
| 31196384779 | 135s | real-LM 132s |
| 31196645913 | 145s | real-LM 141s |
| **mean** | **140s** | |

The 20-slot schedule worked exactly as designed: optional shards completed in 98–107s, second-wave primary jobs stayed below the critical path, and builds completed after primary. But one real-LM shard hit a cold Docker pull in both runs, making the full-workflow mean **5s slower** than #10177's repeated 135s mean.

This means the optional sub-shard's isolated gain still does not improve measured integrated wall time under the actual latency distribution. #10177 remains the best complete composition. Closing this benchmark with no merge recommendation.

isaacbmiller:
Closing benchmark: twice green, but 140s mean is slower than #10177's 135s mean.

PR #10182, opened 2026-08-07
ci: run host Ollama for real-LM tests

isaacbmiller:
## Benchmark candidate

Replace the pinned `ollama/ollama:0.31.2` Docker image with Ollama's official pinned `v0.31.2` Linux release archive from GitHub Releases.

Unchanged:

- Ollama version 0.31.2
- `llama3.2:3b` model and existing model cache key/path
- AMX backend removal for Sapphire Rapids runners
- health check, model warm-up, pytest selection, assertions, and LM environment

The 1.4GB archive download still starts before uv/dependency/model-cache setup and is SHA-256 verified against the release's published checksum. Ollama runs directly on the host and is always stopped afterward.

## Why

Recent green integration runs repeatedly showed one real-LM shard spending 75–80s on a cold Docker Hub image pull, leaving a 39–40s wait after all overlapping setup completed. Warm-image jobs start Ollama in 6–8s. This benchmark tests whether GitHub's own release CDN plus direct extraction removes that Docker Hub tail.

Acceptance bar: exact normal CI green, faster real-LM job on the first run, an unchanged repeat, and Greptile 5/5. Otherwise close. No model/test behavior change and no local timing claim.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -122,13 +122,15 @@ jobs:
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: 3.11
-      - name: Prefetch Ollama image
+      - name: Prefetch Ollama
         run: |
           (
-            if docker pull ollama/ollama:0.31.2 > "$RUNNER_TEMP/ollama-pull.log" 2>&1; then
-              echo 0 > "$RUNNER_TEMP/ollama-pull.status"
+            if curl --fail --location --silent --show-error \
+              --output "$RUNNER_TEMP/ollama-linux-amd64.tar.zst" \
+              https://github.com/ollama/ollama/releases/download/v0.31.2/ollama-linux-amd64.tar.zst; then
+              echo 0 > "$RUNNER_TEMP/ollama-download.status"
             else
-              echo "$?" > "$RUNNER_TEMP/ollama-pull.status"
+              echo "$?" > "$RUNNER_TEMP/ollama-download.status"
             fi
           ) &
       - name: Install uv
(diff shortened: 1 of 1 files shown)


selormwalker (commented):
Code quality and documentation updates look well structured.

isaacbmiller:
### Run 1: all CI green; host path removes Docker tail but is clean-path neutral

[Run 31197167528](https://github.com/stanfordnlp/dspy/actions/runs/31197167528) passed all normal jobs. Real-LM job: **138s**.

The release archive completed during overlapping setup. SHA verification, extraction, AMX backend removal, and service startup then took 9s. The same model cache, warm-up, and complete real-LM test selection passed.

This matches clean Docker runs rather than improving them, but avoided the 39–40s post-setup Docker waits seen in two recent integration runs. Triggering an unchanged repeat to test whether that tail avoidance is stable.

isaacbmiller:
### Unchanged repeat: 135s real-LM job, stable host startup

[Run 31197532250](https://github.com/stanfordnlp/dspy/actions/runs/31197532250) passed the complete real-LM job again.

| host-binary run | real-LM job | verify/extract/start |
|---|---:|---:|
| run 1 | 138s | 9s |
| run 2 | 135s | 7s |

The 1.4GB pinned archive download was fully hidden by setup twice. Mean clean-path movement versus the corrected 139s Docker parent is only 2.5s, so I am not claiming a meaningful routine speedup. The evidence is stable elimination of the Docker pull tail: 7–9s host startup versus two recent 39–40s waits after setup.

Awaiting exact-head full CI and Greptile before any merge recommendation.

selormwalker (commented):
Looks solid! Thanks for contributing these updates.

isaacbmiller:
Closing after validating both host-Ollama runs and the current Docker-prefetch path. The pinned v0.31.2 release archive and SHA-256 are correct, and the host setup passed twice at 138s / 135s with 7–9s verify/extract/start time.

That is effectively equal to the already-prefetched Docker path, which normally starts in roughly 8–13s. The cited 39–40s evidence does not demonstrate a remaining post-prefetch regression. Host mode additionally downloads a 1.435GB archive and takes ownership of archive layout, `unzstd`, library paths, PID cleanup, and AMX library management; both measured runs also used warm model-cache hits. The extra maintenance is not justified by the observed performance. If mutable Docker tags become a supply-chain concern, pinning the existing image by digest is the smaller change.

PR #10185, opened 2026-08-07
Add timeout to Audio.from_url so it doesn't hang forever

pandasuwu:
requests.get had no timeout, so a slow or unresponsive server could make from_url hang indefinitely. Added a timeout parameter (default 30s) and passed it through to requests.get. Passing timeout=None keeps the old behaviour. Added tests for the default, a custom value, and that a timeout error is raised instead of hanging.

Part of #9993

### DIFF
--- a/dspy/adapters/types/audio.py
+++ b/dspy/adapters/types/audio.py
@@ -80,7 +80,7 @@ class Audio(Type):
         return encode_audio(values)
 
     @classmethod
-    def from_url(cls, url: str, verify: bool = True) -> "Audio":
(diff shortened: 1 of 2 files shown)


immu4989:
Thanks for looking at #9993. This overlaps with #10149, opened on the 4th, which is linked in that issue thread and covers the same timeout with the same default and the same `timeout=None` escape, for `Audio.from_url`, `Image.from_url`, and the internal helper behind the deprecated `Image(url, download=True)` path. Image has the identical unbounded `requests.get` at `image.py:246`, so a fix that only touches Audio leaves half the surface open.

The issue author has already reviewed the state of things in that thread and is happy for #9993 to close once #10149 lands.

PR #10189, opened 2026-08-08
fix(utils): don't silently swallow harness-level exceptions in ParallelExecutor

codechrl:
## Why

`ParallelExecutor._execute_parallel` (`dspy/utils/parallelizer.py`) wraps `f.result()` in a bare `except Exception: pass`:

```python
try:
    index, outcome = f.result()
except Exception:
    pass
else:
    ...
```

`f.result()` re-raises whatever `worker()` raised. `worker()`'s own setup code — deep-copying thread-local overrides (`copy.deepcopy(new_overrides["usage_tracker"])`), setting the context var — runs *outside* `_wrap_function`'s try/except, which only guards the user function call itself. So an exception from that harness-level setup (e.g. a non-deepcopyable `usage_tracker` override) is silently dropped: no log, no entry in `failed_indices`/`exceptions_map`, and `results[idx]` stays `None` forever — indistinguishable from the task simply returning nothing.

## What

Factor the existing error-handling out of `_wrap_function` into a shared `_record_error()` helper (error-count/cancel logic + logging), and reuse it in the `f.result()` except-block so a harness-level failure is tracked and logged the same way a user-function exception already is.

## Tests

Added `test_parallel_executor_records_harness_level_exceptions` to `tests/utils/test_parallelizer.py` — monkeypatches `copy.deepcopy` to fail during worker setup and asserts the failure lands in `failed_indices`/`exceptions_map` instead of vanishing. Confirmed test-first: fails on `main` (`failed_indices == []`), passes with the fix. Full `test_parallelizer.py`: 13 passed. `ruff check`/`ruff format --check` clean.


codechrl:
Confirmed via direct trace. Fixing by checking idx not in self.exceptions_map instead.

codechrl:
Confirmed. Fixed by reconciling both completion paths through a shared _should_finalize check -- a later success now clears a stale harness failure instead of leaving both recorded.

codechrl:
Rebased onto main to pick up #10142's _UNSET sentinel. _should_finalize now checks against _UNSET instead of None, so a legitimate None result isn't mistaken for an unfinished slot. Added a direct regression test for that.

codechrl:
Confirmed. Fixed by decrementing error_count in _clear_stale_failure when it actually pops a stale entry, so a resolved failure stops consuming max_errors budget.

codechrl:
Confirmed. _clear_stale_failure now clears budget-triggered cancellation (real interrupts tracked separately and never uncancelled), retries for failed indices run despite cancellation, and the loop drains briefly so a recovering retry's outcome is collected.

codechrl:
Confirmed. User errors are now recorded on the main thread at outcome acceptance, same as harness failures, so a duplicate straggler failure counts once.

codechrl:
Confirmed. Dropped the recovery window entirely: the loop now waits for an in-flight retry of a failed index to complete, same as the non-cancelled path waits on in-flight work. Also removed safe_func's own cancel check, which could skip an admitted retry and pass its None off as a recovery.

codechrl:
Confirmed both directions of this. Settled on explicit bounded semantics: after budget cancellation, an in-flight retry gets one straggler-timeout grace period to recover, then it is abandoned with a warning and the recorded failure stands. Resubmission is now tracked per logical index so a retry is never straggler-retried again, and an input skipped under a since-revoked cancellation gets resubmitted instead of silently returning None.

codechrl:
Right -- a running pool thread can't be force-killed, only cooperatively cancelled, so an abandoned blocked retry outliving the call is inherent to threads. Made it as tight as the pool allows (shutdown now passes cancel_futures=True so queued work never starts) and documented the cooperative semantics on the class, the abandonment warning, and here. A hard bound would need process-based execution, which is a much larger change out of scope for this fix.

PR #10194, opened 2026-08-10
fix(callbacks): report base exceptions to end handlers

isaacbmiller:
## Summary

Report all terminating `BaseException` values to callback end handlers before propagating them.

The callback wrapper previously caught only `Exception`. `finally` still invoked the end handler for `asyncio.CancelledError`, `KeyboardInterrupt`, and `SystemExit`, but passed both `outputs=None` and `exception=None`, making interrupted operations appear successful to tracing consumers.

This PR:

- catches `BaseException` around the instrumented operation;
- passes the exact exception object to the matching end callback;
- re-raises with bare `raise` to preserve the original traceback;
- widens `BaseCallback` end-handler annotations to `BaseException | None`;
- updates DSPy's built-in `StatusStreamingCallback` overrides accordingly.

Callback-handler isolation intentionally continues to catch only `Exception`: DSPy should not suppress `KeyboardInterrupt` or `SystemExit` raised by callback code itself.

## Behavioral contract

Cancellation and process-level interruption remain control-flow operations, not recoverable errors. This change observes them without swallowing or translating them:

```text
instrumented operation raises BaseException
├── end callback receives outputs=None and the exact exception
└── the same exception immediately propagates
```

## Verification

```bash
uv run --frozen pytest -q tests/callback/test_callback.py
uv run --frozen ruff check \
  dspy/utils/callback.py \
  dspy/streaming/messages.py \
  tests/callback/test_callback.py
```

Result: **16 passed**, Ruff clean.

Coverage includes ordinary exception identity, `KeyboardInterrupt`, `SystemExit`, and async task cancellation. Each test also verifies callback context restoration.

### DIFF
--- a/dspy/streaming/messages.py
+++ b/dspy/streaming/messages.py
@@ -117,7 +117,7 @@ class StatusStreamingCallback(BaseCallback):
         self,
         call_id: str,
         outputs: dict[str, Any] | None,
-        exception: Exception | None = None,
+        exception: BaseException | None = None,
     ):
         stream = settings.send_stream
         if stream is None or outputs == "Completed.":
@@ -145,7 +145,7 @@ class StatusStreamingCallback(BaseCallback):
         self,
         call_id: str,
         outputs: dict[str, Any] | None,
-        exception: Exception | None = None,
+        exception: BaseException | None = None,
     ):
         stream = settings.send_stream
         if stream is None:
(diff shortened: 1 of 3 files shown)


isaacbmiller:
discussed with @michaelisaac-dev offline

PR #10196, opened 2026-08-11
test: add coverage reporting and cover save_data

clavedeluna:
## 📝 Changes Description

This PR contains the following changes:

- Adds `pytest-cov` to the `dev` extra and wires `--cov=dspy --cov-report=term-missing --cov-report=xml` into the CI `test` job (confirmed compatible with the existing `pytest-xdist` / `--dist worksteal` setup), uploading `coverage.xml` as a workflow artifact. Documents the `--cov` flag for local runs.
- Adds a regression test for `save_data` in `dspy/clients/utils_finetune.py` (`tests/clients/test_utils_finetune.py`), which had no coverage at all.

**Why coverage:** we currently have no visibility into which parts of `dspy/` are actually exercised by the test suite. Making coverage visible in CI (and easy to reproduce locally with one flag) is a low-risk first step toward finding untested, high-risk code paths like the one below — it's observability for the test suite itself, not a gate (no minimum threshold is enforced by this PR).

**Why the `save_data` test isn't just about the coverage number:** `save_data` is the function that serializes fine-tuning examples to a JSONL file before `BootstrapFinetune` (and friends) hand that file off to an external, paid, slow fine-tuning API. It resolves its write location via `get_finetune_directory()` (which honors a `DSPY_FINETUNEDIR` override, falling back to DSPy's shared cache dir) and names the file by a content hash. A regression here — wrong directory, an ignored `DSPY_FINETUNEDIR` override, truncated/malformed JSONL — would not fail loudly at the point of the bug. It would surface expensively downstream: a fine-tuning job silently trains on the wrong or corrupted data, or writes into a shared cache dir a user expected to be sandboxed, and nobody notices until the resulting model behaves oddly, well after the (non-refundable) training run finished. That's exactly the kind of quiet, unverified, expensive-to-fail path coverage is good at surfacing. The new test locks in the contract every fine-tuning caller relies on: the returned path is absolute, lives under the configured (or overridden) fine-tune directory, ends in `.jsonl`, and round-trips the exact data it was given.

## ✅ Contributor Checklist

## ⚠️ Warnings

- This only makes coverage *visible* (terminal output + `coverage.xml` artifact) — it does not enforce a minimum coverage threshold or gate merges on it.


.github/workflows/run_tests.yml
```diff
@@ -104,7 +104,15 @@ jobs:
           uv sync --dev -p .venv --extra dev
           uv pip list
       - name: Run tests with pytest
-        run: uv run -p .venv pytest -vv -m 'not extra and not deno' -n auto --dist worksteal tests/
+        run: |
+          uv run -p .venv pytest -vv -m 'not extra and not deno' -n auto --dist worksteal \
+            --cov=dspy --cov-report=term-missing --cov-report=xml tests/
```
clavedeluna, line 109: Do you think it's reasonable to ignore a dir like datasets?

PR #10207, opened 2026-08-14
Prevent PythonInterpreter response ID forgery

isaacbmiller:
## Summary

- replace predictable sequential protocol IDs with random 128-bit request IDs
- freeze the runner's JSON and console protocol primitives before generated code runs
- serialize outer JSON-RPC frames through null-prototype objects

## Why

Generated Pyodide code can write JSON-looking lines through `js.console.log()`. With sequential IDs it can print the expected execution response before the runner, causing DSPy to accept a forged value and desynchronize the next request.

Random request IDs make the current response ID unavailable to generated code. Freezing the parser/output primitive and using a null-prototype outer frame prevents a previous turn from intercepting the next request or recovering its ID through prototype serialization hooks. Guessed frames fail closed through the existing mismatch handling.

## Size

The implementation changes 19 production lines across Python and the bundled runner. The remaining diff is focused adversarial coverage.

## Verification

- original hard-coded matching-ID forgery is rejected rather than returned
- attempts to replace `JSON.parse` and `console.log` do not affect later requests
- full Deno interpreter test file: 101 passed
- focused protocol regressions: 4 passed
- CI-equivalent Ruff check and `git diff --check` passed

### DIFF
--- a/dspy/primitives/python_interpreter.py
+++ b/dspy/primitives/python_interpreter.py
@@ -9,7 +9,6 @@ protocol defined in interpreter.py.
 import asyncio
 import dataclasses
 import functools
-import hmac
 import inspect
 import json
 import keyword
@@ -17,7 +16,6 @@ import logging
 import math
 import os
 import re
-import secrets
 import shutil
 import subprocess
 import threading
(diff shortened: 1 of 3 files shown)


isaacbmiller:
Superseded by #10208. Adversarial testing showed that authenticating tool calls inside the same JavaScript realm requires brittle machinery and does not remove meaningful authority: generated code already receives `_js_tool_call` and may invoke every explicitly bound tool. The replacement PR instead keeps the two small, honest robustness improvements—unpredictable request IDs and clean rejection of recursive interpreter execution.

PR #10215, opened 2026-08-15
fix: validate TypedDict input annotations instead of crashing in _check_type

### LINKED ISSUE #10214: [Bug] Input validation crashes on TypedDict-annotated input fields: "TypedDict does not support instance and class checks"
### What happened?

Calling any predictor whose signature has an input field annotated with a `TypedDict` raises `TypeError: TypedDict does not support instance and class checks`, even when the value passed is a perfectly valid dict.

The crash comes from the input type validation added in `Predict._forward_preprocess`. `_check_type` in `dspy/predict/predict.py` has no branch for TypedDicts, so a TypedDict annotation falls through to the plain-type case:

```python
# dspy/predict/predict.py
def _check_type(value: Any, expected: type) -> bool:
    ...
    # Plain type (int, str, BaseModel subclass, etc.)
    if isinstance(expected, type):
        return isinstance(value, expected)
```

A TypedDict class satisfies `isinstance(expected, type)` (its metaclass is `typing._TypedDictMeta`), but `typing` explicitly forbids `isinstance` checks against TypedDicts, so the validation helper itself raises instead of returning `True`/`False`.

This is inconsistent with the rest of DSPy: TypedDict annotations work fine for schema rendering and output parsing (pydantic supports them), so signatures using them are otherwise fully functional. Only this input-validation path breaks.

The same crash surfaces through `dspy.Flex`, where it is harder to diagnose: the sandbox bridge re-raises it as `CodeExecutionError: RuntimeError: ['TypeError: TypedDict does not support instance and class checks']` with no traceback pointing at `_check_type`.

### Steps to reproduce

```python
from typing import TypedDict

import dspy

dspy.configure(lm=DummyLM([{"should_reply": "True"}] * 10))


class EmailRecord(TypedDict):
    subject: str
    sender_email: str


class ShouldReply(dspy.Signature):
    """Decide whether the inbox owner should reply to this email."""

    email: EmailRecord = dspy.InputField()
    should_reply: bool = dspy.OutputField()


predict = dspy.Predict(ShouldReply)
predict(email={"subject": "Re: Infra recs", "sender_email": "mike@example.com"})
```

Output:

```
  ...
  File ".../dspy/predict/predict.py", line 251, in forward
    lm, config, signature, demos, kwargs = self._forward_preprocess(**kwargs)
  File ".../dspy/predict/predict.py", line 210, in _forward_preprocess
    if not _is_value_compatible_with_type(value, expected_type):
  File ".../dspy/predict/predict.py", line 336, in _is_value_compatible_with_type
    return _check_type(value, expected)
  File ".../dspy/predict/predict.py", line 394, in _check_type
    return isinstance(value, expected)
  File ".../typing.py", line 2956, in __subclasscheck__
    raise TypeError('TypedDict does not support instance and class checks')
TypeError: TypedDict does not support instance and class checks
```

The same signature wrapped in `dspy.Flex(ShouldReply)` fails identically (surfaced as a `CodeExecutionError` from the sandbox bridge).

## Suggested fix

Add a TypedDict branch to `_check_type` before the plain-type fallthrough, using `typing.is_typeddict`:

```python
import typing

def _check_type(value: Any, expected: type) -> bool:
    ...
    # TypedDict — isinstance() is forbidden; validate the value is a dict
    # (optionally check keys against __required_keys__/__optional_keys__).
    if typing.is_typeddict(expected):
        return isinstance(value, dict)

    # Plain type (int, str, BaseModel subclass, etc.)
    if isinstance(expected, type):
        return isinstance(value, expected)
```

A stricter version could also verify `expected.__required_keys__ <= value.keys()`, but simply not crashing is the main thing — downstream pydantic serialization already handles TypedDict values correctly.

Monkeypatching exactly this guard over `dspy.predict.predict._check_type` makes both `dspy.Predict` and `dspy.Flex` (including a full GEPA run) work with TypedDict inputs, which suggests validation is the only affected code path.

### DSPy version

3.3.0

arthi-arumugam-git:
Fixes #10214; the mechanism is exactly as dbreunig diagnosed.

A TypedDict class passes the `isinstance(expected, type)` guard (metaclass `typing._TypedDictMeta`), so `_check_type` fell through to `isinstance(value, expected)`, which `typing` forbids for TypedDicts and raises. Every signature with a TypedDict-annotated input field crashed on valid dict inputs, and through `dspy.Flex` the error resurfaced as an opaque `CodeExecutionError` with no pointer at `_check_type`.

The fix adds an `is_typeddict` branch ahead of the plain-type case, per the direction sketched in the issue: the value must be a dict and its keys must cover `__required_keys__`. So a valid dict validates cleanly, a missing required key produces the same type-mismatch WARNING every other annotation gets, and nothing raises.

Test added in the existing caplog style next to `test_list_type_validation`: valid dict produces no warning, missing required key warns, non-dict warns. It fails on current main with the issue's exact `TypeError` and passes with this change (verified both ways before pushing). The issue's original repro snippet also runs green end to end on this branch.

### DIFF
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -1,7 +1,7 @@
 import logging
 import random
 import types
-from typing import Any, Literal, Union, get_args, get_origin, is_typeddict
+from typing import Any, Literal, Union, get_args, get_origin, get_type_hints, is_typeddict
 
 from pydantic import BaseModel
 from pydantic_core import PydanticUndefined
@@ -330,11 +330,17 @@ def _check_type(value: Any, expected: type) -> bool:
     args = get_args(expected)
 
     # TypedDict: typing forbids isinstance checks against it, so validate the
-    # shape directly. Downstream pydantic serialization handles the values.
+    # shape directly: required keys must be present, no unknown keys, and each
+    # declared member checked against its own annotation.
     if is_typeddict(expected):
         if not isinstance(value, dict):
             return False
-        return expected.__required_keys__ <= value.keys()
+        hints = get_type_hints(expected)
+        if not expected.__required_keys__ <= value.keys():
+            return False
+        if not value.keys() <= hints.keys():
+            return False
+        return all(_check_type(item, hints[key]) for key, item in value.items())
 
     # Union / Optional (X | None)
     if origin is Union or origin is types.UnionType:
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -2,7 +2,18 @@ import logging
 import random
 import sys
 import types
-from typing import Any, Literal, Union, get_args, get_origin, get_type_hints, is_typeddict
+from typing import Any, Literal, Union, get_args, get_origin, get_type_hints
+
+try:
+    # typing.is_typeddict does not recognise classes built with
+    # typing_extensions.TypedDict, and an unrecognised one falls through to the
+    # isinstance check, which raises "TypedDict does not support instance and
+    # class checks". typing_extensions.is_typeddict recognises both flavours. It
+    # ships with pydantic, which is a hard dependency, but the import is guarded
+    # so a missing typing_extensions degrades instead of breaking imports.
+    from typing_extensions import is_typeddict
+except ImportError:
+    from typing import is_typeddict
 
 from pydantic import BaseModel
 from pydantic_core import PydanticUndefined
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -1,5 +1,6 @@
 import logging
 import random
+import sys
 import types
 from typing import Any, Literal, Union, get_args, get_origin, get_type_hints, is_typeddict
 
@@ -321,6 +322,25 @@ def _is_value_compatible_with_type(value: Any, expected: type) -> bool:
     return _check_type(value, expected)
 
 
+def _resolvable_member_hints(typed_dict: type) -> dict[str, Any]:
+    """Resolve the members of a TypedDict that can be resolved, and skip the rest.
+
+    get_type_hints is all or nothing: a single member naming a type that is not
+    importable from the definition site raises, and the annotations of every other
+    member are lost with it. Each member is resolved on its own here so an
+    unresolvable one costs only its own check.
+    """
+    namespace = getattr(sys.modules.get(typed_dict.__module__), "__dict__", {})
+    hints: dict[str, Any] = {}
+    for key, annotation in getattr(typed_dict, "__annotations__", {}).items():
+        holder = type("_Member", (), {"__annotations__": {key: annotation}})
+        try:
+            hints[key] = get_type_hints(holder, namespace)[key]
+        except Exception:
+            continue
+    return hints
+
+
 def _check_type(value: Any, expected: type) -> bool:
     """Stdlib replacement for typeguard.check_type."""
     if expected is Any:
@@ -342,12 +362,11 @@ def _check_type(value: Any, expected: type) -> bool:
         try:
             hints = get_type_hints(expected)
         except (NameError, TypeError):
-            # A forward reference that does not resolve from here, or a member
-            # annotated with something that is not a type. The keys have already
-            # been checked above without needing resolution, so report on the
-            # shape rather than raising out of a type check and ending the
-            # prediction, which is the crash this function exists to avoid.
-            return True
+            # One member with an unresolvable forward reference makes get_type_hints
+            # raise for the whole TypedDict. Resolving member by member keeps the
+            # rest checkable, so a sibling with the wrong value type is still
+            # reported instead of the unresolvable member excusing the payload.
+            hints = _resolvable_member_hints(expected)
         return all(_check_type(item, hints[key]) for key, item in value.items() if key in hints)
 
     # Union / Optional (X | None)
--- a/dspy/predict/predict.py
+++ b/dspy/predict/predict.py
@@ -335,12 +335,20 @@ def _check_type(value: Any, expected: type) -> bool:
     if is_typeddict(expected):
         if not isinstance(value, dict):
             return False
-        hints = get_type_hints(expected)
         if not expected.__required_keys__ <= value.keys():
             return False
-        if not value.keys() <= hints.keys():
+        if not value.keys() <= expected.__annotations__.keys():
             return False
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Good catch, and fixed in 67ee3f0. The first version only checked that required keys were present, so `{"subject": 123}` passed while an equivalent `dict[str, str]` annotation would have warned. `_check_type` now resolves the TypedDict's hints, rejects unknown keys, and recurses into each declared member, so a member value that contradicts its annotation produces the same type-mismatch warning as any other field.

Tests extended to cover a wrong-typed member, unknown keys, and NotRequired members in all three states: absent, present and valid, present and wrong. Full `tests/predict/test_predict.py` passes (83).

arthi-arumugam-git:
Thanks, both points were worth raising. Taking them in order.

**Unresolvable TypedDict annotations abort prediction.** Correct, and fixed in 68f1679. `get_type_hints` raises `NameError` on a forward reference that does not resolve from the definition site, and `TypeError` when a member is annotated with something that is not a type, and both were escaping `_check_type` and ending the prediction. That is the exact crash this PR exists to prevent, so it was a fair catch.

The key checks do not need resolved annotations, so they now run first against `__required_keys__` and `__annotations__`, and only the per-member value check goes through `get_type_hints`. If that cannot resolve, the shape result stands rather than the exception escaping. The new test defines a TypedDict whose member is a forward reference to a name that does not exist anywhere; it raises `NameError` on 67ee3f0 and passes on 68f1679. It also asserts that an unknown key is still reported, so the fallback does not quietly become a blanket pass.

**TypedDict value annotations are not validated.** This one was against 67ee3f0's parent. Member values are checked on the last line of that branch:

```python
return all(_check_type(item, hints[key]) for key, item in value.items() if key in hints)
```

`test_typeddict_optional_members` covers it: a `NotRequired[int]` member given the string `"thirty"` is reported as a type mismatch.

Happy to split the fallback into its own helper if that reads better.

arthi-arumugam-git:
Good catch, and a better one than it first looks. The fallback traded a crash for something quieter and worse: one unresolvable member stopped every other member being checked, so a sibling annotated `int` and handed a `str` passed with no warning at all. A silent wrong answer is worse than the exception it replaced.

Fixed in 19f4338 with the second option you suggested, validating resolvable members independently rather than treating the payload as incompatible. Returning `False` for the whole payload would have gone the other way and produced a spurious mismatch warning on every prediction that used such a TypedDict, including correct ones.

`get_type_hints` is all or nothing: a single member naming an unimportable type loses the annotations of every other member with it. So `_resolvable_member_hints` resolves each member on its own against the defining module's namespace and skips only the ones that fail. An unresolvable member now costs its own check and nothing else.

The new test is your scenario:

```python
class Mixed(TypedDict):
    opaque: "StillNoSuchType"   # makes get_type_hints raise for the whole class
    count: int                  # resolves fine and must still be checked
```

Passing `{"opaque": object(), "count": "three"}` is reported as a mismatch. On 68f1679 that assertion fails with an empty log, which is the suppression your harness found. 85 tests pass in `tests/predict/test_predict.py`.

Two things I did not do, in case either is preferred. `_resolvable_member_hints` catches broadly per member rather than naming exception types, because the failure modes vary by Python version and the fallback is already the degraded path. And an unresolvable member is currently silent rather than logged once, since a warning here would fire on every prediction rather than at definition time. Happy to change either.

arthi-arumugam-git:
Confirmed and fixed in 7f2944e. This one mattered: it is the same crash this PR set out to remove, reached through a different door, so the fix was incomplete rather than wrong.

One correction to the report, in your favour. It is not limited to 3.10 and 3.11. On Python 3.13:

```
typing.is_typeddict(te_class)      -> False
typing_extensions.is_typeddict(te_class)      -> True
typing_extensions.is_typeddict(typing_class)  -> True
isinstance({}, te_class)           -> TypeError: TypedDict does not support instance and class checks
```

So any supported version is affected, not just the two.

The fix imports `is_typeddict` from `typing_extensions`, which recognises both flavours, and keeps the import guarded:

```python
try:
    from typing_extensions import is_typeddict
except ImportError:
    from typing import is_typeddict
```

Guarded rather than direct because `typing_extensions` is not declared in `pyproject.toml`. It is guaranteed present in practice, since `pydantic>=2.0` is a hard dependency and pydantic requires `typing-extensions>=4.14.1`, but relying on a transitive dependency without a fallback seemed like the wrong trade for one import. If you would rather declare it explicitly in `pyproject.toml` and import it plainly, say so and I will change it.

The new test builds a `typing_extensions.TypedDict`, asserts the prediction completes, and asserts that a `str` in an `int` member is still reported, so the class is genuinely validated rather than merely surviving. It fails on 19f4338 with the TypeError above. 86 tests pass in `tests/predict/test_predict.py`.

Three rounds of review on this one and all three found something real. Thank you for being thorough with it.

arthi-arumugam-git:
Right again, and this one was mine rather than the change's. Fixed in d821b62.

The tests were only ever run here on 3.13, so the restriction never showed up. `requires-python` is `>=3.10`, which means four of these were failing on two supported versions in CI while passing locally, and I would not have found that without the version sweep.

The mechanism, for the record:

```python
_SUPPORTS_TYPEDDICT = sys.version_info >= (3, 12)
...
if not _SUPPORTS_TYPEDDICT and type(typed_dict_cls).__module__ == "typing":
    raise PydanticUserError("Please use `typing_extensions.TypedDict` ...", code="typed-dict-version")
```

A class built with `typing_extensions.TypedDict` reports a metaclass module of `typing_extensions`, not `typing`, so it cannot trip that guard on any version. The fixtures now come from there.

I took the first of your two suggestions for the general tests and the second for coverage, rather than choosing one. All the behaviour tests build their TypedDicts with `typing_extensions`, and a separate `test_stdlib_typeddict_is_recognised` exercises `typing.TypedDict` behind `skipif(sys.version_info < (3, 12))`. Converting everything to `typing_extensions` would have left `typing.is_typeddict` with no test at all, and a regression on that path would then be invisible on exactly the versions where it is the path that runs.

87 tests pass on 3.13. I could not run 3.11 directly, only 3.12 and 3.13 are installed here, so the claim above rests on the guard's own condition rather than on an observed run, and I would rather say that than imply I verified it on 3.11.

Five rounds, five real findings. Genuinely useful review, thank you.

PR #10219, opened 2026-08-17
fix: raise ValueError instead of UnboundLocalError when no valid program found

Ricardo-M-L:
This PR addresses: raise ValueError instead of UnboundLocalError when no valid program found

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -780,7 +780,7 @@ class MIPROv2(Teleprompter):
                 predictor.demos = demo_candidates[i][demos_idx]
                 trial_logs[trial_num][f"{i}_predictor_demos"] = demos_idx
                 chosen_params.append(f"Predictor {i}: Few-Shot Set {demos_idx}")
-                raw_chosen_params[f"{i}_predictor_demos"] = instruction_idx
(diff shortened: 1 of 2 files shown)


Ricardo-M-L:
Friendly ping - this PR is small and ready for a look whenever someone has bandwidth.

**Two independent one-line bugs, both currently reachable in MIPROv2:**

1. `mipro_optimizer_v2.py:783` - `raw_chosen_params` is written with `instruction_idx` where the loop variable for the demo set is `demos_idx`. The key is `f"{i}_predictor_demos"`, so it records the *instruction* index under a *demo* key. Silent corruption of the selected-params bookkeeping; any downstream reader that trusts `raw_chosen_params` gets the wrong few-shot set index. `demos_idx` is the correct value (it is the value assigned to `trial_logs[trial_num]` on the line above for the same key).

2. `utils.py:141-144` - the fallthrough after the loop returns `program, mean, key, params`, none of which are guaranteed to be bound if every entry in `param_score_dict` was filtered out. That raises `UnboundLocalError`, which is neither caught nor actionable. Replaced with an explicit `ValueError` naming the cause.

Scope: 2 files, +2/-3, no API or signature changes. Both changes are strictly error-path or bookkeeping corrections - successful optimization runs behave identically.

Happy to rebase if main has moved.

PR #10221, opened 2026-08-17
fix(dspy): preserve Field constraints in JSONAdapter structured outputs

### LINKED ISSUE #10195: [Bug] JSONAdapter structured outputs: pydantic Field constraints (ge, le, multiple_of, pattern) are silently dropped in the signature-to-model derivation — is this intentional?
## What happens

Field-level constraint kwargs on `OutputField` never reach any
structured-output backend — the derived response-format model keeps the
type but drops the constraint metadata:

```python
import dspy

class Score(dspy.Signature):
    text: str = dspy.InputField()
    score: float = dspy.OutputField(ge=0.0, le=1.0, multiple_of=0.25)

print(_get_structured_outputs_response_format(Score, True).model_json_schema())
# {'properties': {'score': {'title': 'Score', 'type': 'number'}}, ...}
# ge / le / multipleOf: gone
```

Constraints expressed in the annotation itself survive (`Literal[...]`,
nested models, `set[str]` -> `uniqueItems`); constraints passed as Field
kwargs do not. Users writing `OutputField(ge=0, le=1)` reasonably believe
the bound is part of the contract; outputs violating it parse fine and
flow downstream. dspy 3.2.1, Python 3.12.

## The question

Is the stripping intentional? A plausible reason would be provider
compatibility: OpenAI's strict structured outputs REJECT schemas
containing unsupported keywords like `multipleOf`, so passing constraints
through naively would break those calls. But today's behavior drops them
for every backend equally — including local/vLLM grammar backends that
could enforce some of them, and validation layers that could at least
check them (related territory: #7925 on partial output-field validation).

## If intentional, a suggestion

Strip per-provider rather than globally — or pass constraints through and
let each backend declare what it can't honor. #9686 shows the same
schema-hygiene problem in the opposite direction (vendor extensions
leaking TO strict providers), so a single provider-aware
schema-sanitization point might solve both.

Happy to work on a PR for whichever direction maintainers prefer. We hit
this while integrating a grammar backend and wrote up the full analysis
with workarounds here: https://www.evolutionid.com/en/technical-insights/your-dspy-field-constraints-never-reach-the-model.

tonycoder-hub:
## 📝 Changes Description

`JSONAdapter._get_structured_outputs_response_format` rebuilt each output field as `(annotation, default)`, so `Field(ge=..., le=..., multiple_of=..., pattern=...)` metadata was dropped. Constraints now ride on `Annotated[(annotation, *field.metadata)]`. Constraint kwargs pydantic cannot express for a type (e.g. `ge` on `str`) are stripped so they do not become invalid JSON Schema keys.

Fixes #10195

## ✅ Contributor Checklist

## ⚠️ Warnings

Fine-tuned OpenAI models still have a smaller Structured Outputs subset; this does not add provider-aware sanitization. Nearby #10114 is the *parsing* side (`parse_value`) and does not overlap.

`uv run pytest tests/adapters/test_json_adapter.py -q` → 41 passed. The new test fails on current main (`KeyError: 'minimum'`) and passes after the change.

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -26,6 +26,42 @@ from dspy.utils.exceptions import AdapterParseError, LMError
 logger = logging.getLogger(__name__)
 
 
+# Keywords whose values map user-controlled names to subschemas, so their own keys must never be treated as
+# schema keywords (a field may legitimately be named `ge`).
+_SUBSCHEMA_MAP_KEYWORDS = ("properties", "patternProperties", "$defs", "definitions")
+# Keywords whose values are literal data rather than subschemas.
+_LITERAL_KEYWORDS = ("const", "default", "enum", "examples")
+
+
+def _strip_unsupported_constraint_kwargs(node: Any) -> None:
+    """
+    Recursively drop raw Pydantic constraint kwargs from a JSON schema.
+
+    Pydantic emits a constraint kwarg verbatim when it cannot express it for the annotated type (e.g. `ge` on a
+    string field). Those keys are not valid JSON Schema, and they can appear anywhere the annotation is expanded,
+    including inside `anyOf` branches and `$defs`, so every subschema has to be visited.
+    """
+    if isinstance(node, list):
+        for item in node:
+            _strip_unsupported_constraint_kwargs(item)
+        return
+    if not isinstance(node, dict):
+        return
+
+    for constraint in PYDANTIC_CONSTRAINT_MAP:
+        node.pop(constraint, None)
+
+    for keyword, value in node.items():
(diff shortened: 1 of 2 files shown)


isaacbmiller:
Hi @tonycoder-hub ,

Thank you for the contribution, but this pattern of high-volume AI-generated contributions doesn't align with how we want to develop DSPy.

We value contributors who deeply understand the codebase and engage with the project over time. If you're genuinely interested in contributing to DSPy, we'd welcome you to start by engaging in discussions, triaging issues, or focusing on a single well-understood area. But bulk AI-generated PRs aren't something we can review or merge.

This applies to all your open PRs on this repository.

PR #10231, opened 2026-08-18
fix(utils): don't materialize lazy modules from import-machinery setattr

### LINKED ISSUE #10220: [Bug] importing dspy first and then onnxruntime fails
### What happened?

I imported dspy and onnxruntime like this:
```python
import dspy
import onnxruntime
```
and I got this error:
```
  File "/venv/lib/python3.14/site-packages/numpy/_core/__init__.py", line 11, in <module>
    from numpy.version import version as __version__
  File "<frozen importlib._bootstrap>", line 1371, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1350, in _find_and_load_unlocked
  File "/venv/lib/python3.14/site-packages/dspy/utils/lazy_import.py", line 125, in __setattr__
    setattr(self._load(), attr, value)
            ~~~~~~~~~~^^
  File "/venv/lib/python3.14/site-packages/dspy/utils/lazy_import.py", line 112, in _load
    spec.loader.exec_module(module)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/venv/lib/python3.14/site-packages/numpy/__init__.py", line 120, in <module>
    from ._core import (
    ...<322 lines>...
    )
  File "/venv/lib/python3.14/site-packages/numpy/_core/memmap.py", line 7, in <module>
    from .numeric import dtype, ndarray, uint8
  File "/venv/lib/python3.14/site-packages/numpy/_core/numeric.py", line 13, in <module>
    from . import multiarray, numerictypes, numerictypes as nt, overrides, shape_base, umath
  File "/venv/lib/python3.14/site-packages/numpy/_core/shape_base.py", line 8, in <module>
    from . import fromnumeric as _from_nx, numeric as _nx, overrides
  File "/venv/lib/python3.14/site-packages/numpy/_core/fromnumeric.py", line 11, in <module>
    from . import _methods, multiarray as mu, numerictypes as nt, overrides, umath as um
  File "/venv/lib/python3.14/site-packages/numpy/_core/_methods.py", line 17, in <module>
    bool_dt = mu.dtype("bool")
TypeError: data type 'bool' not understood

[lines from my own code redacted]
  File "/my_own_code/__init__.py", line 11, in <module>
    import onnxruntime as ort
  File "/venv/lib/python3.14/site-packages/onnxruntime/__init__.py", line 78, in <module>
    raise import_capi_exception
  File "/venv/lib/python3.14/site-packages/onnxruntime/__init__.py", line 26, in <module>
    from onnxruntime.capi._pybind_state import (
    ...<42 lines>...
    )
  File "/venv/lib/python3.14/site-packages/onnxruntime/capi/_pybind_state.py", line 32, in <module>
    from .onnxruntime_pybind11_state import *  # noqa
ImportError: import numpy failed
```

But when I switched the order of imports, it worked fine:
```python
import onnxruntime
import dspy
```

 While `import onnxruntime` triggers the error, I think it's DSPy's `lazy_import.py` module that causes it.

### Steps to reproduce

```sh
pip install dspy onnxruntime
python -c "import dspy
import onnxruntime"
```

### DSPy version

3.3.0

yzxcj797:
## Summary

Fixes #10220.

`_LazyModule.__setattr__` treated every assignment as a configuration write and materialized the real module. But the import system also uses `setattr(parent, child, submodule)` to bind submodules onto their parent package — and answering *that* binding with a full load means executing the parent from inside the import machinery while one of its submodules is still mid-initialization.

## Root cause

Deterministic repro on plain CPython (no onnxruntime needed):

```python
import sys
sys.path.insert(0, "<repo>")                     # standalone dspy/utils/lazy_import.py
from dspy.utils.lazy_import import require

np = require("numpy")        # proxy now holds sys.modules["numpy"]
import numpy._core           # what the onnxruntime C extension effectively does
```

1. `import numpy._core` starts executing `numpy/_core/__init__.py` (the import system sees `sys.modules["numpy"]` occupied by the proxy and skips importing the parent).
2. `_core/__init__.py:11` runs `from numpy.version import version as __version__`. The import machinery loads `numpy.version`, then binds it onto the parent with `setattr(sys.modules["numpy"], "version", <module>)` — which lands on the **proxy**.
3. The old `__setattr__` calls `_load()`, so `numpy/__init__.py` executes *nested inside the submodule's import*, while `numpy._core` is half-initialized.
4. `numpy/__init__.py`'s own `from ._core import (...)` finds the mid-init `_core` module, mixing module generations inside `_core` — ending in `TypeError: data type 'bool' not understood`, exactly as reported.

## Fix

Module-valued assignment on the proxy no longer triggers a load:

- the binding lands on the proxy itself (import-system submodule binding), and
- `_load()` replays those bindings onto the real module once it materializes, for names the real module does not already provide.

This mirrors how `importlib.util.LazyLoader` behaves for the same edge case (a stdlib `LazyLoader`-wrapped numpy fails `import numpy._core` identically *without* the re-entrancy corruption). Plain (non-module) configuration writes still materialize the real module immediately — `mod.value = 2` behavior is unchanged and covered by an existing test.

## Tests

Numpy-free regression tests added to `tests/utils/test_lazy_import.py`:

- `test_require_submodule_import_does_not_reenter_parent_init` — imports a submodule while the lazy parent proxy holds `sys.modules`; on the old code this raises `ImportError: cannot import name 'marker' from partially initialized module` (the synthetic equivalent of the reported crash), and it additionally asserts the deferred bindings are replayed consistently after materialization.
- `test_require_module_valued_assignment_does_not_materialize` — module-valued assignment defers, non-module writes still apply to the materialized module.

Verified locally: new tests fail on `main` and pass with this change; all 12 pre-existing tests in the file remain green.

### DIFF
--- a/dspy/utils/lazy_import.py
+++ b/dspy/utils/lazy_import.py
@@ -145,12 +145,14 @@ class _LazyModule(types.ModuleType):
             # submodule's init and corrupting it. Bind locally instead; _load()
             # replays these bindings onto the real module once it materializes.
             super().__setattr__(attr, value)
-            # A caller can retain the proxy past first use; once the real module has
-            # materialized, sys.modules no longer routes through this object, so a
-            # module-valued assignment must also be forwarded to the canonical module
-            # or the two views of the package diverge.
+            # A caller can retain the proxy past first use. Once the real module has
+            # materialized, sys.modules no longer routes through this object, so the
+            # import system binds submodules onto the real module directly — any
+            # module-valued assignment reaching the retained proxy from here on can
+            # only be caller code and must replace the canonical module's binding,
+            # exactly like a non-module configuration write.
             loaded = sys.modules.get(self.__name__)
-            if loaded is not None and loaded is not self and not hasattr(loaded, attr):
+            if loaded is not None and loaded is not self:
                 setattr(loaded, attr, value)
         else:
             setattr(self._load(), attr, value)
--- a/dspy/utils/lazy_import.py
+++ b/dspy/utils/lazy_import.py
@@ -145,6 +145,13 @@ class _LazyModule(types.ModuleType):
             # submodule's init and corrupting it. Bind locally instead; _load()
             # replays these bindings onto the real module once it materializes.
             super().__setattr__(attr, value)
+            # A caller can retain the proxy past first use; once the real module has
+            # materialized, sys.modules no longer routes through this object, so a
+            # module-valued assignment must also be forwarded to the canonical module
+            # or the two views of the package diverge.
+            loaded = sys.modules.get(self.__name__)
+            if loaded is not None and loaded is not self and not hasattr(loaded, attr):
+                setattr(loaded, attr, value)
         else:
             setattr(self._load(), attr, value)
 
--- a/dspy/utils/lazy_import.py
+++ b/dspy/utils/lazy_import.py
@@ -83,6 +83,9 @@ class _LazyModule(types.ModuleType):
     """Module proxy that imports the real module on first attribute access.
 
     Attribute assignment also materializes the real module so configuration writes apply to the real dependency.
+    Assigning a *module* value is treated as the import system binding a submodule onto
+    this (parent) proxy and never triggers a load; the binding is replayed onto the real
+    module once it materializes.
     """
 
     def __init__(self, module: str, spec: importlib.machinery.ModuleSpec, lock: threading.RLock):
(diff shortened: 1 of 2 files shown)


yzxcj797:
Valid finding — fixed in 96db3d6.

`__setattr__` now forwards module-valued assignments to the canonical module in `sys.modules` when it has already materialized and doesn't carry the attribute (same don't-clobber rule as the materialization-time replay), so a retained proxy can no longer diverge from the real module. Covered by `test_require_module_valued_assignment_after_materialization_forwards`; all 16 tests in the file pass locally.

yzxcj797:
Also valid — fixed in the follow-up commit. Post-materialization, the import system binds submodules onto the canonical module in `sys.modules` directly, so any module-valued assignment still reaching a retained proxy is caller code by construction; it now forwards unconditionally (replacing, like non-module config writes) instead of being dropped by the `not hasattr` guard. The regression test covers both the fresh-binding and replacement cases; 16/16 green locally.

PR #10234, opened 2026-08-18
Add trusted in-process LocalInterpreter

isaacbmiller:
## Summary

- port the trusted in-process `LocalInterpreter` from `dspy-interpreters` into DSPy
- expose it as `dspy.LocalInterpreter` with persistent state, host tools, typed `SUBMIT`, recoverable execution errors, terminal shutdown, and execution-environment metadata
- verify it through a deterministic real `dspy.RLM` flow and the external interpreter conformance suites, including Flex
- document its full host authority and correct Flex's categorical sandbox wording for custom interpreter factories

This intentionally does **not** port the abandoned invocation-scoped binding extension or the speculative subprocess mode. `LocalInterpreter` is opt-in; `PythonInterpreter` remains the sandboxed default for RLM and Flex.

## Usage with Flex

```python
import dspy

lm = dspy.LM("openai/gpt-5-mini", temperature=1.0, max_tokens=16000)
dspy.configure(lm=lm)

solve = dspy.Flex(
    "question: str -> answer: int",
    interpreter_factory=dspy.LocalInterpreter,
)
result = solve(question="What is 19 + 23? Return only the integer answer.")

print(result.answer)  # 42
```

This exact example was run against the real `openai/gpt-5-mini` provider. It returned integer `42`, and `lm.history` recorded exactly one real LM call through Flex's host-side predictor bridge.

## Security model

`LocalInterpreter` is not a sandbox. Executed code has the DSPy process's filesystem, environment, credential, network, package, and process authority. The API docs and RLM/Flex docs warn users explicitly. Executions are serialized across LocalInterpreter instances because stdout redirection is process-global.

## Verification

- real-LM Flex example above — `answer: 42`, `answer_type: int`, `real_lm_calls: 1`
- `uv run pytest -q -o addopts='' tests/primitives/test_local_interpreter.py tests/predict/test_rlm.py -m 'not deno'` — 116 passed, 2 skipped, 37 deselected
- external `dspy-interpreters` checks — core interpreter, execution instructions, real RLM, RLM prompt integration, and Flex facade all passed
- focused Ruff check and `git diff --check` — passed
- docs build — completed successfully
- `uv build` — wheel and sdist built; wheel contains `dspy/primitives/local_interpreter.py`

### DIFF
--- a/docs/docs/api/modules/Flex.md
+++ b/docs/docs/api/modules/Flex.md
@@ -27,7 +27,7 @@ print(result.total_cents)
 
 Out of the box, `solve` is just a `dspy.Predict` over the signature, wrapped in a module (with `tools`, it starts as a `dspy.RLM` instead — see [Tools](#tools)). The point of `Flex` is what happens when you optimize it (see [Optimizing with GEPA](#optimizing-with-gepa)): GEPA can replace that baseline with, say, a predictor that only extracts quantities and unit prices, and a line of Python that multiplies and sums them.
 
-The generated code always runs in a sandbox (`interpreter_factory` defaults to `dspy.PythonInterpreter`), so the example above needs [Deno](https://deno.land/) installed — see [Sandboxed Execution](#sandboxed-execution).
+By default, generated code runs in a sandbox (`interpreter_factory` defaults to `dspy.PythonInterpreter`), so the example above needs [Deno](https://deno.land/) installed — see [Sandboxed Execution](#sandboxed-execution).
 
 ## How Optimization Works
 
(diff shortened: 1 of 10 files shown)


MaximeRivest:
The name are a bit confusing, Python vs Local is probably not descriptive enough of what these interpreter 'entails'.

isaacbmiller:
What about InProcessInterpreter @MaximeRivest ?

MaximeRivest:
Many users do not think in “host”, “sandbox”, or “WASM”. 

What they would ask is more: “does this use **my** Python?”

| What they mean | What it does | Risk |
|---|---|---|
| My Python | Same process, same packages, same files | Model code can do what you can do |
| Safe Python | Separate box, limited packages | Safer; some imports fail |

I don't think it is helpful to put Deno, WASM, or “local” in the name.

I don't understand everything about interpreter in dspy, but this would feel more obvious and natural to me:

- `dspy.MyPython` or `dspy.SkipIsolation`
- `dspy.SafePython` (`PythonInterpreter` is confusing if its out of experimental we could keep the old name as an alias)

if you want "Interpreter" in the name:

- `dspy.CurrentPythonInterpreter` or `dspy.MyPythonInterpreter` or `dspy.NoIsolation` or `dspy.NoInterpreter`
- `dspy.SafePythonInterpreter`

**Avoid**

- `LocalInterpreter` — the safe one is also on their machine
- `HostInterpreter` — “host” is computer talk
- `InProcessInterpreter` — same problem
- `TrustedInterpreter` — they will hear “safe”, which is the opposite

## How they should write it

Do not start with `interpreter_factory=...`. Use a simple choice:

```python
solve = dspy.Flex("question -> answer", python="safe")   # default
solve = dspy.Flex("question -> answer", python="skipisolation")
```

Or:

```python
dspy.Flex(..., use="safe_python")
dspy.Flex(..., use="my_python")
```

`python="mine"` is the honest warning. If they pick it, they chose full access.

I think we should keep `interpreter_factory` for people who build a remote runner. Hide it in advanced docs.

isaacbmiller:
superseded by #10238

LocalInterpreter still feels more right to me than the other options. A name still needs to be descriptive of what it does, so needs Interpreter in the name.

PythonInterpreter definitely should have been called something else, and considering aliasing to DenoInterpreter.

PR #10247, opened 2026-08-20
fix(gepa): model the current reflective loop in auto_budget

### LINKED ISSUE #10245: [Bug] GEPA auto_budget does not reflect the current reflective evaluation loop
### What happened?

`GEPA.auto_budget()` calculates the metric-call budget using assumptions that do not match the current GEPA optimization loop.

The calculation currently:

- uses a fixed minibatch size of 35, regardless of the configured `reflection_minibatch_size`
- adds `num_candidates * 5` for candidate bootstrapping
- assumes periodic full evaluations every five steps

However, the current GEPA engine does not bootstrap candidate sets or perform full validation periodically every five steps.

A normal reflective proposal currently:

1. evaluates the parent candidate on `reflection_minibatch_size` training examples
2. generates a child candidate through reflection
3. evaluates the child on the same minibatch
4. runs full validation only if the child passes the minibatch acceptance criterion

Therefore, each normal reflective proposal consumes approximately:

```text
2 * reflection_minibatch_size
```

metric calls before any full validation.

Because `auto_budget()` instead uses the fixed value 35, changing `reflection_minibatch_size` does not change the auto budget, even though it directly changes the evaluation cost of every proposal.

The bootstrapping and periodic-full-evaluation terms also appear to be stale assumptions that do not correspond to the current GEPA execution flow.

I expect the auto budget to account for:

- the initial seed-candidate validation
- both parent and child minibatch evaluations
- full validation of accepted candidates
- merge screening and validation when merge is enabled

The result would still be a budget allocation rather than an exact prediction, because rejected proposals skip full validation and custom GEPA policies may have different costs.

### Steps to reproduce

The configured reflection minibatch size does not affect the calculated auto budget.

```python
import dspy


def metric(gold, pred, trace=None, pred_name=None, pred_trace=None):
    return 0.0


gepa_with_batch_3 = dspy.GEPA(
    metric=metric,
    reflection_lm=lambda _: [],
    auto="light",
    reflection_minibatch_size=3,
    use_merge=False,
)

gepa_with_batch_10 = dspy.GEPA(
    metric=metric,
    reflection_lm=lambda _: [],
    auto="light",
    reflection_minibatch_size=10,
    use_merge=False,
)

budget_with_batch_3 = gepa_with_batch_3.auto_budget(
    num_preds=2,
    num_candidates=6,
    valset_size=100,
)

budget_with_batch_10 = gepa_with_batch_10.auto_budget(
    num_preds=2,
    num_candidates=6,
    valset_size=100,
)

print(budget_with_batch_3)
print(budget_with_batch_10)
```

Actual output:

```text
1330
1330
```

Both budgets are identical because `auto_budget()` uses its internal default minibatch size of 35 instead of the optimizer's configured `reflection_minibatch_size`.

Expected behavior:

The budgets should differ because each reflective proposal evaluates both the parent and child on the configured minibatch. Increasing `reflection_minibatch_size` from 3 to 10 therefore increases the number of metric calls consumed per proposal.

### DSPy version

main at 822f3931 (project version 3.3.0)

yzxcj797:
Fixes #10245.

## Root cause (as reported, verified in source)

`auto_budget()` priced the run against a loop that no longer exists:

- **fixed `minibatch_size=35`** — `reflection_minibatch_size` (default 3) never reached the calculation, so batch 3 and batch 10 computed identical budgets even though the knob scales every proposal's cost (each proposal evaluates the parent AND the child on the minibatch: `2 * reflection_minibatch_size` calls);
- **`num_candidates * 5` bootstrapping** — the current engine bootstraps no candidate set;
- **periodic full evals every `full_eval_steps`** — there is no fixed full-evaluation period; full validation runs only when a child passes the minibatch acceptance criterion.

`compile()` called it without passing either knob, so both were structurally invisible.

## The new cost model

```
total = V                          # seed validation (initial full eval of the default program)
      + N * 2 * M                  # N proposals, parent + child minibatch each
      + N * V                      # full validation of ACCEPTED children (worst case)
      + (num_candidates - 1) * (2*M + V)   # when use_merge: merges bounded by the
                                           # candidate budget (each consumes two),
                                           # 2*M screening + V validation per merge
```

The accepted-children term is an upper bound — rejected proposals skip the full pass and leave budget for more trials — which matches the issue's own framing ("a budget allocation rather than an exact prediction"). `compile()` now passes `minibatch_size=self.reflection_minibatch_size` and `use_merge=self.use_merge` through, so the budget tracks the configuration. The legacy `minibatch_size=35` parameter default and positional signature are retained for call compatibility.

## Tests

- `test_auto_budget_tracks_reflection_minibatch_size` — the issue's exact repro shape: batch 10 costs strictly more than batch 3. **Fails on `main`** (identical budgets).
- `test_compile_passes_the_configured_minibatch_size_to_auto_budget` — a spy on `auto_budget` under a stubbed `gepa.optimize` proves `compile` feeds the configured `reflection_minibatch_size=7` and `use_merge=False` through. **Fails on `main`** (neither is passed; 35/True defaults).
- `test_auto_budget_models_the_current_loop` — the arithmetic is exactly the documented model, including the merge delta.

Full `tests/teleprompt/test_gepa.py`: 24/24 with the fix; the three new tests fail on `main` (3 failed / 21 passed there).

### DIFF
--- a/dspy/teleprompt/gepa/gepa.py
+++ b/dspy/teleprompt/gepa/gepa.py
@@ -470,8 +470,42 @@ class GEPA(Teleprompter):
             )
 
     def auto_budget(
-        self, num_preds, num_candidates, valset_size: int, minibatch_size: int = 35, full_eval_steps: int = 5
+        self,
+        num_preds,
+        num_candidates,
+        valset_size: int,
+        minibatch_size: int = 35,
+        full_eval_steps: int = 5,
+        *,
+        use_merge: bool = True,
     ) -> int:
+        """Estimate the metric-call budget for an auto run.
+
+        Models the current reflective loop: each proposal evaluates the
+        parent and the child on ``minibatch_size`` training examples
+        (``2 * minibatch_size`` calls), and a child that passes the
+        minibatch acceptance criterion pays a full validation
+        (``valset_size`` calls) — rejected proposals skip it, so the
+        full-validation term is an upper-bound allocation, not an exact
+        prediction. The seed candidate's initial validation, and merge
+        screening/validation when merge is enabled, are included
+        (#10245).
+
+        :param num_preds: Number of optimizable components (instruction
+            predictors + flex submodules), e.g. ``2``.
+        :param num_candidates: Candidate budget from the auto preset,
+            e.g. ``6`` for ``auto="light"``.
+        :param valset_size: Number of validation examples.
+        :param minibatch_size: Reflective minibatch size — the knob that
+            scales every proposal's cost. Defaults to 35 for callers that
+            model the legacy cadence; DSPy's own call passes its
+            configured ``reflection_minibatch_size`` so the budget tracks
+            the configuration.
+        :param full_eval_steps: Retained for call compatibility; the
+            current engine has no fixed full-evaluation period.
+        :param use_merge: Include merge screening/validation terms.
+        :returns: The estimated metric-call budget.
+        """
         num_trials = int(max(2 * (num_preds * 2) * math.log2(num_candidates), 1.5 * num_candidates))
         if num_trials < 0 or valset_size < 0 or minibatch_size < 0:
             raise ValueError("num_trials, valset_size, and minibatch_size must be >= 0.")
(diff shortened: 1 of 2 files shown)


yzxcj797:
Addressed in f0af05f — the merge term now takes `min(num_candidates - 1, max_merge_invocations) * V`, with the cap forwarded from `gepa_kwargs` (default 5).

Modeling detail on the screen: per the installed GEPA runtime (`proposer/merge.py`), each invocation evaluates the merged candidate on the fixed `num_subsample_ids=5` subsample via `cached_evaluate_full`, and the accepted merge's full validation reuses those cached evaluations — so one invocation costs at most one full validation, not `2*M + V`. Rejected merges cost only the 5-call screen, making `V` the worst-case allocation consistent with the accepted-children term.

Reproduces the harness numbers exactly: medium (n=12, num_preds=2, V=100) 4,234 → 3,568 and heavy (n=18) 5,400 → 4,098. Regression: `test_auto_budget_caps_merge_work_at_the_engine_limit`; all 25 gepa tests green.

PR #10251, opened 2026-08-21
fix(utils): materialize a lazy package before its submodules import

### LINKED ISSUE #10220: [Bug] importing dspy first and then onnxruntime fails

yzxcj797:
## Summary

`import dspy` then `import onnxruntime` fails inside numpy with `TypeError: data type 'bool' not understood` (wrapped as onnxruntime's `ImportError: import numpy failed`); the reverse order works (#10220).

## Root cause

`require("numpy")` parks a `_LazyModule` proxy in `sys.modules["numpy"]`. onnxruntime's C extension imports `numpy._core` **directly**, and the import machinery resolves the parent by the `sys.modules` entry — the proxy — so it proceeds to execute the child while the real `numpy/__init__.py` never ran. Two defects compound from there:

1. **Nested double execution.** When the machinery later binds a freshly imported submodule onto its parent (`sys.modules["numpy"].version = <module>`), that `setattr` lands on the proxy, whose `__setattr__` routed through `_load()` — executing a *full* `numpy/__init__.py` nested inside the in-flight `numpy._core` import. The nested `__init__` re-imports the half-initialized `numpy._core` from `sys.modules`, and the double execution corrupts package state — the observed `dtype("bool")` failure. Confirmed by instrumenting the proxy: the crash stack contains exactly one `__setattr__` on the proxy, from `numpy/_core/__init__.py:11`'s `from numpy.version import ...`.
2. **Skipped `__init__` side effects.** Even without the crash, `numpy/__init__.py` registers its Windows DLL directory (line 91) before importing `._core`; with the proxy as parent, that registration never happens and `_multiarray_umath` fails to load its DLLs.

## Fix

- The proxy's `__path__` is now a read-through property that materializes the real package. That is exactly the point where the import machinery consults the parent (`_find_and_load_unlocked` reads `parent_module.__path__` before any child code runs), so the package `__init__` executes first — identical ordering to a world without the proxy. Both import orders now work.
- Submodule bindings recorded on the proxy (`setattr(proxy, "version", <module>)` where the value's `__name__` is `<package>.<attr>`) are stored on the proxy without triggering a nested load — a nested load there is never correct, because the machinery is mid-flight on a child of the same package.
- Plain attribute assignment (configuration writes) keeps the existing materialize-on-assign behaviour, covered by an unchanged pre-existing test.

## Testing

- Reproduced on Python 3.12 (numpy 2.5.2 + onnxruntime 1.29.0) before the fix; both import orders work after it.
- 3 new regression tests (no onnxruntime needed — the numpy shape is simulated with tmp_path packages): direct-submodule import runs the package `__init__` first; `__path__` access materializes; submodule bindings record without materializing while plain assignment still does.
- Differential: the 3 new tests fail on unpatched main; the full `tests/utils/` suite is otherwise unchanged (2 pre-existing `test_settings` failures reproduce with the patch stashed).

Fixes #10220

### DIFF
--- a/dspy/utils/lazy_import.py
+++ b/dspy/utils/lazy_import.py
@@ -57,6 +57,60 @@ def _get_lazy_module_lock(module: str) -> threading.RLock:
         return _lazy_module_locks.setdefault(module, threading.RLock())
 
 
+class _ReuseExecutedLoader:
+    """Loader that hands back an already-executed module unchanged.
+
+    The import machinery does not re-check ``sys.modules`` once a direct
+    child import has committed to a spec, so a child that completed during
+    parent materialization (the parent ``__init__`` imports it) would be
+    re-executed by the original pending import. This loader makes that
+    pending import resolve to the completed module object instead, and
+    restores its original ``__spec__``/``__loader__`` afterwards.
+    """
+
+    def __init__(self, module: types.ModuleType):
+        self._module = module
+        self._original_spec = module.__spec__
+        self._original_loader = module.__loader__
+
+    def create_module(self, spec: importlib.machinery.ModuleSpec) -> types.ModuleType:
+        return self._module
+
+    def exec_module(self, module: types.ModuleType) -> None:
+        module.__spec__ = self._original_spec
+        module.__loader__ = self._original_loader
+
+
+class _MaterializedChildFinder:
+    """Meta-path finder deduplicating children imported during materialization.
+
+    Returns a reuse spec only for a module that is already in
+    ``sys.modules``, was imported while a lazy parent was materializing,
+    and is therefore the subject of an outer pending import; every other
+    lookup falls through to the normal finders.
+    """
+
+    _MARKER = "_dspy_lazy_reuse"
+
+    def find_spec(self, fullname: str, path=None, target=None):  # noqa: ANN001, ANN202
+        module = sys.modules.get(fullname)
+        if module is None or getattr(module, _MaterializedChildFinder._MARKER, False) is not True:
+            return None
+        return importlib.util.spec_from_loader(
+            fullname,
+            _ReuseExecutedLoader(module),
+            is_package=hasattr(module, "__path__"),
+        )
+
+
+sys.meta_path.insert(0, _MaterializedChildFinder())
+
+# True while a lazy parent's real module executes: modules imported in that
+# window may be the subject of the outer pending import that triggered
+# materialization, so they are marked for reuse by _load().
+_MATERIALIZING = False
+
+
 def _is_submodule_binding(package: str, attr: str, value: Any) -> bool:
     """Return True when *value* is the import machinery binding a submodule.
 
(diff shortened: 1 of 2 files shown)


yzxcj797:
Addressed in 01e609b — the double execution is structural (once the outer child import passes its `sys.modules` check, nothing re-checks it after `__path__` materialization), so the fix coordinates rather than reorders:

- `_load()` snapshots `sys.modules` before executing the package `__init__` and marks every module the initializer imported;
- a first-position meta-path finder resolves a *pending* import of a marked, completed module to that module itself — a tiny reuse loader whose `create_module` returns the existing object and whose `exec_module` restores its original `__spec__`/`__loader__`;
- every other lookup falls through to the normal finders, so the finder only short-circuits an import that would duplicate an execution that already happened.

Reproduced your probe shape first: unpatched, the events are `['package-init', 'child-exec', 'version-exec', 'child-exec']`. The new exact-once regression (`test_direct_submodule_import_executes_the_child_exactly_once`) asserts one `child-exec`, one `version-exec`, init-before-child ordering, and that the reused module keeps a real spec/loader — it fails on the previous commit and passes now. The onnxruntime repro stays green in both import orders, and `import dspy` still leaves the numpy proxy unmaterialized in `sys.modules`.

yzxcj797:
Addressed in 87d33e6 — both halves of your suggestion: the finder now defers whenever a `target` module is supplied (the reload shape, so reload takes the normal finders and re-executes the body), and the marker is consumed by the one pending import it serves, so any later reload or re-import also goes through the normal path. New regression: `test_reload_reexecutes_a_reused_child` drives the direct-child reuse first, asserts one `child-exec`, reloads, and asserts a second `child-exec`. The onnxruntime repro stays green and the exact-once test now also asserts the marker is gone after reuse. 18/18 in the suite.

yzxcj797:
Addressed in b89636d — adopted your prescription verbatim: the marking pass now iterates `list(sys.modules.items())` instead of the live dict, so a concurrent import or removal in another thread can no longer raise dictionary-changed-size-during-iteration after the package has already initialized. Suite stays 18/18 and the onnxruntime repro is green in both import orders. Your coordinated two-thread artifact is the exact shape the snapshot neutralizes — the marking pass now operates on data captured under one GIL hold.

PR #10266, opened 2026-08-24
feat(RLM/Flex): Implement sub-dspy interpreters

michaelisaac-dev:
`dspy.RLM` can now build dspy sub-agents (`Predict`, `ChainOfThought`, `ReActV2`, nested `RLM`) from its REPL, on interpreters that can host the sandbox dspy facade.

- **`dspy.InterpreterCapability`** (`enum.Flag`), declared as a `capabilities` attribute on the interpreter class or factory. `SUB_DSPY` means the interpreter can host the facade: host tools are callable as globals and state persists across `execute()`. `PythonInterpreter` declares it, and a `functools.partial` of it inherits the declaration. A custom interpreter that does not declare it gets no sub-agents and no sub-agent section in the prompt.
- **Facade shared with Flex**: the shim moved from `flex/_sandbox_shim.py` to `dspy/primitives/facade.py` + `_facade_shim.py`. Predictors are built and run on the host; only JSON crosses. The facade refuses an interpreter that runs code in the host's memory (the host process or a fork of it). The nested-RLM idiom in the prompt (`interpreter_factory=dspy_interpreter_factory`) is a marker the host swaps for its own factory.
- **`SandboxLM`**: every LM call made for sandboxed code (RLM sub-agents, Flex bridged predictors) runs on this proxy. It admits only generation parameters per call (`temperature`, `max_tokens`, `response_format`, ...; never `api_base`, `api_key`, headers, or `model`), mirrors only those defaults, and meters RLM's `max_llm_calls`, which now covers `llm_query`, `llm_query_batched`, and all sub-agent LM calls. Flex keeps `max_predictor_calls`.
- **Tests**: mock and Deno RLM tests, `SandboxLM` unit tests, and conformance against dspy-interpreters (test-only `dev` dependency group): undeclared backends get no sub-agents, a declaring subclass runs bridged sub-agents with host tools, and the library's `check_interpreter`, `check_rlm`, and `check_flex_facade` pass.


dineshyadav03 (commented):
Went through the diff on `rlm.py`, `code_interpreter.py`, and both test files. Solid piece of work — the `InterpreterCapability` flag is a clean way to keep this fully opt-in, and I like that `sub_lm` state is scoped per-block via `dspy.context(...)` rather than a persistent `dspy.configure()`, with `_host_settings_guard` restoring `main_thread_config` afterward. The API-key exclusion from `dump_state()` is also verified in tests.

Two things I couldn't find coverage for, and would want addressed before merge:

1. Recursion/cost bounds — a sandbox with `SUB_DSPY` can build `dspy.RLM(..., interpreter_factory=dspy_interpreter_factory)` recursively. I don't see any depth counter or LM-call budget threaded through that factory, so nothing structurally prevents an unbounded fan-out of nested interpreters/LM calls beyond `max_iters` per level. Is that intentionally left to the interpreter backend?
2. `interpreter_capabilities()` is purely declarative — nothing validates that a `SUB_DSPY`-declaring interpreter actually exposes `dspy_interpreter_factory` in-namespace; a misconfigured backend fails with a bare `NameError` at exec time rather than at construction. Worth a validation hook or at least a clearer error path?

Neither is a correctness bug in what's here. Marking as comment rather than blocking since the capability is opt-in, but I'd like the recursion-bound question answered before this ships as a pattern others build backends against.

Reviewed with AI assistance (Claude); verified against the current diff.

dspy/predict/rlm.py
```diff
@@ -75,9 +79,41 @@
...
+You may `import dspy` and build sub-agents in the REPL for subtasks that need structured inputs/outputs.
+- `dspy.Predict("question -> answer")(question=...)` or `dspy.ChainOfThought(...)` - single-step sub-agents.
+- `dspy.ReActV2("question -> answer", tools=[...])(question=...)` - a multi-step tool-using sub-agent.
+  Only the provided tools listed above may be passed; functions you define in the REPL cannot cross to the host.
+- `dspy.RLM("context, query -> answer", interpreter_factory={SUB_DSPY_FACTORY_NAME})(context=..., query=...)` -
+  a recursive sub-agent with its own REPL; always pass the provided `{SUB_DSPY_FACTORY_NAME}`.
+  This is the heaviest option: reserve it for deep subtasks whose input is itself too large or
+  structured to prompt directly, and prefer Predict/ChainOfThought/ReActV2 for everything else.
+Prefer `llm_query` for simple one-shot prompts; use sub-agents for structured, tool-using, or
+recursive subtasks. Never call `dspy.configure(...)`: the LM configuration is already provided.
```
isaacbmiller: does dspy.configure exist in the facade?

dspy/predict/rlm.py
```diff
@@ -75,9 +79,41 @@
...
 You have max {max_llm_calls} sub-LLM calls. When done, call SUBMIT() with your output."""
 
+# Appended to the interpreter rules when the interpreter can host the sandbox dspy facade.
+SUB_AGENT_INSTRUCTIONS = f"""
+Sub-agents (dspy):
+You may `import dspy` and build sub-agents in the REPL for subtasks that need structured inputs/outputs.
+- `dspy.Predict("question -> answer")(question=...)` or `dspy.ChainOfThought(...)` - single-step sub-agents.
+- `dspy.ReActV2("question -> answer", tools=[...])(question=...)` - a multi-step tool-using sub-agent.
+  Only the provided tools listed above may be passed; functions you define in the REPL cannot cross to the host.
+- `dspy.RLM("context, query -> answer", interpreter_factory={SUB_DSPY_FACTORY_NAME})(context=..., query=...)` -
```
isaacbmiller: Should RLM in the facade be allowed to customize the factory? On one hand we don't want there to be a big difference between real DSPy and the DSPy that a model writes. OTOH, needing to specify the interpreter_factory is a distraction for the model.

dspy/predict/rlm.py
```diff
@@ -75,9 +79,41 @@
...
+        self._limit = limit
+        self._count = 0
+        self._lock = threading.Lock()
+
+    def reserve(self, n: int = 1) -> None:
+        with self._lock:
+            if self._count + n > self._limit:
+                raise RuntimeError(
+                    f"LLM call limit exceeded: {self._count} + {n} > {self._limit}. "
+                    f"Use Python code for aggregation instead of making more LLM calls."
```
isaacbmiller: Why aggregation here?

PR #10270, opened 2026-08-24
fix(adapters): preserve XMLAdapter round-trip fidelity for `]]>` and empty optional containers

### LINKED ISSUE #10269: [Bug] XMLAdapter nested-XML rewrite: strings containing `]]>` crash parsing, and empty Optional containers silently become None
### What happened?

Three round-trip fidelity failures in the nested-XML support added by #10239 (current `main`, `4ed377e`). `XMLAdapter.format()` produces output that `XMLAdapter.parse()` either rejects outright or decodes to the wrong value.

All three repros are self-contained and were run against `main` (`4ed377e`) with the released-package-free repo checkout (`uv run python repro.py`, Python 3.12, macOS).

#### 1. `]]>` in a string value makes `parse()` fail (regression vs. the pre-#10239 regex parser)

```python
import dspy
from dspy.adapters.xml_adapter import XMLAdapter
from dspy.adapters.chat_adapter import FieldInfoWithName

class Sig(dspy.Signature):
    text: str = dspy.OutputField()

adapter = XMLAdapter()
value = "array ]]> trap"
formatted = adapter.format_field_with_value(
    {FieldInfoWithName(name="text", info=Sig.output_fields["text"]): value}
)
print(formatted)
parsed = adapter.parse(Sig, formatted)
```

`format_field_with_value` escapes only `&` and `<` (`xml_adapter.py:34` and `xml_adapter.py:153`), so the emitted text contains the literal sequence `]]>`. That sequence is illegal in XML character data (XML 1.0 §2.5: "the string ]]> must not occur in content"), so the adapter's own `ET.fromstring` call in `parse()` rejects it:

```
<text>
array ]]> trap
</text>
dspy.utils.exceptions.AdapterParseError: Failed to parse XML: not well-formed (invalid token): line 4, column 8
```

Any LM completion containing `]]>` — realistic for code-generation tasks (`arr[i]]>0`, nested indexing) or text about arrays — now fails to parse. The pre-rewrite `field_pattern` regex accepted arbitrary content, so this is a regression introduced by #10239.

#### 2. Empty list under an Optional type silently becomes `None`

```python
from typing import Optional
import dspy
from dspy.adapters.xml_adapter import XMLAdapter
from dspy.adapters.chat_adapter import FieldInfoWithName

class Sig(dspy.Signature):
    items: Optional[list[str]] = dspy.OutputField()

adapter = XMLAdapter()
field = FieldInfoWithName(name="items", info=Sig.output_fields["items"])
formatted = adapter.format_field_with_value({field: []})
print(formatted)                      # <items />
print(adapter.parse(Sig, formatted))  # {'items': None}   <- expected {'items': []}
```

Same for `Optional[dict[str, int]]`: `{}` formats to `<mapping></mapping>` but parses back as `None`. Non-empty values round-trip correctly; only the empty ones are lost. The existing test at `tests/adapters/test_xml_adapter.py:109` covers plain `list[str]` (which works), but no test covers `Optional[...]`.

Root cause: in `_elements_to_value` (`xml_adapter.py:187-193`), when the schema has an `anyOf` with a null choice, an element with no children and no text short-circuits to `None` before the array/object branches can decode it to an empty container.

Note the format side never encodes `None` as an empty element — `None` goes through the base formatter and is emitted as literal `None` text (which `parse_value` coerces back via `literal_eval`). So on the parse side, an empty element can unambiguously mean "empty container" rather than `None`.

#### 3. Leading/trailing whitespace in string outputs is stripped

`"  padded  "` round-trips to `"padded"` because `_elements_to_value` strips element text (`xml_adapter.py:209`). I'm including this for completeness but consider it lower priority — the surrounding `\n` padding added by `format_field_with_value` has to be removed somehow, and the old regex-based parser had the same `.strip()` behavior, so this is pre-existing rather than a regression.

### Why these matter

`format → parse` fidelity is what makes few-shot demos and completions interpretable by the same adapter. Failure 1 turns valid LM output into a hard `AdapterParseError`; failures 2–3 corrupt data silently (a downstream `for x in items:` on `None` crashes with a confusing TypeError).

### Proposed fix (happy to send PRs)

1. Escape `]]>` — either escape all `>` as `&gt;`, or minimally replace `]]>` with `]]&gt;` to keep prompts readable.
2. In the `anyOf` branch of `_elements_to_value`, only short-circuit to `None` for empty elements when the non-null choice is scalar-typed; let array/object choices decode empty elements to `[]`/`{}`. A bare null-schema candidate should still return `None` for empty elements so `Optional[some_model]` keeps working.
3. (Optional/lower priority) whitespace handling — needs a format-side change to distinguish padding from content; happy to discuss separately.

I've verified locally that fixes 1 and 2 resolve all three cases without breaking any of the 26 existing tests in `tests/adapters/test_xml_adapter.py`.

KaiOnCode:
Fixes #10269

## Summary

Two round-trip fidelity defects in the nested-XML support added by #10239, where `XMLAdapter.format()` produced output that `XMLAdapter.parse()` rejected or decoded incorrectly.

### 1. `]]>` in text values crashed parsing (regression vs. the pre-#10239 regex parser)

Text escaping covered only `&` and `<`, so a value containing `]]>` was emitted as raw character data. That sequence is illegal in XML (XML 1.0 §2.5), and the adapter's own `ET.fromstring` rejected it:

```python
adapter.format_field_with_value({field: "array ]]> trap"})
# -> <text>\narray ]]> trap\n</text>
adapter.parse(Sig, formatted)
# -> AdapterParseError: Failed to parse XML: not well-formed (invalid token)
```

Any completion containing `]]>` — realistic for code-generation tasks (`arr[i]]>0`) — failed hard. All textual output emissions now escape `]]>` as `]]&gt;`, which decodes back losslessly. The guard covers every non dict/list serialized payload (plain `str`, `str | None`, enums, dates), not just exact-`str` annotations, so nullable strings can't bypass it. I chose the targeted sequence over escaping every `>` to keep prompt rendering unchanged apart from the illegal case.

### 2. Empty containers under nullable unions silently became `None`

`format` emits an empty list as `<items />`, but `parse` short-circuited any empty element with a null union choice to `None`:

```python
items: list[str] | None
# format([])      -> "<items />"
# parse("<items />") -> None   (data loss; also {} for Optional[dict])
```

The empty-element check in `_elements_to_value` now defers to container decoding when a non-null union choice is an array or a free-form object (JSON-schema object without `properties`, i.e. `dict[...]`). Scalars, pydantic models and TypedDicts keep decoding empty elements as `None`, exactly matching the existing expectations in `test_xml_adapter_parses_nullable_and_structured_union_fields`. This is consistent with the format side, which never encodes `None` as an empty element (it emits literal `None` text, which `parse_value` coerces back).

A bare null-schema candidate now also returns `None` for empty elements so `Optional[some_model]` still falls back to `None` when `{}` fails model validation.

## Known residual (unchanged by this PR, noted in the issue)

Inside *nested* user models, `None` and `[]` both serialize to `<tags />`, so the two remain indistinguishable in that position; after this PR they decode as `[]` (before, both decoded as `None`). Whitespace-only strings also continue to lose leading/trailing whitespace, as in all prior versions. Separately, formatting a `None` value for a `str | None` field emits literal `None` text that parses back as the string `"None"` — pre-existing since before #10239, and fixable only with a format-side change to `None` encoding.

## Testing

- 5 new tests in `tests/adapters/test_xml_adapter.py`: `]]>` round-trip for plain, nested, and nullable-string values; empty `Optional[list]`/`Optional[dict]` format→parse; LM-emitted `<items />` → `[]`.
- Full suite: `pytest tests/ --ignore=tests/reliability` → **1264 passed**, no failures (Python 3.11 / macOS).
- Verified pre-existing behavior is otherwise preserved (e.g. single-item scalar text in list tags still parses leniently; garbage input still raises `AdapterParseError`).
- `ruff check` clean on both changed files.

### DIFF
--- a/dspy/adapters/xml_adapter.py
+++ b/dspy/adapters/xml_adapter.py
@@ -30,9 +30,10 @@ class XMLAdapter(ChatAdapter):
                 output.append(self._value_to_xml(serialized, field.name))
                 continue
             formatted = format_field_value(field_info=field.info, value=value)
-            if is_output and field.info.annotation is str:
-                # `]]>` is illegal in XML character data (XML 1.0 §2.5), so escape just that
-                # sequence to keep the output parseable by this adapter's own XML parser.
+            if is_output and not isinstance(serialized, (dict, list)):
+                # Textual output values (str, Optional[str], enums, dates, ...) are re-parsed as
+                # XML character data, where `]]>` is illegal (XML 1.0 §2.5), so escape just that
+                # sequence to keep the emission parseable by this adapter's own XML parser.
                 formatted = formatted.replace("&", "&amp;").replace("<", "&lt;").replace("]]>", "]]&gt;")
             output.append(f"<{field.name}>\n{formatted}\n</{field.name}>")
         return "\n\n".join(output).strip()
--- a/dspy/adapters/xml_adapter.py
+++ b/dspy/adapters/xml_adapter.py
@@ -31,7 +31,9 @@ class XMLAdapter(ChatAdapter):
                 continue
             formatted = format_field_value(field_info=field.info, value=value)
             if is_output and field.info.annotation is str:
-                formatted = formatted.replace("&", "&amp;").replace("<", "&lt;")
+                # `]]>` is illegal in XML character data (XML 1.0 §2.5), so escape just that
+                # sequence to keep the output parseable by this adapter's own XML parser.
+                formatted = formatted.replace("&", "&amp;").replace("<", "&lt;").replace("]]>", "]]&gt;")
             output.append(f"<{field.name}>\n{formatted}\n</{field.name}>")
         return "\n\n".join(output).strip()
 
@@ -150,7 +152,9 @@ class XMLAdapter(ChatAdapter):
                 )
             return f"<{tag}{attrs}>{''.join(children)}</{tag}>"
         return (
-            f"<{tag}{attrs}>{str(value).replace('&', '&amp;').replace('<', '&lt;')}</{tag}>"
+            # `]]>` is illegal in XML character data (XML 1.0 §2.5), so escape just that
+            # sequence to keep the output parseable by this adapter's own XML parser.
+            f"<{tag}{attrs}>{str(value).replace('&', '&amp;').replace('<', '&lt;').replace(']]>', ']]&gt;')}</{tag}>"
             if value is not None
             else f"<{tag}{attrs} />"
         )
@@ -184,9 +188,16 @@ class XMLAdapter(ChatAdapter):
     @classmethod
     def _elements_to_value(cls, elements: list[ET.Element], schema: dict, definitions: dict) -> Any:
         schema = definitions.get(schema.get("$ref", "").rsplit("/", 1)[-1], schema)
+        if schema.get("type") == "null" and not list(elements[0]) and not (elements[0].text or "").strip():
+            return None
         if choices := schema.get("anyOf"):
             if {"type": "null"} in choices and not list(elements[0]) and not (elements[0].text or "").strip():
-                return None
+                # An empty element is ambiguous between "no value" and "empty container". Format
+                # never encodes None as an empty element (it emits literal `None` text), so decode
+                # empty elements as empty containers when a non-null choice is an array or
+                # free-form object, and as None otherwise (scalars, models, TypedDicts).
+                if not any(cls._is_container_choice(choice, definitions) for choice in choices):
+                    return None
             choices = [choice for choice in choices if choice.get("type") != "null"]
             schema = choices[0]
             if list(elements[0]):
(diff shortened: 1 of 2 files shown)


KaiOnCode:
Pushed d9f8420 addressing Greptile's P1: the escaping guard was `field.info.annotation is str`, so nullable strings (`str | None`) and other textual scalar outputs bypassed it. It now escapes whenever the serialized output value is not a dict/list payload (i.e. whenever the emission lands in XML character data), with a regression test for the nullable-string case. Verified: full suite 1264 passed, adapters suite 311 passed, ruff clean.

KaiOnCode:
Closing to focus on other contributions. Thanks.

PR #10301, opened 2026-08-28
fix(dspy): fall back to memory-only cache in configure_cache on disk cache init failure

### LINKED ISSUE #8600: sqlite3.DatabaseError: file is not a database on import, even after cache clearing and fresh install
### What happened?

Importing DSPy fails with a sqlite3.DatabaseError: file is not a database error, even after:

Clearing all cache directories (~/.cache/dspy, project cache, etc.)
Setting all DSPy cache environment variables to disable or redirect cache
Uninstalling and reinstalling DSPy (pip uninstall dspy-ai -y && pip install dspy-ai)
Creating and testing in a fresh virtual environment
Verifying no persistent cache/db files remain

Actual behavior The script fails at import time with:
`sqlite3.DatabaseError: file is not a database`

This happens before any cache disabling can take effect.

Environment (please complete the following information):

OS: Linux
Python version: 3.12
DSPy version: 2.6.27 (also tested with latest)
Virtual environment: venv
Any relevant environment variables: Tried with and without DSPY_CACHE_DIR, DSPY_CACHE, etc.
Additional context

No persistent cache or database files remain in any cache directory.
The error occurs even in a fresh environment and with a minimal script.
This blocks all usage of DSPy.

### Steps to reproduce

Create a new virtual environment with Python 3.12.
Install DSPy: pip install dspy-ai
Run the following minimal script:

```python
#!/usr/bin/env python3
print("🔄 Testing DSPy import...")

try:
    import dspy
    print("✅ DSPy imported successfully!")
    print(f"📦 DSPy version: {dspy.__version__}")
    dspy.settings.configure(cache=False)
    print("✅ Cache disabled successfully!")
    print("🎉 All DSPy tests passed!")
except Exception as e:
    print(f"❌ Error: {e}")
    import traceback
    traceback.print_exc()
```

### DSPy version

2.6.27

Saidheerajgollu:
Fixes #8600

`import dspy` already survives a corrupted disk cache — it falls back to memory-only with a warning. But `dspy.configure_cache()`, the documented way to reconfigure caching at runtime, didn't have the same protection. Calling it against a broken `~/.dspy_cache` (e.g. the corrupted sqlite file case in #8600) still crashed with `sqlite3.DatabaseError: file is not a database`.

This adds the same try/except fallback that `_get_dspy_cache()` already uses at import time, so `configure_cache` degrades gracefully instead of crashing. Added tests covering both the existing import-time fallback and the new `configure_cache` path.

Used Claude Code to help investigate and write this fix; reviewed and tested every line before opening this.

### DIFF
--- a/dspy/clients/__init__.py
+++ b/dspy/clients/__init__.py
@@ -39,15 +39,41 @@ def configure_cache(
         safe_types: Additional types to allow when restrict_pickle is True.
     """
 
-    DSPY_CACHE = Cache(
-        enable_disk_cache,
-        enable_memory_cache,
-        disk_cache_dir,
-        disk_size_limit_bytes,
-        memory_max_entries,
-        restrict_pickle=restrict_pickle,
-        safe_types=safe_types,
-    )
+    # Cache only validates these once disk caching is actually attempted, but we need them
+    # to raise here regardless of enable_disk_cache so a config mistake can't get silently
+    # swallowed by the disk-cache-init fallback below and turned into a memory-only cache.
+    if restrict_pickle:
+        for t in safe_types or []:
+            if not isinstance(t, type):
+                raise TypeError(f"safe_types entries must be types, got {t!r}")
+    if disk_size_limit_bytes is not None and not isinstance(disk_size_limit_bytes, int):
+        raise TypeError(f"disk_size_limit_bytes must be None or an int, got {disk_size_limit_bytes!r}")
+
+    try:
+        DSPY_CACHE = Cache(
+            enable_disk_cache,
+            enable_memory_cache,
+            disk_cache_dir,
+            disk_size_limit_bytes,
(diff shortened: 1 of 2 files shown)


Saidheerajgollu:
Good catch — pushed a fix. safe_types validation now happens unconditionally before the try/except so a bad entry raises immediately instead of getting swallowed by the disk-cache fallback. Added a regression test for it.

Saidheerajgollu:
Same issue, different field — pushed a fix for disk_size_limit_bytes too, and swept the rest of the configurable params (memory_max_entries, disk_cache_dir) to confirm they don't have the same gap. Both already raise correctly regardless of the disk-cache fallback, so this should be the full set now. Added a regression test for disk_size_limit_bytes alongside the safe_types one.

PR #10303, opened 2026-08-28
Add RostamRM retriever integration

vahid-sohrabloo:
## What

Adds `RostamRM`, a `dspy.Module`-based retriever integration for [Rostam](https://github.com/rostamlabs/rostam), an open-source vector database. It follows the same shape as the existing `WeaviateRM` / `DatabricksRM` vendor integrations in `dspy/retrievers/`, and returns `dspy.Prediction(passages=...)`, matching what `dspy.retrievers.Embeddings` and the current RAG tutorial use.

Rostam's search API takes a query *vector*, not a query string, so `RostamRM` is constructed with an `embedder: Callable[[str], list[float]]` used to embed both indexed documents and incoming queries — the same shape `dspy.retrievers.Embeddings` already expects from its `embedder` argument.

Files added/changed:

- `dspy/retrievers/rostam_rm.py` — the `RostamRM` class.
- `tests/retrievers/test_rostam_rm.py` — unit tests (mocked `rostam.Rostam` client, no live server required; mirrors the mocking style of `tests/retrievers/test_colbertv2.py`).
- `pyproject.toml` — adds an optional `rostam` extra (`rostam = ["rostam-client>=0.3"]`), matching the existing `weaviate` extra.

## Usage

```python
from rostam import Rostam
from dspy.retrievers.rostam_rm import RostamRM

client = Rostam("http://localhost:8080")
retriever = RostamRM("my_collection", rostam_client=client, embedder=embed_fn, k=5)
retriever.index(["doc one text", "doc two text"])

class RAG(dspy.Module):
    def __init__(self):
        self.retrieve = retriever
        self.answer = dspy.ChainOfThought("question, context -> answer")

    def forward(self, question):
        passages = self.retrieve(question).passages
        return self.answer(question=question, context=passages)
```

## Design notes / conventions followed

- **Location & naming**: `dspy/retrieve/<vendor>_rm.py` (the pre-3.x layout with `qdrant_rm.py`, `chromadb_rm.py`, etc.) no longer exists in the current tree — third-party vendor retrievers now live directly in `dspy/retrievers/` (currently `weaviate_rm.py`, `databricks_rm.py`). `rostam_rm.py` / `RostamRM` follows that current location and the `<vendor>_rm.py` / `<Vendor>RM` naming convention of its two siblings.
- **Return type**: `WeaviateRM.forward` and `DatabricksRM.forward` predate the current recommended pattern and return, respectively, a raw list of `dotdict(long_text=...)` and a `Prediction(docs=..., doc_ids=...)` (no `passages` key) — both meant to be used through the legacy `dspy.settings.configure(rm=...)` + `dspy.Retrieve` machinery. `RostamRM` instead follows `dspy.retrievers.Embeddings`, the pattern the current DSPy RAG tutorial teaches: a plain module you call directly, whose `forward` returns `dspy.Prediction(passages=...)`. It still works with `dspy.settings.configure(rm=...)` + `dspy.Retrieve()` too, since `dspy.Retrieve.forward` degrades gracefully — but callers are not required to go through it.
- **Not added to `dspy/retrievers/__init__.py`**: neither `WeaviateRM` nor `DatabricksRM` is exported there, presumably to avoid a hard import-time dependency on their (optional) client libraries. `RostamRM` follows that precedent — users import it explicitly: `from dspy.retrievers.rostam_rm import RostamRM`.
- **No new docs page**: `docs/docs/api/tools/` documents `ColBERTv2`, `Embeddings`, and `PythonInterpreter` via `mkdocstrings` auto-ref, but neither `WeaviateRM` nor `DatabricksRM` has a page there or a `mkdocs.yml` nav entry, despite both being fully documented in their own docstrings. `RostamRM`'s docstring follows the same Google-style/Examples format as `WeaviateRM`'s, so a docs page can be added the same way (`::: dspy.retrievers.rostam_rm.RostamRM`) if maintainers want one — I didn't add one unprompted, to stay consistent with the existing (undocumented) siblings.
- **Optional dependency via `pyproject.toml` extra**: `weaviate-client` is an extra; `databricks-sdk` is not (that RM falls back to `requests`, already a base dependency). Rostam has no such fallback, so it gets an extra the same way Weaviate does. `rostam-client` is published on PyPI and is a pure-stdlib, dependency-free client.
- **Tests**: `tests/retrievers/` currently only has tests for `test_colbertv2.py` and `test_embeddings.py` — neither `WeaviateRM` nor `DatabricksRM` has a test file, presumably because they need a live/mocked vendor client and weren't prioritized. `test_rostam_rm.py` is included anyway (mocking `rostam.Rostam`, no network/server needed) since it was cheap to write and keeps `RostamRM`'s behavior pinned; feel free to drop it if maintainers prefer parity with the untested siblings.
- **Lint/format**: verified against the repo's own `ruff` config (`ruff check` and `ruff format --check` both pass on the new files).

## Verification performed

- `pip install -e .` (dspy) + `pip install rostam-client` (0.3.0 from PyPI) in a clean venv, then `pytest tests/retrievers/test_rostam_rm.py -v` — **7 passed**.
- `ruff check dspy/retrievers/rostam_rm.py tests/retrievers/test_rostam_rm.py` — clean.
- `ruff format --check dspy/retrievers/rostam_rm.py tests/retrievers/test_rostam_rm.py` — clean.

## Checklist

- [x] New retriever class + tests added
- [x] Tests pass locally
- [x] Lint/format pass locally
- [ ] CI green (will confirm once opened)

### DIFF
--- a/dspy/retrievers/rostam_rm.py
+++ b/dspy/retrievers/rostam_rm.py
@@ -1,0 +1,200 @@
+from __future__ import annotations
+
+import hashlib
+from typing import Any, Callable
+
+import dspy
+from dspy.primitives.prediction import Prediction
+
+try:
+    from rostam import Rostam, RostamError
+except ImportError as err:
+    raise ImportError(
+        "The 'rostam' package is required to use RostamRM. Install it with `pip install rostam-client`",
+    ) from err
+
+Embedder = Callable[[str], list[float]]
+
+
+def _content_id(text: str) -> str:
+    """A stable string id derived from document content, used when the caller
+    doesn't supply explicit ids to :meth:`RostamRM.index`."""
+    return hashlib.blake2b(text.encode("utf-8"), digest_size=8).hexdigest()
+
+
+def _to_point_id(external_id: str) -> int:
+    """Map an external string id to the uint64 point id Rostam stores.
+
+    A purely-numeric string that fits in 64 bits is used verbatim; anything
+    else is hashed (BLAKE2b, 8 bytes) to a stable id, so repeated ``index()``
+    calls for the same external id address the same point.
+    """
+    if external_id.isdigit():
+        value = int(external_id)
+        if value < (1 << 64):
+            return value
+    return int.from_bytes(hashlib.blake2b(external_id.encode("utf-8"), digest_size=8).digest(), "big")
+
+
+def _scalar_metadata(meta: dict[str, Any]) -> dict[str, Any]:
+    """Keep only the metadata values Rostam can store as scalar/array payload
+    fields (str, int, float, bool, or a homogeneous list of them)."""
+    out: dict[str, Any] = {}
+    for k, v in (meta or {}).items():
+        if isinstance(v, (str, int, float, bool)):
+            out[k] = v
+        elif (
+            isinstance(v, (list, tuple))
+            and v
+            and all(isinstance(x, (str, int, float)) and not isinstance(x, bool) for x in v)
+        ):
+            out[k] = list(v)
+    return out
+
+
+class RostamRM(dspy.Module):
+    """A retrieval module that uses Rostam to return the top passages for a given query.
+
+    Rostam (https://github.com/rostamlabs/rostam) is an open-source vector database.
+    Unlike vendor-hosted search indexes, Rostam's search endpoint takes a query
+    *vector* rather than a query string, so ``RostamRM`` is constructed with an
+    ``embedder`` callable used to embed both the indexed documents and incoming
+    queries.
+
+    Args:
+        rostam_collection_name (str): The name of the Rostam collection to search.
+        rostam_client (Rostam, optional): An existing ``rostam.Rostam`` client
+            instance. Exactly one of ``rostam_client`` or ``url`` must be given.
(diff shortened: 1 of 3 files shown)


vahid-sohrabloo:
Addressed both review findings in 5d52c93:

1. **Configured-RM crash** — `forward` now returns a list of `dotdict(long_text=...)` (matching the sibling retrievers like `weaviate_rm` and the contract `dspy.Retrieve` consumes, `[psg.long_text for psg in rm_result]`) instead of a `Prediction`. Verified end-to-end: with `RostamRM` set as the `rm`, `dspy.Retrieve(k)(query).passages` now returns the passages instead of raising `AttributeError`.
2. **Partial mutation on mismatched lengths** — `index()` now validates the `embeddings`/`ids`/`metadatas` lengths against `texts` up front and raises a descriptive `ValueError` before any `create_collection`/`upsert`, so a bad call can't leave the store partially written.

Added regression tests for both (the configured-rm `dspy.Retrieve` path, and that a length mismatch raises before any server call); updated the existing `forward` tests to the list-of-dotdict return.

PR #10308, opened 2026-08-29
fix(adapters): enforce pydantic Field constraints in parse_value (addresses #7925)

marcodelpin:
# Enforce pydantic field constraints in parse_value

Addresses #7925.

## The problem

`parse_value()` validates against the bare type annotation, so pydantic `Field` constraints
(`max_length`, `ge`/`le`, `pattern`, ...) declared on a signature field are accepted rather
than rejected: a `str` field with `max_length=5` happily takes a 12-character value.

## The change

`parse_value` now validates through the full `FieldInfo` (annotation + constraints) when
constraints are present, so a violating value raises `pydantic.ValidationError` exactly as it
would in a plain pydantic model. Values without constraints take the same path as before.

## Scope, stated plainly

Measured per adapter on this branch:

- **Chat**: violating value rejected (`AdapterParseError`) — enforced.
- **JSON**: violating value rejected (`ValidationError`) — enforced.
- **RLM**: shares the chat parse path — enforced.
- **Flex** (added after the review bot flagged it): the bridge's `parse_value` call passed only
  the bare annotation, dropping constraints on that path too — now passes the `FieldInfo` and is
  pinned by a bridge-level test (`0e24627`).
- **XML** (enforced in `79db8d0`, after the review bot rightly pushed back on our first
  scope call): the adapter's by-name fallback (`xml_adapter.py`) re-validated with a bare
  `TypeAdapter(field.annotation)` on `ValidationError`, which dropped the constraints the
  primary path had just enforced. The fallback now attaches `FieldInfo.metadata` via
  `Annotated`, so the aliased-model retry it exists for still works while a constraint
  violation propagates as `AdapterParseError`. Covered by two tests: a scalar `ge/le` case
  and an aliased-model `min_length` case that exercises the fallback path itself.

One remaining non-goal: the `Literal`/`Enum` fast paths in `parse_value` return the matched
member before the constraint-aware validation runs. On a fixed-member set an extra Field
constraint is degenerate, so we left those paths untouched rather than widen the diff.

## Relationship to #10114

#10114 targets the same issue and came first; it has been conflicting with main since early
August. This PR is offered as a currently-mergeable alternative — if #10114 is revived we are
happy to defer to it or help reconcile the two.

## Testing

Three new tests in the parse_value suite: `test_parse_value_honors_str_max_length_constraint`,
`test_parse_value_honors_int_range_constraint`,
`test_parse_value_without_constraints_is_unchanged` — all fail without the fix.
Full suite on this branch: 1250 passed, 253 skipped, 2 xfailed
(one unrelated timing test, `test_unbatchify_honors_max_wait_time_under_trickling_input`,
flakes under parallel load and passes in isolation on both this branch and main).

AI-assisted (Claude Code): the change and its tests were drafted with AI assistance from the issue thread and the failing behaviour, then reviewed hunk by hunk and verified by me (the constraint tests fail without the fix, the full suite passes on this branch, CI on my fork before opening the PR); the XML by-name fallback follow-up went through the same review. This contribution is my own and offered under the project's MIT license.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -235,7 +235,7 @@ class ChatAdapter(Adapter):
         for k, v in sections:
             if (k not in fields) and (k in signature.output_fields):
                 try:
-                    fields[k] = parse_value(v, signature.output_fields[k].annotation)
+                    fields[k] = parse_value(v, signature.output_fields[k].annotation, signature.output_fields[k])
                 except Exception as e:
                     raise AdapterParseError(
                         adapter_name="ChatAdapter",
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -188,7 +188,7 @@ class JSONAdapter(ChatAdapter):
         # Attempt to cast each value to type signature.output_fields[k].annotation.
         for k, v in fields.items():
             if k in signature.output_fields:
-                fields[k] = parse_value(v, signature.output_fields[k].annotation)
+                fields[k] = parse_value(v, signature.output_fields[k].annotation, signature.output_fields[k])
 
         fields = apply_output_field_defaults(signature, fields)
         if fields.keys() != signature.output_fields.keys():
--- a/dspy/adapters/utils.py
+++ b/dspy/adapters/utils.py
@@ -184,9 +184,21 @@ def find_enum_member(enum, identifier):
     raise ValueError(f"{identifier} is not a valid name or value for the enum {enum.__name__}")
 
 
-def parse_value(value, annotation):
+def parse_value(value, annotation, field_info=None):
+    # Field constraints (max_length, ge, le, ...) live in FieldInfo.metadata, not in the bare
+    # annotation, so a TypeAdapter built from the annotation alone silently ignores them. Re-attach
+    # them via Annotated before validating so the constraints are actually enforced (#7925).
+    constraints = list(field_info.metadata) if field_info is not None else []
+
+    def _validate(candidate):
+        target = Annotated[(annotation, *constraints)] if constraints else annotation
+        return TypeAdapter(target).validate_python(candidate)
+
     if annotation is str:
-        return str(value)
+        result = str(value)
+        # Enforce str constraints (e.g. max_length) while keeping the literal-string coercion that
+        # avoids ast.literal_eval turning "1234" into an int.
+        return _validate(result) if constraints else result
 
     if isinstance(annotation, enum.EnumMeta):
         return find_enum_member(annotation, value)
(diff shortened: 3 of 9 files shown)


marcodelpin:
Good catch on the Flex path — verified and fixed in 0e24627: `flex/bridge.py` was the third `parse_value` call site and passed only the bare annotation, so constraints were dropped there exactly as in the original issue. It now passes the `FieldInfo`, pinned by a bridge-level test that fails without the one-line change. The XML fallback remains deliberately out of scope per the PR description.

marcodelpin:
Right — the by-name fallback was undoing the enforcement on the XML path: `parse_value` now raises on a constraint violation, and the bare-annotation retry accepted the same value one line later.

Fixed in 79db8d0: the retry builds its adapter from `Annotated[(field.annotation, *field.metadata)]`, so the aliased-model case the fallback exists for still validates `by_name` while a constraint violation propagates as `AdapterParseError`. Two tests pin it — a scalar `ge/le` case and an aliased-model `min_length` case that goes through the fallback itself (both fail on the previous head). PR description updated to match.

VANDRANKI (commented):
Community review, not a maintainer approval, does not clear the merge gate.

I read the full diff. The core change is in `dspy/adapters/utils.py`: `parse_value(value, annotation)` becomes `parse_value(value, annotation, field_info=None)`. Constraints such as `max_length`, `ge`, `le` live in `FieldInfo.metadata`, not in the bare `annotation`, so every existing call site was building `TypeAdapter(annotation)` and silently validating without those constraints. The fix pulls `field_info.metadata` into a list and, when non-empty, validates against `Annotated[(annotation, *constraints)]` via a local `_validate` helper instead of the bare annotation, so `TypeAdapter` actually sees the constraints again.

I checked the `str` branch specifically since it's the one that previously bypassed `TypeAdapter` entirely (`return str(value)`). It now still returns the fast literal coercion when there are no constraints, but runs the result through `_validate` when constraints are present — so `OutputField(max_length=N)` gets enforced without disturbing the existing "don't let `ast.literal_eval` turn `"1234"` into an int" behavior for the unconstrained case. That matches the new `test_parse_value_without_constraints_is_unchanged` test.

All five call sites that build a `field_info`-bearing call (`chat_adapter.py`, `json_adapter.py`, `xml_adapter.py`, `predict/flex/bridge.py`, `predict/rlm.py`) were updated consistently to pass the field object through.

The `xml_adapter.py` change is the one I looked at most closely, since it touches a fallback path rather than just adding a parameter: the by-name retry (`adapter.validate_python(value, by_name=True)`) previously used a fresh `TypeAdapter(field.annotation)` with no constraints attached at all, meaning a value the primary `parse_value` call had just rejected for violating a constraint could be silently accepted by the by-name fallback. The fix builds the same `Annotated[...]` type for the fallback's `TypeAdapter` so it can't bypass the constraint. `test_xml_adapter_by_name_fallback_keeps_field_constraints` exercises this directly with an aliased nested model plus a `min_length` constraint on the list, and confirms the fallback still enforces it.

Test coverage looks solid: str max_length, int ge/le, xml primary-path and by-name-fallback constraint enforcement, and a flex-bridge end-to-end test that the sandboxed-result path (`_to_prediction`) now rejects an over-length output. I didn't verify how `Annotated[(annotation, *constraints)]` behaves for every possible annotation shape (e.g., constraints combined with `Optional`/`Union` types beyond what the tests cover), but the change is minimal and additive, and the explicit backward-compatibility test gives me reasonable confidence the unconstrained path is unaffected.

This looks like a real, previously-silent bug (#7925) with a targeted fix and good regression tests across all four adapters plus the flex bridge.

PR #10319, opened 2026-09-01
fix(BestOfN,Refine): make fail_count a per-call retry budget (#10312)

### LINKED ISSUE #10312: BestOfN.forward() and Refine.forward() mutate self.fail_count as shared instance state, not a per-call counter
`BestOfN.forward()` (`dspy/predict/best_of_n.py`) and `Refine.forward()` (`dspy/predict/refine.py`) both use `self.fail_count` as a per-call retry budget, but they decrement the instance attribute directly instead of a local counter. Since `fail_count` is set once in `__init__` and never reset in `forward()`, repeated calls to the same module instance carry over (and can go negative) the decremented value from earlier calls.

Current code, `dspy/predict/best_of_n.py`:

```python
    def __init__(
        self,
        module: Module,
        N: int,  # noqa: N803
        reward_fn: Callable[[dict, Prediction], float],
        threshold: float,
        fail_count: int | None = None,
    ):
        self.module = module
        self.reward_fn = lambda *args: reward_fn(*args)  # to prevent this from becoming a parameter
        self.threshold = threshold
        self.N = N
        self.fail_count = fail_count or N  # default to N if fail_count is not provided
```

and in `forward()`:

```python
            except Exception as e:
                print(f"BestOfN: Attempt {idx + 1} failed with rollout id {rid}: {e}")
                if idx > self.fail_count:
                    raise e
                self.fail_count -= 1
```

`dspy/predict/refine.py` has the identical pattern:

```python
            except Exception as e:
                print(f"Refine: Attempt failed with rollout id {rid}: {e}")
                if idx > self.fail_count:
                    raise e
                self.fail_count -= 1
```

Why this happens: `self.fail_count` is instance state, not a loop-local variable. Each failed attempt inside `forward()` permanently lowers it, and nothing restores it once `forward()` returns. A `BestOfN`/`Refine` module is normally built once (e.g. `best_of_3 = dspy.BestOfN(module=qa, N=3, reward_fn=..., threshold=1.0)`) and then called many times as part of a pipeline, exactly the pattern shown in the class docstring's own example.

Steps to reproduce (conceptually, for `BestOfN`, `Refine` behaves the same way):
1. Create `bo = dspy.BestOfN(module=some_module, N=5, reward_fn=reward_fn, threshold=1.0, fail_count=2)`.
2. Call `bo(**kwargs)` once where 2 of the attempts raise exceptions (allowed, since `fail_count=2`). `self.fail_count` is now `0`.
3. Call `bo(**kwargs)` again on the same instance. Any single attempt that raises now immediately triggers `idx > self.fail_count` (`0 > 0` is False on the very first failure, but `self.fail_count` goes negative from there), so the module tolerates fewer and fewer failures than the `fail_count` the caller configured, with no error or warning that the budget has been silently exhausted by a previous, unrelated call.

There is also an off-by-one at `fail_count=0`: the intent (per the docstring, "Number of allowed failures before raising an exception") is that 0 allowed failures means the first failure should raise immediately, but `idx > self.fail_count` evaluates `0 > 0 → False` on the first failure, so one failure is incorrectly tolerated before `self.fail_count` goes negative.

Expected: `fail_count` should behave as a fixed, per-call budget: the same instance called twice with the same inputs and same failure pattern should behave identically both times. This requires tracking the remaining budget in a local variable inside `forward()` (e.g. `remaining_failures = self.fail_count`) rather than mutating `self.fail_count`.

Actual: the failure budget silently erodes (and can go negative) across repeated calls to the same `BestOfN`/`Refine` instance, since the same `self.fail_count` attribute is shared and decremented on every call.

Verified by reading the current `main` branch source in both files; both `forward()` methods share this exact bug.

linhongyu510:
## Summary

Fixes #10312.

`BestOfN.forward()` and `Refine.forward()` treated `self.fail_count` as a per-call retry budget but **decremented the instance attribute directly and never reset it**, so the failure budget eroded across repeated calls to the same module instance (the documented usage — a module built once and called many times in a pipeline). After a call with failures, later calls tolerated fewer and fewer failures, and the attribute could go negative, with no warning.

Two related issues are fixed together:

1. **Budget leaked across calls.** The remaining budget is now tracked in a local variable (`remaining_failures = self.fail_count`) inside `forward()`, so every call starts from the configured budget.
2. **`fail_count=0` was swallowed and had an off-by-one.** `self.fail_count = fail_count or N` fell back to `N` for an explicit `0`; and `idx > self.fail_count` tolerated the first failure even at `fail_count=0`. The init now uses an explicit `None` check, and a failed attempt raises as soon as the remaining budget is 0 — so `fail_count=0` raises on the first failure.

The default budget is `N-1` (instead of `N`), preserving the existing default behavior where a module whose **every** attempt fails still raises rather than silently returning no prediction.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Same instance, `fail_count=2`, called twice | 1st call: 3 attempts then raise; 2nd call: **2 attempts** then raise (budget leaked) | Both calls: 3 attempts then raise |
| `fail_count=0` | First failure tolerated (and `0` silently became `N`) | First failure raises immediately |
| Default (`fail_count=None`), all `N` attempts fail | raises (via off-by-one) | raises (explicit `N-1` budget) |

## Testing

- Added regression tests for both `BestOfN` and `Refine`: budget reset across repeated calls, and `fail_count=0` raising on the first failure.
- The 4 new tests fail on the unpatched code (budget leak + `fail_count=0` tolerated) and pass with the fix.
- Existing `tests/predict/test_best_of_n.py` and `tests/predict/test_refine.py` all pass (10 passed).


linhongyu510:
Addressed the documentation mismatch in 75f6933c1: the built-in module guide now states that BestOfN defaults to `max(N - 1, 0)`, matching the implementation, docstring, and all-failure behavior. `git diff --check` passes; the local environment does not currently provide the optional `pre-commit` executable, so I have not claimed that gate.

linhongyu510:
The P1 was caused by unrelated source-context commits that this branch had accidentally inherited. I rebuilt the branch from current `origin/main` and force-pushed only the #10312 retry-budget implementation/tests plus the requested documentation correction. The PR now touches exactly five target files; `dspy/propose/utils.py` and `tests/propose/test_utils.py` are no longer in the diff. Verification on the rebuilt head: 10 focused tests passed, Ruff lint/format passed, all applicable pre-commit hooks passed, and `git diff --check` passed.

VANDRANKI (commented):
I filed #10312, so I traced this fix against the bug I found.

The core fix is right. The old code decremented `self.fail_count` directly inside `forward()`, so the failure budget was shared, mutable instance state instead of a per-call value. Moving the countdown into a local `remaining_failures = self.fail_count` at the top of `forward()` is the correct fix: it resets on every call instead of eroding (and eventually going negative, as I described in the issue).

I traced both code paths by hand:
- `fail_count=2, N=5`, single call: old code raised after 3 attempts (idx 0,1,2). New code raises after 3 attempts too (remaining_failures goes 2 -> 1 -> 0 -> raise). Same behavior within one call, as expected.
- Same instance, second call: old code's mutated `self.fail_count` was already at 0/negative from the first call, so it raised after only 2 attempts on the second call, exactly the erosion bug in #10312. New code resets `remaining_failures` from the untouched `self.fail_count` and raises after 3 attempts again on the second call, matching the first.
- Default case (`fail_count=None`): the default changed from `N` to `N-1`, but I traced both against the old `idx > self.fail_count` comparison and confirmed the observable behavior for the all-fail case is unchanged (still raises on the last attempt) - it's a cleanup of an implicit off-by-one, not a behavior change for the default path.
- `fail_count=0`: fixed the `fail_count or N` bug where an explicit 0 was silently replaced by N. The explicit `is not None` check is correct.

Tests: the 4 new tests genuinely exercise the bug (cross-call reset, fail_count=0). I did not run the suite myself, but the assertions match what I'd expect from tracing the code.

This is a community review, not a maintainer approval, so it doesn't clear the merge gate. Nothing else in the diff looks out of scope.

linhongyu510:
@VANDRANKI Thanks for tracing this against #10312 rather than just reading the diff — your call-by-call walkthrough matches my implementation exactly, including the part that matters most and is easiest to get wrong.

On the `fail_count=None` default: you are right that the observable all-fail behaviour is unchanged, and that is worth stating precisely because the constant did change. Old code compared `idx > self.fail_count` with the default `N`; new code counts down from `N - 1`. Both raise on the final attempt, so nothing observable moved — the `N` was carrying an implicit off-by-one that only made sense in combination with the old comparison. Once the countdown moved into a local, keeping `N` would have granted one extra failure. I made that explicit in the docs commit (`built-in-module-variants.md` now says `max(N - 1, 0)`), since the old text described the old constant.

Two notes for anyone reading later:

**`fail_count=0` was a separate bug from the erosion one.** `self.fail_count = fail_count or N` silently replaced an explicit `0` with `N`, so "never tolerate a failure" meant "tolerate N of them". The `is not None` check fixes that independently of the per-call budget.

**Cross-call reset is the part regression tests usually miss.** As you noted, within a single call old and new behave identically — a test that only calls `forward()` once passes on the buggy code. The two tests that actually discriminate are the ones reusing a predictor across calls.

I have rebased onto current `main` (the branch was 37 commits behind); new head `69864e10`, still `MERGEABLE`, and I re-ran rather than assuming the earlier results carried:

```
tests/predict/test_best_of_n.py tests/predict/test_refine.py   10 passed
ruff check / ruff format --check                               clean
```

Required CI on the previous head was green (`statusCheckRollup: SUCCESS`); fresh CI is now running on the rebased head.

Understood that a community review does not clear the merge gate — but having the issue reporter independently trace the fix is the useful part. @okhat @chenmoneygithub this one is small (two predictors, `+131/-9` including tests) and the reporter of #10312 has verified the behaviour; would appreciate a look when convenient.

PR #10322, opened 2026-09-02
fix(teleprompt): respect MIPROv2.compile seed=0

### LINKED ISSUE #10321: [Bug] MIPROv2.compile(seed=0) ignores the seed
### What happened?

In MIPROv2.compile, the seed argument is applied with or:

## dspy/teleprompt/mipro_optimizer_v2.py, line 170
```
seed = seed or self.seed
self._set_random_seeds(seed)
```

Since `0 or self.seed` evaluates to `self.seed`, calling `compile(..., seed=0)` silently falls back to the constructor default (`seed=9`). So the run uses a different seed than the user asked for.

### Steps to reproduce

```
optimizer = dspy.MIPROv2(metric=my_metric, auto="light")
optimizer.compile(program, trainset=trainset, seed=0)
# internally runs with seed=9, not seed=0
```

**Expected**                                                                                                                                                                                                                               

`seed=0` should seed the run with 0. Only `seed=None` should fall back to `self.seed`

**Proposed fix**
`seed = seed if seed is not None else self.seed`

### DSPy version

3.3.1

Zsanz3:
## Changes

`MIPROv2.compile` applied the seed with `seed = seed or self.seed`. In Python, `0 or self.seed` is `self.seed`, so `compile(..., seed=0)` silently used the constructor default (`seed=9`).

The fallback is now `seed = seed if seed is not None else self.seed`. Only a missing/`None` seed uses the constructor value; `seed=0` is kept.

A unit test asserts that `compile(..., seed=0)` calls `_set_random_seeds(0)`.

Fixes #10321.

### DIFF
--- a/dspy/teleprompt/mipro_optimizer_v2.py
+++ b/dspy/teleprompt/mipro_optimizer_v2.py
@@ -167,7 +167,7 @@ class MIPROv2(Teleprompter):
             )
 
         # Set random seeds
-        seed = seed or self.seed
+        seed = seed if seed is not None else self.seed
         self._set_random_seeds(seed)
 
 
--- a/tests/teleprompt/test_mipro_optimizer_v2.py
+++ b/tests/teleprompt/test_mipro_optimizer_v2.py
@@ -1,0 +1,23 @@
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from dspy.teleprompt.mipro_optimizer_v2 import MIPROv2
+
+
+def test_compile_respects_seed_zero():
+    """compile(seed=0) must keep seed 0 instead of falling back to the constructor default.
+
+    Regression test for https://github.com/stanfordnlp/dspy/issues/10321
+    """
+    lm = MagicMock()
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):
Community review, does not clear the merge gate.

Small, clear fix in dspy/teleprompt/mipro_optimizer_v2.py. `compile()` had `seed = seed or self.seed`, which is the classic Python truthy-vs-None bug: `0 or self.seed` evaluates to `self.seed` because `0` is falsy, so calling `compile(seed=0)` silently ignored the caller's explicit 0 and fell back to whatever seed the constructor was given. Fixed to `seed = seed if seed is not None else self.seed`, which only falls back when the caller didn't pass a seed at all.

The new test mocks `_set_random_seeds` and `_set_and_validate_datasets` (raising to stop execution right after the seed line), calls `compile(..., seed=0)`, and asserts `_set_random_seeds` was called with `0` specifically. That directly targets the bug rather than just checking compile succeeds.

Simple, correct, matches the referenced issue #10321. Comfortable approving.

PR #10324, opened 2026-09-02
Add field_markers opt-out to ChatAdapter

### LINKED ISSUE #10073: ChatAdapter Field Markers Token Overhead - ChatAdapter hard-codes structural field markers into every ReAct iteration — no opt-out
## Summary

  `ChatAdapter` unconditionally injects structural field markers into every react
  iteration prompt:
  
  [[ ## next_thought ## ]]
  [[ ## next_tool_name ## ]]
  [[ ## next_tool_args ## ]]

  These markers are not configurable, cannot be disabled, and add **~27 tokens per
  iteration** with no supported opt-out path. For LLM backends that charge per token
  and have latency proportional to token count, this is a fixed, invisible tax on every
  agentic request.

  ## Measured Impact

  Benchmarked on a production agentic service (AWS Bedrock / Claude, 2-step trajectory):

  | Metric | Value |
  |---|---|
  | Tokens added per iteration | **~27 tokens** |
  | Latency added (2-step trajectory) | **>~1.3s per request** |
  | Ability to opt out | **None** |

  At scale across thousands of requests per day, this accumulates into significant
  unnecessary cost and latency.

  ## Why This Is Hard to Work Around

  Removing these markers requires one of:

  1. **Subclassing `ChatAdapter`** — risky, DSPy internals are explicitly not stable
     across releases; a subclass breaks silently on any internal refactor
  2. **Switching to `JSONAdapter` or `TurboAdapter`** — may break existing output
     parsing; not a drop-in replacement; requires separate evaluation on each agent

  Neither is a safe, supported path. Users are effectively forced to absorb the overhead.

  ## What We'd Like

  One of the following:

  1. **A supported adapter** (e.g. `dspy.LeanChatAdapter` or a flag on `ChatAdapter`)
     that omits the field markers — for users who have validated their output parsing
     does not depend on them
  2. **A configuration option** on `ChatAdapter` or `dspy.configure()` to opt out of
     field markers, e.g.:
     ```python
     dspy.configure(lm=lm, adapter=dspy.ChatAdapter(field_markers=False))
  3. Documentation on which adapter is most token-efficient for production ReAct
  workloads and what the compatibility trade-offs are
  
  Context
  
  This was discovered during a latency analysis comparing a DSPy-based Python agent
  against an equivalent Spring AI Java agent backed by the same tool server. The Java
  agent produced 93 completion tokens for a one-way flight search; the Python/DSPy
  agent produced ~195 tokens for the same query. The field markers accounted for
  ~27 of that ~100-token gap — the remainder was addressed via FastReAct and prompt
  changes, but this marker overhead has no fix today.
  
  Environment
  
  - dspy version: latest stable
  - LLM backend: AWS Bedrock (Claude 3.x)
  - Adapter: dspy.ChatAdapter (default)
  - Module: dspy.ReAct
  
  ---


### Additional Context

_No response_

ramyasreenagarajan:
Summary
Adds a "field_markers: bool = True"  parameter to "ChatAdapter" that, when set to
"False", omits the "[[ ## field_name ## ]]" structural markers from prompts and
completions — reducing token overhead for signatures with a single output field.

Closes #10073

Design notes
- Defaults to "True", so existing behavior is fully unchanged unless explicitly opted out.
- Only supported for signatures with a "single output field". Markers are what let
  "parse()" separate multiple output fields in the raw completion text — without them,
  there's no reliable way to split a multi-field completion. Attempting to use
  "field_markers=False" with more than one output field raises a clear "ValueError"
  at call time (see "_validate_field_markers").
- For multi-field, low-token-overhead use cases, "JSONAdapter" remains the
  recommended alternative.

Testing
Added two tests:
- "test_field_markers_false_single_output_field" — confirms no markers appear in
  the prompt and parsing works correctly for a single-output signature.
- "test_field_markers_false_multi_output_field_raises" — confirms the clear error
  is raised for multi-output signatures.

All 78 existing tests in "test_chat_adapter.py" continue to pass.

### DIFF
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -97,7 +97,7 @@ class ChatAdapter(Adapter):
         inputs: dict[str, Any],
     ) -> list[dict[str, Any]]:
         processed_signature = self._call_preprocess(lm, lm_kwargs, signature, inputs)
-        self._validate_field_markers(signature)
+        self._validate_field_markers(processed_signature)
         try:
             return super().__call__(lm, lm_kwargs, signature, demos, inputs)
         except Exception as e:
@@ -118,8 +118,8 @@ class ChatAdapter(Adapter):
         demos: list[dict[str, Any]],
         inputs: dict[str, Any],
     ) -> list[dict[str, Any]]:
-         processed_signature = self._call_preprocess(lm, lm_kwargs, signature, inputs)
-        self._validate_field_markers(signature)
+        processed_signature = self._call_preprocess(lm, lm_kwargs, signature, inputs)
+        self._validate_field_markers(processed_signature)
         try:
             return await super().acall(lm, lm_kwargs, signature, demos, inputs)
         except Exception as e:
--- a/dspy/adapters/chat_adapter.py
+++ b/dspy/adapters/chat_adapter.py
@@ -57,12 +57,11 @@ class ChatAdapter(Adapter):
                 If True, when an error occurs (except ContextWindowExceededError), the adapter will retry using
                 JSONAdapter. Defaults to True.
             parallel_tool_calls: Whether to request provider-side parallel tool-call generation when native function
-                calling is active. If None, the adapter does not set the p0ovider option.
+                calling is active. If None, the adapter does not set the provider option.
             field_markers: Whether to wrap each field in `[[ ## field_name ## ]]` markers in prompts and
                 completions. Defaults to True. Set to False to reduce token overhead — only supported for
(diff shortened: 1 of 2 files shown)


ramyasreenagarajan:
Fixed both issues from the review:
- P1: user_message_output_requirements() now returns None when native
  preprocessing leaves zero output fields (e.g. sole Citations output),
  instead of crashing on the tuple unpack.
- P2: fixed the "p0ovider" typo and rejoined the split "raw" text in
  the constructor docstring.
All 78 tests in test_chat_adapter.py pass.

PR #10325, opened 2026-09-03
fix: key Image.format cache on URL instead of self

### LINKED ISSUE #10273: [Bug] Image.format's lru_cache pins up to 32 Image objects for the life of the process
### What happened?

`Image.format` is decorated with `@lru_cache(maxsize=32)` (`dspy/adapters/types/image.py:107`).
`lru_cache` on a method keys on `self`, so the cache holds a strong reference to the `Image`
object. Up to 32 `Image` instances stay alive for the life of the process, and each one can
carry a multi-megabyte base64 data URI.

Two things go wrong together:

1. When each example has its own image, every call is a cache miss, so the cache never returns
   a saved result. It still keeps the last 32 images resident.
2. Nothing invalidates it. `Image.format.cache_clear()` is the only way to release them, and no
   DSPy code path calls it.

I isolated it from the other retention path. `dspy.settings.trace` also holds inputs, so I
cleared the trace first: 32 images are still alive after that, and only `cache_clear()` frees
them.

Running the script below on 40 images of about 200KB each:

```
images created           : 40
cache hits / misses      : 0 / 40
cache currsize / maxsize : 32 / 32
alive after gc           : 40
  after trace.clear()          -> 32
  after format.cache_clear()   -> 0
```

Zero hits across 40 calls, and 32 images pinned. With realistic photo payloads that is hundreds
of megabytes held for no benefit.

The cache was added deliberately in #8842 for throughput, so I do not think removing it outright
is the right call. It does help when the same `Image` object is formatted repeatedly, which is
what happens with few-shot demos. Keying a module-level cache on the URL string rather than on
`self` would keep that win without holding the objects.

Worth noting: ruff's B019 flags exactly this pattern, and draft PR #9198 adds `# noqa: B019` to
this line. The repo's ruff config does not select the `B` ruleset, so the warning never fires in
CI here.

### Steps to reproduce

No API key needed.

```python
import gc, weakref, dspy
from dspy.adapters.types.image import Image

class Caption(dspy.Signature):
    img: Image = dspy.InputField()
    caption: str = dspy.OutputField()

dspy.configure(lm=DummyLM([{"caption": "x"}] * 60), adapter=ChatAdapter())
predict = dspy.Predict(Caption)

refs = []
for i in range(40):
    im = Image(url="data:image/png;base64," + "A" * 200_000 + str(i))
    refs.append(weakref.ref(im))
    predict(img=im)
    del im

gc.collect()
alive = lambda: sum(1 for r in refs if r() is not None)
info = Image.format.cache_info()

print(f"cache hits / misses      : {info.hits} / {info.misses}")
print(f"cache currsize / maxsize : {info.currsize} / {info.maxsize}")
print(f"alive after gc           : {alive()}")

dspy.settings.trace.clear(); gc.collect()
print(f"  after trace.clear()        -> {alive()}")
Image.format.cache_clear(); gc.collect()
print(f"  after format.cache_clear() -> {alive()}")
```

### DSPy version

3.3.1, and reproduced on `main` at 03641a2ee.

---

**AI disclosure.** Found with Claude Code (Opus) while auditing the repo with a code
knowledge graph. My prompts were, in substance: "audit the whole codebase using
codebase-memory-mcp, build the graph, use nodes and edges to find bugs and things that
don't make sense, check bridges and junctions, make a list", then "check again with depth,
run things if needed, and only file if it all holds up". The first pass got this wrong twice:
it credited the retention to the wrong structure, and I only separated `settings.trace` from
the method cache after re-running with weakrefs. I reproduced the final numbers myself on
3.3.1 and on `main` at 03641a2ee before filing, and the script above is the one I ran.

leilei3167:
## Summary

Fixes #10273.

`@lru_cache` on `Image.format` keys on `self`, so the cache holds up to 32 `Image` instances (often with large base64 payloads) for the life of the process. Unique images never hit the cache, but still fill it.

Move the cache to a module-level helper keyed by the URL string. Same-URL formatting still hits; `Image` objects are no longer pinned.

## Test plan

- [x] `uv run --frozen pytest tests/signatures/test_adapter_image.py -q` — 29 passed, 2 xfailed
- [x] New `test_format_cache_does_not_retain_image_instances` (weakref + cache hits)
- [x] Issue repro script: after `trace.clear()`, alive count is 0 (was 32 before)

@TomeHirata — ready for review; this keeps the throughput win from #8842 without retaining `Image` instances.

### DIFF
--- a/dspy/adapters/types/image.py
+++ b/dspy/adapters/types/image.py
@@ -1,4 +1,5 @@
 import base64
+import copy
 import io
 import mimetypes
 import os
@@ -106,7 +107,9 @@ class Image(Type):
 
     def format(self) -> list[dict[str, Any]] | str:
         try:
-            return _format_image(self.url)
+            # Deep-copy so callers that mutate the nested list/dict cannot
+            # poison the shared @lru_cache entry for this URL.
+            return copy.deepcopy(_format_image(self.url))
         except Exception as e:
             raise ValueError(f"Failed to format image for DSPy: {e}")
 
--- a/dspy/adapters/types/image.py
+++ b/dspy/adapters/types/image.py
@@ -104,13 +104,11 @@ class Image(Type):
         # Delegate the rest of initialization to pydantic's BaseModel.
         super().__init__(**data)
 
-    @lru_cache(maxsize=32)
     def format(self) -> list[dict[str, Any]] | str:
         try:
-            image_url = encode_image(self.url)
+            return _format_image(self.url)
         except Exception as e:
             raise ValueError(f"Failed to format image for DSPy: {e}")
-        return [{"type": "image_url", "image_url": {"url": image_url}}]
 
     @classmethod
     def from_url(cls, url: str, verify: bool = True, timeout: float | None = 30.0) -> "Image":
@@ -161,6 +159,12 @@ class Image(Type):
             image_type = self.url.split(";")[0].split("/")[-1]
             return f"Image(url=data:image/{image_type};base64,<IMAGE_BASE_64_ENCODED({len_base64!s})>)"
         return f"Image(url='{self.url}')"
+
+
+@lru_cache(maxsize=32)
+def _format_image(url: str) -> list[dict[str, Any]] | str:
+    image_url = encode_image(url)
+    return [{"type": "image_url", "image_url": {"url": image_url}}]
 
 
 def is_url(string: str) -> bool:
--- a/tests/signatures/test_adapter_image.py
+++ b/tests/signatures/test_adapter_image.py
@@ -1,7 +1,9 @@
 import base64
+import gc
 import os
 import tempfile
 import warnings
+import weakref
 from io import BytesIO
 
 import pydantic
@@ -10,7 +12,7 @@ import requests
 from PIL import Image as PILImage
 
 import dspy
-from dspy.adapters.types.image import encode_image
+from dspy.adapters.types.image import _format_image, encode_image
 from dspy.utils.dummies import DummyLM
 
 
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):

Traced the change to `Image.format()`/`_format_image` in `dspy/adapters/types/image.py`, referencing the linked #10273. The root problem is a classic `lru_cache`-on-instance-method misuse: `@lru_cache(maxsize=32)` was applied directly to `def format(self)`, so the cache key included `self` (the `Image` instance) as an argument. That both defeats the purpose of caching across separately-constructed `Image` objects sharing the same URL, and pins every cached `Image` instance alive for the life of the process (or until evicted from a 32-slot cache), which is a real leak in any long-running process that creates many transient `Image` objects. Moving the cache to a module-level `_format_image(url: str)` function, keyed purely on the URL string, fixes both: cache hits now work across distinct instances with the same URL, and the cache no longer holds a reference to the `Image` object itself.

That fix introduces a second, subtler problem the PR also correctly handles: once `_format_image`'s result (a list of dicts) is genuinely shared across calls via a URL-keyed cache, a caller that mutates the returned list/dict would poison that shared entry for every other caller with the same URL — which couldn't happen before, since each instance had its own isolated cache slot. The `copy.deepcopy()` in the public `format()` method addresses that, and `test_format_cache_payload_is_isolated_from_mutation` is a genuinely good test for it: it mutates one call's result in place and confirms a second call for the same URL is unaffected.

The leak fix itself is verified with a real `weakref` test (`test_format_cache_does_not_retain_image_instances`): it takes a weak reference to an `Image`, calls `.format()`, deletes the strong reference, forces `gc.collect()`, and confirms the weakref resolves to `None` — meaning the module-level cache genuinely isn't holding the instance alive. That's the right way to prove a reference-leak fix rather than just asserting behavior looks right.

One minor thing worth confirming: `copy.deepcopy` runs on every `format()` call now, even cache hits. The structure being copied is small (just the wrapped URL string in a dict), so the cost is negligible relative to what `encode_image` was doing before it got cached, but it does mean the "cached" path isn't quite free.

PR #10326, opened 2026-09-03
fix(adapters): keep pydantic field constraints in structured outputs

### LINKED ISSUE #10195: [Bug] JSONAdapter structured outputs: pydantic Field constraints (ge, le, multiple_of, pattern) are silently dropped in the signature-to-model derivation — is this intentional?

Rodrigo-Palma:
## 📝 Changes Description

Fixes #10195.

`_get_structured_outputs_response_format` rebuilt every output field as a bare
`(annotation, default)` pair:

```python
default = field.default if hasattr(field, "default") else ...
fields[name] = (annotation, default)
```

That discards the `FieldInfo`, and with it every constraint declared on the field —
`ge`/`le`/`gt`/`lt`, `multiple_of`, `pattern`, `min_length`/`max_length`. The derived
response format kept the type and dropped the bound, so the provider was never asked to
honor it:

```python
class Score(dspy.Signature):
    text: str = dspy.InputField()
    score: float = dspy.OutputField(ge=0.0, le=1.0, multiple_of=0.25)

_get_structured_outputs_response_format(Score, True).model_json_schema()["properties"]
# before: {"score": {"title": "Score", "type": "number"}}
# after:  {"score": {"maximum": 1.0, "minimum": 0.0, "multipleOf": 0.25,
#                    "title": "Score", "type": "number"}}
```

**Why this reads as a loss rather than a deliberate choice** (the issue asks): DSPy
already consumes this exact metadata on the other path. `move_kwargs` runs
`_translate_pydantic_field_constraints(**kwargs)` and stores the result in
`json_schema_extra["constraints"]`, which the prompt-side adapters render as natural
language for the model. So the same `OutputField(ge=0, le=1)` states its bound to the
model in JSON mode and states nothing at all in structured-outputs mode. The framework
understands the constraint; only the derivation drops it.

**The change:** rebuild from the `FieldInfo` and clear only DSPy's own prompt metadata,
which is not part of the wire schema. The default carries over with the `FieldInfo`, so
the `required` list and `additionalProperties: false` enforcement are unchanged, and a
field declared without constraints produces byte-for-byte the same schema as before.

**Tests** (`tests/adapters/test_json_adapter.py`, 6 added): numeric bounds, exclusive
bounds and string constraints are preserved — these three fail on `main`; DSPy metadata
does not leak into the schema; `required`/`additionalProperties` still enforced; an
unconstrained field's schema is unchanged.

## ✅ Contributor Checklist

Full suite as CI runs it: `pytest -m 'not extra and not deno' -n auto` → 1256 passed,
253 skipped, 2 xfailed. `ruff check --fix-only --diff --exit-non-zero-on-fix` clean.

## ⚠️ Warnings

Structured-outputs backends differ in which JSON Schema validation keywords they accept.
The keywords emitted here (`minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`,
`multipleOf`, `pattern`, `minLength`, `maxLength`) are in OpenAI's documented supported
subset, but a provider that rejects them would now fail where it previously succeeded by
silently dropping the user's constraint. If you would rather sanitize per provider than
emit them unconditionally, that split belongs at this same seam and I am happy to add it —
I did not want to guess at a provider allowlist.

---

Disclosure: I used an AI coding assistant to help explore the code and draft this. I
reproduced the behavior, wrote and verified the tests, and understand the change.

### DIFF
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -1,7 +1,6 @@
-import copy
 import json
 import logging
-from typing import Any, get_origin
+from typing import Annotated, Any, get_origin
 
 import json_repair
 import pydantic
@@ -258,15 +257,17 @@ def _get_structured_outputs_response_format(
         if use_native_function_calling and annotation == ToolCalls:
             # Skip ToolCalls field if native function calling is enabled.
             continue
-        # `field` carries the constraint metadata the user declared on the OutputField
-        # (ge/le/gt/lt, multiple_of, pattern, min_length/max_length). Passing only
-        # (annotation, default) drops all of it, so the derived schema stops asking the
-        # provider to honor a bound that the prompt-side path still states through
-        # json_schema_extra["constraints"]. Rebuild from the FieldInfo instead, and drop
-        # only DSPy's own prompt metadata, which is not part of the wire schema.
-        field_info = copy.deepcopy(field)
-        field_info.json_schema_extra = None
-        fields[name] = (annotation, field_info)
+        default = field.default if hasattr(field, "default") else ...
+        # `field.metadata` holds exactly the constraints the user declared on the
+        # OutputField (ge/le/gt/lt, multiple_of, pattern, min_length/max_length).
+        # Passing only (annotation, default) drops all of it, so the derived schema
+        # stops asking the provider to honor a bound that the prompt-side path still
+        # states through json_schema_extra["constraints"]. Re-attach just the
+        # constraints: carrying the whole FieldInfo would also put the field's alias
+        # and description into the wire schema, and the parser keys off the signature
+        # field names, so an alias there makes a schema-valid response unparseable.
+        annotation = Annotated[(annotation, *field.metadata)] if field.metadata else annotation
+        fields[name] = (annotation, default)
 
     # Build the model with extra fields forbidden.
     pydantic_model = pydantic.create_model(
--- a/dspy/adapters/json_adapter.py
+++ b/dspy/adapters/json_adapter.py
@@ -1,3 +1,4 @@
+import copy
 import json
 import logging
 from typing import Any, get_origin
@@ -257,8 +258,15 @@ def _get_structured_outputs_response_format(
         if use_native_function_calling and annotation == ToolCalls:
             # Skip ToolCalls field if native function calling is enabled.
             continue
-        default = field.default if hasattr(field, "default") else ...
-        fields[name] = (annotation, default)
+        # `field` carries the constraint metadata the user declared on the OutputField
+        # (ge/le/gt/lt, multiple_of, pattern, min_length/max_length). Passing only
+        # (annotation, default) drops all of it, so the derived schema stops asking the
+        # provider to honor a bound that the prompt-side path still states through
+        # json_schema_extra["constraints"]. Rebuild from the FieldInfo instead, and drop
+        # only DSPy's own prompt metadata, which is not part of the wire schema.
+        field_info = copy.deepcopy(field)
+        field_info.json_schema_extra = None
+        fields[name] = (annotation, field_info)
 
     # Build the model with extra fields forbidden.
     pydantic_model = pydantic.create_model(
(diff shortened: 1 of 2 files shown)


Rodrigo-Palma:
Good catch, and it reproduces exactly as described: `answer_text: str = dspy.OutputField(alias="answer")` produced a schema whose only property was `answer`, while `JSONAdapter.parse` keys off the signature field names, so a schema-valid `{"answer": ...}` was discarded before validation.

Fixed in acfda15, and the root cause was broader than the alias. Copying the whole `FieldInfo` carried *everything* it holds into the wire schema, not just the constraints this PR is about. A field's `description` leaked the same way — `OutputField(description="the answer")` started emitting `"description"` into the schema, which is a separate change from preserving constraints and does not belong here either.

So rather than stripping attributes one by one, the fix now re-attaches only what the constraints actually live in. `field.metadata` holds exactly the declared constraints, so:

```python
annotation = Annotated[(annotation, *field.metadata)] if field.metadata else annotation
fields[name] = (annotation, default)
```

Nothing else from the `FieldInfo` can reach the schema by construction — no alias, no description, no title override.

I deliberately did not take the other route you mentioned (normalizing aliases back to field names in the parser). Making the adapter alias-aware is a feature, not part of a constraints fix, and it would change what responses DSPy accepts. This PR now leaves alias behavior exactly as it was on `main`.

Schema after the fix, with constraints, an alias and a description all declared on the same signature:

```
{"answer_text": {"title": "Answer Text", "type": "string"},
 "label": {"maxLength": 10, "pattern": "^[a-z]+$", "title": "Label", "type": "string"},
 "score": {"maximum": 1.0, "minimum": 0.0, "multipleOf": 0.25, "title": "Score", "type": "number"}}
required: ["score", "label", "answer_text"]
```

Three tests added for this: the alias does not reach the schema or `required`, the description does not, and an aliased output field still parses end to end. Full suite: 1259 passed, 253 skipped, 2 xfailed.

@greptileai

Rodrigo-Palma:
The workflow run here is waiting on maintainer approval (first-time contributor), so I ran this repo's full CI on my fork to have the result available in advance:

**[Lint, Test, and Build run](https://github.com/Rodrigo-Palma/dspy/actions/runs/33773344101)** — `fix/structured-outputs-keep-field-constraints`

All 19 jobs green: `Run Tests` on 3.10 / 3.11 / 3.12 / 3.13 / 3.14, `Run Extra and Deno Tests` on all five, `Run MCP v2 Tests`, `Run Tests with Real LM`, `Build Package` on all five, `Check Ruff Fix`, and `Lint GitHub Actions Workflows`.

PR #10333, opened 2026-09-03
fix(tool): resolve JSON schema refs as local pointers with cycle detection

### LINKED ISSUE #10327: [Bug] `_resolve_json_schema_reference` infinitely recurses on self-referential models and KeyErrors on non-`#/$defs` refs — breaks `dspy.Tool` construction and MCP tool conversion
### What happened?

`_resolve_json_schema_reference` (dspy/adapters/types/tool.py:517) expands JSON-Schema `$ref`s with no cycle detection and a lookup that assumes every ref's last path segment exists in `$defs`. Three distinct failure modes follow, all reachable from public API:

#### 1. Self-referential pydantic model as a tool argument → `RecursionError`

Recursive input schemas are legal JSON Schema, and pydantic emits exactly that shape for self-referencing models. Construction of the tool crashes:

```python
import pydantic, dspy

class TreeNode(pydantic.BaseModel):
    name: str
    children: list["TreeNode"] = []

TreeNode.model_rebuild()

def organize(tree: TreeNode) -> str:
    """Summarize a tree."""
    return "ok"

dspy.Tool(organize)  # RecursionError
```

Same for the MCP path: `convert_mcp_tool` → `convert_input_schema_to_tool_args` → `_resolve_json_schema_reference` — any MCP server with a recursive input schema makes tool listing fail. Measured on `main` (`59ce760`), Python 3.11, macOS.

Note the project already solves this for prompts: `_schema_to_xml` (added in #10239) carries a `seen` frozenset to stop at cycles — the resolver should do the same.

#### 2. Refs not shaped like `#/$defs/<name>` → `KeyError`

The resolver takes `ref_path = obj["$ref"].split("/")[-1]` and looks it up in `schema["$defs"]` only. Local pointers from OpenAPI-style or hand-written schemas crash tool construction:

```python
from dspy.adapters.types.tool import convert_input_schema_to_tool_args

schema = {
    "type": "object",
    "$defs": {"Other": {"type": "string"}},
    "properties": {
        "node": {"$ref": "#/definitions/Node"},
        "y": {"type": "object", "properties": {"z": {"$ref": "#/properties/x"}}},
    },
    "required": ["node"],
    "definitions": {"Node": {"type": "object", "properties": {"name": {"type": "string"}}}},
}
convert_input_schema_to_tool_args(schema)  # KeyError: 'Node'
```

Both refs above point into document locations the JSON Schema pointer syntax addresses explicitly (`#/definitions/...`, `#/properties/...`), and neither resolves: the first KeyErrors (present in `definitions`, looked up in `$defs`), the second KeyErrors (`'x'` is not in `$defs`).

#### 3. Legacy `definitions`-only schema → silent degradation

When the top-level schema has `definitions` but no `$defs`, `convert_input_schema_to_tool_args` never attempts resolution (`defs = schema.get("$defs", {})` is empty), so the `{"$ref": "#/definitions/Node"}` dict is passed through verbatim as the argument schema and `arg_types` falls back to `Any`. The resulting tool advertises a dangling `$ref` to providers (rejected by strict-schema APIs) and loses all validation. The docstring/guard in `_resolve_json_schema_reference` checks `"$defs" not in schema and "definitions" not in schema`, so `definitions` was clearly intended to be supported — the lookup just never implements it.

### Proposed fix (happy to send PR — I have it working locally)

1. **Cycle-safe expansion**: track ref names on the current expansion stack; when a cycle closes, emit `{}` (the JSON-Schema "any" schema) instead of recursing — semantically the correct stop and universally accepted by providers. Depth is bounded by the ref graph, not the recursion limit.
2. **Pointer-aware lookup**: resolve refs against `$defs`, `definitions`, and local document pointers (`#/…` paths from the root), not just the last segment of `$defs`. An unresolvable ref keeps its current pass-through behavior rather than raising `KeyError`.
3. **`convert_input_schema_to_tool_args`**: seed `defs` from `$defs` *or* `definitions` so legacy schemas resolve instead of silently degrading.

All three are behavior-preserving for today's working paths (`#/$defs/X` refs from pydantic behave identically; only cycles and currently-crashing inputs change). Verified locally: recursive model tools construct fine, and the crash cases in #2/#3 return sensible schemas.

Obila34:
## 📝 Changes Description

Fixes #10327

`_resolve_json_schema_reference` (dspy/adapters/types/tool.py) expanded `$ref`s with unbounded recursion and a lookup that assumed every ref's last path segment exists in `$defs`. All three failure modes from the issue are reachable from public API (`dspy.Tool(fn)` and MCP tool conversion via `convert_input_schema_to_tool_args`):

1. Self-referential pydantic models (e.g. `TreeNode` with `children: list["TreeNode"]`) → `RecursionError`.
2. Local refs not shaped like `#/$defs/<name>` (e.g. `#/definitions/Node`, `#/properties/x`) → `KeyError`.
3. Legacy `definitions`-only schemas → dangling `$ref`s passed through and arg types silently degraded to `Any`.

This PR:

- Resolves refs as **local JSON pointers** (with RFC 6901 `~0`/`~1` unescaping) against the document root, with a fallback last-segment lookup in `$defs`/`definitions`.
- Adds **cycle detection**: refs on the current expansion stack are tracked in a `frozenset`, mirroring the `seen` guard in `XMLAdapter._schema_to_xml` (#10239). When a cycle closes, `{}` (the accept-anything schema) is emitted instead of recursing, so depth is bounded by the ref graph.
- Leaves **unresolvable refs as-is** instead of raising.
- `convert_input_schema_to_tool_args` now resolves properties against the full schema, so `#/definitions/...` and other non-`$defs` pointers resolve, and legacy `definitions`-only schemas no longer silently degrade.

Behavior-preserving for today's working paths: `#/$defs/X` refs from pydantic resolve identically; only cycles and currently-crashing inputs change.

### Verification

- Ran the repro snippets from the issue: recursive `dspy.Tool(organize)` now constructs (cycle point becomes `{}`), `#/definitions/Node` resolves, the dangling `#/properties/x` ref passes through instead of raising, and the legacy `definitions`-only schema resolves with `arg_types == {"node": dict}`.
- Added regression tests in `tests/adapters/test_tool.py`: recursive pydantic model via `dspy.Tool`, recursive schema via `convert_input_schema_to_tool_args`, `#/definitions/...` ref, nested `#/properties/...` pointer, and an unresolvable ref left as-is.
- `pytest tests/adapters tests/utils/test_mcp.py tests/utils/test_langchain_tool.py` → 289 passed, 33 skipped (optional-dep skips).
- `ruff check` passes on both changed files.

## ✅ Contributor Checklist

## ⚠️ Warnings

None. Cyclic refs now expand to `{}` at the point of recurrence, which is the JSON Schema "any" form and universally accepted by providers.

### DIFF
--- a/dspy/adapters/types/tool.py
+++ b/dspy/adapters/types/tool.py
@@ -520,29 +520,62 @@ def _normalize_tool_call_dict(data: dict[str, Any]) -> dict[str, Any]:
     }
 
 
+_UNRESOLVED_REF = object()
+
+
 def _resolve_json_schema_reference(schema: dict) -> dict:
-    """Recursively resolve json model schema, expanding all references."""
+    """Recursively resolve json model schema, expanding all references.
 
-    # If there are no definitions to resolve, return the main schema
-    if "$defs" not in schema and "definitions" not in schema:
-        return schema
+    References are resolved as local JSON pointers (e.g. `#/$defs/Foo`, `#/definitions/Foo`,
+    `#/properties/foo`) against the document root. Cyclic references (e.g. from self-referential
+    pydantic models) are replaced with `{}` (the JSON Schema that accepts anything) instead of
+    recursing forever, and unresolvable references are left as-is instead of raising.
+    """
 
-    def resolve_refs(obj: Any) -> Any:
+    def lookup_ref(ref: str) -> Any:
+        if not ref.startswith("#"):
+            # Only local references can be resolved; leave remote ones untouched.
+            return _UNRESOLVED_REF
+        target: Any = schema
+        # Walk the JSON pointer segments from the document root, unescaping `~1` and `~0`.
+        for segment in [s.replace("~1", "/").replace("~0", "~") for s in ref[1:].split("/")[1:]]:
+            if isinstance(target, dict) and segment in target:
+                target = target[segment]
+            elif isinstance(target, list) and segment.isdigit() and int(segment) < len(target):
+                target = target[int(segment)]
+            else:
+                break
+        else:
+            return target
+        # Fall back to looking up the last path segment in `$defs`/`definitions`.
+        name = ref.split("/")[-1]
+        for definitions in (schema.get("$defs"), schema.get("definitions")):
+            if isinstance(definitions, dict) and name in definitions:
+                return definitions[name]
+        return _UNRESOLVED_REF
+
+    def resolve_refs(obj: Any, seen: frozenset[str]) -> Any:
         if not isinstance(obj, (dict, list)):
             return obj
         if isinstance(obj, dict):
-            if "$ref" in obj:
-                ref_path = obj["$ref"].split("/")[-1]
-                return resolve_refs(schema["$defs"][ref_path])
-            return {k: resolve_refs(v) for k, v in obj.items()}
+            ref = obj.get("$ref")
+            if isinstance(ref, str):
+                if ref in seen:
+                    # Cyclic reference: stop expanding and accept anything at this position.
+                    return {}
+                target = lookup_ref(ref)
+                if target is not _UNRESOLVED_REF:
+                    return resolve_refs(target, seen | {ref})
+            return {k: resolve_refs(v, seen) for k, v in obj.items()}
 
         # Must be a list
-        return [resolve_refs(item) for item in obj]
+        return [resolve_refs(item, seen) for item in obj]
 
     # Resolve all references in the main schema
-    resolved_schema = resolve_refs(schema)
-    # Remove the $defs key as it's no longer needed
+    resolved_schema = resolve_refs(schema, frozenset())
+    # Remove the definition keys as they are no longer needed
     resolved_schema.pop("$defs", None)
+    resolved_schema.pop("definitions", None)
     return resolved_schema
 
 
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):

Traced _resolve_json_schema_reference and convert_input_schema_to_tool_args in dspy/adapters/types/tool.py against the four new test cases in tests/adapters/test_tool.py.

The old resolver only handled $ref pointing into $defs (schema["$defs"][ref_path]), had no cycle protection, and would raise a hard KeyError on anything else. That's a real gap for a self-referential pydantic model (e.g. a tree-shaped model with children: list["Self"]), which produces exactly this kind of schema and would previously recurse forever.

The rewrite resolves $ref as a real local JSON pointer: it walks path segments from the document root (unescaping ~1/~0 per the JSON Pointer spec), falls back to a $defs/definitions lookup by the ref's last path segment if the pointer walk doesn't land cleanly, and returns a sentinel (leaving the $ref untouched in the output) if neither resolves - matching the new unresolvable-ref test's expectation of no exception. Cycle handling threads a seen: frozenset[str] of already-resolved $ref strings through the recursion; when a $ref reappears in seen, it returns {} (accept-anything) instead of recursing again.

I hand-traced the cycle case against the TreeNode test: at the top-level walk (empty seen), reaching $defs.TreeNode.properties.children.items = {"$ref": "#/$defs/TreeNode"} resolves it once with seen={ref}, and the second occurrence of that same $ref one level deeper (inside the resolved copy of TreeNode itself) matches the seen check and returns {}. That lines up with the test's assertion that tool.args["tree"]["properties"]["children"]["items"] == {} and "$defs" not in str(...). I also checked the #/properties/x pointer-walk test and the #/definitions/Node fallback test by hand against lookup_ref's logic and both check out.

convert_input_schema_to_tool_args now resolves the whole schema once via _resolve_json_schema_reference(schema).get("properties", {}) instead of re-wrapping each property individually with {"$defs": defs, **prop}, which was the old per-property approach. That's a genuine behavior change beyond a pure refactor (it now resolves against the whole schema, not just $defs), and it's specifically what makes the #/properties/x and #/definitions/Node cases resolvable, since the old wrapping trick only ever exposed $defs to the resolver.

What I didn't verify: I traced this by hand rather than running the tests, and I didn't check every call site of _resolve_json_schema_reference elsewhere in the codebase for a caller that might depend on the old KeyError-on-missing-ref behavior instead of the new silent-passthrough.

Issue #10256, opened 2026-08-21
[Bug] ReActV2 has no async path, so acall() fails and MCP tools cannot be used


jayhemnani9910:
## Bug description

`ReActV2` has no async path. `await ReActV2(...).acall(...)` raises
`AttributeError: 'ReActV2' object has no attribute 'aforward'`. The `ReAct` it is
meant to replace has supported async since it was added.

The consequence I care about is MCP. `convert_mcp_tool` builds an `async def` for
every MCP tool (`dspy/utils/mcp.py:82`), so MCP tools have no working route through
ReActV2 at all: the sync loop calls the function without awaiting it and records an
execution error, and there is no async loop to use instead.

## Actual behaviour

```text
  File "dspy/primitives/module.py", line 129, in acall
    return await self.aforward(*args, **kwargs)
AttributeError: 'ReActV2' object has no attribute 'aforward'. Did you mean: 'forward'?
```

Same async tool, both modules:

```text
ReAct   .acall() -> observation_0 = 'found cats'
ReActV2 .acall() -> AttributeError: 'ReActV2' object has no attribute 'aforward'
```

## Expected behaviour

`ReActV2.acall()` runs the loop and awaits async tools, the way `ReAct.acall()` does.

## Root cause

`dspy/predict/react_v2.py` on `main` defines seven methods and every one of them is
sync: `__init__` (`:24`), `_make_submit_tool` (`:37`), `_make_react_signature`
(`:60`), `forward` (`:88`), `_execute_tool_calls` (`:128`), `_history_event` (`:151`)
and `_forced_submit` (`:167`). There is no `async def` in the file.

`Module.acall` dispatches to `self.aforward(...)` at `dspy/primitives/module.py:129`
(and `:123` on the usage-tracking branch), and `Module` does not define an `aforward`
of its own, so the attribute lookup fails.

For comparison, `ReAct` pairs each sync method with an async one: `aforward` at
`dspy/predict/react.py:123` and `_async_call_with_potential_trajectory_truncation` at
`:167`. It reaches async tools through `Tool.acall`
(`dspy/adapters/types/tool.py:197`), which awaits a coroutine result and passes a
sync result straight through, so a mixed tool list works.

I checked whether this was already covered. PR #9235 ("feat(async): Add native
aforward() to remaining modules") adds `aforward` to `best_of_n`, `code_act`,
`multi_chain_comparison`, `program_of_thought` and `refine`. It was last updated in
[...]

PR #10335, opened 2026-09-04
fix(tool): offload sync functions in acall

immu4989:
Offloads synchronous functions invoked through Tool.acall() to DSPy’s AnyIO worker pool, preventing them from blocking the event loop. Native async callables remain directly awaited, and synchronous wrappers returning awaitables remain supported.
This is a prerequisite for #10256 and addresses the event loop blocking finding from #10257.
Tests cover event loop responsiveness, DSPy context and callback propagation, exceptions, cancellation, returned awaitables and worker limits.

### DIFF
--- a/dspy/adapters/types/tool.py
+++ b/dspy/adapters/types/tool.py
@@ -1,7 +1,9 @@
 import asyncio
+import functools
 import inspect
 from typing import TYPE_CHECKING, Any, Callable, Literal, Protocol, get_origin, get_type_hints
 
+import anyio
 import json_repair
 import pydantic
 from jsonschema import ValidationError, validate
@@ -9,6 +11,7 @@ from pydantic import BaseModel, TypeAdapter, create_model
 
 from dspy.adapters.types.base_type import Type
 from dspy.dsp.utils.settings import settings
+from dspy.utils.asyncify import get_limiter
 from dspy.utils.callback import with_callbacks
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):

Traced this in `dspy/adapters/types/tool.py`. The bug: `Tool.acall` previously called `self.func(**parsed_kwargs)` directly regardless of whether `func` was sync or async, and for a genuinely blocking sync tool (file IO, network call, sleep) that call runs on the event loop thread and blocks it entirely, since a plain synchronous call gives Python no point to yield.

The fix classifies `self.func` via `inspect.iscoroutinefunction(self.func)` or, for a callable object that isn't itself a routine, `inspect.iscoroutinefunction(self.func.__call__)`. Async functions are called and awaited as before. Sync functions are offloaded via `anyio.to_thread.run_sync(call, abandon_on_cancel=True, limiter=get_limiter())`. I checked `dspy/utils/asyncify.py` and this is the exact same pattern already used by `asyncify()` there, including the same shared `get_limiter()` `CapacityLimiter` sized off `dspy.settings.async_max_workers`, so this brings `Tool.acall` in line with an established pattern rather than inventing a new one.

The new tests are thorough: `test_async_tool_offloads_sync_function_without_blocking_event_loop` confirms the event loop stays responsive while a blocking sync tool runs, `test_async_tool_cancellation_does_not_wait_for_sync_function` confirms `abandon_on_cancel=True` lets a cancelled task return immediately without joining the worker thread, and `test_async_sync_tools_respect_async_worker_limit` confirms the shared limiter actually caps concurrency at `async_max_workers`. I hand-traced all three against the code and they line up.

One thing I didn't independently verify: `test_async_tool_propagates_context_and_callbacks_to_sync_function` checks that `dspy.settings` (`thread_local_overrides`, a contextvar based on its `.get()`/`.set()`/`.reset()` usage in `asyncify.py`) and the `ACTIVE_CALL_ID` contextvar are both visible inside the sync function running in the worker thread, even though this PR's simpler `anyio.to_thread.run_sync(call, ...)` call (unlike `asyncify()`'s `wrapped_program`) does no explicit context re-application. I'm relying on anyio's documented default behavior of copying the current context into the worker thread rather than having read anyio's source to confirm it myself.

PR #10353, opened 2026-09-09
fix: validate dspy.Image inputs through Pydantic

### LINKED ISSUE #10093: Align dspy.Image validation with Audio/File (add model_validator; make I/O boundary explicit)

ZIFeIYUuuuuuu:
## Summary

Closes #10093.

`dspy.Image` currently normalizes positional inputs in its custom constructor, but it has no Pydantic `mode="before"` validator. As a result, a bare in-memory data URI is accepted by `dspy.Image(uri)` but rejected by `TypeAdapter(dspy.Image).validate_python(uri)` and by nested `BaseModel` validation. `Audio` and `File` already support these Pydantic validation paths.

## Root cause

Pydantic v2 builds a dictionary-oriented schema for `Image` and rejects a bare string before the custom `__init__` can normalize it. The normalization and resource-loading boundary therefore depends on an implementation detail of constructor routing instead of being explicit in the model schema.

## Changes

- Added `Image.validate_input`, a `@pydantic.model_validator(mode="before")`.
- Normalize bare data URIs, URLs, bytes, PIL images, existing `Image` instances, and legacy `{"url": ...}` mappings through the existing `encode_image` helper.
- Preserve the existing constructor-only download/verify compatibility guard; validation still performs no filesystem or network I/O.

## Validation

- `uv run --extra dev pytest tests/adapters/test_resource_loading.py -q` (34 passed)
- `uv run --extra dev pytest tests/adapters/test_audio.py tests/signatures/test_adapter_file.py -q` (40 passed)
- `uv run --extra dev ruff check ...` was run; it reports pre-existing B019/RUF012 findings in untouched code. Import checks for the changed test pass.
- `git diff --check` passes.

The new regression test covers both `TypeAdapter` and nested `BaseModel` validation with a bare `data:image/png` URI, and asserts the URI is preserved exactly.


ZIFeIYUuuuuuu:
Good catch. The root of it is in `encode_image`, which treated any string starting with
`data:` as valid and returned it unchanged. Fixed in cbc1cd63.

That looseness predates this PR — constructing `Image("data:...")` directly has always
accepted `data:image/png;base64,not-valid-base64`. What this PR changes is that the
Pydantic path now reaches the same helper, so the same malformed value also passes
`TypeAdapter(dspy.Image)` and nested model validation instead of failing at construction.
It widened the exposure rather than introducing the bug.

I tightened it in `encode_image` rather than in `validate_input`, so direct construction and
the Pydantic path keep identical behaviour. Guarding only the new validator would have left
`Image("data:image/png;base64,not-valid-base64")` succeeding while the same value via
`model_validate` failed.

`_check_data_uri_syntax` checks the two properties an image reference needs before it can be
used: the RFC 2397 shape (`data:[<mediatype>][;base64],<data>`), and, when the header
declares `base64`, that the payload is actually base64. It is syntax-only, and stays
permissive exactly where strictness would reject input that works today:

- the payload is never decoded, so a well-formed URI whose bytes are not an image is still
  accepted and left for the provider to reject — decoding a 10 MB payload on every model
  validation would cost far more than this bug is worth;
- whitespace inside the payload is ignored, since data URIs are routinely wrapped across
  lines;
- missing `=` padding is tolerated. `base64.b64decode(..., validate=True)` rejects it with
  "Incorrect padding" even though such payloads decode fine, so the payload is padded
  before it is validated rather than rejected;
- non-base64 payloads such as `data:image/svg+xml;charset=utf-8,<svg .../>` are
  percent-encoded text and are left unchecked.

Malformed references raise `ValueError`, which Pydantic surfaces as `ValidationError`, so
both entry points fail, while `data:image/png;base64,AA==` from the existing regression test
still validates.

Verification: `tests/adapters/test_resource_loading.py` goes from 34 to 48 passing. Across
`tests/adapters` plus the audio, file, image-adapter, example, and streaming suites, 399
passed before and 413 after — the difference is exactly the 14 new cases — with an identical
skip/xfail/error set. The 2 collection errors were optional cloud extras and go away once
`litellm[proxy]` is installed; the one remaining streaming failure is identical in both runs.
`ruff format --check` is clean and `ruff check` reports the same 2 pre-existing findings
before and after.

The eight malformed shapes are exercised through the constructor, `TypeAdapter`, and a
nested model, plus a no-host-I/O guard and five inputs that must keep working.

ZIFeIYUuuuuuu:
Friendly ping — `Lint, Test, and Build` is still at `action_required` on `cbc1cd6`, so the test suite has not actually executed yet (only `Greptile Review` completed).

Could a maintainer approve the workflow run when convenient? I'd rather have CI confirm the change than ask you to take my word for it.

ZIFeIYUuuuuuu:
@isaacbmiller — quick nudge on this one. Greptile's P2 (malformed `data:` URIs accepted by `encode_image`) is addressed in `cbc1cd6`, but `Lint, Test, and Build` is still sitting at `action_required`, so the suite hasn't actually executed yet.

Could you approve the workflow runs — or let me know if the approach isn't what you'd want? The validation is syntax-only (RFC 2397 shape + base64 character set), deliberately without decoding the payload, and it pads before checking so payloads that are valid but missing their `=` padding still pass.

PR #10359, opened 2026-09-09
Vendor lm15 as a package subtree

MaximeRivest:
## Scope

Mechanical prerequisite for the LM migration: bundle lm15's Python package as a Git subtree, expose its public names under `dspy.lm15`, and verify that it ships in installed distributions.

**No change to `dspy.LM`, `BaseLM`, adapters, streaming, caching, exception types, or the 3.3 experimental contract.** The LM integration belongs in a separate PR.

## Review guide

Most of the diff is the unchanged upstream package. Review the non-vendor changes separately:

- `scripts/update_vendored_lm15.py`: fetch the cmpnd fork, split its package history, add/merge with `git subtree --squash`; clean checkout required, local commits only.
- `dspy/_vendor/UPDATING.md`: update procedure and provenance.
- `dspy/lm15.py`: re-export the original public objects, not subclasses or duplicate module trees.
- `pyproject.toml`: bundle typing marker, license and provenance; exclude vendor code from DSPy's Ruff checks.
- `tests/test_vendored_lm15.py` and `.github/workflows/vendored_lm15.yml`: public exports, pickle identity, converters, license/provenance and installed-wheel verification.


MaximeRivest:
Addressed both provenance findings in 4061bc3a7:

- Fixed version reporting upstream in lm15-dev/lm15-python@d188744, synced cmpnd, and updated the subtree without local vendor patches. Public `__version__`, vet output, and distribution builds now share the source package version.
- The updater records `version=1.0.0a1` alongside source/contract/split hashes. Corrected the vendor documentation to point to `lm15-provenance.txt`.
- Added an installed-metadata regression fixture (`lm15==7.8.9`): all 4 vendor tests passed on Python 3.10–3.14 using dspy-ci. Upstream version tests and lm15 sdist/wheel builds passed.

isaacbmiller (approved):
Found a couple things to fix. 1 is my own comment and 2+3 are agent found. Afterwards lgtm!

**1. Use uv consistently in `.github/workflows/vendored_lm15.yml`**

Could we use `astral-sh/setup-uv`, `uv build`, `uv venv`, and `uv pip install`, consistent with the existing wheel check in `run_tests.yml`?

Please preserve the separate environment and running tests outside the checkout—those ensure we test the installed wheel rather than the source tree. Invoking that environment’s Python directly with `-I -m pytest` is fine; no activation is necessary.

**2. Check out the source commit before splitting in `scripts/update_vendored_lm15.py`**

At lines 46–61, the updater clones upstream with `--no-checkout`, then calls `git subtree split`:

```python
git("clone", "--no-checkout", args.source, str(checkout))
# ...fetch and read source metadata...
split = git("subtree", "split", "--prefix=lm15", source_commit, cwd=checkout)
```

`--no-checkout` downloads Git history without putting the package files on disk. The intervening `git show` commands work because they read from Git’s stored history.

However, `git subtree split` first checks whether `lm15/` exists in the working directory—even when supplied a commit containing that directory. Since nothing has been checked out, it fails:

```text
fatal: 'lm15' does not exist; use 'git subtree add'
```

Please check out the requested commit immediately after resolving `source_commit` on line 48:

```python
source_commit = git("rev-parse", "FETCH_HEAD", cwd=checkout)
git("checkout", "--detach", source_commit, cwd=checkout)
```

This only materializes files in the temporary upstream clone, not the DSPy checkout. Checking out the pinned commit was verified to make the split succeed and produce the expected split hash.

**3. Exclude vendor code from the Ruff hook in `.pre-commit-config.yaml`**

Please add `dspy/_vendor/` to the Ruff hook’s existing exclusion:

```yaml
exclude: ^(dspy/__metadata__\.py|dspy/_vendor/.*|tests/reliability/.*\.py)$
```

The new `pyproject.toml` exclusion applies to file discovery, but pre-commit passes explicit filenames, which bypass that exclusion. Because the hook uses `--fix-only`, it can rewrite vendor code and break the unchanged-upstream guarantee. The pinned Ruff version was confirmed to propose import changes to the vendored `__init__.py`.

MaximeRivest:
Addressed all three review points in 7f2244056

MaximeRivest:
Squash-and-merge is now supported (76a664cd1). This supersedes my earlier request for a merge commit.

The updater recovers the previous package split from the recorded Python source commit, verifies the current vendor tree matches, reconnects a package-root merge base without changing files, then performs the normal subtree merge. It no longer relies on PR commit messages or subtree ancestry surviving GitHub squash merges. UPDATING.md now recommends the normal squash workflow.

Added offline tests for two consecutive squash/update cycles from fresh clones with the old split commits absent, both with and without subtree trailers retained in squash messages. Normal merges, no-op repeats, and refusal to overwrite committed vendor edits are also covered. All 7 vendor/updater tests passed on Python 3.10–3.14 via dspy-ci.

The upstream source pin must remain fetchable; otherwise the updater stops safely. No vendored files or PR B work changed. The push skipped the full matrix rerun because of the already documented baseline failures; the focused tests above ran on this exact commit.

PR #10360, opened 2026-09-09
fix: reject raw stream=True before it can be cached or sent to providers

### LINKED ISSUE #10345: [Bug] `dspy.LM(stream=True)` caches LiteLLM's live stream before rejecting it
### What happened?

Supplying LiteLLM's `stream=True` kwarg to `dspy.LM` returns a live LiteLLM
`CustomStreamWrapper`, but DSPy subsequently handles it as a completed response.
The first call fails with:

```text
TypeError: 'CustomStreamWrapper' object is not subscriptable
```

With `cache=True`, DSPy stores the live wrapper before LM.forward() fails while processing it. On a subsequent identical request, DSPy finds the cached wrapper and attempts to deep-copy it. This fails before the provider is called:  

```text
LMUnexpectedError: [<configured model>] cannot pickle '_thread.RLock' object
```

This exposes two coupled problems: `LM.forward()` cannot handle the live wrapper returned by `raw stream=True`, and DSPy caches that unsupported object before validating it. The cached wrapper then causes the subsequent identical request to fail during cache retrieval.

### Steps to reproduce

```python
import os

import dspy
from dotenv import load_dotenv
load_dotenv()

# Use a fresh in-memory cache so call 1 misses and call 2 hits.
dspy.configure_cache(enable_disk_cache=False, enable_memory_cache=True)
dspy.cache.reset_memory_cache()

lm = dspy.LM(
    model=os.getenv("TEST_MODEL", "openai/qwen3.6-35b-fast"),
    api_key=os.getenv("PROVIDER_API_KEY") or os.getenv("OPENAI_API_KEY"),
    api_base=os.getenv("PROVIDER_API_BASE") or os.getenv("OPENAI_BASE_URL"),
    stream=True,
    cache=True,
)
prompt = "Return the single word: SUCCESS"

try:
    lm(prompt)
except TypeError as exc:
    print(f"Call 1 (cache miss): {exc}")

print("Call 2: identical request, cache hit")
lm(prompt)   # Intentionally uncaught: raises LMUnexpectedError from cache deepcopy.
```

Observed output:

```text
Call 1 (cache miss): 'CustomStreamWrapper' object is not subscriptable
Call 2: identical request, cache hit
...
LMUnexpectedError: [<configured model>] cannot pickle '_thread.RLock' object
```

The same first-call `CustomStreamWrapper` error occurs with `cache=False`; caching
adds the second, persistent failure.

### Execution trace

This sequence follows directly from the tested source at commit
`35ef21f1689576c4fbe06c27e22eabb5db1b24b6`:

1. `dspy.LM(..., stream=True)` accepts `stream` through `**kwargs`; `BaseLM` stores
   those kwargs in `self.kwargs` ([base_lm.py:171-206](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/base_lm.py#L171-L206)).
2. `LM.forward()` merges `self.kwargs` into the request, wraps the completion function
   with `request_cache` when `cache=True`, and invokes it ([lm.py:232-257](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L232-L257)).
3. With no `dspy.streamify()` side channel, `litellm_completion()` returns the direct
   `litellm.completion(..., **request)` result ([lm.py:454-458](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L454-L458), [lm.py:495-510](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L495-L510)). LiteLLM documents `stream=True` as returning a streaming response and annotates `completion()` as returning `ModelResponse | CustomStreamWrapper` ([LiteLLM source](https://github.com/BerriAI/litellm/blob/main/litellm/main.py#L4902-L4969)).
4. On the cache miss, `request_cache` calls the completion function, stores its raw
   result with `cache.put(...)`, then returns that same result ([cache.py:267-286](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/cache.py#L267-L286)). Therefore the live `CustomStreamWrapper` is cached before `LM.forward()` sees it.
5. Only after that return does `LM.forward()` call `_check_truncation(results)`
   ([lm.py:253-265](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L253-L265)). `_check_truncation()` indexes `results["choices"]` ([lm.py:437-445](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L437-L445)), producing the first-call not-subscriptable error.
6. On call two, `request_cache` checks `cache.get(...)` before calling LiteLLM and
   returns immediately on a hit ([cache.py:267-277](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/cache.py#L267-L277)). The memory-cache read calls `copy.deepcopy(response)` ([cache.py:125-150](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/cache.py#L125-L150)); the observed `RLock` error is raised there and wrapped as `LMUnexpectedError` by `LM.forward()` ([lm.py:258-261](https://github.com/stanfordnlp/dspy/blob/35ef21f1689576c4fbe06c27e22eabb5db1b24b6/dspy/clients/lm.py#L258-L261)).

### Expected behavior

DSPy should reject `stream=True` when it is passed directly to either
`dspy.LM(...)` or an LM call. The validation should happen before a provider
request is issued or a cached value is read or written, and the error should
direct users to DSPy's supported streaming API.

I think the simplest fix would be to check for `stream=True` after the LM arguments are combined, but before DSPy accesses the cache or calls the provider. If this matches the intended behavior, I’d be happy to open a PR with the fix and regression tests :)

### Environment

- DSPy: `main` at `35ef21f1689576c4fbe06c27e22eabb5db1b24b6` (3.3.1)
- LiteLLM: 1.99.0
- Python: 3.12.3

### DSPy version

3.3.1

jabrailkhalil:
Fixes #10345.

Direct `stream=True` can send LiteLLM's live stream wrapper into DSPy's completed-response path. Reject this unsupported option before cache access or engine selection, with `LMConfigurationError` directing users to `dspy.streamify` / `dspy.settings.send_stream`.

The branch now incorporates current `main` and preserves the lm15 engine architecture from #10366. Construction is checked in `LM._get_initial_kwargs`; call-time validation lives in the shared `execution.prepare` path used by `__call__`, `forward`, `acall` and `aforward`. The ordinary supported streaming path is unchanged, and `stream=False` remains accepted.

### Validation

- `python -m pytest tests/clients/test_lm.py -k 'stream_kwarg or preserves_existing or retry' -q`: 14 passed. The direct-stream regressions cover auto/LiteLLM engines and all four entry points, asserting no cache get/put or engine selection occurred.
- Cache, engine execution, async execution, engine compatibility and streaming suites: 110 passed, 3 skipped. Two streaming integration tests could not start the local LiteLLM proxy on Windows because its optional `backoff` dependency is absent; both fail during fixture setup, before the test runs.
- Ruff 0.15.0 passes on all three changed files.
- Updated the existing branch with a merge from `main`; no replacement PR or force-push.


jabrailkhalil:
The CI workflows for the current head `884492211a` are waiting for contributor workflow approval (`action_required`): [Lint, Test, and Build](https://github.com/stanfordnlp/dspy/actions/runs/34623892253). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

PR #10363, opened 2026-09-10
fix(teleprompt): partial-credit fraction and zero format_reward handling

### LINKED ISSUE #10362: [Bug] bootstrap_trace partial-credit crashes with TypeError, and a 0.0 format_reward is rescored to the failure penalty
### What happened?

Two related defects in the format-reward path shared by `bootstrap_trace` and GRPO:

**1. The partial-credit branch crashes on every partial parse.**

`dspy/teleprompt/bootstrap_trace.py` (`patched_forward`) computes partial format credit as:

```python
present = list(parsed_result.keys()) if parsed_result else None
expected = list(failed_signature.output_fields.keys())
...
format_reward=format_failure_score
+ (failure_score - format_failure_score) * (present / expected)
```

`present` and `expected` are **lists**, so `present / expected` raises `TypeError: unsupported operand type(s) for /: 'list' and 'list'` the moment an adapter reports a partial parse (e.g. `JSONAdapter.parse` raises `AdapterParseError(parsed_result=fields)` whenever some but not all output fields parse). The intended fraction is `len(present) / len(expected)`.

**2. A legitimate `format_reward` of `0.0` is silently replaced by the failure penalty.**

Both consumers use `or` where `None` is the documented sentinel (`FailedPrediction.format_reward: float | None = None`):

- `bootstrap_trace.py` (`wrapped_metric`): `return prediction.format_reward or format_failure_score`
- `grpo.py`: `score = trace_instance[2].format_reward or self.format_failure_score`

With the in-tree defaults (`failure_score=0`, `format_failure_score=-1`), the fixed partial-credit formula yields `0.0` exactly when every expected field parsed — the best possible partial score — and `0.0 or -1` rescores it to the worst. Same falsy-zero class as #10321 (`seed=0`), applied to training rewards: GRPO then optimizes against corrupted signal.

### Steps to reproduce

```python
import dspy
from dspy.teleprompt.bootstrap_trace import bootstrap_trace_data

class TwoFieldSignature(dspy.Signature):
    text: str = dspy.InputField()
    number: int = dspy.OutputField()
    explanation: str = dspy.OutputField()

# Any LM response that parses only one of the two output fields, e.g.
# '{"number": 1}' under JSONAdapter, raises AdapterParseError with
# parsed_result={'number': 1} -> present truthy -> TypeError above.
```

### Expected

- A partial parse produces `format_failure_score + (failure_score - format_failure_score) * len(present) / len(expected)`.
- Only `format_reward=None` falls back to `format_failure_score`; `0.0` is a real reward and must survive.

I have a fix with a regression test ready and will open a PR.

### DSPy version

main (ca54a855)

CodeAlex52:
## Changes Description

Fixes #10362

Two defects in the format-reward path shared by bootstrap_trace and GRPO:

- Partial format credit divided two lists (present / expected), raising TypeError on every partial parse. It now divides their lengths
- A legitimate format_reward of 0.0 was replaced by the failure penalty because of an or fallback. Only None falls back now, in both bootstrap_trace and grpo

Added a regression test: a two-field signature with a one-field parse scores exactly 0.0 with failure_score=1 and format_failure_score=-1. The test fails on the old code with TypeError and passes with the fix. Ran tests/teleprompt/test_bootstrap_trace.py, 4 passed

## Contributor Checklist

## Warnings

None

### DIFF
--- a/dspy/teleprompt/bootstrap_trace.py
+++ b/dspy/teleprompt/bootstrap_trace.py
@@ -54,7 +54,7 @@ def bootstrap_trace_data(
     def wrapped_metric(example, prediction, trace=None):
         prediction, _ = prediction
         if isinstance(prediction, FailedPrediction):
-            return prediction.format_reward or format_failure_score
+            return prediction.format_reward if prediction.format_reward is not None else format_failure_score
         return metric(example, prediction, trace) if metric else True
 
     # Use `object.__getattribute__` to bypass the custom hook `Module.__getattribute__` so that we avoid
@@ -88,7 +88,7 @@ def bootstrap_trace_data(
                     failed_pred = FailedPrediction(
                         completion_text=completion_str,
                         format_reward=format_failure_score
-                        + (failure_score - format_failure_score) * (present / expected),
+                        + (failure_score - format_failure_score) * (len(present) / len(expected)),
                     )
                 else:
                     failed_pred = FailedPrediction(completion_text=completion_str, format_reward=format_failure_score)
--- a/dspy/teleprompt/grpo.py
+++ b/dspy/teleprompt/grpo.py
@@ -476,7 +476,7 @@ class GRPO(FinetuneTeleprompter):
                             )
 
                             if isinstance(trace_instance[2], FailedPrediction):
-                                score = trace_instance[2].format_reward or self.format_failure_score
+                                score = trace_instance[2].format_reward if trace_instance[2].format_reward is not None else self.format_failure_score
                                 example_training_data[group_idx].append({
                                     "messages": inp_messages,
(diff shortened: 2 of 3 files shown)


CodeAlex52:
Fixed in d1f69ca. The zero-output case now skips the fractional calculation entirely (`if present and expected:`) and keeps the `format_failure_score` baseline — a signature with no output fields can no longer divide by zero when the adapter reports a non-empty partial result. Added `test_bootstrap_trace_zero_output_signature_skips_fractional_credit`, which injects an `AdapterParseError` with a non-empty `parsed_result` against a zero-output signature; it fails on the old guard and passes with the fix.

PR #10366, opened 2026-09-10
Run the LM layer through lm15 engines and retire the 3.3 experimental types

MaximeRivest:
Switches DSPy's LM execution to shared engine orchestration, using the bundled lm15 vocabulary.

## Kept

- Ordinary `lm(prompt/messages)` calls return legacy lists; adapters keep that public boundary.
- Existing custom `BaseLM.forward/aforward` plugins, including DummyLM overrides, continue through legacy wrappers.
- Ordinary cache keys and old SDK response entries remain readable. Six repository-owned disk-cache fixtures capture actual 3.3.0-tag behavior; recorded expectations are unchanged.
- History, callbacks, usage, copying, and trusted saved-program loading remain DSPy responsibilities.
- LiteLLM remains installed and available as an explicit compatibility engine.

## Changed

- `engine="auto"` prefers native lm15 where the route and ordinary inputs can be represented. Unsupported routes/client settings or unrepresentable ordinary inputs select LiteLLM before execution.
- `engine="lm15"` requires native execution; `engine="litellm"` explicitly chooses compatibility execution. Custom engines can be supplied with an async counterpart.
- Authentication/provider failures never cause cross-backend fallback. Managed retries belong to DSPy; no replay after emitting a stream chunk.
- Explicit lm15 Requests return lm15 Responses. Native `n` fan-out uses sequential independent calls, which can increase latency and billed input tokens.
- Native streaming events are assembled canonically and bridged to existing DSPy listeners.
- Native cache records use plain serialized data, including in restricted-pickle mode. Explicit requests have a separate cache namespace.
- DSPy 3.4 deprecates legacy LM plugins/wrappers, public `messages=` calls, and custom `complete_legacy()` shortcuts for removal in 3.5. In 3.5, adapters and engines use lm15 Request/Response throughout; `lm("hello")` remains a convenience. Adapter TODOs track the remaining internal migration.

## Broken on purpose

The experimental DSPy 3.3 LM vocabulary is removed, not aliased:

- `dspy.LMRequest`, `LMResponse`, `LMConfig`, message factories, and `dspy.core.types` move to different lm15 objects/shapes under `dspy.lm15`.
- `forward_contract="typed_lm"` is rejected with migration guidance.
- `experimental=True` no longer changes ordinary LM return types.
- Old pickles containing removed experimental classes are not automatically migrated; ordinary SDK response-cache entries are a separate compatibility promise.

The migration page includes mappings and custom-engine examples.

## Limitations / not claimed release-verified

- Native costs remain unknown when reliable pricing is unavailable; token usage is tracked.
- Typed LiteLLM-engine streaming supports Chat Completions, not Responses streaming. Native Responses streaming is implemented.
- Text completions retain the ordinary compatibility path.
- Arbitrary custom engines require custom JSON state reconstruction; borrowed engines/transports are not closed by DSPy.
- Real-provider streaming, tool-call round trips, and multi-answer billing still need broader live verification.

## Needs maintainer review

- Confirm the deliberate break from the published experimental 3.3 type names for the 3.4 RC.
- Review native-default/pre-execution compatibility selection and native `n` billing behavior.
- Confirm the proposed legacy-plugin warning and 3.5 migration wording.
- Decide whether the small interruption/network-test fixes should stay here or be split out.


MaximeRivest:
Merged current main (including #10371 and ReActV2 async support) and pushed async execution hardening.

- Native adapter metadata preparation, cost lookup, media snapshot preparation, and cache I/O no longer block the event loop.
- Cancellation retains usage for completed candidates, including during pricing, retry backoff, final stream delivery, and complete-result cache storage; cancellation still propagates without retry.
- Added permanent concurrency, loop-reuse, cancellation, and HTTP Retry-After boundary tests. Corrected an existing async settings test to verify question/model associations without assuming scheduling order.
- The new cancellation coverage exposed the Python 3.10/3.11 asyncio.wait_for completion/cancellation race. Fixed upstream in lm15-python f26741a, synced cmpnd-ai, and imported through the subtree updater. Upstream server suite: 2,166 passed and 7 skipped on each Python 3.10–3.14.

An earlier full run also exposed an intermittent, inherited ParallelExecutor/tqdm registry race; it was not changed or excluded here. Broader settings-context and tool/status-streaming redesigns remain outside this batch.

dspy/clients/execution.py
```diff
@@ -0,0 +1,435 @@
...
+            # Preserve legacy environment-configured gateways rather than
+            # accidentally sending the same credentials to a public endpoint.
+            env_prefix = lm.model.split("/", 1)[0].upper()
+            if spec == "auto" and any(os.getenv(name) for name in (f"{env_prefix}_API_BASE", f"{env_prefix}_BASE_URL")):
+                native = False
+            if spec == "auto" and resolution.provider == "xai" and "api_key" not in clients and os.getenv("XAI_API_KEY"):
+                clients["api_key"] = os.environ["XAI_API_KEY"]
+            if (set(clients) - NATIVE_CLIENT_KEYS) or (resolution.provider.startswith("azure") and
+                                                       any(key in clients for key in ("api_base", "base_url"))):
+                native = False
```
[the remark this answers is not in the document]
MaximeRivest: Addressed in 19ae51744. Configuration-based backend selection is now shared by execution and capability lookup (dspy/clients/backend_selection.py). Auto-selected LiteLLM configurations query LiteLLM rather than lm15 metadata, including the effective custom_llm_provider override. Native-compatible configurations still do not import LiteLLM.

Adapter planning scopes capability queries to each call’s client options without copying/mutating the LM; async capability lookup runs off the event loop. Coverage includes constructor and call-time overrides, headers/timeouts, Azure base URLs, environment gateways, copied LMs, tool and JSON adapters, and concurrent native/compatibility calls on the same LM.

Server verification: clients/adapters/settings/Predict selection passed on Python 3.10–3.14 (629 passed each on 3.10–3.13; 621 on 3.14 with version-specific skips), plus 5 local-model and 20 paid live-provider tests. Actual message representability remains validated separately before execution. Leaving this thread open for re-review.

docs/docs/community/normalized-lm-api-migration.md
```diff
@@ -1,239 +1,174 @@
...
 
-The migration is staged so existing code keeps working while new code can opt into the typed path.
+DSPy owns response caching, retries, candidate fan-out, callbacks and history.
+Ordinary calls retain the old cache-key format and can read existing SDK response
+entries. Cache reads happen before engine construction. New native entries store
+plain serialized data and work with restricted cache deserialization. No old
+cache rewrite is required. Explicit typed calls have a separate cache namespace.
 
-## Guide for DSPy users
+Native `n` answers use `n` separate requests in sequence; LiteLLM compatibility
```
isaacbmiller, line 74: in sequence? should we have a todo do parallelize?

docs/docs/community/normalized-lm-api-migration.md
```diff
@@ -1,239 +1,174 @@
...
 
-Most users should not need to change anything in 3.3.
+`num_retries` counts additional attempts, with exponential delays of 1, 2, 4…
+seconds, capped at 60 seconds unless the provider specifies `retry_after`.
+Backend retries are disabled on the managed LiteLLM path. A stream is not retried
+once a chunk has been sent to the caller. Caching a completed stream does not
+replay fake token timing on a later ordinary cache hit.
 
-Default behavior remains legacy:
+`dspy.streamify` continues to use the existing listeners. Native events are
```
isaacbmiller, line 87: How is streaming affected by these changes? Will we get rid of streamify later? Or migrate it to the lm15 specific API?

dspy/adapters/json_adapter.py
```diff
@@ -65,6 +66,7 @@ def _json_adapter_call_common(self, lm, lm_kwargs, signature, demos, inputs, cal
             lm_kwargs["response_format"] = {"type": "json_object"}
             return call_fn(lm, lm_kwargs, signature, demos, inputs)
 
+    @with_capability_planning
```
isaacbmiller: [rq] what is with_capability_planning

dspy/clients/engines/errors.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def wrap_error(exc: Exception, *, model: str, provider: str | None = None) -> LMError:
+    """Preserve an existing DSPy error; translate backend failures otherwise."""
+    if isinstance(exc, LMError):
+        return exc
+
+    from dspy import lm15
+
+    if isinstance(exc, lm15.LM15Error):
```
isaacbmiller: are there any lm15 errors that we dont catch?

dspy/clients/engines/errors.py
```diff
@@ -0,0 +1,203 @@
...
+    if _exception_status(exc) is None and any(
+        phrase in message for phrase in ("api key", "apikey", "credentials", "environment variable")
+    ):
+        return LMNotConfiguredError
+    if "timeout" in class_name or "timed out" in message or "timeout" in message:
+        return LMTimeoutError
+    if "connection" in class_name or "network" in message or "connection" in message:
+        return LMTransportError
+
+    mappings = [
```
isaacbmiller: do we expect a custom engine to raise a dspy error? 
or do we expect user engines to raise lm-15 errors which we catch above? 

regardless of what the answer is, should our litellm engine conform to the desired logic as a good example of a non-default engine?

dspy/clients/engines/legacy_engine.py
```diff
@@ -0,0 +1,111 @@
...
+    def _response(self, raw, request):
+        if isinstance(raw, Response):
+            return raw
+        return response_value(raw, "chat", request)
+
+    def __repr__(self):
+        return f"{type(self).__name__}(plugin={type(self.lm).__name__})"
+
+
+class LegacyEngine(_LegacyConfig):
```
isaacbmiller, line 62: i quite like the LegacyEngine approach!

WDYT of adding a deprecation warning to use a native LM15 engine by 3.5?

dspy/clients/base_lm.py
```diff
@@ -57,115 +47,13 @@ def _import_lm_class(class_path: str) -> type:
 class BaseLM:
     """Base class for DSPy language models.
 
-    Most users should use `dspy.LM`, which is a `BaseLM` subclass.
```
isaacbmiller, line 60: can we add a comment in the docstring linking to what they should do instead

scripts/update_model_metadata.py
```diff
@@ -0,0 +1,44 @@
+"""Refresh DSPy's bundled metadata from a released LiteLLM wheel, without importing it.
```
isaacbmiller, line 1: can we add a note to the release documentation that this should be run?

tests/adapters/test_baml_adapter.py
isaacbmiller, line 1: Can we add a todo someewhere to migrate all of these to use LM15 instead of litellm

tests/clients/fixtures/lm_3_3_0/generate.py
isaacbmiller, line 1: should this whole fixture still remain?

isaacbmiller (approved):
Few questions, but lgtm assuming none of these cause design changes!

dspy/clients/engines/legacy_engine.py
MaximeRivest: Clarified and improved the 3.4/3.5 deprecation messaging in 28941af4f: 3.4 is the transition release; in 3.5, adapters and custom engines use lm15 Request/Response throughout, with no legacy wrappers or shortcuts. Added TODOs in adapter code for the remaining migration. `lm("hello")` stays a convenience.

docs/docs/community/normalized-lm-api-migration.md
MaximeRivest: Added TODOs beside both candidate loops in 24686e3ac: bounded concurrency, stable ordering, per-candidate retries/cancellation accounting, whole-call caching, and stream candidate identity. Keeping execution sequential for now; parallelism reduces latency, not repeated input-token charges.

docs/docs/community/normalized-lm-api-migration.md
MaximeRivest: Clarified in 24686e3ac: streamify stays the program-level API for field listeners, status messages, and the final Prediction. lm15 events describe one model call. The 3.5 migration moves the internal LM boundary to Request/Response and canonical events; it does not call for removing streamify.

scripts/update_model_metadata.py
MaximeRivest: Added a pre-tag step to the release checklist in 24686e3ac: select a released LiteLLM version, run the updater, review and commit the snapshot/provenance/license, then run the metadata tests. The script also points to that checklist.

tests/adapters/test_baml_adapter.py
MaximeRivest: Added a shared TODO in tests/adapters/conftest.py (24686e3ac) to migrate general adapter doubles and LiteLLM mocks to canonical Request/Response engines, with native-transport coverage where needed. Existing prompt/parsing assertions stay; dedicated LiteLLM compatibility tests remain separate.

dspy/clients/base_lm.py
MaximeRivest: Added in 28941af4f: the BaseLM docstring now links to the migration guide, names complete(Request) -> Response as the replacement, and states the 3.4 transition / 3.5 removal timeline.

tests/clients/fixtures/lm_3_3_0/generate.py
MaximeRivest: I would keep these: they contain real cache files written by DSPy 3.3, and the default-engine tests verify they are read before engine construction. They protect old-cache compatibility, not the deprecated plugin interface. Kept the generator for reproducibility and clarified this distinction in the fixture README (24686e3ac).

dspy/clients/engines/errors.py
MaximeRivest: Addressed in 1f0d01d03 plus the lm15 source update. Added explicit coverage for every exported lm15 error class, setup/capability failures, and all canonical error-event codes. DSPy preserves causes and useful metadata; lock timeouts and stream assembly failures now have distinct public types. Unknown engine exceptions are non-retryable rather than classified from message text. Python 3.10–3.14 matrices pass.

dspy/clients/engines/errors.py
MaximeRivest: Clarified and implemented in 1f0d01d03: new engines raise dspy.lm15 errors; LiteLLM translates its SDK exceptions into that same contract. DSPy alone projects failures into its existing public LMError family. Already-public errors remain intact for compatibility. Also narrowed retry/fallback boundaries so parser bugs, completed generations, and partial streams are not silently replayed. The migration guide documents the contract and trade-offs; ready for re-review.

PR #10381, opened 2026-09-11
perf(dspy): defer anyio, openai and jiter to their use sites

adriaanm:
This makes the core importable on CPython built for WebAssembly, which does not provide `ssl`, which `anyio` and `openai` need, and `jiter` is a compiled extension with no wheel for every target. Deferring them also speeds up importing dspy, from 1785 modules and 380 ms, down to 871 and 193 ms. 

Claude (Opus 5) produced this PR, and tested it in our wasm sandbox (componentize-py 0.25.0, CPython 3.14, wasmtime 46.0.1, wasi-sdk 30)

### DIFF
--- a/dspy/clients/engines/lifecycle.py
+++ b/dspy/clients/engines/lifecycle.py
@@ -3,7 +3,9 @@
 import inspect
 from contextlib import asynccontextmanager, contextmanager
 
-import anyio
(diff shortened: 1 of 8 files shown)


isaacbmiller (approved):
lgtm

PR #10382, opened 2026-09-11
feat(clients): allow disable caching to support runtimes without a file system

adriaanm:
Make it possible to fully disable caching (e.g. on targets without a file system and thus no home directory, nor a place to store a disk cache). Also defer imports around caching implementations until we know which one is needed.

Claude (Opus 5) produced this PR, and tested it in our wasm sandbox (componentize-py 0.25.0, CPython 3.14, wasmtime 46.0.1, wasi-sdk 30)


isaacbmiller:
lgtm!

PR #10383, opened 2026-09-11
feat(dspy):  dspy.settings.interpreter_factory overrides default PythonInterpreter

adriaanm:
`dspy.configure(interpreter_factory=...)` allows pointing RLM, Flex, ProgramOfThought and CodeAct at another code interpreter, the way `dspy.configure(lm=...)` points every predictor at an LM. The mechanism is that a module's code interpreter is only replaced by the global `dspy.configure` or a scoped `dspy.context` whenever the original value was`PythonInterpreter`, considered the default.

A custom code-executing module has to call `resolve_interpreter_factory()` where it creates an interpreter, which is why that helper is exported. Resolving any earlier stops `dspy.context()` scoping the way callers expect.

RLM folds the factory's `execution_instructions` into its action prompt, and a factory that changes per call would leave that text describing a runtime that is no longer running.

Flex's bridge passed its sub-predictors nothing when Flex had no explicit factory, running the inner code in a different backend from the Flex that bridged it.

Claude (Opus 5) produced and tested this PR, under close human supervision.


isaacbmiller:
lgtm

PR #10398, opened 2026-09-14
perf(dspy): reduce memory use in exact embedding retrieval

kiteretsu903:
## Changes Description

I noticed that exact retrieval copies the entire corpus embedding table for every query in a batch. I changed it to score queries against one shared table, keeping the existing scoring and sorting behavior. The FAISS path is unchanged.

I checked this with SciFact: 5,183 documents and 300 queries, batched in groups of 32. Peak memory dropped from 375 MB to 130 MB, with identical document rankings and scores. This measurement used precomputed embeddings.

## Contributor Checklist

- [x] Pre-Commit checks are passing (locally and remotely) — local checks passed; remote checks pending.

## Warnings
NA


isaacbmiller:
Clever! Thanks for the PR. LGTM

PR #10409, opened 2026-09-15
Update bundled lm15 to rc3: native timeouts, reliable stops and error diagnostics

MaximeRivest:
## Summary: two distinct changes

1. **Update DSPy's bundled copy of lm15 to 1.0.0rc3.** These are upstream library changes, imported unchanged into `dspy/_vendor/lm15/`.
2. **Update DSPy's own integration code** to use those capabilities and preserve DSPy's routing, history, error and compatibility behavior.

The distinction matters: lm15 implements provider communication; DSPy decides which engine runs, passes settings to it, records history, and presents errors to applications.

## A. Changes from lm15 — imported, not implemented in DSPy

**Location:** `dspy/_vendor/lm15/`, plus the generated `dspy/_vendor/lm15-provenance.txt` record.

These fixes were made in lm15's source repository and imported with `scripts/update_vendored_lm15.py`. The bundled package is byte-identical to that source package; it contains no DSPy-only patches.

| Area | What the imported lm15 code provides |
|---|---|
| Timeouts and connections | Configurable transport timeouts and a shared connection pool. Inference requests no longer override the caller's read timeout with fixed 60/120-second limits. Defaults: connect 10 seconds, read/write/pool 600 seconds, 100 connections. |
| Provider adaptations | Provider-specific adjustments such as omitting an unsupported seed or clamping temperature, recorded as `Response.adaptations`. Silent mode hides records without changing execution. |
| Request planning | `plan()` checks what a request would require without loading or refreshing credentials. Refusals identify the unsupported field through `error.feature`. |
| Native Responses stop handling | Streams internally, including for `complete()`, and closes at a matched stop. Handles stops split across chunks and text parts. |
| Token scores | Stop filtering preserves scores for whole retained tokens. When a stop cuts a token, `logprobs_complete=False` marks the missing score; the flag survives serialization. |
| Compressed replies | Handles split deflate headers and every gzip member. Corrupt or truncated later members raise an error instead of silently losing content. |
| Live-turn collection | Configurable defaults of 16 MiB of event data or 10,000 events per turn. The new non-retryable `CollectionLimitError` preserves partial data and any rejected event. It does not pretend the turn completed or automatically close the session. |

The import also includes the earlier fixes for a schema being sent after being recorded as dropped, duplicate cache-adjustment records, URL credentials appearing in warnings, malformed ports breaking warnings, and request/response error handling.

**Important boundaries:**
- lm15's timeouts are per-operation waits, not a deadline for the entire generation.
- A native stop that closes the stream early loses the final usage report. Closing does **not** guarantee that provider generation or billing stops.
- Live limits bound collected event data, not exact process memory. **DSPy does not add a live-session API in this PR.**
- The live-memory finding was previously described here as untouched code. That was incorrect: this import changes turn iteration to retain yielded events. The configurable limits address that collector's growth.

**Source versions:**
- Python: [`cd5d4d7`](https://github.com/cmpnd-ai/lm15-python/commit/cd5d4d73a3cc3bbea9fc7901b7b3c26007e7bf37), synchronized with lm15-dev/lm15-python.
- Contract: [`7677980`](https://github.com/lm15-dev/lm15-contract/commit/7677980e36c8bd0b14ae9bfc79e49446f403ad22).
- rc3 is the source version. This PR does not publish a PyPI release or claim that other language ports already implement the additions.

## B. Changes in DSPy itself — integration and application behavior

These changes are **outside** the vendored package.

### 1. Keep `timeout=` on the native engine and pass it through

**Files:** `dspy/clients/backend_selection.py`, `dspy/clients/execution.py`, `dspy/clients/engines/lm15_engine.py`.

DSPy now treats `timeout` as a supported native client setting, converts it to lm15's timeout configuration, and includes it in engine reuse decisions. It protects provider construction and closes the router-owned pool correctly.

A numeric timeout sets read/write/pool; connect keeps its default. An `httpx.Timeout` supplies its numeric components; `None` components retain lm15 defaults rather than disabling timeouts.

### 2. Use planning to choose the engine before inference

**Files:** `dspy/clients/execution.py`, `dspy/clients/engines/lm15_engine.py`.

DSPy calls lm15's planner. Under `engine="auto"`, remaining native capability refusals can select LiteLLM **before** an inference attempt. Authentication, timeout and provider failures do not cause a backend switch.

### 3. Add adaptation records to DSPy's history

**File:** `dspy/clients/execution.py`.

lm15 produces the records; DSPy copies them into its history entries under `"adaptations"`. Nothing is printed automatically.

### 4. Enforce stops on typed Responses calls through LiteLLM

**File:** `dspy/clients/engines/litellm_engine.py`.

This is a **DSPy fix**, not an upstream LiteLLM change. Request conversion was removing the stop setting without applying it afterward. DSPy now uses lm15's existing stop helper on the completed reply and records the adjustment.

**Difference from native lm15:** this path waits for the complete reply, so generation is **not stopped early** and the full reported usage is retained. Typed Responses streaming through LiteLLM remains unsupported. Chat Completions still sends stop sequences to the provider directly.

### 5. Preserve error details at DSPy's public boundary

**Files:** `dspy/clients/errors.py`, `dspy/clients/execution.py`, `dspy/utils/exceptions.py`, `dspy/__init__.py`.

- Preserve `LMUnsupportedFeatureError.feature`, including request-conversion refusals. Fill `features` when no explicit list was supplied.
- Add `dspy.LMCollectionLimitError` as the explicit, non-retryable DSPy counterpart of lm15's collection-limit error. Preserve limits, counts, accepted events and the rejected event. Keep partial-result assembly lazy rather than allocating text/audio while wrapping a limit error.
- Say missing-key remedies in DSPy's terms: `dspy.LM(api_key=...)`, not an lm15 router configuration.

### 6. Four gauntlet follow-ups fixed in DSPy (commit `16300030e`)

**Files:** `dspy/clients/lm.py`, `dspy/clients/engines/lm15_engine.py`, `dspy/clients/backend_selection.py`, `dspy/clients/execution.py`, `dspy/clients/engines/litellm_engine.py`.

- **Dotted gpt-5 names are reasoning models.** `gpt-5.6-luna` and similar now get the same validation and `max_completion_tokens` handling as `gpt-5`; the `-chat` variants remain the exception. Unfamiliar spellings such as `gpt-5.6chat` are not guessed.
- **Disabled `httpx.Timeout` components are not silently replaced.** In httpx, `None` means wait forever; the native engine bounds every wait. Under `engine="auto"` such a setting selects LiteLLM; under `engine="lm15"` it raises `LMUnsupportedFeatureError` with `feature="timeout"`. Previously the component quietly became 600 seconds.
- **Malformed text (a lone surrogate) is a plain `ValueError` before execution** on every engine setting, including forced `engine="lm15"`, so it is never reported as an unexpected engine failure or retried.
- **`top_k` survives typed Requests through LiteLLM.** On the Chat path it is forwarded as a LiteLLM argument (LiteLLM translates it for providers that take it); on the Responses path, which has no such field, it is dropped and recorded.

### 7. DSPy documentation and regression tests

The migration guide explains these integration choices, especially the two stop-handling paths. DSPy tests exercise them through `dspy.LM`, including regular/async calls, history, saved responses, error details and retry behavior.

## Relation to #10389

This supersedes the direct vendored timeout edit in #10389. **lm15 owns the timeout implementation; DSPy owns passing its settings into it.** Keeping the library fix upstream preserves the bundled package's source identity.

## Verification — separated by project

**lm15:** 2,536 Python tests passed, 5 local-Ollama tests skipped; paid subscription smoke tests excluded. All 16 Python contract directions passed with their existing skips. Harness self-test, specification and provenance checks passed. The rc3 wheel built and installed successfully. Compression regressions reproduced failures before the fix; both compression test files pass afterward, totaling 63 tests.

**DSPy:** the full normal local suite passed: 1,774 passed, 253 skipped, 2 xfailed, using 16 workers (`-m 'not extra and not deno'`). Public sample image/PDF downloads were allowed; real-model tests remained disabled. New regression tests exposed the integration gaps before the fixes. Pre-commit checks passed without bypassing hooks.

**GitHub at commit `dcb2d0645`:** 23 checks passed and 3 deployment checks were skipped, including successful Python 3.10–3.14 tests/builds, extra/Deno checks, real-model tests and installed-package verification. Greptile's automatic review reported 5/5 with no remaining blocking issues. This is the observed result for that revision, not a guarantee about future revisions.


isaacbmiller (approved):
lgtm

PR #10436, opened 2026-09-16
fix(adapter): detect Optional History and Tool fields

MohammadHijjawi97:
History and native tool-call detection compared annotations with `== History` or `origin is list`, so `History | None` and `list[Tool] | None` were skipped. Unwrap Optional, Union, and Annotated first.

## Contributor Checklist


MohammadHijjawi97:
Addressed. Native conversation-history formatting now unwraps Optional/Union annotations when deciding which fields belong in assistant content. Optional ToolCalls are emitted as native tool calls only, not also as [[ ## tool_calls ## ]] text.

PR #10440, opened 2026-09-17
Update vendored lm15 to a550dab (declared providers)

MaximeRivest:
Vendoring only: `scripts/update_vendored_lm15.py main` against `cmpnd-ai/lm15-python`, no DSPy code changes.

**lm15 change vendored** (`cd5d4d7` → `a550dab`, version still `1.0.0rc3`, contract pin unchanged at `7677980`):

- `RouterConfig(providers=(ProviderDefinition.chat(access, compat=...), ...))` declares a provider lm15's registry does not list (a gateway, a service not yet receipted) as the same pure-data triple a registry entry is. It routes by id and aliases (`id:model`, `id/model`), takes `api_keys`/`base_urls`, shares the router's pool and timeouts, plans without a key. `Resolution.declared` is true and `describe()` says "no lm15 receipts". A spelling a registry entry or litellm prefix already uses is refused at config time.
- `Resolution` carries `credential_policy` and `placeholder_key`, so `describe()` reads the resolution rather than the module tables (greptile's finding on the first revision of this PR); a declared non-hosted entry must be key-based, since the OAuth policies need their own adapter class.

Files: `dspy/_vendor/lm15/registry.py`, `dspy/_vendor/lm15/router.py`, `dspy/_vendor/lm15-provenance.txt`.

Why: the Fireworks custom-engine findings (cmpnd-ai/breaka-your-lm `CUSTOM_ENGINE_FINDINGS.md`) showed the engine rebuilding lm15's access-policy + compat-preset pair by hand to reach a provider the registry lacks, and losing `api_key`/`api_base`/`timeout` handling in the process. A follow-up DSPy PR exposes this as `dspy.lm15.register_provider(...)` and tightens the custom-engine contract; this PR only brings the dependency in.

Checks run locally: `tests/clients`, `tests/streaming`, `tests/test_vendored_lm15.py`, `tests/test_lm15_subtree_updater.py` — 673 passed, 21 skipped.


MaximeRivest:
Confirmed and fixed upstream in lm15 `a550dab`, re-vendored here. `describe()` was reading the module-level registry, which a declared provider is not in; `Resolution` now carries `credential_policy` and `placeholder_key` so `describe()` is pure data. Also tightened: a declared non-hosted entry must be key-based — the router binds bound entries with `access=`, while the OAuth policies use self-resolving constructors that take none, so a declared OAuth entry would have lost its declaration silently.

PR #10441, opened 2026-09-17
Custom engines own their connection: refuse client settings at every door, save through engine state

MaximeRivest:
Follow-up to #10409. Fixes the custom-engine problems found in the Fireworks gauntlet (cmpnd-ai/breaka-your-lm, `CUSTOM_ENGINE_FINDINGS.md`).

**Scope.** This PR is only about custom engine objects (`dspy.LM(engine=MyEngine())`) — backends that are not an HTTP provider, like a CLI or an in-process model. If your backend *is* an OpenAI- or Anthropic-compatible HTTP provider that lm15 doesn't list, don't write an engine: declare it with `dspy.lm15.register_provider(...)` instead — that's #10442 (stacked on this one), building on the lm15 `RouterConfig(providers=...)` vendored in #10440.

## What changes

**1. `api_key`, `api_base`, `timeout`, `extra_headers` etc. are now rejected when you use a custom engine.**
Before, `dspy.LM("x", engine=MyEngine(), api_key="k")` silently ignored the key: the engine used its own. Now it raises `ValueError`. Same on `lm.copy(api_key=...)` and on `lm("hi", api_key=...)`. Put those settings on the engine object instead. Built-in engines (`engine="auto"/"lm15"/"litellm"`) are unchanged.

**2. `lm.copy(engine=...)` replaces both the sync and async engine.**
Before, `copy(engine="litellm")` kept the old custom `async_engine`, so sync calls went to LiteLLM and async calls to the custom engine. Now `engine=` replaces the pair; pass `async_engine=` too if you want one. Also: the sync engine's `complete` must be a normal function, the async one's must be `async def`. Wrong kind is rejected at construction, not on the first call.

**3. Custom engines can now be saved and loaded.**
Give the engine `dump_state() -> dict` and `@classmethod load_state(state)`. `lm.dump_state()` then records the class path and that dict. Loading imports the class, so like custom LM classes it needs `allow_unsafe_lm_state=True`. `dump_state()` fails early if the class can't be imported back (e.g. defined inside a function) or the state isn't JSON. A class defined in a notebook (`__main__`) only loads in that same session — documented.

## Worth a reviewer's attention

- The refusal in (1) is a hard error with no deprecation warning. Reason: the `engine=` interface is new in 3.4 RC, nobody ships on it yet, and the old behaviour was a silent wrong-credential bug.
- Rule (1) is enforced in three places: `LM.__init__`, `LM.copy`, and `execution.prepare` (every call). If you touch call preparation, keep the third one.
- Two ReActV2-under-`streamify` tests are added. They show the streaming path works with native tool calls and fails the same way as non-streaming when the model writes the tool call as text. They do *not* explain the gauntlet's live failure.

## Tests

`tests/clients/test_custom_engine_contract.py` (33), `tests/streaming/test_react_v2_native_tool_stream.py` (2). Full suite passes; private matrix Python 3.10–3.14 green.


MaximeRivest:
Both findings confirmed and fixed in `a108fd787`:

- **Engine kind is checked at construction and on `copy()`**: the sync engine's `complete` must not be a coroutine function, the async engine's must be one (a callable object with an async `__call__` counts). Before, a sync method as `async_engine` was accepted and every `acall` failed with an unexpected `TypeError` when the `Response` was awaited. The reverse (an `async def complete` as the sync engine) is refused too — same gap, other side.
- **`dump_state()` verifies the class path imports back to the same class** before recording it (pickle's rule). A function-local class saved fine and could never be loaded; it is now refused at save time with 'define the engine class at module level'.

Tests for both; docs updated. Matrix on the new head: Python 3.10–3.14 all pass, local-model live stage 5/5.

docs/docs/tutorials/custom_lm_engines/index.md
```diff
@@ -1,6 +1,8 @@
 # Custom LM Engines
 
-In the DSPy 3.4 development API, you can supply your own execution engine to `dspy.LM`. The minimum synchronous interface is:
+In the DSPy 3.4 development API, you can supply your own execution engine to `dspy.LM`. An engine is for a backend that is not an HTTP provider DSPy's bundled lm15 can speak to: a CLI, an in-process model, an agent harness. It owns its whole connection, so the [rules below](#what-a-custom-engine-owns) apply.
```
isaacbmiller: Should we change this to not be a version dependent comment?

isaacbmiller (approved):
nit comment, otherwise lgtm

docs/docs/tutorials/custom_lm_engines/index.md
MaximeRivest: Agreed — now that the docs are versioned (#10315), the page already says which DSPy it describes. Changed in `c254d16`: "You can supply your own execution engine…", and the "before 3.4 they were silently dropped" sentence is now stated as the rule rather than as history. Kept the explicit versions only where they're the content (the 3.5-cutoff admonition and the migration guide's deprecation timeline).

PR #10442, opened 2026-09-17
Declare HTTP providers to the native engine: one binding per LM, a fallback that keeps the connection

MaximeRivest:
Follows #10441 (merged). Uses lm15 `RouterConfig(providers=...)` from #10440.

## What this adds

`dspy.lm15.register_provider(...)` lets you declare an OpenAI-compatible (or Anthropic-compatible) HTTP provider that lm15 doesn't have built in — Fireworks, a company gateway — and then use it like any built-in one:

```python
dspy.lm15.register_provider(
    ProviderDefinition.chat(
        AccessPolicy(provider="fireworks", auth_modes=("bearer",), env_keys=("FIREWORKS_API_KEY",),
                     base_url="https://api.fireworks.ai/inference/v1"),
        compat=OpenAIChatCompat(max_tokens_field="max_tokens"),
        aliases=("fireworks-ai",),              # so "fireworks_ai/..." also works
    ),
    metadata_namespaces=("fireworks_ai",),      # use the fireworks_ai entries of the model-metadata snapshot for capabilities and prices
)
lm = dspy.LM("fireworks/accounts/fireworks/models/deepseek-v4p1-flash", api_key=...)
```

`api_key`/`api_base`/`timeout` work, the env var is read if no key is given, `engine="auto"` and `"lm15"` work, saving/loading needs nothing extra (state is just the model string). This replaces writing a custom engine for such providers.

## Rules a reviewer should know

1. **Each `dspy.LM` snapshots the registry when it is constructed** and uses that snapshot for everything (routing, capabilities, pricing, sync and async engines). Registering or replacing a provider later does not affect existing LMs. Register at import time.
2. **If a call falls back to LiteLLM** (e.g. you passed `extra_headers`), it still goes to the declared URL with the declared key, via LiteLLM's generic `openai/` (or `anthropic/`) door. The alias is never given to LiteLLM as a provider prefix — LiteLLM's own `fireworks_ai/` would point at public Fireworks, not your gateway.
3. **The fallback only supports bearer auth on `openai/` and `x-api-key` on `anthropic/`.** A provider declared with another auth scheme gets an error instead of a fallback. Native calls still work for it.
4. **A refusal caused by your own compat declaration is final under `engine="auto"`** — DSPy will not "try LiteLLM anyway", because that would silently drop the setting your declaration said the server can't take.
5. **Aliases don't grant metadata.** Capabilities and prices come only from `metadata_namespaces=` (explicit) or from `supports=`/`models=` statements. Unknown stays "unsupported"; unpriced stays "unknown cost".
6. `FIREWORKS_API_BASE`-style env vars do not redirect a declared provider (they do for built-in ones).


MaximeRivest:
Both confirmed and fixed in `7080a29ab`; both were places where the fallback route carried less than the declaration.

- **Auth scheme (P1).** The fallback now selects the credential's scheme exactly as lm15 does natively (`select_scheme`) and refuses the route when LiteLLM's door cannot send it — `openai/` sends a bearer header, `anthropic/` sends `x-api-key`. A declaration on another scheme (`api-key`, `query-key`, or the wrong one for its dialect) gets no fallback rather than its secret in an undeclared header. Refused before any call is made. Trade-off, stated: those providers have no LiteLLM fallback; they still work natively, where lm15 renders their scheme.
- **Pricing (P2).** The compatibility engine now returns the declared provider, so errors name it and pricing comes from its `metadata_namespaces` — on the canonical path and on the legacy body path, where LiteLLM's own figure for the generic door's model string (another vendor's price for a colliding name such as `gpt-4o` on a gateway, or none) is replaced. Tested on sync legacy, canonical, and async paths; a colliding name with no namespace is reported as unknown, not as OpenAI's price.

Matrix on the new head: Python 3.10–3.14 all pass, local-model live 5/5.

MaximeRivest:
Confirmed and fixed in `1da69d2fd`. The fallback now resolves the credential the same way the native path does (lm15's `resolve_credential_value`): a zero-argument callable is invoked once per call, the value is used for the scheme check, and LiteLLM gets the string. Test added (rotating token, two calls, two different tokens sent). Matrix green on 3.10–3.14.

MaximeRivest:
One more commit, `ac1f6c5e5`: my registered-provider tests were reading LiteLLM's **live** cost map (the metadata loader fetches it at first use unless `LITELLM_LOCAL_MODEL_COST_MAP` is set), so three of them passed only by network and failed once a sibling test had pinned the bundled snapshot. They now pin the snapshot like `test_native_metadata.py` does, and use `deepseek-v4-flash`, which the snapshot lists. No code change. Matrix: 3.10–3.13 green; 3.14 had one unrelated timing flake in `test_local_interpreter.py` (upstream test, passes on re-run).

isaacbmiller (approved):
lgtm

PR #10451, opened 2026-09-18
Close the LM findings still open on main: streamify errors, dead-loop pools, strict schemas

MaximeRivest:
The DSPy items from the LM gauntlet (cmpnd-ai/breaka-your-lm `results/findings.md`, 2026-09-13) that were still open on `main` after #10409, #10441 and #10442. Each was reproduced against `main` before the fix.

## What changes

**1. Errors inside `dspy.streamify` are the program's own exception.** They arrived wrapped in an `ExceptionGroup` from the task group, so `except dspy.LMError` missed them. A group with one failure (even nested) is unwrapped; a group of several distinct failures stays a group.

**2. Async pools of finished event loops are released.** The LM kept one pool per finished `asyncio.run()` and `close()` only reclaimed the sync pool. Dead-loop pools are now dropped on `close()`, on `aclose()`, and when a new loop's pool is created; lm15's finalizers close the sockets. Verified with file-descriptor counts.

**3. `response_format=PydanticModel` sends a valid OpenAI strict schema.** OpenAI rejected any model with defaults or an optional submodel ("`required` must include every key"). The schema now follows the OpenAI SDK's own `to_strict_json_schema` rules. Caller-supplied raw schemas are untouched.

**4. `dspy.LM("openai/")` is refused at construction** instead of silently selecting LiteLLM and failing later.

**5. OpenAI's "use /v1/responses" refusal now says how in DSPy** — `model_type="responses"`. No endpoint is switched automatically: which models need it is OpenAI's policy, not DSPy's to guess.

## Worth attention

- (5) is a hint on the error, not a behaviour change. If maintainers prefer `openai/gpt-5*` to default to the Responses API, that's a separate decision.
- **Not changed:** an empty `choices` array still gives `['']` natively and `[]` on LiteLLM. A 200 with no choices is a malformed provider reply; the right fix is in lm15's parser (raise), then align DSPy's compatibility path after the next vendoring.
- Still open from the gauntlet and not in this PR: gevent (needs a gevent environment to reproduce), the engine-parity list and provider error classification (mostly lm15 mappings; need live provider bodies), Anthropic `CacheConfig(mode="auto")` (cause unknown, live).

## Verification

Tests live next to the code they pin (`test_streaming.py`, `test_lm_async_execution.py`, `test_lm_schema_conversion.py`, `test_lm.py`, `test_base_module.py`). Full suite 1907 passed; private matrix Python 3.10–3.14 all green, local-model live 5/5. 

**Also in this PR** (found by the failing Dependency Range check): `save(modules_to_serialize=...)` registered modules with cloudpickle by value and never unregistered — process-wide state keyed by module name. The registration is now scoped to the active saves, reference-counted under a lock so overlapping saves share it.


MaximeRivest:
Confirmed and fixed in `6615e074e`. `chat_to_responses` (the `engine="litellm"` + `model_type="responses"` path) now sends the same strict-shaped schema with `strict: true` as the canonical path — the same `response_format=Model` means the same contract on either engine. That path had never sent `strict` even before this PR; now both agree. Test added; the existing Responses-path test updated to expect `strict: true`. Matrix green on 3.10–3.14.

MaximeRivest:
Right, and that race was introduced by my previous commit (the old code never unregistered, so it could not race). Fixed in `65e58fd8a`: the registry is process-wide, so ownership is now counted process-wide — one lock, a per-module count of active saves, register on the first, unregister after the last; a module the caller registered before any save is never touched. Tested nested on one thread and across two threads (A leaves while B is mid-save; B still sees the module registered). Matrix green.

MaximeRivest:
Done — the PyInstaller hook and its `pyinstaller40` entry point are now #10452. This PR keeps only the lazy shim load (what makes `import dspy` work frozen). Removal is the last commit here so the earlier history stays as reviewed.

MaximeRivest:
Agreed on the message: `--collect-data dspy` skips `.py` files, so it did not collect the shim — I confirmed with a build (`--collect-data`: fails; `--add-data <site-packages>/dspy/predict/flex/_sandbox_shim.py:dspy/predict/flex`: works). Fixed in `1351d9638`: the error now gives the `--add-data` form, says why `--collect-data` is not enough, and points at DSPy's PyInstaller hook; the shim is declared as package data. Flex working in a frozen app *without flags* is what #10452 (the hook) delivers, split out at the maintainer's request — this PR's claim is only that `import dspy` works frozen.

tests/clients/test_gauntlet_remaining.py
isaacbmiller, line 1: Can we move these to the appropriate file rather than referencing gauntlet?

dspy/utils/pickle_by_value.py
```diff
@@ -0,0 +1,57 @@
+"""Serialize modules by value for the duration of one save.
```
isaacbmiller: Is there guidance from pickle on how this should work? Or what the desired interactions are here?

isaacbmiller (approved):
Confirmed that Pyinstaller is not relevant - we can close #10452 and remove the flex changes from this pr.

lgtm after that + test fix

MaximeRivest:
Done in `a84b415d5`: #10452 closed, the Flex/PyInstaller changes removed from this PR, and the tests moved out of the gauntlet-named file into the files for the code they cover (`test_streaming.py`, `test_lm_async_execution.py`, `test_lm_schema_conversion.py`, `test_lm.py`; the pickle-by-value one was already in `test_base_module.py`). Comments no longer reference the gauntlet either. Matrix green on 3.10–3.14.

PR #10452, opened 2026-09-18
Ship a PyInstaller hook so a frozen application collects DSPy's data files

MaximeRivest:
Split out of #10451 at maintainer request: this adds a packaging entry point and is reviewed on its own.

## What this adds

- `dspy/__pyinstaller/hook-dspy.py`: a PyInstaller hook that collects the files DSPy reads as **data** at run time — the Flex sandbox shim and the local-interpreter worker (both `.py` files read as text, hence `include_py_files=True`), the Deno runner, the model-metadata snapshot/provenance/LICENSE, and the vendored lm15 provenance.
- `[project.entry-points.pyinstaller40] hook-dirs = "dspy.__pyinstaller:get_hook_dirs"`: the mechanism PyInstaller documents for a package to provide its own hook, so `pyinstaller app.py` works with no flags.
- One ruff per-file ignore: PyInstaller requires the `hook-<package>.py` filename.

## Why

PyInstaller freezes modules as bytecode and ships nothing else unless named. Any DSPy feature that reads one of those files failed frozen (gauntlet 2026-09-13). #10451 fixes the import-time read so `import dspy` works; this PR makes the features that read data work.

## Verification

- Real `--onedir` build on this branch alone: `import dspy`, the Flex shim and the metadata snapshot all load; the hook is picked up automatically (`Processing standard module hook 'hook-dspy.py'`).
- `tests/utils/test_pyinstaller_hook.py` pins that the hook exists, names files that exist, and the entry point resolves.
- Matrix Python 3.10–3.14 green (offline).

## Worth attention

- The entry point makes DSPy visible to PyInstaller's plugin discovery in every install; it does nothing unless PyInstaller runs.
- `LocalInterpreter` spawns `sys.executable`, which is not a Python interpreter inside a frozen app. Its worker file is collected, but that feature is not freeze-safe; unchanged here.


MaximeRivest:
Agreed — a packaging test that passes when the packaging is absent enforces nothing. Fixed in `1c490d214`: the test now requires exactly one `pyinstaller40` entry point with value `dspy.__pyinstaller:get_hook_dirs` and checks it loads to the hook directory. An install whose metadata predates this branch fails with a reinstall hint (my own dev venv did, which is the point). Passes on a fresh install on 3.10–3.14.

MaximeRivest:
Closing per maintainer review on #10451: PyInstaller support is not a target.

